PR #526, opened 2013-10-17
use full paths in url comparison also provide Date and Server headers

purplefox:


vertx/WebServer.java
```diff
@@ -18,52 +17,71 @@
   private static String helloWorld = "Hello, World!";
   private static Buffer helloWorldBuffer = new Buffer(helloWorld);
   private static String helloWorldContentLength = String.valueOf(helloWorldBuffer.length());
+  private static DateFormat DATE_FORMAT = new SimpleDateFormat("EEE, dd MMM yyyyy HH:mm:ss z");
```
ddierickx, line 20: There is a redundant `y` here?

PR #833, opened 2014-04-17
Update JsonController.cs

branciat:
Method already exist in the base controller mvc


msmith-techempower:
Closing due to inactivity - please rebase off master and submit a new pull request if you would like to make this change.

Issue #707, opened 2014-01-09
Validation Failure - play-activate-mysql


msmith-techempower:
https://github.com/TechEmpower/TFB-Round-9/blob/master/i7/verify-test-1/latest/logs/play-activate-mysql/out.txt

<b>DB</b>
Invalid response:
expecting `{"id":3684,"randomNumber":8967}`
saw `[{"id":3684,"randomNumber":8967}]`

LadyMozzarella:
play-activate-mysql is passes/warns on master. Good to close?
![play-activate-mysql](https://cloud.githubusercontent.com/assets/4926965/6472220/ab6ae26e-c1a5-11e4-86da-96c5ab282df7.png)

Issue #659, opened 2013-12-10
play-activate-mysql: Review multi-query approach


bhauer:
The play-activate-mysql test permutation may be ignoring the spirit of the multiple query test.  My understanding is that Activate is an STM implementation and for the purposes of this test, it is effectively an in-memory cache.  This aligns Activate with (future) test type 7 from #133, which is a caching test.

As it is, the existing implementation yields the following outcome on EC2:

![ec2-mysql-multi-query](https://f.cloud.github.com/assets/777068/1717547/54f54cba-61d1-11e3-8ae4-3108b85a640f.png)

This is showing just MySQL tests (Mongo tests are not shown).

Put another way, if one to run a profiler on the MySQL database, all implementations should look the same—each web request should yield 20 database queries.  I suspect that play-activate-mysql is not actually running 20 queries per request, at least not consistently.

msmith-techempower:
play-activate seems to have been removed entirely.

PR #880, opened 2014-06-22
add zoot-activate framework

fwbrasil:
This PR introduces the zoot-activate stack as the replacement for play-activate-mysql.

Related issues:

#707 - Fix the return json for single db query.
#659 - Disable the cache and use eager queries, producing the same amount of database roundtrips as the other implementations.

Permutations:

finagle-mongo
finagle-mysql
finagle-postgres
spray-mongo
spray-mysql
spray-postgres

Related frameworks:

http://github.com/fwbrasil/zoot
http://github.com/fwbrasil/activate
http://github.com/twitter/finagle
http://github.com/spray/spray


LadyMozzarella:
Closing as it appears that this contributor's associated TFB repository has been removed, so we can no longer access the commits.

PR #920, opened 2014-07-11
[In progress] Use python logging framework consistently

hamiltont:
This is basically a forward-looking commit. Currently some modules use logging, some dont, some use 
print, etc. This just standardizes us on python's logging and provides enough logging code to allow 
people to have an idea of what is expected from their additions

Replaces `--verbose` flag with `--log` flag

I'd propose some minimal guidelines for logging: 
* Use info to report standard, expected system operation e.g. method entry
* Don't use info on methods you expect to be called more than 10 times (e.g. get_foo())
* avoid re-writing code just to add logs e.g. don't add `else` statements just to log

Obviously these are just some general thoughts, feedback welcome

### DIFF
--- a/toolset/benchmark/benchmarker.py
+++ b/toolset/benchmark/benchmarker.py
@@ -2,6 +2,7 @@ from setup.linux.installer import Installer
 from setup.linux import setup_util
 
 from benchmark import framework_test
+from utils import WrapLogger
 
 import os
 import json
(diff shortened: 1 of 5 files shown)


msmith-techempower:
Ugh, this is over a year old and DEFINITELY needs to be rebased off master.

@ssmith-techempower See what you can do on a new branch with these changes... probably a lot of merge conflicts. Maybe it makes more sense to open a new PR with the ideas set forth here (using `logging` rather than file handles); though, if you do please close this in favor of your new one.

msmith-techempower:
Okay, I'm closing this pull request. We are still using simple file handles for writing out/err logging, but this pull request is dead.

PR #921, opened 2014-07-11
New benchmark: ULib framework

stefanocasazza:
Hi,

I updated my pull request again to make it apply more cleanly with the new model of installation.

I save the building output on:
$HOME/FrameworkBenchmarks/installs/ulib/ULIB_BUILD_OUTPUT-$DATE.txt
and the program output on:
$HOME/FrameworkBenchmarks/installs/ulib/ULIB_SERVER_OUTPUT.txt

If there are other problems please give me the possibility to look at the content.

Thanks all for the time dedicating

Stefano

### DIFF
--- a/ULib/bash_profile.sh
+++ b/ULib/bash_profile.sh
@@ -1,0 +1,20 @@
+if [ -z "${FWROOT}" ]; then
+  export FWROOT=${HOME}/FrameworkBenchmarks
+fi
+#------------------------------------------------------------ --------------------------------------------
+# IROOT - Path of this test's install directory ($FWROOT/installs or $FWROOT/installs/pertest/<test-name>)
+#------------------------------------------------------------ --------------------------------------------
+if [ -z "${IROOT}" ]; then
+  export IROOT=${FWROOT}/installs
+fi
+
+# Set the root of our ULib installation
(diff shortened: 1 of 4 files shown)


bhauer:
Hi @stefanocasazza. We finally have Round 10 preview data available.  If you are available, could I solicit your opinion in the following thread at the Google Group?

https://groups.google.com/d/msg/framework-benchmarks/yyK8VclJ6Zc/Y4YRCWPUEJMJ

This is pertaining to the inclusion of a SQLite test which, upon further consideration, and in large part due to how it skews the results, I am considering hiding by default.  Assuming you have a dissenting opinion--which would be completely understandable--I would like to have that captured so that we can factor that into our future planning.

Thank you!

stefanocasazza:
Hi <https://github.com/brianhauer,>https://github.com/bhauer.

first, I'd like to thank you and the team for the amazing work on these
benchmarks.
Regarding the inclusion of a SQLite choice, my vote would actually go for
hiding by default.
I also think it's better to differentiate to avoid  how it skews the
results...

Congratulations again for the great job!

Stefano

2015-02-25 20:50 GMT+01:00 Brian Hauer <notifications@github.com>:

PR #938, opened 2014-07-23
More Erlang frameworks

b0oh:
I added three more Erlang frameworks: Mochiweb, Yaws and Misultin. Will be glad if someone can run benchmark, because I haven't possibility to run it. Thanks and have a nice day.


msmith-techempower:
Need to retest this since it's been over half-a-year and we may have addressed the `rebar` issues.

b0oh:
@msmith-techempower nice to hear that

msmith-techempower:
I am working on a branch that simplifies the setup process quite a bit, but I'll pull this branch and see if I can't get it working.

msmith-techempower:
Pinging @waiteb3 as our resident Erlang expert to try and update these new frameworks to the new suite.

msmith-techempower:
@ssmith-techempower We have fixed a lot of our issues with Erlang and Rebar, could you pull these changes, fix the merge conflicts, and test to see if we are ready to go here?

b0oh:
sure, will give a look

ssmith-techempower:
Looks like these haven't been updated since we moved away from using Python scripts to stop/start the frameworks. At least the following needs to be done:

* Move each framework into `FrameworkBenchmarks/frameworks/Erlang/<framework-name>`
* Rename the `benchmark_config` files to `benchmark_config.json`
* Consolidate the `install.sh`, `setup.py`, `start.sh` scripts into a single `setup.sh` bash script. This should have the `fw_depends` line, building the app, and running it. Frameworks don't need to be explicitly stopped anymore so that code can just be omitted.

You can look at `cowboy` for an example of an Erlang framework that works with the current tooling.

nbrady-techempower:
Unfortunately, rebar now breaks with the version of yaws attempted here, but I will keep working to find another solution. Closing this now in favor of #1849 and #1850.

b0oh:
@ssmith-techempower, @nbrady-techempower I lost my interest in benching erlang frameworks and as you used my code, I would like you to copy a erl_bench dependency as well that I will be able to remove it from my account. Thanks.

PR #1101, opened 2014-10-17
Rust Iron framework

torhve:


Ap0ph1s:
any updates on merging this?

hamiltont:
@Ap0ph1s I think the main concern is this line `curl -s https://static.rust-lang.org/rustup.sh | sudo sh`, becuase it was regularly returning a different version of rust and breaking this PR. I think @torhve updated this multiple times because Rust was moving so quickly. 

There's not much value in merging something that's only stable for a week, so IIRC the consensus is that this needs to be linked to one version, but it's unclear which version. If you're interested in seeing Rust results ASAP, feel free to pull this branch into your own repo, fix it to a single worthy version, and send in an updated PR

hamiltont:
I see this timeline, perhaps this could be fixed to the beta when it's out?

```
Rust 1.0.0-alpha – Friday, Jan 9, 2015
Rust 1.0.0-beta1 – Week of Feb 16, 2015
Rust 1.0.0 – One or more six-week cycles later
```

hamiltont:
Oh, I also see that the alpha candidate is supposedly feature complete, so it is not too unreasonable to fix this to the alpha if the beta is taking too long as the upgrade should be painless

Ap0ph1s:
Rust is improving every day, whether performance or features, or even both, so I think It can wait until beta1, or the actual 1.0, every version that comes out is more worthy than the last.

fbernier:
Rust 1.0 has now been released. I guess it would be worth putting some more work into this.

steveklabnik:
oh, i missed this PR, and wrote up https://github.com/TechEmpower/FrameworkBenchmarks/pull/1636 . Time to compare!

msmith-techempower:
Closing in favor of #1638

PR #1153, opened 2014-10-30
Force LF newlines in some files

### LINKED ISSUE #1147: Vagrant development setup xplatform improvement
Currently, when you clone the project to a Windows machine using the favored "repo has *nux line-endings, I'll use my system's line-endings", the code exists locally with Windows-flavor line-endings.

When a user tries to `vagrant up`, the code is copied into the vagrant instance with the Windows-style line-endings which causes everything to break.

There should probably be a stage of this copy where line endings are transformed from ANY-style to *nux-style during the `vagrant up` call.

hamiltont:
Closes #1147 by ensuring that certain files always use line feeds (not CRLF)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -15,3 +15,4 @@
 #
 *.sh text eol=lf
 *.py text eol=lf
+config/benchmark_profile text eol=lf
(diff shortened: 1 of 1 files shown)


nbrady-techempower:
This unfortunately does not solve any of the windows/symbolic link line ending problems. Will continue to look into it though.

PR #1164, opened 2014-11-02
Go: Load balance between multiple connection pools

tuxychandru:
Instead of using a single connection pool for all requests, this change
uses as many pools as CPUs available. This will reduce contention when
borrowing a connection for request.

I see an improvement of around 16% on my quad core laptop. The
improvement may be significantly higher on Peak hardware.

### DIFF
--- a/frameworks/Go/go/src/hello/hello.go
+++ b/frameworks/Go/go/src/hello/hello.go
@@ -10,6 +10,7 @@ import (
 	"runtime"
 	"sort"
 	"strconv"
+	"sync/atomic"
 
 	_ "github.com/go-sql-driver/mysql"
 )
@@ -28,6 +29,12 @@ type Fortune struct {
 	Message string `json:"message"`
 }
 
+type atomicCount uint64
+
+func (i *atomicCount) next() uint64 {
+	return atomic.AddUint64((*uint64)(i), 1)
+}
+
 const (
 	// Database
 	connectionString   = "benchmarkdbuser:benchmarkdbpass@tcp(localhost:3306)/hello_world"
@@ -41,13 +48,16 @@ const (
 )
 
 var (
+	npool   = runtime.NumCPU()
+	lbIndex atomicCount
+
 	// Templates
 	tmpl = template.Must(template.ParseFiles("templates/layout.html", "templates/fortune.html"))
 
 	// Database
-	worldStatement   *sql.Stmt
-	fortuneStatement *sql.Stmt
-	updateStatement  *sql.Stmt
+	worldStatements   = make([]*sql.Stmt, npool)
+	fortuneStatements = make([]*sql.Stmt, npool)
+	updateStatements  = make([]*sql.Stmt, npool)
 
 	helloWorldBytes = []byte(helloWorldString)
 )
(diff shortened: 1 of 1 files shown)


tuxychandru:
`sql.DB` itself can be shared between goroutines and is protected by a mutex, so switch to `db.Query` is not likely to reduce contention.

In this benchmark the database is local, so roundtrips will not be expensive. Also, I'd not use `fmt.Sprintf` in my production applications for the fear of SQL injection and I am certain that using prepared statements is common practice in most Go applications. Thus, using `Sprintf` in the benchmark will not reflect real world usage.

Reducing contention by using multiple smaller pools instead of one large pool seems like a clean solution to this problem, that can be sufficiently isolated and adopted in real applications with very little drawback.

methane:
@tuxychandru This benchmark, db is not local, but on same LAN.  I've tested on placement group and c3.8xlarge.  So this benchmark result is very similar to PEAK environ.  (low latency, 10Gbit Ethernet).

As you can see in [this stackdump](https://gist.github.com/methane/dd3c3f8de8128730bd63), lock contention is happen for Stmt.mu, not DB.mu.
`db.Query()` locks DB.mu as you said, but I think it should not be bottleneck since code between lock may not take long time.

I'll make another pull request using `db.Query()` with benchmark result in this month.
And I'll try to fix golang/go#9484 in Go 1.5.

methane:
Benchmark updated: https://gist.github.com/methane/1341a204256e5dab52ac
Go 1.5 will be nice if my patch is accepted.  And Sprintf is fastest on -c128.

methane:
@tuxychandru 

> Also, I'd not use fmt.Sprintf in my production applications for the fear of SQL injection and I am certain that using prepared statements is common practice in most Go applications. Thus, using Sprintf in the benchmark will not reflect real world usage.

When writing Go program without ORM, I use prepared statement for most case.  But sometimes uses `db.Query(stmt, params...)` for queries with variable length argument (e.g. `WHERE id IN (?, ?, ?, ?)`).

Since fmt.Sprintf with "%d" is safe placeholder replacement for integers, I use it in real world without  fear of SQL injection.  But I'm not sure it is a common practice.

> Reducing contention by using multiple smaller pools instead of one large pool seems like a clean solution to this problem, that can be sufficiently isolated and adopted in real applications with very little drawback.

OK, I agree.  I'm +1 now since I will be able to see difference between Go 1.4 and 1.5 on one large pool in other benchmarks like `Go/gin`.

Could you add comment and use smaller pools like:

```go
db.SetMaxIdleConns(maxConnectionCount/npool + 1)
```

methane:
@tuxychandru I've added `interpolateParams=true` option to `go-sql-driver/mysql`.
See go-sql-driver/mysql#309 and go-sql-driver/mysql#318.

So now we have another option: Stop using `DB.Prepare()` and use `interpolateParams=true`.

hamiltont:
Hey @methane and @tuxychandru - thanks for the progress so far, this appears to be yielding large speedups! If I'm reading the thread correctly, this PR is indeed something we are interested in merging, perhaps with a nice comment in the codebase summarizing the main points covered in this thread (and mentioning that this may be repaired by Go 1.5). 

@methane - you're my trusted expert for understanding Go's ecosystem, so please comment if you disagree with my short summary above, or think of somehting I've missed

methane:
I'm not familiar with Go community much. But I saw some article about prepared-statement vs interpolation recently.
So both of prepared statement and interpolation are common practice.

On the other hand, I feel multi pool and Sprintf are not common practice nor preferred way.
So I'm +1 on using `interapolateParams=true` option.

I'll create another pull request to use it.

hamiltont:
Great. Given the speedups I'd be comfortable merging this, but I think it absolutely needs a big comment in the source code explaining what's happening, especially if the reason for multiple connection pools is to work around limitations of some part of the Go ecosystem e.g. `sql.DB`. A link back to this thread would be good to

methane:
I've created #1353

hamiltont:
Just re-reviewed this and I'm confused - are @tuxychandru and @methane are in agreement on the direction the go framework should be taking? I'm eager to get any fixes necessary included in R10, but I'm no Go expert so I need to hear from you

It looks like #1353 uses this `db.Query` stuff, but in the thread above @tuxychandru says that won't create the performance benefits of using multiple connection pools. So, is #1353 intended to replace this PR as a stopgap until the inherent problem in `Go` is solved, or is #1353 intended to be merged in addition to this PR? The benchmark results above are excellent - has #1353 been tested similarly? If not, benchmarking #1353 would probably add valuable information to this debate - if it achieves similar results in a much more conventional manner, then by all means let's go with that, but if it doesn't achieve similar results then this PR has a lot of value IMO

methane:
I intended to #1353 replaces this PR.
While I hasn't benchmarked it yet, it may be similar to `Sprintf` in [this benchmark](https://gist.github.com/methane/1341a204256e5dab52ac).

LadyMozzarella:
Closing. I merged #1353 as replacement, as @methane intended. Please update and reopen if we should readdress this pull request.

PR #1175, opened 2014-11-04
Update Nimrod to 0.10.0

idlewan:
Update **Nimrod**, **Jester** and **Nawak** to their latest versions.
- Nawak incorporated its stdlib changes upstream, and is now using and dependent on the nimble package manager and the nim-zmq and nim-uuid package. It was also made compatible with the latest Nim/Nimrod.
- Jester is using its new async (unfortunately there is no scgi support for now).

N.B.: Nimrod the programming language is being renamed Nim when the official 0.10 comes out.

### DIFF
--- a/toolset/setup/linux/frameworks/jester.sh
+++ b/toolset/setup/linux/frameworks/jester.sh
@@ -5,4 +5,5 @@ RETCODE=$(fw_exists jester)
 
 git clone git://github.com/dom96/jester.git jester
 cd jester
+# most recent commit in branch new-async as of 2014-10-31
 git checkout 3e77ce753293b523004dfdaebd2fb195217f30f0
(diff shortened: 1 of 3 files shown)


Vikaton:
Nim has been updated to 0.11.2, what does it take to update it?

idlewan:
It's probably alright for jester, however nawak_mongrel is not yet compatible with it.

I guess you could take a look at https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/languages/nim.sh and have two versions of that file for Nim 0.10.2 and 0.11.2  (like python2/3 or java7/8 have) until nawak gets an update.

Vikaton:
I also think it;s a good diea to have bare Nim http server, which is very performent in 0.11.2

Issue #1124, opened 2014-10-21
Quiet down the bash scripts


hamiltont:
If we pass some "debug level" stuff from python down into the bash scripts, we can selectively enable or disable certain types of printing. 

At the moment I think two prime targets are bash tracing (set -x), which prints at least one line of output for each line of bash, and apt-get, which prints something like 1000 lines for most of our commands. It's very simple to create variables such as `$APT_OPTS=-q` or `$TRACE=false` and use these throughout our bash scripts, and then set these variables from python before calling into bash. I'm partial to always setting apt to quiet mode when inside travis, as it's not outputting high value stuff (the only high value stuff is errors, which quiet mode doesn't affect) and you only get 10k lines before Travis stops updating their "terminal". Currently we use something like 3-6k of these lines on nothing more than apt. 

Perhaps this is something that can be included in #920

nbrady-techempower:
Moving to the roadmap

PR #1196, opened 2014-11-06
use epoll for mojolicious (should other Perl frameworks do so too?)

jberger:
Just a quick config change that should hopefully improve mojolicious performance. If it does so, I feel conscientiously bound to mention that other Perl frameworks might also benefit from the change if done in their `bash_profile.sh` files too.

### DIFF
--- a/frameworks/Perl/mojolicious/cpanfile.snapshot
+++ b/frameworks/Perl/mojolicious/cpanfile.snapshot
@@ -8,6 +8,14 @@ DISTRIBUTIONS
     requirements:
       ExtUtils::MakeMaker 0
       common::sense 0
+  IO-Socket-IP-0.36
+    pathname: P/PE/PEVANS/IO-Socket-IP-0.36.tar.gz
+    provides:
+      IO::Socket::IP 0.36
+    requirements:
+      IO::Socket 0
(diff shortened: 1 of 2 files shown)


hamiltont:
@jberger still working on this one, or can I close it down for now?

jberger:
Travis choked on it last time or I would have encouraged merging it. Give
me a moment to see if the situation is improved or else I will close.

hamiltont:
No worries! BTW, perhaps it would be useful to follow in the footsteps of #1205 and scatter `2>&1 | tee $IROOT/perl-install.log | awk '{ if (NR%100 == 0) printf "."}'` inside of perl.sh, if we can safely assume that the build/install log output of perl is something we're not really interested in seeing fully most of the time

jberger:
OK to merge

hamiltont:
LGTM! Re your comment on IRC:

> because it took 2160 lines just to GET to that point

I know :-( Totally aware that the bash scripts are insanely verbose ATM, see #1124. In my view, fixing this is delayed a bit by #920, which is itself delayed by R10, but it's on my personal wish plate - I don't like linking people to travis and telling them "oh skim until line 5000 something"

PR #1205, opened 2014-11-07
Reduce verbosity of python and mono

aschneider-techempower:
Python and Mono are both established projects with little to no possibility of errors/issues during compilation and install. It really bugs me that in Travis, we run up the 10k line limit with them. Until we develop some way to use the caching mechanism on Travis, this hack should work.


hamiltont:
Sorry this one took so long to merge @aschneider-techempower - it's a great addition!

PR #1262, opened 2014-12-18
add ruby/ngx_mruby framework

kbrock:
Hello,

I was able to get this installed and running in a vagrant dev env.
It does exit after running the test, so there may be a bug in my code or the current suite.

Any pointers would be appreciated.

thanks


matsumoto-r:
Any update on this?

msmith-techempower:
I tested this out and every test seems to be failing.

msmith-techempower:
Trying to run it manually, I see [this](https://gist.github.com/msmith-techempower/e71d281e7f294cee8257) error.

matsumoto-r:
@msmith-techempower thanks for your check!

@kbrock Could you fix this?

kbrock:
Thanks @msmith-techempower 
I added `hiredis` to the list of package installations in [prerequisites.sh:44](https://github.com/kbrock/FrameworkBenchmarks/blob/nginx_mruby/toolset/setup/linux/prerequisites.sh#L44). Surprised this would fail for you.

@matsumoto-r I'll rebuild my vm and see if I can reproduce.
So hard to know if the travis failures are mine or random errors.

msmith-techempower:
Ah, I hadn't seen the updated `prerequisites.sh`, I will test again and make sure that is run.

msmith-techempower:
Hmm, that didn't seem to resolve the issue.

msmith-techempower:
MUCH BETTER! That was the trick. Okay, I have tested it and all tests pass; LGTM!

kbrock:
@msmith-techempower @hamiltont Thank you guys so much for helping here. And for this really cool project in general.

Now that I have one under my belt, I have a few others I want to add :smile:

matsumoto-r:
Great work! Thanks a lot!

msmith-techempower:
@kbrock No problem; glad like you like the project!

Just one thing to remember (and this is not yet a hard rule), we are trying to remain apt-get agnostic where possible, so if there is a distribution-agnostic way of installing a software, we would prefer that moving forward. If the software is only available via apt-get, then that will have to do.

Keep 'em coming and we'll keep measuring 'em!

hamiltont:
Thanks all, glad to get this merged in!

PR #1287, opened 2014-12-22
Fix ASP.NET

pdonald:
Plaintext, JSON and database tests should all work except for the fortunes tests.

Had to downgrade to ASP.NET MVC 4. When mono is released with [these changes](https://github.com/mono/mono/pull/1163), we can upgrade to ASP.NET MVC 5.2.2 and the fortunes tests will also work.

When I launch nginx myself, everything seems to work. But `toolset/run-tests.py --mode verify --install server --test aspnet-mono` doesn't work for some reason.

```
vagrant@TFB-all:~/FrameworkBenchmarks$ cat results/ec2/latest/logs/aspnet-mono/out.txt
test.os.lower() = linux  test.database_os.lower() = linux
self.results['frameworks'] != None: True
test.name: aspnet-mono
self.results['completed']: {u'aspnet-mono': u'port 8080 is not available before start'}
```


toolset/setup/linux/languages/mono.sh
```diff
@@ -1,30 +1,53 @@
...
+
+# temp dir for extracting archives
+TEMP=$IROOT/mono-snapshot-${SNAPDATE}-temp
+
+# start fresh
+rm -rf $TEMP && mkdir -p $TEMP
+rm -rf $MONO_HOME && mkdir -p $MONO_HOME
+
+# download .debs and extract them into $TEMP dir
+fw_get http://jenkins.mono-project.com/repo/debian/pool/main/m/mono-snapshot-${SNAPDATE}/mono-snapshot-${SNAPDATE}_${SNAPDATE}-1_amd64.deb
```
msmith-techempower: @pdonald 

http://jenkins.mono-project.com/repo/debian/pool/main/m/mono-snapshot-20141220092712/mono-snapshot-20141220092712_20141220092712-1_amd64.deb does not seem to resolve any more... is there a more permanent location?

PR #1299, opened 2015-01-02
Cleanup Nim tests

judofyr:
* Rename Nimrod to Nim
* Use Nim 0.10.2
* Upgrade to latest Nawak and Jester
* Move framework setup from toolset/ into install.sh
* Link Mongrel2 and Nawak with ZeroMQ properly
* Add proper .gitignore files
* Run servers in the background in setup.sh
* Remove bash_profile.sh

Nawak and Jester now works fine in my Vagrant installation.


judofyr:
Feel free to cancel the previous commit: https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/45682753 (I forgot to change Nimrod to Nim in .travis.yml).

frameworks/Nim/nawak/setup.sh
```diff
@@ -0,0 +1,24 @@
...
+sed -i 's|host=.* port=5432|host='"${DBHOST}"' port=5432|g' model_postgre.nim
+
+echo "** Compiling app"
+nim c --threads:on -d:release -d:postgre_model -l:-Wl,-rpath,$IROOT/zeromq-4.0.3/lib -o:nawak_postgre app.nim
+
+echo "** Starting mongrel2"
+cd conf
+mkdir -p run logs tmp
+m2sh load -config mongrel2.conf
+m2sh start -name test &
```
idlewan: Travis is failing for nawak and I could find no error in its log file, but the tests timeout (which is consistent with either mongrel2 or nawak not being started/not finding each other).

As I can see nawak's startup log lines, I'm suspecting mongrel2 might not have started correctly?

The following are the [only lines that are output by mongrel2](https://s3.amazonaws.com/archive.travis-ci.org/jobs/45683237/log.txt), and it's from the `m2sh load -config mongrel2.conf` line.
```bash
INFO:run-ci:stop: Unknown instance: 
INFO:run-ci:Fri, 02 Jan 2015 21:29:55 GMT [WARN] (errno: None) No option --db given, using "config.sqlite" as the default.
```
Do you know why can't we see anything that `m2sh start -name test &` should output? There should be many lines starting with `$date GMT [INFO] ` like the following I usually get when starting mongrel2 without sudo on my local machine:
```bash
Fri, 02 Jan 2015 22:41:06 GMT [INFO] (src/mime.c:49) MAX limits.mime_ext_len=128
Fri, 02 Jan 2015 22:41:06 GMT [INFO] (src/host.c:74) MAX limits.url_path=256, limits.host_name=256
Fri, 02 Jan 2015 22:41:06 GMT [INFO] (src/handler.c:343) MAX limits.handler_stack=102400
Fri, 02 Jan 2015 22:41:06 GMT [INFO] (src/config/config.c:85) Loaded handler 1:tcp://127.0.0.1:9999:0138a43-micro-nimrod:tcp://127.0.0.1:9998:
Fri, 02 Jan 2015 22:41:06 GMT [INFO] (src/config/config.c:227) Loaded route 1:/:handler for host 1:localhost
Fri, 02 Jan 2015 22:41:06 GMT [INFO] (src/config/config.c:312) Loaded 1 hosts for server 1:bd81667c-505a-41da-952c-1e672338db8a
Fri, 02 Jan 2015 22:41:06 GMT [INFO] (src/server.c:445) LOADING Handler tcp://127.0.0.1:9999
[...]
```

All your changes look good by the way.

msmith-techempower: Logging might be getting hidden because `m2sh` is being run in the background via the ampersand, maybe try the following:
`m2sh start -name test 1>$OUT 2>$ERR &`
which should force stdout to the normal log file and stderr to the normal err file (results/latest/logs/[test]/err.log).

dom96:
Looks good. Thank you :)

toolset/setup/linux/webservers/mongrel2.sh
```diff
@@ -23,7 +22,14 @@ fw_get https://raw.github.com/zedshaw/mongrel2/9b565eeea003783c47502c2d350b99c96
 mv -f zmq_compat.h mongrel2-install/src/
 
 cd mongrel2-install
-PREFIX=${IROOT}/mongrel2 make clean all
-PREFIX=${IROOT}/mongrel2 make install
 
-touch ${IROOT}/mongrel2.installed
\ No newline at end of file
+# Do this in a subshell to avoid leaking env variables
```
msmith-techempower: This technique is interesting, but ultimately meaningless in the context of this project - the `testrunner` user that invokes this script is run from a brand new shell, so we don't care if we leak environment variables into that shell. Specifically, once the test is completed, we nuke that shell anyway.

frameworks/Nim/nawak/setup.sh
msmith-techempower: I pulled this and tested it manually, it seems to be working after a few tweaks, but I'm not sure why Travis thinks that it isn't.

msmith-techempower:
I think that we are almost ready to merge this in, but there seems to be a bug with `fortunes` for `nawak-redis`:

```
Accessing URL http://localhost:6767/fortunes
  Response (trimmed to 40 bytes): "<!DOCTYPE html>
<html>
  <head><title>Fo"
   FAIL for http://localhost:6767/fortunes
     Invalid according to FortuneHTMLParser
   FAIL for http://localhost:6767/fortunes
     `<tr><td>11</td><td>&lt;script&gt;alert(This should not be displayed in a browser alert box.);&lt;/script&gt;</td></tr>` should be `<tr><td>11</td><td>&lt;script&gt;alert(&quot;This should not be displayed in a browser alert box.&quot;);&lt;/script&gt;</td></tr>`
Failed verify!
```

Looks like the quotation marks inside the "alert" are simply being stripped out instead of being turned into `&quot;`. If you can fix this one, I would be comfortable merging it in.

msmith-techempower:
Not sure why Travis doesn't like this; I just pulled a clean installation and it ran and passed everything; LGTM!

hamiltont:
Yea, unfortunately travis's predone setup, while great for everyone else, tends to wreak havoc with a lot of our projects. A big part of the run-ci.py file is actually unloading travis software so we are closer to a true clean state, but that's an error-prone process... :-/

PS - could you create an issue where we can track the things that travis fails on but local tests succeed? That would help me keep track fo what needs to be done on that front

msmith-techempower:
I can probably get a full list of that after the preview run that should finish later today.

hamiltont:
Cool. No rush

idlewan:
Thanks for merging!

@msmith-techempower : I checked what was going on with the fortune on redis: it's not specific to nawak, it's a bug in the insertion of redis data that I fixed with #1318.

msmith-techempower:
@idlewan I see, the quotation mark never made it in at all. Thanks for the update.

idlewan:
It was there at first, but an update to the file made the quotation disappear.

PR #1306, opened 2015-01-09
Attempt to fix onion (again)

msmith-techempower:
Should only require calling `make clean` to fix up onion.


msmith-techempower:
I believe this is due to `gcc` not being properly installed by default... ???

Every test I have run suggests that this is working appropriately. LGTM and fix that travis issue later.

PR #1307, opened 2015-01-12
Update Lwan tests

lpereira:
This updates Lwan to a newer version (which fixes a few bugs that might affect the benchmark results). As a byproduct, `libluajit-5.1-dev` is also installed as a dependency.

It also defines a `"default"` test, for `/plaintext` and `/json` endpoints.


hamiltont:
LGTM, thanks for the clean PR!

PR #1308, opened 2015-01-13
Updated activejdbc code

ericbn:
Also updated vagrant-production/Vagrantfile for virtualbox to allow creating virtual machines with different IPs each.


hamiltont:
Nice - one main question: 

Can you explain why you're advocating to drop the activeweb-raw configuration, and move many of it's tests into the activeweb configuration? Would it not make more sense to maintain the activeweb-raw and just add the new test types into activeweb? 

Before this PR, I see this verification summary: 

```
| Test: activeweb
|       plaintext   : PASS
|       json        : PASS
| Test: activeweb-raw
|       query       : PASS
|       db          : WARN
|       update      : PASS
|       fortune     : PASS
```

Whereas now I see this: 

```
| Test: activeweb
|       fortune     : FAIL
|       plaintext   : PASS
|       db          : PASS
|       update      : PASS
|       json        : PASS
|       query       : PASS
```

FYI @msmith-techempower the Vagrant changes included here LGTM

frameworks/Java/activeweb/benchmark_config
```diff
@@ -4,40 +4,23 @@
...
-      "setup_file": "setup",
       "db_url": "/activeweb/db",
       "query_url": "/activeweb/queries?queries=",
       "fortune_url": "/activeweb/fortunes",
       "update_url": "/activeweb/updates?queries=",
+      "plaintext_url": "/activeweb/plaintext",
       "port": 8080,
       "approach": "Realistic",
-      "classification": "Platform",
+      "classification": "Full-stack",
```
hamiltont: ping @msmith-techempower - can you take a look at all the changes to these files? I think you're better at understanding the intention behind the keys than I am, and multiple keys are being changed. Of note, I've never seen `platform="servlet"` that I can recall, I am unsure if jdbc (which uses activerecord concept I think) is a micro ORM or a full ORM, I still stink at distinguishing what the intention is behind platform vs full-stack, etc. Basically, I'm totally lost as to if this file is correct ;-)

hamiltont:
Also, any idea why the fortune test (which appears to have been passing before this PR - see [here](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/47840060#L4950) ) has become broken in the latest Travis-CI run on this PR?

frameworks/Java/activeweb/benchmark_config
msmith-techempower: I think specifying `servlet` as the platform is appropriate. We have Gemini listed as a "Fullstack" framework running on the "Servlet" platform.

Classifying something as a "Platform" is more like `Undertow` - while it implements all the tests, it is really not a framework per se.

msmith-techempower:
Not really sure - it looks like the index.html was simply minified and not really altered otherwise. The results from travis show that the `<script>` tags are not escaped properly: https://gist.github.com/msmith-techempower/19b38309f1c0d4dd9ec6

ericbn:
Thank you for following this up. I'll update the tests and also the dependency to our latest release of ActiveWeb and ActiveJDBC. The escaping of HTML characters was fixed in the release.

frameworks/Java/activeweb/benchmark_config
ericbn: I'll update `"classification"` to  `"Fullstack"`.

ericbn:
Corrections done.

ericbn:
@hamiltont, @msmith-techempower, this is ready to be merged now.

hamiltont:
This LGTM! @msmith-techempower I don't see any issues merging this while R10 is ongoing, but I'll leave it up to you

msmith-techempower:
@hamiltont Merge at your convenience.

hamiltont:
@ericbn The second preview for R10 is already underway I believe, so this won't be represented there, but it should make it into R10 official

ericbn:
Thank you, @hamiltont!

PR #1312, opened 2015-01-21
Replace Python Setup Files For Play2 Scala

LadyMozzarella:
Setup files replaced for play2-scala.

PR #1310, opened 2015-01-20
Fix HHVM

jmaddux-techempower:

PR #1302, opened 2015-01-06
make changes for EvHttp-sharp to use bash setup script and updated mono

lgratrix-techempower:
fix EvHttp-sharp

PR #1314, opened 2015-01-21
Remove bash profile

lgratrix-techempower:
Still a work in progress. The following tests currently include failing tests or failure of the framework to launch
All CSharp tests (awaiting merge of pull request #1302 and fixCSharp branch)
yesod
lapis
openresty
nawak
hhvm (possibly fixed in #1310)
laravel
pimf
silex-orm
silex
silica
django
pyramid
play-scala-mongodb (possibly fixed in 1313)
play2 (possibly fixed in #1312)


hamiltont:
Looking good so far! Don't forget to remove the bash_profile stuff from the [installer](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/installer.py#L102) and from the [framework_test](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/benchmark/framework_test.py#L165), and any references from the README files

@msmith-techempower what do you think about using this PR to also remove the [test-specific lines](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/config/benchmark_profile#L21-40) from benchmark_profile and rename it to `benchmark_profile.sh` so text editors can pick up the extension for coloring and indention presets

msmith-techempower:
@hamiltont I would be down with that, but we would not merge that in until after round 10 completes (it's going to break SO MUCH).

hamiltont:
ah truth. Perhaps that a different PR, let's just get this finished and merge it

msmith-techempower:
We can make this perfect - this is for after round 10 (which is imminent).

lgratrix-techempower:
In Reviewing the Travis results, I have sorted the failures into categories: Unaddressed in this change, Addressed by another PR, Works on local, and True issue. For those that work on my local I am asking bmazza to run a selection of those tests to see if they also work for her (which would lead to a suspected travis issue).
Unaddressed in this change:
weber, ngx-mruby

Addressed by other PR or Issue:
aspnet, evhttp-sharp, nancy, servicestack - 1287 and 1270
yesod - 1316
play2, play-scala-mongodb - 1313
hhvm - 1310
laravel - 1319
django, pyramid - 1317
silica - 1320

Works on local:
onion (works outside travis confirmed by msmith)
wt
dart-stream
cowboy
elli
wai
cakephp (works outside travis confirmed by bmazza)
codeigniter
php-phalcon (works outside travis confirmed by bmazza)
pimf
silex (some -raw tests tests fail, as they did on round 9, because methods being used to connect to db are incorrect, needs the attention of someone with silex experience to resolve)
silex-orm (some tests fail, as they did on round 9. Needs the attention of someone with silex experience to resolve)
symfony2
yii2
phreeze
grape
padrino
rack
rails
sinatra
urweb

True Issue
Nawak (I believe this stems from the renaming on Nimrod/Nim)
php-phpixie (change already submitted)
play-activate-mysql

lgratrix-techempower:
After removing Nimrod Folder, Nawak now runs on local

toolset/setup/linux/frameworks/wt.sh
```diff
@@ -37,4 +37,7 @@ cmake .. -DWT_CPP_11_MODE=-std=c++0x -DCMAKE_BUILD_TYPE=Release \
   -DCMAKE_CXX_COMPILER=$(which g++-4.8) -DDESTDIR=${IROOT}/wt \
   -DWEBUSER=$(id -u -n) -DWEBGROUP=$(id -g -n)
 make
-make install
\ No newline at end of file
+make install
+
+cd ..
+touch duda-0.23.installed
```
msmith-techempower: This should be `touch wt.installed` right?

frameworks/Ruby/sinatra/install.sh
```diff
@@ -10,5 +10,6 @@ rvm jruby-1.7.8 do bundle install --gemfile=$TROOT/Gemfile
 
 export LC_ALL=en_US.UTF-8 
 export LANG=en_US.UTF-8
+sudo apt-get install llvm-dev
```
msmith-techempower: This should probably be moved to the `prerequisites.sh` file.

frameworks/Ruby/rails/install.sh
```diff
@@ -10,5 +10,6 @@ rvm jruby-1.7.13 do bundle install --gemfile=$TROOT/Gemfile
 
 export LC_ALL=en_US.UTF-8 
 export LANG=en_US.UTF-8
+sudo apt-get install llvm-dev
```
msmith-techempower: Same here - `prerequisites.sh`

frameworks/Ruby/rack/install.sh
```diff
@@ -10,5 +10,6 @@ rvm jruby-1.7.8 do bundle install --gemfile=$TROOT/Gemfile
 
 export LC_ALL=en_US.UTF-8 
 export LANG=en_US.UTF-8
+sudo apt-get install llvm-dev
```
msmith-techempower: ditto

frameworks/Ruby/padrino/install.sh
```diff
@@ -10,5 +10,6 @@ rvm jruby-1.7.8 do bundle install --gemfile=$TROOT/Gemfile
 
 export LC_ALL=en_US.UTF-8 
 export LANG=en_US.UTF-8
+sudo apt-get install llvm-dev
```
msmith-techempower: ditto

frameworks/Ruby/grape/install.sh
```diff
@@ -10,5 +10,6 @@ rvm jruby-1.7.8 do bundle install --gemfile=$TROOT/Gemfile
 
 export LC_ALL=en_US.UTF-8 
 export LANG=en_US.UTF-8
+sudo apt-get install llvm-dev
```
msmith-techempower: ditto

frameworks/PHP/php-silex/setup_raw.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
+export PHP_HOME=${IROOT}/php-5.5.17
+export PHP_FPM=$PHP_HOME/sbin/php-fpm
+export COMPOSER_HOME=${IROOT}/php-composer
+export NGINX_HOME=${IROOT}/nginx
 
-sed -i 's|192.168.100.102|'"${DBHOST}"'|g' web/index_raw.php
+sed -i 's|172.16.98.120|'"${DBHOST}"'|g' web/index_raw.php
```
msmith-techempower: This should probably be changed to a regular expression that doesn't rely on whatever IP is in the file.

frameworks/Lua/lapis/setup.sh
```diff
@@ -4,6 +4,8 @@ sed -i 's|DBHOSTNAME|'"${DBHOST}"'|g' config.lua
 sed -i 's|DBHOSTNAME|'"${DBHOST}"'|g' config.moon
 sed -i 's|DBHOSTNAME|'"${DBHOST}"'|g' nginx.conf
 
+export OPENRESTY_HOME=${IROOT}/openresty-1.7.7.1
+export LAPIS_OPENRESTY=${IROOT}/openresty-1.7.7.1
```
msmith-techempower: Does this need to be set? It's the same value as `$OPENRESTY_HOME` - if it's required, please add a comment as to why it's the same value.

frameworks/C++/wt/setup_postgres.sh
```diff
@@ -1,7 +1,13 @@
...
+export BOOST_INC=${BOOST_ROOT}/include
+export BOOST_LIB=${BOOST_ROOT}/lib
 
-# This has to be included here since it is one of the vars NOT copied from the original environment
-export LD_LIBRARY_PATH="${BOOST_LIB}:${WT_LIB}:${LD_LIBRARY_PATH}"
+export WT_ROOT=${IROOT}/wt
+export WT_LIB=${WT_ROOT}/lib
+export WT_INC=${WT_ROOT}/include
+
+export LD_LIBRARY_PATH="${BOOST_LIB}:${WT_LIB}:${LD_LIBRARY_PATH}"pi
```
msmith-techempower: guessing this `pi` at the end of the line is a mistake?

hamiltont:
This seems to be looking great! Am I reading correctly that "True Issue" is the only category that's preventing a merge for this PR, and the only things remaining in that category are php-phpixie and play-activate-mysql?

LadyMozzarella:
@hamiltont I just wanted to point out/note that php-phixie is running successfully [(at least on Travis)](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/48659815) in @lgratrix-techempower's last commit (88f9a20: fix typos for wt setup files, move required ...). Also, play-activate-mysql is [fixed](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/52481535) in Pull Request #1324 and potentially fixed (I haven't tested it out) in Pull Request #1064.

hamiltont:
@LadyMozzarella thanks for the update. Seems like this right after R10 we have a symphony of merges ready to go

frameworks/Groovy/grails/setup.sh
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
-
 sed -i 's|jdbc:mysql://.*:3306|jdbc:mysql://'"${DBHOST}"':3306|g' hello/grails-app/conf/DataSource.groovy
 
+export RESIN_HOME=${IROOT}/resin-4.0.41
```
hamiltont: Typo here - forgot to add in the line `export JAVA_HOME=/usr/lib/jvm/java-1.7.0-openjdk-amd64`

hamiltont:
Note: We should restart the travis build on this to get a current snapshot of this PR versus master

toolset/setup/linux/installer.py
```diff
@@ -98,16 +98,6 @@ def __install_server_software(self):
       previousDir = os.getcwd()
       os.chdir(test_dir)
 
-      # Load profile for this installation
-      profile="%s/bash_profile.sh" % test_dir
-      if not os.path.exists(profile):
-        profile="$FWROOT/config/benchmark_profile"
-      else:
-        logging.info("Loading environment from %s (cwd=%s)", profile, test_dir)
-      setup_util.replace_environ(config=profile,
```
hamiltont: This is incorrect IMO. We still need to load the `benchmark_profile` file, so this should read: 

```
profile="$FWROOT/config/benchmark_profile"
setup_util.replace_environ(config=profile, 
  command='export TROOT=%s && export IROOT=%s' %
  (test_dir, test_install_dir))
```

toolset/setup/linux/languages/dart.sh
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash
 
-RETCODE=$(fw_exists dart-sdk)
+RETCODE=$(fw_exists ${IROOT}/dart-sdk.installed)
```
hamiltont: Thanks for updating all these to use the .installed convention

toolset/setup/linux/prerequisites.sh
```diff
@@ -60,4 +60,7 @@ RETCODE=$(fw_exists ~/.bash_profile.bak)
 
 sudo sh -c "echo '*               -    nofile          65535' >> /etc/security/limits.conf"
 
+#Ensure llvm-dev is installed (required for ruby to install correctly)
+sudo apt-get install llvm-dev
```
hamiltont: This can be added to the sudo apt-get above like so: 

```
cloc dstat                        `# Collect resource usage statistics` \
llvm-dev                          `# Needed for ruby`
```

toolset/setup/linux/systools/composer.sh
```diff
@@ -4,12 +4,16 @@ RETCODE=$(fw_exists ${IROOT}/php-composer.installed)
...
 mkdir -p php-composer
+cd php-composer
+
+fw_get https://getcomposer.org/installer -O composer-installer.php
 
-# Use the PHP from our IROOT directory
-PHP_HOME=${IROOT}/php-5.5.17
-${PHP_HOME}/bin/php composer-installer.php --install-dir=$IROOT
+# Use the PHP and composer from our PHP_HOME directory and 
+# COMPOSER_HOME directories (should be specified in frameworks
```
hamiltont: Instead of saying "should be specified in ...", either a) throw a big warning that this variable is required to depend on PHP, or b) just provide a default on the variable using something like this. That way the requirement is baked into the codebase instead of just as a comment

```
PHP_HOME=${PHP_HOME:-${IROOT}/php-5.5.17}
```

See [here](http://stackoverflow.com/questions/2013547/assigning-default-values-to-shell-variables-with-a-single-command-in-bash) for more info

frameworks/Ruby/README.md
```diff
@@ -12,8 +12,8 @@ In your framework's `bash_profile.sh`, put this:
 Because TFB uses Python's `subprocess` module, which runs 
 all shell processes in a non-login mode, you must source the 
 `rvm` script before using `rvm` anywhere. Placing the `source`
-call in your `bash_profile.sh` ensures that it will be called 
-before running `install.sh` and before running `setup.py`
+call in your `insatll.sh` before other commands ensures that it
```
LadyMozzarella: typo - insatll -> install

PR #1317, opened 2015-01-22
Fix django and pyramid

jmaddux-techempower:
The __init__.py's added back here are necessary for Python to recognize their respective directories as modules.


frameworks/Python/pyramid/frameworkbenchmarks/views.py
```diff
@@ -21,7 +21,7 @@ def test_2(request):
     """
     Test type 2: Single database query
     """
-    num = randint(1, 10001)
+    num = randint(1, 10000)
```
msmith-techempower: Is `randint` inclusive-inclusive? I feel like this was changed to `randint(1, 10001)` because it's inclusive-exclusive.

jmaddux-techempower: I tried ```randint(0, 2)``` in python 2.7 and 3.4, and both versions gave results of 0, 1, and 2.  It looks like the other python tests were using ```randint(1, 10000)``` as well.

msmith-techempower: All I needed to hear; LGTM!

PR #1318, opened 2015-01-23
Fix fortune insertion for redis

idlewan:
One of the fortunes didn't validate because of missing quotes in the alert.


msmith-techempower:
Tested this locally - looks good; LGTM!

PR #1324, opened 2015-01-29
Clean up play-activate-mysql (rename to play2-activate-mysql)

LadyMozzarella:
It looks like these two generally weren't passing travis, so I have updated them to do so (or almost do so in the case of play-activate-mysql). I didn't do any special cleaning up, as I noticed that was started in the [PR #1064: Rename Play 2 Activate and ReactiveMongo tests](https://github.com/TechEmpower/FrameworkBenchmarks/pull/1064). Hopefully this will allow @richdougherty to update/complete his pull request.

* Updated the setup scripts to match the other play2 setup scripts.
* ~~play-scala-mongodb passes both tests on travis now. [See travis test for this.](https://travis-ci.org/LadyMozzarella/FrameworkBenchmarks/jobs/48798450)~~
* play-activate-mysql now warns and passes for everything except the /db route. [See travis tests for this.](https://travis-ci.org/LadyMozzarella/FrameworkBenchmarks/jobs/48798449)


LadyMozzarella:
Note: Overall this branch won't pass Travis because of the failed /db test on play-activate-mysql.

msmith-techempower:
@LadyMozzarella If you change the controller code and `benchmark_config` to have separate URIs for db and queries, I think you'll be there.

Basically, `play-activate-mysql` currently has `/db` and `/db?queries=` for DB and QUERIES respectively. If you changed the controller code so that `db` did not require a query param and added a new end-point for queries (`/queries?queries=` for instance) and copy/pasted the current `db` controller code and renamed it `queries`, I think it will work.

LadyMozzarella:
This appears to be failing on and off in Travis now and I believe that it is because sbt needs more memory to run properly.

Every once and a while I will run a test on my local fork and the framework loads fine. I don't completely get it.

This is the error that keeps coming back:
```bash
ERROR:root:Failure running setup.sh
  File "/home/travis/build/TechEmpower/FrameworkBenchmarks/toolset/benchmark/framework_test.py", line 232, in start
    shell=True, stderr=errout, stdout=out)
  File "/opt/python/2.7.9/lib/python2.7/subprocess.py", line 540, in check_call
    raise CalledProcessError(retcode, cmd)
CalledProcessError: Command 'sudo -u testrunner -E -H bash -e setup.sh' returned non-zero exit status 137
ERROR:root:/home/travis/build/TechEmpower/FrameworkBenchmarks/installs/sbt/bin/sbt-launch-lib.bash: line 47: 13804 Killed                  "$@"
```

Any advice on how to handle this?

msmith-techempower:
If it works locally, then that's good enough; I'll test and merge if I'm satisfied.

richdougherty:
@LadyMozzarella, thanks for looking at this.

If sbt needs more memory then you can try passing a JVM options with something like `SBT_OPTS="-Xmx<some amount of memory>"`.

I fixed the db/queries issue in my PR, so I don't think that was what caused the failing tests for me. I actually have no idea what caused the failing tests because the tests worked locally for me. Maybe I was also getting a memory problem with sbt on Travis?

richdougherty:
@msmith-techempower, if you're happy to test locally then you might want to look at #1064. That PR fixes up lots of stuff, including moving these two Play tests into the play2 directory so that all Play tests use the same directory structure and setup scripts. That PR seemed to work for me locally, I wasn't sure about what was causing the Travis failure. I think PR #1064 would conflict with this PR though so it's up to you which one you want to use.

LadyMozzarella:
Thanks @richdougherty! @msmith-techempower - This PR isn't absolutely essential, as the tests are running fine locally. If you want to merge PR [#1064](https://github.com/TechEmpower/FrameworkBenchmarks/pull/1064) to get the extra tests, I can come back and look at it afterward and see if/what may need to be changed in order to get it passing Travis.

LadyMozzarella:
Opened PR #1388 because play2-scala-mongodb began actually failing. So, I'm separating play-activate-mysql and play-scala-mongodb pull requests. This PR now only contains changes for play-activate-mysql.

LadyMozzarella:
Since this is basically one PR or the other, I went ahead and moved the play-activate-mysql under play2-scala and renamed the test to play2-scala-activate-mysql (just in case).

LadyMozzarella:
Closing in favor of #1410, which incorporates changes (and original commit) from #1064.

PR #1326, opened 2015-01-30
Run rack-unicorn, rails-stripped-ruby, rails-unicorn, sinatra-unicorn, and grape-trinidad locally

LadyMozzarella:
rack-unicorn, rails-stripped-ruby, rails-unicorn, and sinatra-unicorn needed correct paths within nginx to their install route in order to run locally.


LadyMozzarella:
These don't pass Travis CI because all of the Ruby tests are looking for rvm in /home/testrunner/ and probably a couple of other things, but they run locally.

LadyMozzarella:
All of these tests run locally (unfortunately, not on Travis yet) with this pull request:
* grape-trinidad
* grape-unicorn
* rack-unicorn
* rails-unicorn
* rails-stripped-ruby

LadyMozzarella:
I just wanted to update on the status of all of the Ruby tests on Travis CI with this pull request now. I have about everything mostly working (every once and a while some of the tests don't pass) except for anything with rbx. I'm still looking into that.

All tests are passing for:
* ngx_mruby
* rails-stripped

Tests failing for:
* grape-puma-rbx (in grape)
* padrino-puma-rbx (in padrino)
* rack-puma-rbx (in rack)
* rails-puma-rbx (in rails)
* sinatra-puma-rbx (in sinatra)

msmith-techempower:
@LadyMozzarella I (finally) pulled your branch and tested out the original issue (`rack-unicorn` was my first test) and it threw an error :-(
`ERROR:root:nginx: [emerg] open() "/usr/local/nginx/conf/mime.types" failed (2: No such file or directory) in /home/techempower/FrameworkBenchmarks/frameworks/Ruby/rack/config/nginx.conf:40`

I think you need to add a line to the startup script (probably for the others as well) along the lines of:
```bash
sed -i 's|include .*/nginx/conf/mime.types|include '"${IROOT}"'/nginx/conf/mime.types|g' config/nginx.conf
```

Additionally, looking at that nginx.conf, it looks like the `root` is being set to `/path/to/app/current/public`. Now, the comments suggest that this is simply for static files, so it might not matter, but there is a chance that the startup of nginx will error on that not existing just as the mime.times file not existing. Therefore, you will probably want to clean that up in the start scripts as well:
```bash
sed -i 's|root .*;|root '"${TROOT}"'/public;|g' config/nginx.conf
```

LadyMozzarella:
@msmith-techempower Are you sure that you tested my branch correctly? I deleted everything, re-cloned, and tested my branch locally and am not seeing those errors. They also aren't appearing on Travis. I'll double check from my home computer when I get home, but I'm not getting the same errors. I've been using vagrant. Do you think that makes a difference?

LadyMozzarella:
ruby-tests-pass

msmith-techempower:
Okay, this is now passing locally for me; LGTM!

PR #1327, opened 2015-02-02
PHP Slim - Turn off 'beautiful' column names

LadyMozzarella:
The test 'slim' now warns/passes with this pull request. RedBeanPHP automatically converts table names that are camelCase to snake_case.
Turn this off in order to get the correct "randomNumber" table name for testing.


hamiltont:
Nice!

PR #1329, opened 2015-02-04
PyPy 2.5.0

methane:


hamiltont:
LGTM, and travis confirms that no Python tests were affected. Thanks for another good PR @methane

PR #1333, opened 2015-02-12
Boost python3 performance

GMLudo:
Hi,

I added `--with-computed-gotos` to compile Python3, because, at least on my computer, it increases +/- 5%.
Normally, you don't need to enable this option, ./configure detects that, but apparently, it doesn't work all the time.

Regards.


hamiltont:
Cool. Also, this is mentioned in the [official Py3 optimization documentation](https://docs.python.org/3.1/whatsnew/3.1.html#optimizations). Travis passes all green, this LGTM

PR #1336, opened 2015-02-19
Gemini postgres

knewmanTE:
This separates Gemini into non-db and db tests and adds a Postgres test.


frameworks/Java/gemini/source_code
```diff
@@ -27,9 +27,9 @@
 ./gemini/Docroot/WEB-INF/lib/mustache-compiler-0.8.13.jar
```
msmith-techempower: @knewmanTE We need to add the postgresql jar here as well.

msmith-techempower:
@knewmanTE I'm not sure why the postgresql tests are failing, can you check this out on a completely blank repository clone to see what's missing?

knewmanTE:
Yeah, I'll give that  shot.

msmith-techempower:
Closed in favor of #1344

Issue #1119, opened 2014-10-20
Gemini startup appears to be unreliable (at least in Travis-CI)


hamiltont:
Looking at some recent travis builds e.g. 447-458 from [here](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds) I noticed that gemini seems to cycle between pass/fail quite a lot - it's almost 50/50. There's no logic changes that I can see - nothing in git indicates major changes that should cause this, and the failed runs have messages such as this: 

```
INFO:run-ci:Server: Resin/4.0.36
INFO:run-ci:Content-Type: text/html; charset=utf-8
INFO:run-ci:Content-Length: 489
INFO:run-ci:Date: Sun, 19 Oct 2014 11:40:50 GMT
INFO:run-ci:
INFO:run-ci:<html><head><title>Temporarily Unavailable</title><style>body { background-color: white; color: black; }p { font-family: Arial, Helvetica, Sans-serif; font-size: 12px; }h2 { font-family: Arial, Helvetica, Sans-serif; font-size: 14px; font-weight: bold; }pre { font-size: 9px; }</style></head><body><h2>Temporarily Unavailable</h2>
INFO:run-ci:<p>This web site is temporarily unavailable due to maintenance work.  Please check back later.</p>
INFO:run-ci:<!-- Application not yet initialized. -->
INFO:run-ci:</body></html>
```

My only thought is that either a) there needs to be some "am I alive" check before you return from start in `setup.py` because the 60 second wait is not enough, or different travis workers have different properties that can cause gemini to not be installed or started correctly, such as less RAM.

msmith-techempower:
I believe this is resolved.

Issue #1050, opened 2014-08-22
Reduce project size and remove files from past commits


hamiltont:
There are a number of huge files/folders that should be downloaded from somewhere, not stored in the project tree. Some offenders: 
- Cakephp appears to be entirely stored within the PHP/cake/lib folder. 
- Scala & SBT stuff takes up > 400 MB across the project
- servicestack includes 50MB of library files

others :-(

The sbt issue is mostly that it is not centralized to the `installs` folder yet. Traditionally you can `rm -rf installs` to clean out all installation files, but the root `sbt` folder is a nasty exception to this rule

hamiltont:
Ok, this is partially done in the case of PHP. However, at some point there should probably be a rebase on master to rewrite history and remove some of the huge files from this repo's commit history. 

Two useful links: https://rtyley.github.io/bfg-repo-cleaner/ and http://stackoverflow.com/a/2158271/119592

hamiltont:
Ok, some hard numbers on the largest objects in our git tree are actually *not* showing any majorly-huge files in the git pack - the largest is 4MB. Translation: There may not be much more for us to improve upon unless we want to strip hundreds of "medium-sized" files e.g. 2-4MB files. Looking at these however, they seem needed to run the code e.g. not some useless huge log files.

Note that git automatically compresses files into a big binary blob sometimes - this is called the pack. So the actual space used when downloading or storage on disk, whereas the physical size is the uncompressed size. If you sort by the physical size (my first run does this), the largest file is an ~80MB log file and the second is an ~8MB code file

I used [this](https://gist.github.com/nk9/b150542ef72abc7974cb) script. 

```
Hamiltons-MacBook-Pro:FrameworkBenchmarks hamiltont$ python largestFiles.py -p -c 30
Finding the 30 largest objects…
Finding object paths…

All sizes in kB. The pack column is the compressed size of the object inside the pack file.

size   pack  hash                                      path
80962  441   0dda18b60f9857dbc0bbb84b5f323e0e6089de0e  php-fuel/fuel/app/logs/2013/10/02.php
8650   2143  c68b27a8ace847c03cd93e36552db961edcaddeb  frameworks/Go/revel/bin/benchmark
6098   621   cef975890ffec5393b56499ddcff86f20f218c79  aspnet/lib/EntityFramework.6.0.0-alpha3/lib/net45/EntityFramework.xml
[...]

PR #1344, opened 2015-02-25
reimplemented Gemini-Postgres due to Travis error

knewmanTE:
See the original [merge request](https://github.com/TechEmpower/FrameworkBenchmarks/pull/1336)

Broke off a clean branch from master and reimplemented the changes for the Gemini Postgres test after tests passed locally but failed on Travis.


hamiltont:
@knewmanTE @msmith-techempower Is there any way to download the contents of the WEB-INF/lib directory instead of storing them into this repository? One of the things we found over in #1050 was that these 0-4MB binary files are the majority of the repository's 120MB `.git` folder, so I'd advocate for always requiring PRs to download such dependencies

msmith-techempower:
@hamiltont The answer is "yes" for everything except the gemini jar-file, since it is not open source. I could spend some time and convert everything to maven and only include a `POM.xml` and 1 ~4mb jar file.

msmith-techempower:
This still seems to be failing. I'll check it out and take a look.

msmith-techempower:
I have no idea why Travis is failing on this - I created a new VM image and tested `gemini-postgres` and it worked fine. LGTM!

hamiltont:
Linking to #1119 - perhaps related?

msmith-techempower:
OH, that's a good point. Almost certainly related to #1119

PR #1346, opened 2015-02-25
Fix uwsgi errorlog is written to `--ini` file.

methane:
uwsgi's `-d` option requires argument specifying file to write errorlog.
Currently, errorlog is written into file named `--ini`. So I cannot see any messages in [err.txt](https://github.com/TechEmpower/TFB-Round-10/blob/master/peak/linux/results-2015-01-21-peak-preview1/latest/logs/bottle-nginx-uwsgi/err.txt).


msmith-techempower:
This is awesome! LGTM!

PR #1347, opened 2015-02-25
Upgrade Boot and customize Undertow

bclozel:
* upgrade to Spring Boot 1.2.1
* customize Undertow container


msmith-techempower:
Pulled and tested locally - this is all working; LGTM!

PR #1348, opened 2015-02-25
Update Lwan to a version that supports request pipelining.

lpereira:
After seeing the abysmal results from Lwan in the preview round, I had to either pull Lwan from the `plaintext` test or lobby @bhauer to make pipelining optional per framework. The first wouldn't be fair to Lwan, the second wouldn't be fair to all of the other frameworks.

So I implemented pipelining support last night. It wasn't that difficult after all; the rule clarification regarding pipelined requests served as a motivator.

This pull requests updates Lwan to the latest version.

### DIFF
--- a/frameworks/C/lwan/install.sh
+++ b/frameworks/C/lwan/install.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-REV='8ae98d2b6e146948a7ed13e76141e0f69228d690'
+REV='49607addb31879e2aa2b701317773674662315aa'
 
 INSTALLED_FILE="${IROOT}/lwan-${REV}.installed"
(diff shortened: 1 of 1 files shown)


msmith-techempower:
I love how simply this change appears in the commit. I will wait for travis' okay, then merge.

lpereira:
Please do not merge this yet.

lpereira:
The bug I've been chasing the past week has been finally squished, so I've uploaded this pull request.  Feel free to merge once Travis tells its OK to do so.

hamiltont:
LGTM, and because it's a "fix" I think we're good to merge. Don't think this will be included int he next round of previews because they are already started IIRC, but it'll make it into official R10

lpereira:
Awesome, thanks!

PR #1349, opened 2015-02-25
Update netty to chase the performance enhancements made since the latest release

lneves:


lneves:
Can this make the cut for round 10? There are some relevant performance improvements in the latest netty version.

hamiltont:
LGTM. 

@lneves the second preview for R10 has already started IIRC so this won't be represented in those results, but it will make it into the final R10

lneves:
Nice. Thanks!

PR #1350, opened 2015-02-25
update to wildfly 8.2

codylerum:


codylerum:
@hamiltont  Possible to get this in for 10?

hamiltont:
Yup, this LGTM and is a reasonably small PR to fit between preview and official round. I think the second preview has already started so this won't be represented there, but it will make it into official R10

PR #1353, opened 2015-02-26
go: Use interpolateParams=true for MySQL

methane:


methane:
Same pull request passed travis-ci https://github.com/methane/FrameworkBenchmarks/pull/8

PR #1354, opened 2015-02-27
remove need for sudo in setup.sh; and update ringojs runtime

oberhamsi:


oberhamsi:
added 67be6cb to adress travis failing https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/52435887#L2996

i don't see the output for the language installer ringojs.sh in travis but hopefully this was it

hamiltont:
@oberhamsi Ringo's output for `install.sh` is [here](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/52522823#L2656) (note that it's hidden by default, you need to click the "expand" arrow next to the lin `/toolset/run-ci.py install "$TESTDIR"`). 

Something went really wrong with that entire travis-CI run - things like java and wget failed to install properly. Travis-CI has an open issue that `apt-get update` occasionally fails, that may have occurred here. If your update doesn't pass, ping myself or one of the TEchempower team and we can restart the build if similar errors happen again

hamiltont:
@oberhamsi Once again, the travis failure may not have been your fault. [A previous failure with the postgres installation](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/52530067#L2595) seems to have completely broken dpkg, whcih means that a number of the commands failed. Specifically, this lead to the error [ERROR:root:setup.sh: line 5: rp: command not found](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/52530067#L3006)

See #1369 for the suspected error. I'll see if something can be done about that and then your travis run can be restarted.

toolset/setup/linux/languages/ringojs.sh
```diff
@@ -1,12 +1,14 @@
...
 
-RETCODE=$(fw_exists ${IROOT}/ringojs_0.10.installed)
+RETCODE=$(fw_exists ${IROOT}/ringojs_0.11.installed)
 [ ! "$RETCODE" == 0 ] || { return 0; }
 
-fw_get http://www.ringojs.org/downloads/ringojs_0.10-1_all.deb
+fw_get https://github.com/ringo/ringojs/releases/download/v0.11.0/ringojs_0.11_all.deb -O ringojs_0.11_all.deb
 sudo apt-get install -y jsvc
-sudo dpkg -i ringojs_0.10-1_all.deb
+sudo dpkg -i ringojs_0.11_all.deb
```
hamiltont: FYI, the final goal is to remove sudo from these scripts as well. It would be much better to install ringo into a prefix folder (such as `$IROOT/ringojs`) and completely avoid modifying the base OS image. If you can add that to this PR it would be a huge step in teh right direction

hamiltont:
Blocking on #1468

oberhamsi:
> completely avoid modifying the base OS image

i would see this PR as a step in that direction while also upgrading our runtime to the newest version. 

i doubt i have to time to prepare & test the $IROOT prefix solution in time for round11.

hamiltont:
> doubt i have to time to prepare & test the $IROOT prefix solution in time for round11

Coolio. Just so we're on the same page - this PR is destined for round11, whihc is at least 3 months away. The currently running round (round10) has already had multiple previews and should be released very soon, so this PR isn't targeting that. 

FWIW, the super easy way to remove sudo here is to use the `dpkg -x` option, which allows you to extract the deb into a prefix directory. So adding something like this: 

```
dpkg -x ringojs_0.11_all.deb $IROOT/ringojs
```

And then replacing references to `ringo` and `rp` with `$IROOT/ringojs/<normal install path>/ringo` is all that's necessary

oberhamsi:
ah! I did mix up round11 and round10. And thanks for the -x tip.

oberhamsi:
i'll file a new PR for the language update on top of the remove-install-scripts branch... later.

thanks @msmith-techempower, looks much nicer

PR #1355, opened 2015-02-27
Add support for PyParallel

tpn:
The ghaffarian.net site appears to only keep around the latest installers,
which causes installer.ps1 to blow up when it gets a 404 on the download
link.  I've downloaded 7u75 to a permanent (bare metal) server that I maintain
and can ensure that the download link stays live.

(I've verified this change installs the JDK successfully on an Azure Windows 2012 instance.)


tpn:
Ugh, I didn't realize github would glom the PyParallel pull request into the existing one I raised for the JDK.  I've updated the title of this PR to reflect the PyParallel intent.

LadyMozzarella:
Postponed until windows support returns.

msmith-techempower:
Okay, this has been alive long enough; when we have official Windows support in the future we will accept a PR to add this framework, but this PR is so out of date that I am forced to close it until such time.

PR #1356, opened 2015-02-28
Restart Cassandra at the beginning of benchmark run

marko-asplund:
- restart Cassandra at the beginning of benchmark run
- refactor Cassandra installation
- upgrade to Cassandra 2.0.12


toolset/setup/linux/database.sh
```diff
@@ -175,11 +184,10 @@ for i in {1..45}; do
 done
 nc -z localhost 9160
 if [ $? -eq 0 ]; then
-  cat cassandra/cleanup-keyspace.cql | apache-cassandra-$CASS_V/bin/cqlsh 127.0.0.1
+  cat cassandra/cleanup-keyspace.cql | /opt/apache-cassandra-$CASS_V/bin/cqlsh 127.0.0.1
   python cassandra/db-data-gen.py > cassandra/tfb-data.cql
-  apache-cassandra-$CASS_V/bin/cqlsh -f cassandra/create-keyspace.cql 127.0.0.1
-  apache-cassandra-$CASS_V/bin/cqlsh -f cassandra/tfb-data.cql 127.0.0.1
-  rm -rf apache-cassandra-*-bin.tar.gz cassandra
+  /opt/apache-cassandra-$CASS_V/bin/cqlsh -f cassandra/create-keyspace.cql 127.0.0.1
```
hamiltont: Given that the tar above used sudo to bring stuff into `/opt`, would the current non-sudo user have access to execute `/opt/apache-cassandra-$CASS_V/bin/cqlsh` here?

hamiltont:
Looks slick. One thing I'm not sure of - are you still storing cassandra's database inside `/ssd`? 

My understanding of how the official rounds are run is that `/ssd` is a location guaranteed to be mounted on the SSD, for the express purpose of allowing any database software to use that if they so wish.

toolset/setup/linux/database.sh
marko-asplund: yes, it should. I've tested this on the vagrant development deployment and it does work.

marko-asplund:
yes, the plan is to use /ssd for storing the data.
please see the cassandra configfile : https://github.com/marko-asplund/FrameworkBenchmarks/blob/master/config/cassandra/cassandra.yaml

hamiltont:
Great, all my concerns are addressed.

@msmith-techempower this LGTM, but I'm leaving all merges to your team while R10 is ongoing

msmith-techempower:
@hamiltont This fixes a bug, yes? If so, LGTM

marko-asplund:
@msmith-techempower Prior to this change, Cassandra was not restarted (like the other databases) before running the benchmark, so this part aims to fix a bug. This change also refactors the Cassandra deployment model to make starting and stopping Cassandra easier and more reliable. It also upgrades Cassandra from 2.0.7 to the latest stable version 2.0.12.

hamiltont:
@marko-asplund I think the second preview round is already running, so the effects of this may not show up there unless mike pulls it in, but with this merge it should be present in the final R10

marko-asplund:
@hamiltont thanks!

How many previews do you expect there to be for R10?

My big concern (it's been a long wait :-) is that since based on the preview 1 logs/data it's not possible to tell for certain what was causing servlet3-cass test implementation to fail, it's thus not certain if this PR will fix the test implementation. The benchmark seemed to start in my Vagrant based test deployment, but I'd like to be able to make 100% sure the test implementation will run during the actual benchmarking run :-)

Is there anything else I could do to help ensure test setup will be successful for servlet3-cass in R10?

Any chance of getting someone to review/merge #1357 to make diagnosing database connectivity errors easier?

hamiltont:
> How many previews do you expect there to be for R10?

I'm not sure - I suppose two but @msmith-techempower is the person thta would know

> My big concern (it's been a long wait :-) is that since based on the preview 1 logs/data it's not possible to tell for certain what was causing servlet3-cass test implementation to fail, it's thus not certain if this PR will fix the test implementation. The benchmark seemed to start in my Vagrant based test deployment, but I'd like to be able to make 100% sure the test implementation will run during the actual benchmarking run :-)

I totally understand, unfortunately I'm powerless to help you here. Best bet is to hop onto IRC and see if @msmith-techempower might be willing to run just this framework on the official hardware after he's done with the second preview

> Is there anything else I could do to help ensure test setup will be successful for servlet3-cass in R10?

Besides my suggestion above to bug mike, I think this and #1357 are all I can think of

> Any chance of getting someone to review/merge #1357 to make diagnosing database connectivity errors easier?

Sure, I'll take a look now

msmith-techempower:
I can certainly attempt a simple run on a blank VM to test this out.

marko-asplund:
@msmith-techempower that would be awesome! thanks!

msmith-techempower:
This looked good to me; LGTM.

Issue #1103, opened 2014-10-17
verify database connectivity once for all databases before running any of the tests


hamiltont:
the toolset alreday does some DB setup before each test, like restarting or something. It should probably also verify database connectivity too, using something like `nc -vz <db host> <db port>`. This would be really helpful for cassandra and mongodb, who are not ready to start handling connections just because `service <foo> restart` has returned

See #1082

marko-asplund:
@hamiltont please see my take on implementing this feature in PR #1357

hamiltont:
Fixed in #1357, thanks @marko-asplund!

PR #1357, opened 2015-02-28
verify database connectivity once for all databases before running tests

marko-asplund:
see issue #1103

### DIFF
--- a/toolset/benchmark/benchmarker.py
+++ b/toolset/benchmark/benchmarker.py
@@ -6,6 +6,7 @@ from benchmark.test_types import *
 from utils import header
 from utils import gather_tests
 from utils import gather_frameworks
+from utils import verify_database_connections
 
 import os
 import json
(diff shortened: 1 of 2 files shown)


toolset/benchmark/utils.py
```diff
@@ -138,3 +139,36 @@ def header(message, top='-', bottom='-'):
       else:
         result += "\n%s" % bottomheader
     return result + '\n'
+
+def check_services(services):
+
+  def check_service(address, port):
+    try:
+      s = socket.socket()
+      s.settimeout(5)
```
hamiltont: This probably needs to be at least 15 to accommodate databases like mongodb that are wickedly slow on the restart

toolset/benchmark/benchmarker.py
```diff
@@ -538,6 +539,14 @@ def exit_with_code(code):
           """)
```
hamiltont: Could this PR also include a restart to cassandra here? That way we could address all the database issues with one PR

toolset/benchmark/utils.py
```diff
@@ -138,3 +139,36 @@ def header(message, top='-', bottom='-'):
...
+def verify_database_connections(services):
+  allGo = True
+  for r in check_services(services):
+    if r[1]:
+      print r[0] + ": is GO!"
+    else:
+      print r[0] + ": is _NO_ GO: ERROR: " + r[2]
+      allGo = False
+
+  if allGo:
```
hamiltont: My personal preference would be to avoid printing in utility methods if possible, and return a tuple `(allGo, errMessage)` variable to benchmarker.py who can then issue a print statement

toolset/benchmark/benchmarker.py
hamiltont: Ah ignore that, I missed #1356

marko-asplund:
@hamiltont thanks, PR updated based on your comments.

hamiltont:
LGTM

hamiltont:
Travis run matches recent master run and this is a small+valuable PR, I'm comfortable to merge while R10 is ongoing.

PR #1359, opened 2015-02-28
servlet3-cass updates: use Java 8, fix "source_code" file, upgrade libs

marko-asplund:
use Oracle Java 8, delete unnecessary files from source_code file, upgrade libs

### DIFF
--- a/frameworks/Java/servlet3-cass/bash_profile.sh
+++ b/frameworks/Java/servlet3-cass/bash_profile.sh
@@ -1,2 +1,3 @@
-export JAVA_HOME=/usr/lib/jvm/java-1.7.0-openjdk-amd64
+export JAVA_HOME=/opt/java8
+export JAVA_EXE=$JAVA_HOME/bin/java
 export RESIN_HOME=${IROOT}/resin-4.0.41
--- a/frameworks/Java/servlet3-cass/install.sh
+++ b/frameworks/Java/servlet3-cass/install.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-fw_depends java resin maven
+fw_depends java8 resin maven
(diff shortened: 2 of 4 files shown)


hamiltont:
@marko-asplund great progress lately!

Note that we are removing the `bash_profile.sh` files wherever possible and just inserting their contents into the top of `setup.sh` and `install.sh` - there are various bugs associated with them. By R11 they won't exist, so if you want to go ahead and do that here it would be appreciated

marko-asplund:
@hamiltont thanks! 

I'll update the code accordingly, but before that I'd just like to check that I got this right:
* install.sh is ran to build and package the test implementation. Doesn't necessarily get invoked before each test run.
* setup file is executed before each test run to set up the test implementation. Used to start, stop the test implementation and modifying configuration.

Is this correct?

Are both Python and shell script based setup files both being supported? Will there be any changes to this in the short term?

hamiltont:
> install.sh is ran to build and package the test implementation. Doesn't necessarily get invoked before each test run.

Correct! More precisely,  `install.sh` is run when the command-line includes `--install server` and the test filters (e.g. the `--test foo foo-bar` bit) matches the framework

> setup file is executed before each test run to set up the test implementation. Used to start, stop the test implementation and modifying configuration.

Used to start, and if you require modifying configuration that can happen here as well. To stop we just kill all processes owned by the test runner user (the `--runner-user` argument)

> Are both Python and shell script based setup files both being supported? Will there be any changes to this in the short term?

Only shell-based is supported at the moment. There are still some outdated readme's that mention `setup.py` - these all need to be changed to reference `setup.sh`.

marko-asplund:
@hamiltont thanks! I'll eliminate bash_profile.sh from servlet3-cass after R10.
So, this PR is ready for review.

hamiltont:
LGTM, and as it's a "fix" I'm happy to merge while R10 is ongoing. 

@marko-asplund the second preview round has already begun, so this won't be represented there, but it should be represented in the final R10

PR #1364, opened 2015-03-04
Change Ur/Web default display_name

achlipala:
...to get more uniform naming in results visualizations.

### DIFF
--- a/frameworks/Ur/urweb/benchmark_config
+++ b/frameworks/Ur/urweb/benchmark_config
@@ -2,6 +2,7 @@
   "framework": "urweb",
   "tests": [{
     "default": {
+      "display_name": "urweb",
       "setup_file": "setup",
       "json_url": "/json",
       "plaintext_url": "/plaintext",


frameworks/Ur/urweb/benchmark_config
```diff
@@ -2,6 +2,7 @@
   "framework": "urweb",
   "tests": [{
     "default": {
+      "display_name": "urweb",
```
msmith-techempower: This should be included in every test, not just `default`.

achlipala:
Hm... but the other tests already get the names that I want, in the visualized results on the web.  Is it really still necessary to change them, too?

msmith-techempower:
Perhaps not; I guess it might use `display_name` if available, otherwise it tries `framework`-`testname`, and if it it `default`, it uses... Platform?

Not sure.

achlipala:
I'm happy to take my chances on this cosmetic issue for the next run of results. :)  My theory has been that this PR does accomplish my personal goal for consistent naming.

msmith-techempower:
Well, I've already kicked off the 5th preview and we are still aiming for the final run to start on Friday, so if I merge this in, it will likely be used on the final run and nothing else (just because of the timing). Either way, we will try and make sure that all of the displayed values are `urweb`.

achlipala:
OK.  I'm happy with any result that leads to consistent naming.  Thanks.

PR #1380, opened 2015-03-06
Remove Running PID if Present for All Java Play2 Variations, Resolve #1377

### LINKED ISSUE #1377: Java Play2 - Verification Failed
After running Round 10, Preview 2 this test failed and appears to be failing locally:
* play2-java (json serialization, plaintext)

LadyMozzarella:
When preview 2 of round 10 ran these tests didn't run. It appears to be because there were issues with a running pid already existing. So, we need to remove the running pid if it exists for each test. 

Here is an example of the error from the preview: 
```bash
Play server process ID is 19810
This application is already running (Or delete /home/techempower/FrameworkBenchmarks/frameworks/Java/play2-java/play2-java/target/universal/stage/RUNNING_PID file).
```

The same thing is set up for the Scala Play2 frameworks and they didn't have issues during the preview round.

### DIFF
--- a/frameworks/Java/play2-java/setup_java.sh
+++ b/frameworks/Java/play2-java/setup_java.sh
@@ -1,5 +1,12 @@
 #!/bin/bash
 
 cd play2-java
+
+# If application is running, clear old running app.
+if [ -f ${TROOT}/play2-java/target/universal/stage/RUNNING_PID ]
+then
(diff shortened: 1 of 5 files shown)


msmith-techempower:
I didn't test them all, but I tested a few and they looked good - LGTM!

PR #1386, opened 2015-03-08
Upgrading Curacao benchmark libraries to Jetty 9.2.9.v20150224 and Curacao 2.9-M1

markkolich:
Upgrading the versions of 2 libraries used in the Curacao benchmark:
* Curacao 2.9-M1
* Jetty 9.2.9.v20150224


hamiltont:
LGTM. @markkolich this may not be included in the next preview (I'm unsure when the run date is), but it will be included into R10

markkolich:
@hamiltont no problem, thanks!

Issue #1394, opened 2015-03-09
Python Api-hour - Verification Failed


LadyMozzarella:
After running Round 10, Preview ~~2~~ 3 this test failed and appears to be passing locally:
* [aiohttp.web](https://github.com/TechEmpower/TFB-Round-10/tree/master/peak/linux/results-2015-03-06-peak-preview3/latest/logs/aiohttp.web) (multiple queries, fortunes, and data updates): 
  ```bash
gunicorn.error:ERROR 2015-02-26 14:17:59,868 glogging.py:224 => Error handling request
Traceback (most recent call last):
  File "/home/techempower/FrameworkBenchmarks/installs/py3/lib/python3.4/site-packages/aiohttp/server.py", line 240, in start
    yield from handler
  File "/home/techempower/FrameworkBenchmarks/installs/py3/lib/python3.4/site-packages/aiohttp/web.py", line 1209, in handle_request
    resp = yield from handler(request)
  File "/home/techempower/FrameworkBenchmarks/frameworks/Python/API-Hour/hello/hello/endpoints/world.py", line 23, in db
    return JSON((yield from get_random_record(container)))
  File "/home/techempower/FrameworkBenchmarks/frameworks/Python/API-Hour/hello/hello/services/world.py", line 7, in get_random_record
    pg = yield from container.engines['pg']
  File "/home/techempower/FrameworkBenchmarks/installs/py3/lib/python3.4/asyncio/futures.py", line 390, in __iter__
    return self.result()  # May raise too.
  File "/home/techempower/FrameworkBenchmarks/installs/py3/lib/python3.4/asyncio/futures.py", line 277, in result
    raise self._exception
  File "/home/techempower/FrameworkBenchmarks/installs/py3/lib/python3.4/asyncio/tasks.py", line 236, in _step
    result = next(coro)
  File "/home/techempower/FrameworkBenchmarks/installs/py3/lib/python3.4/site-packages/aiopg/pool.py", line 26, in create_pool
    yield from pool._fill_free_pool(False)
  File "/home/techempower/FrameworkBenchmarks/installs/py3/lib/python3.4/site-packages/aiopg/pool.py", line 159, in _fill_free_pool
    **self._conn_kwargs)
  File "/home/techempower/FrameworkBenchmarks/installs/py3/lib/python3.4/site-packages/aiopg/connection.py", line 51, in connect
    yield from conn._poll(waiter, timeout)
[...]

PR #1407, opened 2015-03-10
Increase pgsql connections

GMLudo:
As promised in #1394, I increase PostgreSQL max_connections.
Maybe it will be necessary to change maximum shared memory on server.

### DIFF
--- a/config/postgresql.conf
+++ b/config/postgresql.conf
@@ -61,7 +61,7 @@ listen_addresses = '*'		# what IP address(es) to listen on;
 					# defaults to 'localhost', '*' = all
 					# (change requires restart)
 port = 5432				# (change requires restart)
-max_connections = 3000			# (change requires restart)
+max_connections = 3500			# (change requires restart)
 # Note:  Increasing max_connections costs ~400 bytes of shared memory per
 # connection slot, plus lock space (see max_locks_per_transaction).
 #superuser_reserved_connections = 3	# (change requires restart)
--- a/config/postgresql.conf
+++ b/config/postgresql.conf
@@ -61,7 +61,7 @@ listen_addresses = '*'		# what IP address(es) to listen on;
 					# defaults to 'localhost', '*' = all
 					# (change requires restart)
 port = 5432				# (change requires restart)
-max_connections = 3500			# (change requires restart)
+max_connections = 5000			# (change requires restart)
 # Note:  Increasing max_connections costs ~400 bytes of shared memory per
 # connection slot, plus lock space (see max_locks_per_transaction).
 #superuser_reserved_connections = 3	# (change requires restart)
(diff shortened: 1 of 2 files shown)


GMLudo:
I've tested your configuration file with my PostgreSQL.
PostgreSQL doesn't start until I change kernel.sem: `sudo sysctl -w kernel.sem="250 32000 256 256"`

Where is your `/etc/sysctl.d/99-boost.conf` file in your repository to change kernel settings ?

Moreover, I've compared your `postgresql.conf` config file with mine, I've three extra parameters:
> shared_buffers = 128MB
> work_mem = 32MB
> maintenance_work_mem = 64MB

msmith-techempower:
I don't know what any of this means >_<

Additionally, we want to touch the system as little as possible with the suite, so I'm unnerved by allowed changes to `/etc/sysctl.d/99-boost.conf`.

GMLudo:
> I don't know what any of this means >_<

If you already have a PostgreSQL with 2000 max_connections, we must already change kernel settings somewhere ;-)

> Additionally, we want to touch the system as little as possible with the suite, so I'm unnerved by allowed changes to /etc/sysctl.d/99-boost.conf.

First, be cool, it's only a benchmark, no human lifes are in the game ;-)
Second, in fact, you already change kernel settings, not with a config file but with a command line, but it's the same effect at the end. It means, you're already agree to do that ;-)
See: https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/benchmark/benchmarker.py#L377
I've added in my PR the changes for that.
I can't test myself, do you plan to relaunch benchmarks before the D-day this Friday  ?

hamiltont:
> Second, in fact, you already change kernel settings, not with a config file but with a command line, but it's the same effect at the end. 

Agreed - this is already happening. 

@GMLudo - IMO two excellent additions to the database sysctl settings would be 1) a link to some documentation for each parameter (or just a human-readable explanation?) and 2) adding something like `echo Printing kernel configuration && sudo sysctl -a` at the end. If you have a moment to add either of those it would help out

GMLudo:
@hamiltont ok, I'll add that tomorrow, I need to sleep right now.
BTW, is it possible to launch something to test this script easily without launch all benchmark suite ?

hamiltont:
Nothing easy. Your best bet is to test it using our [vagrant development setup](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/deployment/vagrant-development) which will launch a single virtual machine and configure it to be the application server, the database server, and the client load generation server.

GMLudo:
Hi,

I've added explanations + add some parameters that seems to improve a little bit performance, at least on my postgresql setup.
I've issues to test that with vagrant, I've tested only on my setup.

Regards.

hamiltont:
This LGTM

GMLudo:
Thanks for the merge.

PR #1411, opened 2015-03-12
Port Mojolicious to Postgres

jberger:
Mojo's old MongoDB driver has been abandoned because they kept changing their wire protocol out from under us. It seems that the mongodb version that broke caught up to this app. I have thus done a quick port to Postgres. The structure of the app is all completely the same.

### DIFF
--- a/frameworks/Perl/mojolicious/app.pl
+++ b/frameworks/Perl/mojolicious/app.pl
@@ -1,5 +1,5 @@
 use Mojolicious::Lite;
-use Mango;
+use Mojo::Pg;
 
 use JSON::XS 'encode_json';
 use Scalar::Util 'looks_like_number';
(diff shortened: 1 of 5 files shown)


msmith-techempower:
Subtle update from mojo 5.8 to 6.0 as well in this PR.

Question: does Mojo's db tests fail in the current preview rounds because of this MongoDB stuff?

jberger:
I can back down to 5.8 if that is a concern. Remember that CPAN normally gives the most recent version of a module anyway, I only pin the versions because I don't want the demo breaking if a new cpan module presents a breaking change. When I have made edits previously I usually release the pin and check if the newest versions work. 

As to the question, yes, MongoDB appears to be completely hosed. The only database test that "works" is updates, and that is only because none of the returned data actually comes from the database. The issue is the wire protocol, so the I'm guessing that the request cycle to the db is working but the data is corrupted somehow. This change in wire protocol was the last straw to make mojo move away from MongoDB ([blog post](http://tempi.re/mojo-news-mango), [official announcement](https://groups.google.com/forum/#!topic/mojolicious/GFsXFQF3t-k)).

msmith-techempower:
This is fine, I'll try and test this out and get it merged today.

jberger:
Thanks!

jberger:
BTW, you don't need that travis run, I have this already: https://travis-ci.org/jberger/FrameworkBenchmarks/jobs/54054696

Issue #250, opened 2013-05-04
Go: Add Go with mgo MongoDB driver


bhauer:
This is just a marker that I've seen a request for a Go implementation that connects to MongoDB using a driver named "mgo".

ebramanti-techempower:
gonna take a crack at this

hamiltont:
@ebramanti-techempower feel free to mark yourself as the Assignee

LadyMozzarella:
Closing. #1418 merged.

PR #1367, opened 2015-03-05
Add .json Extension to benchmark_config Files, Resolve #1095

LadyMozzarella:
Rename benchmark_config files to benchmark_config.json to resolve #1095. 

Everything appears to be running fine in my separate Travis, except for Perl (see issue #1366).

Issue #1259, opened 2014-12-17
Remove bash_profile.sh


hamiltont:
@msmith-techempower and I talked that bash_profile.sh seems to now be an outdated feature now that meganuke is here, as it cannot set some of the variables (like PATH, HOME, etc) because they will be overwritten when we switch into the testrunner user. We could hack around this but there's little point, we can just duplicate the 2-3 LOC from bash_profile.sh into install.sh and setup.sh

hamiltont:
Fixed in master

PR #1418, opened 2015-03-13
New Framework: go-mongodb

ebramanti-techempower:
As requested in #250, I've implemented a Go server for the benchmarks with mgo, a popular Go MongoDB driver. I've implemented it as a separate test due to some different logic for the handlers and structs.

In the future, I might port some of my fixes from the original Go test so that there are no more warns showing on the tests.


toolset/setup/linux/prerequisites.sh
```diff
@@ -42,7 +42,8 @@ sudo apt-get -qqy install -o Dpkg::Options::="--force-confdef" -o Dpkg::Options:
   libpq-dev mlton \
   libjemalloc-dev libluajit-5.1-dev `# Needed by lwan at least` \
   libhiredis-dev                    `# Redis client - Needed by ngx_mruby at least` \
-  cloc dstat                        `# Collect resource usage statistics`
+  cloc dstat                        `# Collect resource usage statistics` \
+  libsasl2-dev                      `# Needed by mgo for go-mongodb test`
```
hamiltont: Everything looks good but this - is there any way to avoid putting yet another test-specific dependency into `prerequisites.sh` - especially one that requires `sudo`?

ebramanti-techempower: @msmith-techempower recommended that I put it here, I'm open to suggestions though!

hamiltont: Thought about it a bit and I guess there is no better way to do this at the moment - and you're right that it's important enough it may be a common prereq in the future. Mind changed - PR LGTM

frameworks/Go/go-mongodb/bash_profile.sh
```diff
@@ -0,0 +1,4 @@
+# Set the root of our go installation
+export GOROOT=${IROOT}/go
```
hamiltont: Let's move these lines into the top of the `install.sh` file - see #1259 for the reason

hamiltont:
I read this a bit too quickly, apologies, but I've gotta pull back my LGTM because we missed something critical - there's no line in `.travis.yml` for this new framework, so it wasn't actually verified :-D

Can you please update this to add that line and address the bash_profile change requested above, and we'll see what travis says.

frameworks/Go/go-mongodb/bash_profile.sh
LadyMozzarella: On that note, we just merged the change to get rid of the need for this file entirely (issue #1478). After moving the lines, please remove bash_profile.sh entirely. :)

LadyMozzarella:
Also, if you could add the .json extensions to your benchmark_config that would be super awesome because we just merged #1367, which does that for all of the other frameworks :)

ebramanti-techempower:
@hamiltont On it,  @LadyMozzarella explained to me how to do this last time I was at work.

@LadyMozzarella will do!

LadyMozzarella:
Looks good now :) (All green except that json header check that was my bad). Merging!

PR #1421, opened 2015-03-15
Remove syslog logging

GMLudo:
On the server where you launch API-Hour, you certainly have a lot of logs in /var/log/*.
This PR disable syslog logging from API-Hour.


hamiltont:
Good catch @GMLudo. [Travis](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/54448263) results check out, this LGTM

PR #1361, opened 2015-03-03
Assign Default Tests

LadyMozzarella:
Cleaning up Frameworks without default tests takes care of issue [#1339](https://github.com/TechEmpower/FrameworkBenchmarks/issues/1339). I made my best guess on some of these as to what the default test should be. Please let me know if any of them don't exactly sound correct and what should be the default instead. Here are the frameworks now with default tests:
* Csharp: 
  * servicestack – 'servicestack-iis-default' test is now default
* Dart: For all of the dart tests I chose the most normal postgres option because there's a [PostgreSQL database driver for Dart](https://pub.dartlang.org/packages/postgresql)
  * redstone – 'redstone-postgresql' test is now default
  * start – 'start-postgresql-raw' test is now default
  * stream – 'stream-postgresql-raw' test is now default
* Java: 
  * play2* – 'play2-java' test is now default
  * wildfly-ee7 – 'wildfly-ee7-mysql' test is now default
* PHP:
  * yaf – 'yaf-raw' test is now default
* Python:
  * pyramid – 'pyramid-py3' test is now default
* Ruby: For all of the ruby tests names I refactored them a bit to not include 'ruby' if it uses straight Ruby, but I left (or added) the specification if it uses Jruby or Rubinius (rbx). Basically, I just tried to make them match all of the other ruby test names in other framework benchmark_config files. 
  * grape – 'grape-puma-mri' test is now default
  * padrino – 'padrino-puma-mri' test is now default
  * rack – 'rack-puma-mri' test is now default
  * rails-stripped – 'rails-stripped-mri' test is now default (used to be 'rails-stripped-ruby')
  * rails – 'rails-puma-mri' test is now default
  * sinatra – 'sinatra-puma-mri' test is now default
* Scala: 
  * play2-scala* – 'play2-scala' test is now default

*Because play2 is in Java and Scala, I went with Java as the default language for play and opted to have Scala be secondary. So, the default tests are called by 'play2' and 'play2-scala', respectively.

Issue #1339, opened 2015-02-23
Change benchmark_configs to include default


hamiltont:
I think it makes sense to ensure every benchmark_config has a `default` entry, so that you can always reliably test a framework using `--test <framework_name>`. Here's a list of offenders - we just need to go thru their benchmark_configs and determine which of the listed tests is the best candidate for being the default test for that framework

```
WARNING:root:Framework play2 does not define a default test in benchmark_config
WARNING:root:Framework sinatra does not define a default test in benchmark_config
WARNING:root:Framework rails does not define a default test in benchmark_config
WARNING:root:Framework rails-stripped does not define a default test in benchmark_config
WARNING:root:Framework rack does not define a default test in benchmark_config
WARNING:root:Framework padrino does not define a default test in benchmark_config
WARNING:root:Framework grape does not define a default test in benchmark_config
WARNING:root:Framework pyramid does not define a default test in benchmark_config
WARNING:root:Framework yaf does not define a default test in benchmark_config
WARNING:root:Framework wildfly-ee7 does not define a default test in benchmark_config
WARNING:root:Framework play2 does not define a default test in benchmark_config
WARNING:root:Framework stream does not define a default test in benchmark_config
WARNING:root:Framework start does not define a default test in benchmark_config
WARNING:root:Framework redstone does not define a default test in benchmark_config
WARNING:root:Framework servicestack does not define a default test in benchmark_config
WARNING:root:Framework play2 does not define a default test in benchmark_config
WARNING:root:Framework sinatra does not define a default test in benchmark_config
WARNING:root:Framework rails does not define a default test in benchmark_config
WARNING:root:Framework rails-stripped does not define a default test in benchmark_config
WARNING:root:Framework rack does not define a default test in benchmark_config
WARNING:root:Framework padrino does not define a default test in benchmark_config
WARNING:root:Framework grape does not define a default test in benchmark_config
WARNING:root:Framework pyramid does not define a default test in benchmark_config
WARNING:root:Framework yaf does not define a default test in benchmark_config
WARNING:root:Framework wildfly-ee7 does not define a default test in benchmark_config
WARNING:root:Framework play2 does not define a default test in benchmark_config
WARNING:root:Framework stream does not define a default test in benchmark_config
WARNING:root:Framework start does not define a default test in benchmark_config
WARNING:root:Framework redstone does not define a default test in benchmark_config
WARNING:root:Framework servicestack does not define a default test in benchmark_config
```

[...]

PR #1422, opened 2015-03-15
API-Hour fine tuning

GMLudo:
Hi,

I've finally found a solution to have a more similar setup as you, this is some performance bug fixes for API-Hour.
Moreover, I've added Redis as a separate DB test.

Regards.


methane:
1. While this is religious problem, I prefer newline at end of text files.

2. Why do you use three ports?
Most frameworks use 8080. Follow this convention make it easier to setup test environment.
Some people may manually open ports used in security group on EC2.

3. Why you use (number of processors * 3) processes?
In classical frameworks which uses blocking network I/O, more processes are required to use full CPU.
Does your benchmark uses any blocking I/O?

GMLudo:
Hi @methane,

1. No problems, I've no religion about that ;-) Your pull request is welcome for this point, or it's mandatory to be accepted in FrameworkBenchmarks ?

2. To be honest with you, for 8008, I use that to distinguish my tests and others tests in my wrk console. I've no problems to switch to 8080 if you prefer. About 8009 and 8011, it's an experimentation to see the framework overhead compare to a pure AsyncIO socket. Because of this, I've flagged both tests as `Stripped` instead of `Realistic`. I can merge 8009 and 8011 if I add a small endpoints router, but I can't merge with 8008, because the socket is handled by aiohttp.web.

3. Good point: for now it's empiric value from my part. When I've implemented the benchmark, the best values with API-Hour was (number of processors * 2) processes. Now I've a test environment more closer with TechEmpower, *3 seems to be better. If you have another experience, I'm interested in.

BTW, @methane, thank you a lot for your constructive and pertinent remarks since several weeks, it's interesting.

Regards.

frameworks/Python/API-Hour/benchmark_config
```diff
@@ -20,8 +20,62 @@
...
+      "language": "Python",
+      "orm": "Raw",
+      "platform": "API-Hour",
+      "webserver": "None",
+      "os": "Linux",
+      "database_os": "Linux",
+      "display_name": "API-Hour+aiohttp.web+redis",
+      "notes": "Python 3 + AsyncIO + aiohttp.web + Redis"
+    }
+    "json": {
```
msmith-techempower: How does this differ from the `default`'s `json` test?

frameworks/Python/API-Hour/benchmark_config
```diff
@@ -20,8 +20,62 @@
...
+      "language": "Python",
+      "orm": "Raw",
+      "platform": "API-Hour",
+      "webserver": "None",
+      "os": "Linux",
+      "database_os": "None",
+      "display_name": "API-Hour+AsyncIO",
+      "notes": "Python 3 + AsyncIO"
+    }
+    "plaintext": {
```
msmith-techempower: And this?

frameworks/Python/API-Hour/benchmark_config
```diff
@@ -20,8 +20,62 @@
       "webserver": "None",
       "os": "Linux",
       "database_os": "Linux",
-      "display_name": "API-Hour",
-      "notes": "Python 3.4 + AsyncIO"
+      "display_name": "API-Hour+aiohttp.web+pgsql",
+      "notes": "Python 3 + AsyncIO + aiohttp.web + PostgreSQL"
+    }
```
LadyMozzarella: Needs a `,`. I just checked out your branch and there are errors because of this when attempting to run these tests.

frameworks/Python/API-Hour/benchmark_config
```diff
@@ -20,8 +20,62 @@
...
+      "framework": "aiohttp.web",
+      "language": "Python",
+      "orm": "Raw",
+      "platform": "API-Hour",
+      "webserver": "None",
+      "os": "Linux",
+      "database_os": "Linux",
+      "display_name": "API-Hour+aiohttp.web+redis",
+      "notes": "Python 3 + AsyncIO + aiohttp.web + Redis"
+    }
```
LadyMozzarella: Also needs a `,`.

frameworks/Python/API-Hour/benchmark_config
```diff
@@ -20,8 +20,62 @@
...
+      "framework": "AsyncIO",
+      "language": "Python",
+      "orm": "Raw",
+      "platform": "API-Hour",
+      "webserver": "None",
+      "os": "Linux",
+      "database_os": "None",
+      "display_name": "API-Hour+AsyncIO",
+      "notes": "Python 3 + AsyncIO"
+    }
```
LadyMozzarella: Also needs a `,`.

LadyMozzarella:
After adding the commas, some of your tests launched for me locally. However, many of them didn't pass. I'm seeing most of the errors coming from the AsyncIO test (I believe with gunicorn):
```bash
gunicorn.error:ERROR 2015-03-24 16:19:44,467 glogging.py:224 => Error handling request
  File "/home/vagrant/FrameworkBenchmarks/installs/py3/lib/python3.4/site-packages/aiohttp/server.py", line 240, in start
    yield from handler
...
psycopg2.OperationalError: could not connect to server: Connection refused
	Is the server running on host "127.0.0.1" and accepting
	TCP/IP connections on port 5432?
```
api-hourverification

frameworks/Python/API-Hour/benchmark_config
GMLudo: As you can see, Redis test doesn't have `json_url`.

frameworks/Python/API-Hour/benchmark_config
GMLudo: For this, I use `AsyncIO` directly, without use `aiohttp.web` framework.

frameworks/Python/API-Hour/benchmark_config
GMLudo: Thank you, I've fixed that in the pull request.
I've checked the syntax with jsonlint.
If for round 11, the file extension should be .json, it will be better to detect issues in IDE directly.

LadyMozzarella: I understand @GMLudo - I have a PR #1367 open for this. :)

GMLudo:
Thank you @LadyMozzarella to launch the test.
For AsyncIO errors, based on your error message, your postgresql server refuse the connections from API-Hour. Could you check postgresql logs ? Maybe you accept ssl connections only, but I disable that explicitly in benchmark tests to improve performances.
For AsyncIO-json and plaintext, I've no ideas about this, maybe you see another error when you'll fix postgresql connections.

BTW, to facilitate your merge, I've just rebased my branch with your master, to resolve conflicts.

LadyMozzarella:
Whoops. Something is definitely wrong with postgres over here - still looking into that. Thanks @GMLudo. I'll work on getting that fixed up and try running it again (unless someone else gets to testing it out or merging first).

frameworks/Python/API-Hour/benchmark_config
```diff
@@ -17,11 +17,65 @@
...
+      "platform": "API-Hour",
+      "webserver": "Gunicorn",
+      "os": "Linux",
+      "database_os": "None",
+      "display_name": "API-Hour+AsyncIO",
+      "notes": "Python 3 + API-Hour + AsyncIO"
+    },
+    "plaintext": {
+      "setup_file": "setup",
+      "json_url": "/plaintext",
```
methane: plaintext_url?

frameworks/Python/API-Hour/hello/hello/servers/yocto_http.py
```diff
@@ -0,0 +1,23 @@
...
+        self.transport.write(generate_http_response(payload))
+
+class YoctoHttpText(asyncio.Protocol):
+    def connection_made(self, transport):
+        self.transport = transport
+
+    def data_received(self, data):
+        # self.transport.write(data)
+        payload = 'Hello, World!'
+        self.transport.write(generate_http_response(payload, 'text/plain; charset=UTF-8'))
```
methane: plaintext test uses pipelined request.
Skipping HTTP may not work.

I feel skipping HTTP is less than "stripped". "stripped" means skip some default middleware in framework.

frameworks/Python/API-Hour/benchmark_config
GMLudo: Nice catch, thank you, I've fixed that.

frameworks/Python/API-Hour/hello/hello/servers/yocto_http.py
GMLudo: > plaintext test uses pipelined request.
> Skipping HTTP may not work.

You're certainly right, I'll dig that. I need to find the `wrk` command line used for this test.

> I feel skipping HTTP is less than "stripped". "stripped" means skip some default middleware in framework.

I'm agree with you if I've categorized this test as micro or full stack framework. But for theses tests, I've flagged that as `Platform`.
At least for me, it's the stripped version for a plateform.
The idea behind theses tests is to see the maximum it should be possible to have with AsyncIO in pure Python, like you have with wsgi, not to be used on production. "Platform" and "Stripped" are here as warnings.

frameworks/Python/API-Hour/benchmark_config
```diff
@@ -17,11 +17,65 @@
...
+      "framework": "AsyncIO",
+      "language": "Python",
+      "orm": "Raw",
+      "platform": "API-Hour",
+      "webserver": "Gunicorn",
+      "os": "Linux",
+      "database_os": "None",
+      "display_name": "API-Hour+AsyncIO",
+      "notes": "Python 3 + API-Hour + AsyncIO"
     }
```
LadyMozzarella: 1. I got plaintext and json running by setting database_os to "Linux" (the error I was getting was `OS or Database OS specified in benchmark_config does not match the current environment. Skipping.`). Maybe we should consider allowing the option "None" to pass through for this sort of purpose, but for now I suggest setting database_os to "Linux."
2. Is there are reason that plaintext and json are completely different tests? I don't really understand why they can't just be combined into one called "stripped," like:

```json
    "stripped": {
      "setup_file": "setup",
      "plaintext_url": "/plaintext",
      "json_url": "/json",
      "port": 8009,
      "approach": "Stripped",
      "classification": "Platform",
      "database": "None",
      "framework": "AsyncIO",
      "language": "Python",
      "orm": "Raw",
      "platform": "API-Hour",
      "webserver": "Gunicorn",
      "os": "Linux",
      "database_os": "Linux",
      "display_name": "API-Hour+AsyncIO",
      "notes": "Python 3 + API-Hour + AsyncIO"
    }
```

GMLudo: 1. I've changed that in the pull request.
2. Because AsyncIO doesn't have routing url feature, I use different port to distinguish the tests.

LadyMozzarella: Ahh okay. Thanks. :+1:

LadyMozzarella:
Okay. I am seeing all green over here now. :+1: 

asynciotests

The only thing that would be super awesome is if you could assign one of the tests as the "default" test again. We're trying to clean up all of the frameworks with missing default tests. (issue #1339, pr #1361)

GMLudo:
@LadyMozzarella Thank you again for your time, I've defined a default test.
Don't hesitate if you want some other modifications or Python support.

GMLudo:
@LadyMozzarella Thank you.

PR #1423, opened 2015-03-16
Dart tweaks

kevmoo:
Use best-practice JSON encoding model

some formatting tweaks

### DIFF
--- a/frameworks/Dart/dart/server.dart
+++ b/frameworks/Dart/dart/server.dart
@@ -16,11 +16,11 @@ final _encoder = new JsonUtf8Encoder();
 /// arguments, as is the number of database connections to be maintained in the
 /// connection pool.
 void main(List<String> args) {
-  var parser = new ArgParser();
-  parser.addOption('address', abbr: 'a', defaultsTo: '0.0.0.0');
-  parser.addOption('port', abbr: 'p', defaultsTo: '8080');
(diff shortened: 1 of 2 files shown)


hamiltont:
Any idea why this would cause the update test to start failing? See [here](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/54906142#L4355). Especially odd is that the other database tests still work ok

kevmoo:
@hamiltont Could it be a fluke? Should I rebase and try again?

hamiltont:
It definitely could just an "oops", although normally when that happens all the tests fail. Rebasing can't hurt - why don't you do that and if it fails again we can take a most serious look

kevmoo:
@hamiltont I installed vagrant locally. Turns out I break the DB serialization by making things final. I think this fixes things.

hamiltont:
wow, how unexpected! Cool that travis caught this, let's see if it passes now

LadyMozzarella:
[Travis build](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/56827422) for reference (Not started yet as of this comment)

LadyMozzarella:
Travis-CI looks happy to me, merging. Tests are 'warn' because the check we're doing for headers is case sensitive, but fixed in a merge after these commits were made. Looks like all of the responses for update actually came back fine.

PR #1424, opened 2015-03-17
Python cherrypy

knewmanTE:
Adds basic plaintext and JSON tests for Python CherryPy framework.

I also have DB, Queries, and Updates tests mostly implemented, but have excluded them from this merge request because CherryPy relies on a database interface or ORM like SQLAlchemy, which I see is already implemented in other tests.

edit: DB, Queries, and Updates added.


knewmanTE:
It looks like CherryPy is passing fine. The failed Travis check seems to be from PHP/hhvm.

hamiltont:
Agreed, both times travis shows cherrypy is fine. This LGTM

PR #1429, opened 2015-03-19
Wheezyweb

knewmanTE:
Adds python wheezy.web framework tests.


methane:
Could you use 4 space for indent? It's de-facto standard of Python community.

Additionally, can you use gunicorn+meinheld or nginx+uWSGI like other Python frameworks?
wsgiref's server is only for development. It's slow and unstable on high load.

knewmanTE:
Will do.

frameworks/Python/wheezyweb/bash_profile.sh
```diff
@@ -0,0 +1,9 @@
+export PY2_ROOT=$IROOT/py2
+export PY2=$PY2_ROOT/bin/python
+export PY2_PIP=$PY2_ROOT/bin/pip
+export PY2_GUNICORN=$PY2_ROOT/bin/gunicorn
+
+export PY3_ROOT=$IROOT/py3
+export PY3=$PY3_ROOT/bin/python3
+export PY3_PIP=$PY3_ROOT/bin/pip3
+export PY3_GUNICORN=$PY3_ROOT/bin/gunicorn
```
LadyMozzarella: Can you move these variables to your setup.sh and/or install.sh and then remove the bash_profile.sh file entirely? We just merged a change to get rid of the need for this file entirely. (issue #1478)

knewmanTE: Sure thing. Just pushed the fix.

PR #1431, opened 2015-03-21
Modify display names for Python tests

methane:
Some Python tests have same display name.
Make them more distinguishable.

Additionally, skip duplicated tests. (`/json` and `/plaintext` in `Tornado-MongoDB` and `Tornado-Postgres`).


frameworks/Python/falcon/benchmark_config
```diff
@@ -9,14 +9,14 @@
       "approach": "Realistic",
       "classification": "Micro",
       "database": "None",
-      "framework": "falcon",
+      "framework": "Flacon",
```
GMLudo: You have a typo in `Falcon` name here and below.

methane: LoL, I'v fixed it.

PR #1436, opened 2015-03-22
Updated Scruffy tests in response to round 10 preview

sksamuel:
I have updated the version of scruffy used in the scruffy tests, plus a couple of other tweaks that improve performance and should stop the errors that occurred in round 10 preview 3.


sksamuel:
Failure just looks like Travis not finding a scala dependency that's on maven central. I get this from time to time myself.

LadyMozzarella:
:+1: . I checked this out and all the tests are passing using vagrant.

sksamuel:
Thanks

Issue #1406, opened 2015-03-10
PHP's Cakephp - Validation Failed 


LadyMozzarella:
Test 'cakephp' failing in preview 3 of round 10 with [json serialization, single query, and multiple queries](https://github.com/TechEmpower/TFB-Round-10/blob/master/peak/linux/results-2015-03-06-peak-preview3/latest/logs/cakephp/out.txt). All pass/warn locally through vagrant.

```bash
--------------------------------------------------------------------------------
  VERIFYING DB
--------------------------------------------------------------------------------
Accessing URL http://10.0.3.2:8080/world/index.json 
Response: 
"HTTP/1.1 500 Internal Server Error
Server: nginx/1.4.1
Date: Sat, 07 Mar 2015 01:41:49 GMT
Content-Type: text/html
Transfer-Encoding: chunked
Connection: keep-alive
X-Powered-By: PHP/5.5.17

"
   FAIL for http://10.0.3.2:8080/world/index.json
     Empty Response
--------------------------------------------------------------------------------
  VERIFYING JSON
--------------------------------------------------------------------------------
Accessing URL http://10.0.3.2:8080/json.json 
Response: 
"HTTP/1.1 500 Internal Server Error
Server: nginx/1.4.1
Date: Sat, 07 Mar 2015 01:41:49 GMT
Content-Type: text/html
Transfer-Encoding: chunked
Connection: keep-alive
X-Powered-By: PHP/5.5.17

"
   FAIL for http://10.0.3.2:8080/json.json
     Empty Response
```

hamiltont:
Am I correct in seeing a pattern that tons of things are running inside vagrant but not running on the official test hardware? If so, what's the root cause?

msmith-techempower:
If I had to guess, I'd say global system installs being broken in fantastic ways (example: Lapis from a few days ago)

hamiltont:
Ugh, gross. So we prioritize #1387 for R11

msmith-techempower:
Context: Lapis is a framework installed via an app called `luarocks` which is installed into the `IROOT`, but runs in tandem with Lua, which is installed via `apt-get`. To the best of my understanding, `luarocks` is a package/lib manager that is configured (hopefully, by default, but maybe overall) to install packages (like `lapis`) into `/usr/local/share/lua`.

[...]

PR #1451, opened 2015-03-27
Fixed validation warnings for CakePHP. #1406

ssmith-techempower:
Fixed validation warnings for CakePHP for single and multi-queries. Part of #1406


hamiltont:
@ssmith-techempower FYI if you use "Fixes #1406" instead of "Part of", then when this PR is merged that issue will automatically be closed

PR #1457, opened 2015-03-31
Warn for test if value of randomNumber is greater than 10,000, Resolve #1142

### LINKED ISSUE #1142: Add a verify warning if randomNumber is not within range
A few tests are accidentally returning randomNumbers that are to big, e.g. 323234. We only require numbers between 1 and 10000. There's no reason to disqualify tests for this, but it would be nice to automatically point out to them that these few extra bytes per request could harm their performance

PS - this check should go at the bottom of `db_type.py`

LadyMozzarella:
Warn for single database query, multiple database queries, and update when a value of randomNumber from the returned json response is greater than 10,000.

### DIFF
--- a/toolset/benchmark/test_types/db_type.py
+++ b/toolset/benchmark/test_types/db_type.py
@@ -99,7 +99,11 @@ class DBTestType(FrameworkTestType):
       response_rn = float(db_object["randomnumber"])
 
       if response_id > 10000 or response_id < 1:
-        problems.append( ('warn', "Response key 'id' should be between 1 and 10,000", url)) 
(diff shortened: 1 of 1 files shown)


hamiltont:
LGTM. Are we good to merge into master now?

msmith-techempower:
Please hold off for a bit.

msmith-techempower:
Actually, go ahead - I just confirmed that `round-10` branch is correct hash.

hamiltont:
Looks good on [travis](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/56590553)

PR #1460, opened 2015-03-31
Use Wildfly 9.0.0.Beta2

stuartwdouglas:


LadyMozzarella:
Wildfly is failing for me with this PR. Looks like the [version directory in the setup.sh script](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/wildfly-ee7/setup.sh#L6) needs to be updated to 9.0.0.Beta2 as well.

stuartwdouglas:
Should be fixed now

LadyMozzarella:
Travis-CI says Wildfly looks good, but the changes are causing Undertow to fail. Did you intentionally add those changes on your last commit?

stuartwdouglas:
Sorry, that was unintentional.

PR #1461, opened 2015-04-01
Fixed PHP install and silex-raw #720

### LINKED ISSUE #720: Validation Failure - silex-raw
https://github.com/TechEmpower/TFB-Round-9/blob/master/i7/verify-test-1/latest/logs/silex-raw

No response bodies are present for any of the tests.

See `err.txt` for the exact errors present.

ssmith-techempower:
Fixed the PHP install so that silex-raw successfully runs. Fixes #720. Also fixed validation warnings for both silex and silex-raw. Since this affects the PHP installation script PHP will have to be reinstalled if you already have it installed. This might also help address some of the issues we've had running other PHP frameworks (see #1406).

### DIFF
--- a/frameworks/PHP/php-silex/benchmark_config
+++ b/frameworks/PHP/php-silex/benchmark_config
@@ -3,8 +3,9 @@
   "tests": [{
     "default": {
       "setup_file": "setup",
+      "json_url": "/json",
       "db_url": "/db",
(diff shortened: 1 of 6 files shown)


LadyMozzarella:
:+1:

msmith-techempower:
This request may also fix a few other PHP frameworks.

PR #1462, opened 2015-04-01
Add phoenix

waiteb3:
Because I promised


toolset/setup/linux/languages/elixir.sh
```diff
@@ -1,16 +1,13 @@
...
 
-# TODO seems to be broken installation
-echo "WARN: Elixir does not install"
-return 1
+RETCODE=$(fw_exists ${IROOT}/elixir.installed)
+[ ! "$RETCODE" == 0 ] || { return 0; }
 
-# fw_exists v0.13.3.tar.gz
-# [ $? -ne 0 ] || { return 0; }
+wget http://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb
```
msmith-techempower: Should this be moved into `toolset/setup/linux/languages/erlang.sh` so other erlang frameworks can simply add it as a dependency if they aren't using elixir?

waiteb3: This is just `erlang.sh` with the elixir install added (line 11).

msmith-techempower: Then why is it here? Specify that `elixir` requires `erlang` via `fw_depends erlang` in an early line.

waiteb3:
Cleaned it up.

toolset/setup/linux/languages/elixir.sh
```diff
@@ -1,16 +1,9 @@
...
-# TODO seems to be broken installation
-echo "WARN: Elixir does not install"
-return 1
+RETCODE=$(fw_exists ${IROOT}/elixir.installed)
+[ ! "$RETCODE" == 0 ] || { return 0; }
 
-# fw_exists v0.13.3.tar.gz
-# [ $? -ne 0 ] || { return 0; }
+fw_depends erlang
+sudo apt-get install -y elixir
```
hamiltont: This should be a download, compile, install into `IROOT` instead - we're trying to completely remove sudo

frameworks/Elixir/phoenix/setup.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# sed -i 's|db_host: "localhost",|db_host: "${DBHOST}",|g' config/config.exs
```
hamiltont: Did you mean to leave this commented out?

hamiltont:
Awesome, thanks! I've left some minor comments

toolset/setup/linux/languages/elixir.sh
msmith-techempower: Agreed.

frameworks/Elixir/phoenix/setup.sh
waiteb3: Yes. It doesn't do any DB tests yet (the ORM is still pretty young).

hamiltont: Gotcha, just checking

hamiltont: Perhaps delete the line if it's unnecessary if you're making more changes to the PR

toolset/setup/linux/languages/elixir.sh
waiteb3: If I promise to figure it out for the next round (erlang as well), can I be lazy today?

LadyMozzarella: This would be merged in for Round 11. Does that qualify as the next round?

waiteb3:
Actually, I'll issue another PR when I get the ORM and other stuff figured out. Kane says it won't get into this round.

msmith-techempower:
kain*

Of course it's not making it into this round [10].

hamiltont:
FWIW, this [passed](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/56779608) all verification checks, so if `elixir.sh` was not using sudo I'd be happy to merge this

hamiltont:
So how about this: 

````
# Download and install without sudo
apt-get download elixir
dpkg -x elixir_*.deb $IROOT/elixir

# Binaries are now located here
EROOT=$IROOT/elixir/usr/local/bin/elixir

# Libraries are here
ELIBS=$IROOT/elixir/usr/local/lib/elixir
```

msmith-techempower:
I like this, but they MUST have a permalink to a deb somewhere outside of using `apt-get`, right?

hamiltont:
Yup, i'm sure that's the case. This is just an easy option to find the current official deb ;-)

Issue #1117, opened 2014-10-20
Cabal occasionally fails on Travis-CI due to low memory


hamiltont:
Is there something that can be done to cause it to use less memory? 

[Here](https://travis-ci.org/hamiltont/FrameworkBenchmarks/jobs/38398368) is an example of a failed build that should pass verification according to multiple prior builds, multiple followup builds, and my local testing. It's dumping a lot of this into `err.txt` when trying to verify yesod

```
INFO:run-ci:cabal: Error: some packages failed to install:
INFO:run-ci:auto-update-0.1.2 failed during the building phase. The exception was:
INFO:run-ci:ExitFailure 9
INFO:run-ci:This may be due to an out-of-memory condition.
INFO:run-ci:bench-0.0.0 depends on system-filepath-0.4.12 which failed to install.
INFO:run-ci:bson-0.3.1 depends on data-binary-ieee754-0.4.4 which failed to install.
INFO:run-ci:byteorder-1.0.4 failed during the building phase. The exception was:
INFO:run-ci:ExitFailure 9
INFO:run-ci:This may be due to an out-of-memory condition.
INFO:run-ci:cereal-0.4.1.0 failed during the building phase. The exception was:
```

Note: Travis workers are currently linux containers with relative limits (e.g. container x gets at least 25% of the total RAM), so it's entirely feasible that some workers have enough memory to complete, whereas other workers are more saturated and don't have that extra memory. Something else to mention in #1070 - two identical jobs can still have different available host resources, which can cause problems like this issue

msmith-techempower:
Sometimes `-j1` helps... sometimes it doesn't.

msmith-techempower:
I BELIEVE this is addressed.

PR #1468, opened 2015-04-02
Do not apt-get upgrade

### LINKED ISSUE #1139: Remove apt-get upgrade from setup scripts
Presumably someone can install a fresh 14.04 image, run `toolset/run-tests.py --install server`, and be sitting on a machine that is configured to run a round. Running apt-get upgrade makes this a lot harder to achieve, and I'm not convinced it offers any tangible benefit. While not running upgrade doesn't offer any guarantees, it reduces the range of differences between two machines from "all possible packages" to "packages that TFB explicitly installs".

Thoughts?

### LINKED ISSUE #1369: Installation's apt-upgrade causes postgres errors
When running the prerequisite installation, the command `sudo apt-get -yq upgrade -o Dpkg::Options::=--force-confdef -o Dpkg::Options::=--force-confold` tries to update all the postgres versions, which causes lots of error output because once the first postgres version is started none of the others can start. 

This appears to also cause errors to be returned from subsequent apt commands as dpkg attempts to recover from it's prior issue. 

```
Setting up postgresql-client-common (166.pgdg12.4+1) ...
Setting up postgresql-client-9.2 (9.2.10-1.pgdg12.4+1) ...
Setting up postgresql-common (166.pgdg12.4+1) ...
 * Starting PostgreSQL 9.3 database server
   ...done.
Setting up postgresql-9.2 (9.2.10-1.pgdg12.4+1) ...
Refused to start PostgreSQL 9.2, because PostgreSQL 9.3 is currently running! You should first stop 9.3 instance...
invoke-rc.d: initscript postgresql, action "start" failed.
dpkg: error processing postgresql-9.2 (--configure):
 subprocess installed post-installation script returned error exit status 13
dpkg: dependency problems prevent configuration of postgresql-contrib-9.2:
 postgresql-contrib-9.2 depends on postgresql-9.2 (= 9.2.10-1.pgdg12.4+1); however:
  Package postgresql-9.2 is not configured yet.
dpkg: error processing postgresql-contrib-9.2 (--configure):
 dependency problems - leaving unconfigured
Setting up postgresql-client-9.4 (9.4.1-1.pgdg12.4+1) ...
No apport report written because the error message indicates its a followup error from a previous failure.
Setting up postgresql-9.4 (9.4.1-1.pgdg12.4+1) ...
Refused to start PostgreSQL 9.4, because PostgreSQL 9.3 is currently running! You should first stop 9.3 instance...
invoke-rc.d: initscript postgresql, action "start" failed.
dpkg: error processing postgresql-9.4 (--configure):
 subprocess installed post-installation script returned error exit status 13
dpkg: dependency problems prevent configuration of postgresql-contrib-9.4:
 postgresql-contrib-9.4 depends on postgresql-9.4 (= 9.4.1-1.pgdg12.4+1); however:
  Package postgresql-9.4 is not configured yet.
dpkg: error processing postgresql-contrib-9.4 (--configure):
 dependency problems - leaving unconfigured
Setting up postgresql-client (9.4+166.pgdg12.4+1) ...
No apport report written because MaxReports is reached already
<snip>
Errors were encountered while processing:
 postgresql-9.2
 postgresql-contrib-9.2
 postgresql-9.4
 postgresql-contrib-9.4
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

hamiltont:
Increases stability of our scripts by avoiding unexpected upgrades

Fixes #1369 
Fixes #1139

### DIFF
--- a/toolset/setup/linux/prerequisites.sh
+++ b/toolset/setup/linux/prerequisites.sh
@@ -17,7 +17,6 @@ echo 'deb http://downloads-distro.mongodb.org/repo/ubuntu-upstart dist 10gen' | 
 # One -q produces output suitable for logging (mostly hides
 # progress indicators)
 sudo apt-get -yq update
-sudo apt-get -yq upgrade -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold"
 
 # WARNING: DONT PUT A SPACE AFTER ANY BACKSLASH OR APT WILL BREAK
 sudo apt-get -qqy install -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" \


msmith-techempower:
Will this leave us in the 'never ran upgrade' position?

hamiltont:
More like a "Only run upgrade manually" situation - the problem is that automatic upgrade of all software on the machine has big implications, and often creates complex problems. This upgrade has causes issues in Travis with mongodb, rsyslog, and postgres, and they are annoyingly complex issues (like upgrade seems to re-install the postgres version that we expressly uninstalled, then try and start it, which fails because an older version is running or something crazy like that)

hamiltont:
Basically, if a human is there to understand what to do when apt-get upgrade fails and dpkg starts refusing to do anything more, then sure you can upgrade. Otherwise....it's a PITA

LadyMozzarella:
[Travis build](https://travis-ci.org/LadyMozzarella/FrameworkBenchmarks/builds/56947388) - not yet started as of this commit.

hamiltont:
This one was actually [scheduled correctly](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/56821392) on the main queue. I think a lot of them were (see [here](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/pull_requests)) they just were not hyperlinked from the pull requests because travis didn't have write permission on the organization

LadyMozzarella:
whoops- thanks @hamiltont

hamiltont:
So travis finished, and the only two differences from the latest full build on master are `Haskell/yesod` and `JavaScript/nodejs`. NodeJS started passing (yay!), and yesod started failing with this: 

```
attoparsec-0.12.1.5 failed during the building phase. The exception was:
ExitFailure 9
This may be due to an out-of-memory condition.
bench-0.0.0 depends on attoparsec-0.12.1.5 which failed to install.
```

`Scala/scruffy` started failing, but that turns out to actually be true (master just hadn't caught up to the failure yet - see #1481). 

I'm hoping to merge this, thoughts?

hamiltont:
As an added bonus, this drops about 2 hours off the total travis build time :-P

msmith-techempower:
`cabal` (the package manager for `haskell`) often runs out of memory in the Travis environment. I would suggest pulling and testing on a VM with a bit more RAM and proving it is a false-positive.

methane:
@msmith-techempower See #1117

hamiltont:
Sadly, I have confirmed that `yesod` is indeed broken in master, and that is not just an effect of removing `apt-get upgrade`. Confirmed via travis and virtualbox. 

So the only test affected by this PR appears to be nodejs, which now passes.

PR #1472, opened 2015-04-02
start distinct Go process per 2 (virtual) cores.

funny-falcon:
It is very similar to usage nodejs cluster package or Ruby Unicorn server, so i think it is legitimate.

It is simple version, just to benchmark, but it could be easily extended to match Ruby Unicorn or nodejs cluster functionality.

I think it is able to greatly improve Go performance on Peak and EC2. Even i7 benchmark can benefit from this change.

### DIFF
--- a/frameworks/Go/go/benchmark_config
+++ b/frameworks/Go/go/benchmark_config
@@ -23,6 +23,29 @@
       "display_name": "go",
       "notes": "",
       "versus": "go"
+    },
+    "prefork": {
+      "setup_file": "setup_prefork",
+      "json_url": "/json",
+      "db_url": "/db",
+      "query_url": "/queries?queries=",
+      "fortune_url": "/fortune",
+      "update_url": "/update?queries=",
+      "plaintext_url": "/plaintext",
+      "port": 8080,
+      "approach": "Realistic",
+      "classification": "Platform",
+      "database": "MySQL",
+      "framework": "go",
+      "language": "Go",
+      "orm": "Raw",
+      "platform": "Go",
+      "webserver": "None",
+      "os": "Linux",
+      "database_os": "Linux",
+      "display_name": "go",
+      "notes": "",
+      "versus": "go"
     }
   }]
 }
--- a/frameworks/Go/go/setup_prefork.sh
+++ b/frameworks/Go/go/setup_prefork.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+sed -i 's|tcp(.*:3306)|tcp('"${DBHOST}"':3306)|g' src/hello/hello.go
+
+# Where to find the go executable
+export PATH="$GOROOT/bin:$PATH"
+
+go get ./...
+
+go run src/hello/hello.go -prefork &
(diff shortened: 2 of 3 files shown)


methane:
While I'm interested in effect of this change, I think preforking is not a usual way in Go's world.
I want to keep our code in good shape: new Gophers can use our code as sample code for writing production ready server.

I want to see future version of Go resolve scaling issue.
Go 1.4 has lock contention problem in database/sql. It is found in this benchmark and I've fixed it.
I'll try to investigate this problem before Go 1.6 code freeze, too.

funny-falcon:
But "prefork" helps even with "plain text" and "json" benchmarks. So it is not tied to database/sql.

If i made it as a public library, will you use it for benchmark? Perhaps, as a separate "framework".

Or as external tool like https://github.com/stripe/einhorn with library as adapter for?

methane:
First of all, this PR may not be merged Round 10 because Round 10 is in final phase for now.
(But I don't know, because I'm not a member of TechEmpower).

[General requirements](http://frameworkbenchmarks.readthedocs.org/en/latest/Project-Information/Framework-Tests/#general-test-requirements) says:

> All test implementations should be production-grade. The particulars of this will vary by framework and platform, but the general sentiment is that the code and configuration should be suitable for a production deployment. 

So I want our code and configuration are good enough for reference.
Using library or not is not a problem. The problem is: Is prefork a recommended / commonly used technique for high load production environment?

If the answer is yes, I'm OK to merge this.
If not, I think preforking benchmark should be separated from normal "go" benchmark. "go-prefork", maybe.

Split benchmark doesn't require split the code. Adding `-prefork` flag is enough, maybe.
I don't like einhorn since it depends Ruby. It makes hard to run test for new contributers.

msmith-techempower:
I agree with @methane

funny-falcon:
I've modified code, so by default it runs in one process.

If flag `-prefork` passed then prefork performed. It will be great, if it is added as `go-prefork`.

hamiltont:
> Split benchmark doesn't require split the code. Adding -prefork flag is enough, maybe.

If this is good with both of you, it's good with our team. You'll need to add the new configuration to the [benchmark_config](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Go/go/benchmark_config#L4), and create a new setup file like `setup_prefork` to do exactly what `setup.sh` does but also pass the `-prefork` flag

funny-falcon:
Ok, i've edit benchmark_config and add setup_prefork.sh . Hope i've done it right.

hamiltont:
@funny-falcon Yup, that looks correct to me! We will now see what travis says

funny-falcon:
looks like travis doesn't mind.

funny-falcon:
:( why results could not be viewed?

methane:
I suppose it because it's too late for Round 10.
Please wait for round 11.

FYI, Go 1.5 improved schedular:
https://github.com/golang/go/commit/f47e581e022a29669d4dd6b62f8bf7d45c8395e5

So performance of Go without prefork will be improved slightly.

PR #1477, opened 2015-04-02
New test implementation: Scala, Spray and Elasticsearch

marko-asplund:


martin-g:
Just to remind that the Spray team requested removing another Spray test some months ago because it wasn't using their best practices.
Since then they didn't provide a better implementation though.

marko-asplund:
that's an interesting approach, thanks for the heads up.

I guess that in the OSS world we should usually address issues of non-conformity or non-optimal implementation with pull requests and not take-down notices issued by *the central committee* :-)

marko-asplund:
Test run on the CI passed:  https://travis-ci.org/marko-asplund/FrameworkBenchmarks/jobs/57043378

hamiltont:
I'm good to merge this. 

> Just to remind that the Spray team requested removing another Spray test some months ago because it wasn't using their best practices.

If I recall correctly, the implementation was quite dated. This implementation should be much easily to submit a PR for if there are concerns. Perhaps someone can cross-post this to the spray mailing list to encourage those concerned to take a look and at least open an issue if they have complaints

marko-asplund:
@hamiltont I've also been thinking about asking spray developers for feedback on this implementation. I'll take the issue up on the list.

hamiltont:
Thanks, I think that's the right approach

Issue #1479, opened 2015-04-03
Download debs instead of compiling from source


hamiltont:
This will be much, much faster than compiling big projects from source, doesn't require sudo, and still installs everything into a prefix folder. I expect we can also find a version of this that works with rpm-based systems. 

Turns out this is really easy to do. Here's a rough pattern: 

```
# Download and install without sudo
apt-get download somepackage
dpkg -x somepackage_*.deb $IROOT/somepackage

# Binaries and Libraries are now stores inside here
# For each somepackage you have to ID where they are precisely
# so you can use them
SOMEROOT=$IROOT/somepackage/usr/local/bin/somepackage
```

Note that `apt-get download` does not also download dependencies. If you need to download dependencies too, use this apt command instead to download all the necessary deb files. 

```
function apt-download { wget -c $(apt-get install --reinstall --print-uris -qq $1 | cut -d"'" -f2); }
```

Thoughts?

hamiltont:
Function credit: http://askubuntu.com/a/30581/91562

msmith-techempower:
I think that it's an interesting idea, but it ultimately doesn't solve the problem that is all `apt-get` are not created equal. For instance, our suite doesn't install properly on Debian because the package names for `apt-get` aren't 100% equal.

The only real solution I can offer is to scour for permalinks to individual `.deb` files as needed, then do the `dpkg` stuff. Yes, this will require each installation to know its own dependencies, but that is probably even better -- we'll know exactly what is required for a given stack.

ssmith-techempower:
I've run into issues with `apt-get` while getting TFB running on our Raspberry Pi 2. Apt will potentially install things in platform-specific locations, which leads to issues caused by frameworks' setup scripts looking for things in platform-specific locations.

The Java frameworks specifically seem to assume `java` is located at:

    /usr/lib/jvm/java-1.7.0-openjdk-amd64/bin/java

On the RPi2 with their flavor of Ubuntu 14.04 `apt-get` installs it here:

    /usr/lib/jvm/java-1.7.0-openjdk-armhf/bin/java

nbrady-techempower:
I think this can be closed for #1500 though I still think we're a ways away from that. I'm working on travis and toolset changes currently, and will see if I can start moving things toward distro agnostic installations.

PR #1482, opened 2015-04-03
Fix ASP.NET 

hamiltont:
Replaces #1287

Introduces a new function `fw_apt_to_iroot` that uses `apt-get` to download a deb file, and `dpkg -x` to extract it, all without using sudo. See #1479

Using this method we can download the latest available binaries for mono

Humorously (after the intense discussion over sudo in #1287) this actually uses sudo, but only for adding the package repository to apt's list of package locations, which is a pretty innocuous change to the base OS. Perhaps a later version of the `fw_apt_to_iroot` function can find a way to use a package repo without using sudo to add it to `/etc/apt/sources.list.d/`


hamiltont:
[my travis run](https://travis-ci.org/hamiltont/FrameworkBenchmarks/jobs/56996386)

hamiltont:
Travis passed, except for mongodb (which failed to start as shown by `mongodb: is _NO_ GO!: ERROR: [Errno 111] Connection refused`).

hamiltont:
ping @pdonald - in case you have any remaining feedback

hamiltont:
Last travis shows full green on aspnet except for what I think is a spurious failure on aspnet-mono-mysql-entityframework's fortunes (which have PASSed multiple times before). Merging as better than current

PR #1483, opened 2015-04-03
go: Fix Content-Type

methane:


LadyMozzarella:
That looks good. I'll merge, but jsyk I all of the headers appear to be missing "Server".

Issue #1473, opened 2015-04-02
Error in toolset when checking for required headers in Travis-CI


LadyMozzarella:
Found in [job #56805539](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/56805539)
```bash
Accessing URL http://127.0.0.1:9000/db
  Response (trimmed to 40 bytes): "{"id":3304,"randomNumber":2139}"
WARNING:root:Verifying test db for play2-scala-activate caused an exception: argument of type 'NoneType' is not iterable
   FAIL for http://127.0.0.1:9000
     Caused Exception in TFB
               This almost certainly means your return value is incorrect, 
               but also that you have found a bug. Please submit an issue
               including this message: argument of type 'NoneType' is not iterable
     Traceback (most recent call last):
       File "/home/travis/build/TechEmpower/FrameworkBenchmarks/toolset/benchmark/framework_test.py", line 279, in verify_type
         results = test.verify(base_url)
       File "/home/travis/build/TechEmpower/FrameworkBenchmarks/toolset/benchmark/test_types/db_type.py", line 31, in verify
         if any(v not in response for v in ('Server','Date','Content-Type: application/json')) \
       File "/home/travis/build/TechEmpower/FrameworkBenchmarks/toolset/benchmark/test_types/db_type.py", line 31, in <genexpr>
         if any(v not in response for v in ('Server','Date','Content-Type: application/json')) \
     TypeError: argument of type 'NoneType' is not iterable
```

hamiltont:
Update on this: In a number of places, the new verification `returns` if the headers are not correct. Can this be easily updated to just append the header warning to the full list of warnings? I've noticed a few tests that should be marked `fail` due to their responses are being marked `warn` because the header verification doesn't pass, but just returns `warn` and therefore is skipping all the further checks

msmith-techempower:
Again, I want to emphasize that the headers should be WARNINGS, not ERRORs.

LadyMozzarella:
It isn't supposed to be erroring. I set it up to return with a warn. Either this check is being done too early, or we should just append a note of some sort (like @hamiltont said) that says they're missing headers. I will add this to my todo list for today and see if we can do this check last or see what I can figure out. :)

hamiltont:
[...]

PR #1485, opened 2015-04-03
Check header responses last, Make case insensitive, Closes #1475 as replacement

LadyMozzarella:
* Moved all checks for required headers to be the last thing checked as discussed in issue #1473
* Made header checks case insensitive to close #1475 as replacement


LadyMozzarella:
@hamiltont - Based off of dart-redstone and nodejs and Travis-CI I think this should be good to merge now :)

hamiltont:
LGTM

PR #1478, opened 2015-04-02
Update: Remove bash profile, Closes #1314 by replacement

LadyMozzarella:
__Replacement for #1314__

Updates made based off of comments made on #1314 on top of @lgratrix-techempower's commits.

PR #1486, opened 2015-04-03
Add Golang Goji (resolves #1452)

### LINKED ISSUE #1452: Go: Add Goji
Goji is an easy-to-use Go framework. It is lightweight and opinionated, allowing for quick use by programmers. Goji should be as fast as other Go frameworks, so the question is how well its HTTP router and other under-the-hood features are implemented.

https://github.com/zenazn/goji

zane-techempower:
Add Goji, a Go web framework. 6/6 tests added and passing.


frameworks/Go/goji/setup.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# sed -i 's|tcp(.*:3306)|tcp('"${DBHOST}"':3306)|g' src/hello/hello.go
+
+# Where to find the go executable
+export PATH="$GOROOT/bin:$PATH"
+
+go get github.com/go-sql-driver/mysql
+go get github.com/zenazn/goji
+go get github.com/zenazn/goji/web
```
methane: `install.sh` is preferred place for `go get`.

frameworks/Go/goji/src/goji/server.go
```diff
@@ -0,0 +1,220 @@
...
+// Sets the content type of response. Also adds the Server header.
+func setContentType(w http.ResponseWriter, contentType string) {
+	w.Header().Set("Server", "Goji")
+	w.Header().Set("Content-Type", contentType)
+}
+
+// Test 1: Json Serialization
+func serializeJson(c web.C, w http.ResponseWriter, r *http.Request) {
+	setContentType(w, "application/json")
+    json.NewEncoder(w).Encode(&Message{helloWorldString})
```
methane: Did you use gofmt?

frameworks/Go/goji/bash_profile.sh
```diff
@@ -0,0 +1,4 @@
+# Set the root of our go installation
+export GOROOT=${IROOT}/go
+
+export GOPATH=${TROOT}
```
LadyMozzarella: Can you move these variables to your setup.sh and/or install.sh and then remove the bash_profile.sh file. We just merged a change to get rid of the need for this file entirely. (issue #1478)

hamiltont: > setup.sh and/or install.sh

`and` is what's needed

LadyMozzarella:
Also, could you please add the .json extension to your benchmark_config. That would be baller. We just merged #1367 that does it for all of the frameworks.

frameworks/Go/goji/src/goji/server.go
zane-techempower: I did not. My gosublime was giving me trouble. I installed gofmt from the command line and applied it. Thanks for the catch.

zane-techempower:
Made the requested changes!

hamiltont:
LGTM

Issue #1343, opened 2015-02-24
Extend benchmark machine's SLOCCount to understand Ur/Web source


achlipala:
The SLOCCount manual suggests modifying a script in `/usr/bin` to add support for new languages.  It seems a bit gross, but also not so bad.  Is it possible to make a modification to `/usr/bin/break_filelist`?

Specifically, near the end of the definition of `%file_extensions`, add this line right before the closing brace:
```
  "ur" => "ml", "urs" => "ml", "urp" => "sh",
```

Something similar ought to be done for all other participating languages that SLOCCount doesn't speak out of the box.

achlipala:
I've learned that TFB is actually using the program CLOC, where there is a nicer way to add a new language!  The following works for me:

 1. Run `cloc --write-lang-def=/tmp/defs.txt`.
 2. Add the following lines to `/tmp/defs.txt`:
```
Ur/Web
    filter call_regexp_common Pascal
    extension ur
    extension urs
    3rd_gen_scale 3.00
    end_of_line_continuation \\$
```
 3. Run CLOC like `cloc --read-lang-def=/tmp/defs.txt`.

Perhaps there should be a custom language-definition file checked into the TFB repo and used by the benchmarker?

hamiltont:
Awesome. So my thought is that the ideal PR would allow a file "cloc_defs.txt" inside each frameworks folder e.g. `frameworks/Ur/urweb/cloc_defs.txt`. If the `count_sloc` finds that file, it will pass it to the `cloc` program. 

I'd prefer this be a magic file inside the framework folders instead of a single file inside `config` because there's a high chance that we will have file extension conflicts somewhere across our ~120 frameworks

achlipala:
Are you saying that I need to make the TFB-level changes and put in a PR if I want to see this implemented? :)

hamiltont:
Yup ;-) 

Here's my [guess](https://gist.github.com/hamiltont/6cfd16fe30bf47389089) on what needs to go into the codebase, so if this works as expected the PR is basically this plus documentation in the readme

achlipala:
Thanks for the code suggestion.  Do you have any tips on the easiest way to test this functionality, ideally without creating a virtual machine, but also without adding complicated configuration to my workstation?  (I've never set up anything close to the full benchmarking environment here.)

hamiltont:
[...]

PR #1493, opened 2015-04-05
Extensible line-counting for measuring source code

achlipala:
Here's my interpretation of @hamiltont's suggestion in #1343.  It allows each framework to specify the appropriate rules for counting its lines of code.

Full disclosure: I haven't tested this code as part of a full benchmarking run, but only by commenting out the code that actually runs the tests and adding some extra debug prints.


hamiltont:
Clean addition, thanks

Issue #1012, opened 2014-08-14
Go framework passes verification with warnings


hamiltont:
As seen [here](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/32507549), go has some minor issues with the queries test. It should be 1) coercing into a number and 2) coercing that number to always be between 1 and 500 if needed.

LadyMozzarella:
Closing. This appears to have been resolved on master.

PR #1495, opened 2015-04-06
Add go env variable exports to go-prefork test

LadyMozzarella:
- Add export variables to go-prefork test
- This should have been done when removing bash_profile, but was missed.


hamiltont:
If travis passes here, can someone take a look to see if #1012 is still an issue?

LadyMozzarella:
Travis-CI has passed and go is no longer warning as described in issue #1012 (queries=foo and queries=0 are now responding as if queries=1 was sent). The responses are looking good now, except that they now all warn for missing a required header.

PR #1501, opened 2015-04-07
Python web2py

knewmanTE:
Adds Python [web2py](http://www.web2py.com/) framework tests.


LadyMozzarella:
It looks like there are a lot of error files at $TROOT/Python/web2py/web2py/applications/app/errors/ that I'm guessing were accidentally included. Are we able to remove them?

LadyMozzarella:
Also, can you rename benchmark_config to benchmark_config.json?

frameworks/Python/web2py/setup.sh
```diff
@@ -0,0 +1,4 @@
+export PY2_ROOT=$IROOT/py2
+export PY2=$PY2_ROOT/bin/python
+
+sudo chmod a+w web2py/httpserver.pid & $PY2 web2py/web2py.py -a '' -i 127.0.0.1 -p 8080  &
```
hamiltont: What's the requirement for sudo here? This script is run as testrunner, who should already have write permissions on the `web2py/` folder. If not, mark the folder as group and/or world writable. This sudo would break as we get closer to R11 and remove sudo permissions from testrunner

knewmanTE: I tried running the tests without that and it seems to work fine, so I'll remove the sudo command. I think it was a remnant of when I was still learning web2py and was having trouble with some of their boilerplate files.

hamiltont:
Glad to see another one of the New Framework issues being tackled. My main concern is that this appears to be way too many files - are you downloading the framework, or have you included the source code for the entire web2py framework into this pull request? If the latter, please switch to downloading it, such as calling git clone inside install.sh

knewmanTE:
Currently I have it cloned. My current understanding of web2py is that it requires a lot of the boilerplate code to be right in the directory, but I can see if there's a way to just install it like the other frameworks. I agree that it makes things ugly and burdensome.

knewmanTE:
On the web2py roadmap, it looks like there isn't currently a pip installable version: https://trello.com/c/VcEHYEol/24-pip-installable-package

hamiltont:
Looking better! To address this: 

> web2py is that it requires a lot of the boilerplate code to be right in the directory

Run a `git clone` from `install.sh`. For example, that's what [silicon's installer does](web2py is that it requires a lot of the boilerplate code to be right in the directory). Obviously we'd prefer to clone inside of `IROOT`, but if that's too hard then for now just run the clone inside `TROOT` (which will be set as the web2py directory)

frameworks/Python/web2py/install.sh
```diff
@@ -0,0 +1,18 @@
...
+fw_depends python2
+
+export PY2_ROOT=$IROOT/py2
+export PY2_PIP=$PY2_ROOT/bin/pip
+
+$PY2_PIP install --install-option="--prefix=${PY2_ROOT}" -r $TROOT/requirements.txt
+
+cd $TROOT
+rm -fr web2py
+git clone --recursive https://github.com/web2py/web2py.git
```
hamiltont: Thanks! Is there any reason we should not pin this to a specific tag e.g. `R-2.10.3`? That would make the version used explicit, so anyone looking at the results in the future would be able to tell exactly what commit we were using

frameworks/Python/web2py/app/app/databases/7858106637800f617850eed16452cf10_fortune.table
```diff
@@ -0,0 +1,41 @@
+(dp1
```
hamiltont: Does this file need to be included in the pull request?

knewmanTE: It doesn't. I removed that and a couple other unnecessary files from the app/ folder.

frameworks/Python/web2py/install.sh
knewmanTE: I added a the most recent tag to the git clone.

Issue #1416, opened 2015-03-12
Don't print response string if test passes verification


hamiltont:
Perhaps still print the headers, but not the entire response string.

hamiltont:
Closed by #1505

PR #1505, opened 2015-04-08
Improve framework server log handling

### LINKED ISSUE #1171: Error output does not include copy/paste code
@hamiltont Error like this (https://gist.github.com/msmith-techempower/65bcff7b8b198cb170cd) produce output that is meant to be copypasteable for the users to try out manually, but `$FWROOT` is never set.

### LINKED ISSUE #1504: Logs not capturing all framework output
This is a pretty critical one. When you fork a process (like we do at the end of every `setup.sh`) there is a 4KB output buffer added to that forked process. If you SIGKILL a process (like we do to stop all frameworks), then that output buffer is destroyed without being flushed. *This means that we are not capturing this output in any log files.* Some examples using the R10 final peak results: 

[lwan output](https://github.com/TechEmpower/TFB-Round-10/blob/master/peak/linux/results-2015-03-24-peak-final/latest/logs/lwan/out.txt#L87) ends with `[100%] Built target techempower`, but if you run it manually you'll see stdout happening after the build stage e.g. 

```
Loading configuration file: techempower.conf.
Using 32 threads, maximum 2000 sockets per thread.
Listening on http://0.0.0.0:8080.
TCP_FASTOPEN not supported by the kernel.
Ready to serve.
``` 

Similarly for haywire, duda, and multiple others - we are losing small bits of output that are probably very critical to the framework devs when trying to debug their system (e.g. `TCP_FASTOPEN not supported by the kernel.` would probbaly be a pretty important message if this wasn't output from travis-CI). 

Granted, this is a small amount of output - no framework I could find is sending a lot of information to stdout - but the information they do send is likely critical to successful debugging.

hamiltont:
A step along the path towards #920. This prints output from `setup.sh` in real time to both the console and to the log file. Log output is much more readable, as each test gets a custom prefix. Also closes #1504 - which I think is a critical bug. Travis-CI no longer has to re-print the `err.txt` and `out.txt` files, so that is a really nice benefit as it's much easier to read. Main negative is that `framework_test#start` has become a more complex method, but that's unfortunately the nature of nonblocking IO with multiple levels of buffering and multiple levels of child processes, so I've just commented the hell out of it. If someone wants to start being awesome and throwing some unit tests against that method be my guest :+1: 

```
Setup lwan: [ 96%] Building C object techempower/CMakeFiles/techempower.dir/array.c.o
Setup lwan: [100%] Building C object techempower/CMakeFiles/techempower.dir/database.c.o
Setup lwan: Linking C executable techempower
Setup lwan: [100%] Built target techempower
Setup lwan: + cd /home/vagrant/FrameworkBenchmarks/installs/lwan/techempower
Setup lwan: + /home/vagrant/FrameworkBenchmarks/installs/lwan/build/techempower/techempower
Setup lwan: Loading configuration file: techempower.conf.
Setup lwan: Using 2 threads, maximum 32767 sockets per thread.
Setup lwan: Listening on http://0.0.0.0:8080.
Setup lwan: Ready to serve.
Setup lwan: setup.sh exited due to bound port
INFO:root:Executed setup.sh, returning 0
INFO:root:Sleeping 10 seconds to ensure framework is ready
INFO:root:Verifying framework URLs
Accessing URL http://127.0.0.1:8080/json
  Response (trimmed to 40 bytes): "{"message":"Hello, World!"}"
   WARN for http://127.0.0.1:8080/json
     Required response header missing.
Accessing URL http://127.0.0.1:8080/plaintext
  Response (trimmed to 40 bytes): "Hello, World!"
   PASS for http://127.0.0.1:8080/plaintext
lwan: Setup.sh and framework processes have terminated
Server lwan: Framework processes have terminated
verify: 100% |######################################################################################################################################################################| Rough Time: 00:00:36
```

Also includes some basic code cleanup and fixes #1171

### DIFF
--- a/toolset/setup/linux/setup_util.py
+++ b/toolset/setup/linux/setup_util.py
@@ -1,7 +1,61 @@
 import re
 import os
+import sys
 import subprocess
 import platform
+
+from threading import Thread
+from Queue import Queue, Empty
+
+class NonBlockingStreamReader:
+  '''
+  Enables calling readline in a non-blocking manner with a blocking stream, 
+  such as the ones returned from subprocess.Popen
+
+  Originally written by Eyal Arubas, who granted permission to use this inside TFB
+  See http://eyalarubas.com/python-subproc-nonblock.html
+  '''
+  def __init__(self, stream, eof_message = None):
+    '''
+    stream: the stream to read from.
+            Usually a process' stdout or stderr.
+    eof_message: A message to print to stdout as soon
+      as the stream's end is reached. Useful if you
+      want to track the exact moment a stream terminates
+    '''
+
+    self._s = stream
+    self._q = Queue()
+    self._eof_message = eof_message
+    self._poisonpill = 'MAGIC_POISONPILL_STRING'
+
+    def _populateQueue(stream, queue):
+      while True:
+        line = stream.readline()
+        if line: # 'data\n' or '\n'
+          queue.put(line)
+        else:    # '' e.g. EOF
+          if self._eof_message:
+            sys.stdout.write(self._eof_message + '\n')
+          queue.put(self._poisonpill)
+          return
+
+    self._t = Thread(target = _populateQueue,
+            args = (self._s, self._q))
+    self._t.daemon = True
+    self._t.start()
+
+  def readline(self, timeout = None):
+    try:
+      line = self._q.get(block = timeout is not None,
+        timeout = timeout)
+      if line == self._poisonpill: 
+        raise EndOfStream
+      return line
+    except Empty:
+      return None
+
+class EndOfStream(Exception): pass
 
 # Replaces all text found using the regular expression to_replace with the supplied replacement.
 def replace_text(file, to_replace, replacement):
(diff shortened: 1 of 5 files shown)


hamiltont:
[My travis queue](https://travis-ci.org/hamiltont/FrameworkBenchmarks/builds/57600758)

toolset/setup/linux/setup_util.py
```diff
@@ -1,8 +1,62 @@
 import re
 import os
+import sys
 import subprocess
 import platform
 
+from threading import Thread
+from Queue import Queue, Empty
+
+class NonBlockingStreamReader:
```
hamiltont: By the way, I emailed the original author o fthis class and he confirmed we are free to use it as part of this project (I had to double-check because there's no license statement on it)

msmith-techempower:
This commit is a massive undertaking.

LadyMozzarella:
The good kind of massive undertaking? Personally, I think this looks good. I think it will be extremely helpful when debugging and also like that the structure appears to flow in the order of what's happening, making it overall easier to read. 

Regarding unit testing, I believe that it would be beneficial to put unit tests on not just this, but all of the methods that are running the benchmarks. I was looking into beginning that or submitting an issue for it,  but wanted to first try and get TFB as cleaned up as possible. Maybe I can swing back around to this at some point. I would like to see this project super solid.

hamiltont:
Re the massive undertaking, this is a finished PR (unless we find bugs), so what you see is what you get here. Basically, this all started as trying to get streaming output from `setup.sh`. Along the way I discovered some bugs and cleaned up the code, but fundamentally all it does is stream the output to the console and to the log file :-P

hamiltont:
Merging this will resolve #1416

msmith-techempower:
Sorry, I comment on these early in the morning when I get in and read my email - my comment was one of awe, not one of criticism. Glad this got merged.

Related - do I need to rebase #1508 after this merge? It seems like it might be affected.

PR #1507, opened 2015-04-08
Remove Weber framework, Resolve #1186

### LINKED ISSUE #1186: Fix Weber (Elixir)
This is the only elixir framework and it is failing.

LadyMozzarella:
I only removed the Weber setup file, and not the Elixir setup file so that #1462 doesn't have an issue when it's reopened (hopefully).


hamiltont:
LGTM

PR #1508, opened 2015-04-08
Add more colors

msmith-techempower:
* Adds corresponding colors to the verification output (in addition the currently implemented summary section).
* Adds the link to our documentation for warn/fail output to better inform contributors.


msmith-techempower:
Example output:
```
INFO:root:Verifying framework URLs
Accessing URL http://127.0.0.1:8080/:
  Response (trimmed to 40 bytes): "{"message":"Hello, World!"}"
   PASS for http://127.0.0.1:8080/
Accessing URL http://127.0.0.1:8080/plaintext:
  Response (trimmed to 40 bytes): "Hello, World!"
   WARN for http://127.0.0.1:8080/plaintext
     Server is returning 1 more bytes than are required.
     This may negatively affect benchmark performance.
verify: 100% |#############################################################################################| Rough Time: 00:00:57
  Verification Summary

| Test: gemini
|       plaintext   : WARN
|       json        : PASS
```

Same colors in the `WARN for http://...` as the summary at the bottom.

toolset/benchmark/framework_test.py
```diff
@@ -289,12 +292,25 @@ def verify_type(test_type):
...
+        print ("   " + color + "%s" + Style.RESET_ALL + " for %s") % (result.upper(), url)
         if reason is not None and len(reason) != 0:
           for line in reason.splitlines():
             out.write("     " + line + '\n')
             print "     " + line
+          if showRules:
+            out.write("     See %s" % specific_rules_url);
+            print "     See %s" % specific_rules_url
 
       [output_result(r1,r2,url) for (r1, r2, url) in results]
```
hamiltont: How about replacing the `showRules` lines above with a single conditional underneath line 315, such as this: 

```
if not test.passed:
  out.write("     See %s" % specific_rules_url)
  print "     See %s" % specific_rules_url
```

So that the rules URL is only printed a single time at the end instead of once per URL tested

toolset/benchmark/framework_test.py
```diff
@@ -289,12 +292,25 @@ def verify_type(test_type):
...
+          showRules = True
+
+        out.write(("   " + color + "%s" + Style.RESET_ALL + " for %s\n") % (result.upper(), url))
+        print ("   " + color + "%s" + Style.RESET_ALL + " for %s") % (result.upper(), url)
         if reason is not None and len(reason) != 0:
           for line in reason.splitlines():
             out.write("     " + line + '\n')
             print "     " + line
+          if showRules:
+            out.write("     See %s" % specific_rules_url);
```
hamiltont: stray `;`

toolset/benchmark/test_types/plaintext_type.py
```diff
@@ -21,11 +21,11 @@ def verify(self, base_url):
     body = body.lower()
 
     if "hello, world!" not in body:
-      return [('fail', "Could not find 'Hello, World!' in response", url)]
+      return [('fail', """Could not find 'Hello, World!' in response.""", url)]
 
     if len("hello, world!") < len(body):
       return [('warn', """Server is returning %s more bytes than are required.
-        This may negatively affect benchmark performance""" % (len(body) - len("hello, world!")), url)]
+This may negatively affect benchmark performance.""" % (len(body) - len("hello, world!")), url)]
```
hamiltont: Does changing the indention work? Thought that would break python syntax

hamiltont:
Great idea

toolset/benchmark/framework_test.py
msmith-techempower: It IS only printed out 1 time.

Also, I'm not certain here, but does "warn" count as `not test.passed`?

toolset/benchmark/framework_test.py
msmith-techempower: FoH - Java is superior!!!!

toolset/benchmark/test_types/plaintext_type.py
msmith-techempower: Yes, this works. The `"""` means to treat tabs/spaces as additional indentation. When I removed them all (as this line does), it lines up with the indentation of the output from the line before. If you look at my comment, it lines up much nicer.

hamiltont: Gotcha - this is better.

toolset/benchmark/framework_test.py
```diff
@@ -289,12 +292,22 @@ def verify_type(test_type):
...
+
+        out.write(("   " + color + "%s" + Style.RESET_ALL + " for %s\n") % (result.upper(), url))
+        print ("   " + color + "%s" + Style.RESET_ALL + " for %s") % (result.upper(), url)
         if reason is not None and len(reason) != 0:
           for line in reason.splitlines():
             out.write("     " + line + '\n')
             print "     " + line
+          if not test.passed:
+            out.write("     See %s" % specific_rules_url)
+            print "     See %s" % specific_rules_url
```
LadyMozzarella: Can we add a line break, so that [this doesn't happen](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/57712944#L6263) and the top header line "-------------------------" gets bumped to a new line?

```bash
INFO:run-ci:   WARN for http://127.0.0.1:8080/entityframework/postgresql?queries=foo
INFO:run-ci:     Top-level JSON is an object, not an array
INFO:run-ci:     See http://frameworkbenchmarks.readthedocs.org/en/latest/Project-Information/Framework-Tests/#specific-test-requirements--------------------------------------------------------------------------------
INFO:run-ci:  VERIFYING UPDATE
INFO:run-ci:--------------------------------------------------------------------------------
```

PR #1510, opened 2015-04-08
Phoenix and Erlang/Elixir without sudo

waiteb3:
@hamiltont 
I would have done an install from the debs, but those assume you're installing to /usr/local, so this gets it done from source.

Also, because I promised Ariana.
```
+      o     +              o    
    +             o     +       +
o          +
    o  +           +        +
+        o     o       +        o
-_-_-_-_-_-_-_,------,      o 
_-_-_-_-_-_-_-|   /\_/\  
-_-_-_-_-_-_-~|__( ^ .^)  +     +  
_-_-_-_-_-_-_-" " "   "      
+      o         o   +       o
    +         +
o        o         o      o     +
    o           +
+      +     o        o      +    
```


LadyMozzarella:
Can you rebase this please? I like just merged removal of the other Elixir framework and I think it caused a conflict.

waiteb3:
@LadyMozzarella yeah, I should actually remember to push the rebase

frameworks/Erlang/elli/bash_profile.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
LadyMozzarella: We're no longer using bash_profile.sh. exports need to be moved to setup.sh/install.sh

frameworks/Erlang/cowboy/bash_profile.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
LadyMozzarella: This bash_profile also needs to be removed.

frameworks/Elixir/phoenix/bash_profile.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
LadyMozzarella: Remove this bash_profile also

hamiltont:
Check out [this](https://github.com/hamiltont/FrameworkBenchmarks/blob/improve-java-setup/toolset/setup/linux/languages/java.sh) and [this](https://github.com/hamiltont/FrameworkBenchmarks/blob/improve-java-setup/frameworks/Java/activeweb/setup.sh#L4)

You may want to copy the idea for erlang - use the `erlang.installed` file to hold all the variables that need to be set in the various `setup.sh` files, such as PATH modifications, specific binary locations, etc

frameworks/Elixir/phoenix/setup.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+export PATH=$PATH:$ERL_BIN:$ELX_BIN
+
+# sed -i 's|db_host: "localhost",|db_host: "${DBHOST}",|g' config/config.exs
```
LadyMozzarella: Is this commented out for a reason?

Issue #1224, opened 2014-11-11
Add 'install' key to benchmark_config.json


msmith-techempower:
When trying to test a given framework under `Python`, I see it installing EVERYTHING python-related because the install scripts cover all requirements. An example of this is with `bottle` which has a `py2`, `py3`, and `pypy` implementation.

Question - should we change up the directory structure a bit for these since it seems like each of those python implementations act like different languages?

~/FrameworkBenchmarks/frameworks/Python2
~/FrameworkBenchmarks/frameworks/Python3
~/FrameworkBenchmarks/frameworks/PyPy

instead of just

~/FrameworkBenchmarks/frameworks/Python

Then, there would be a `bottle` directory under each, but the `benchmark_config`, `install.sh`, and `bash_profile.sh` would only need to set up and use the specific version of python based on the dir.

@hamiltont @methane Thoughts?

hamiltont:
Thought about this some more. 

Some info from `find frameworks/Python -name install.sh -exec grep fw_depends {} \;` shows that most of the python tests have py2 and py3 implementations:

```
fw_depends python3
fw_depends python2 python3 pypy nginx
fw_depends python2 python3
fw_depends python2 python3 pypy
fw_depends python2 python3 pypy nginx
fw_depends python2 python3
fw_depends python2 python3
fw_depends python2 nginx
fw_depends python2 nginx
```

I get the point -  if you just want to test `django-py3` why are you also compiling py2 and pypy, but it seems like it's easier to deal with that burden in this case than to have three directories that all have to be maintained e.g. `frameworks/Python2/django`, `frameworks/Python3/django`, and `frameworks/PyPy/django`

methane:
Additionally, I'll add PyPy3 in Round 11.

msmith-techempower:
lmao I love methane's response.

I believe what @hamiltont is getting at is that there should probably be a single directory `frameworks/Python` and therefore a single test directory (like `frameworks/Python/django`) that has setup scripts for all of the python implementations (`py2`, `py3`, `pypy`, etc).

hamiltont:
[...]

PR #1512, opened 2015-04-09
Improve java setup

hamiltont:
Mostly updates the java setup. The `java7.installed` and `java8.installed` files now contain environment variables, and can be sourced (e.g. `source java7.installed`) to load the proper `JAVA_HOME`, `PATH`, etc. This centralization should help us keep java up to date going forward. 

Also includes a ton of minor fixes that don't belong in other PRs


msmith-techempower:
I **REALLY** like this notion of sourcing the `.installed` files to set environment variables properly.

Should we consider this the expected way moving forward so that every Python `install.sh` and `setup.sh` don't have the same 3-4 lines at the top and instead only have `source $IROOT/python3.installed`?

msmith-techempower:
OH, OR we add a utility function for that it by name that will print helpful debugging info in error cases and the like:

`fw_source java7` for instance, could do `source $IROOT/java7.installed` in the background, and also print out useful stuff in the case where a user is contributing a new language (or w/e) and their `install.sh` didn't properly call `fw_install java7` at all, so the `$IROOT/java7.installed` doesn't exist.

hamiltont:
We absolutely should. I'm trying to determine if that should just be some new function like `fw_setup`, so that you just run `fw_depends foo` in `install.sh` and then `fw_setup foo` in `bash_profile.sh`, which basically just sources all the *.installed files for you. Thoughts? 

Also, I might have time to get to #1224 today, which should really help your python complaint. I'll just add an install key, so each framework test can manually specify a specific installation file

hamiltont:
Travis LGTM - gained `undertow`, all other Java-based tests stayed the same as on master

PR #1514, opened 2015-04-10
Add Clojure-Pedestal (resolves #583)

### LINKED ISSUE #583: Clojure: Add Pedestal
We'd like to have the Clojure framework named Pedestal added.

http://pedestal.io/

zane-techempower:
The changes made to the curl command that TFB uses fixed an odd character encoding problem that I couldn't figure out. Japanese/the dash chars were appearing fine when the server was run normally, but TFB would recieve the special characters differently, as `?`'s and fail the test.


.gitignore
```diff
@@ -3,6 +3,10 @@ installs
 *.log
 node_modules/
 
+# Added for pedestal framework test
```
hamiltont: Instead of adding this to the global `.gitignore` for the entire project, it would be better to create a file at `frameworks/Clojure/pedestal/.gitignore` and add these rules there

hamiltont: Oh...I see you did that already. I guess these couldn't go in there for some reason!

frameworks/Clojure/pedestal/install.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+export JAVA_HOME=/usr/lib/jvm/java-1.7.0-openjdk-amd64
+
+export LEIN_HOME=$IROOT/lein
+
+fw_depends leiningen java
```
hamiltont: Please rebase or merge master, and use `java7` instead of Java. See #1512 for details. Also, you can remove the line `export JAVA_HOME` from this file once you've merged master, installing it will set that

frameworks/Clojure/pedestal/setup.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+export JAVA_HOME=/usr/lib/jvm/java-1.7.0-openjdk-amd64
```
hamiltont: Per the changes in #1512, replace this line with `source $IROOT/java7.installed`

frameworks/Clojure/pedestal/setup.sh
```diff
@@ -0,0 +1,14 @@
...
+export LEIN_HOME=$IROOT/lein
+
+$LEIN_HOME/bin/lein clean
+
+rm -rf target
+# pack all dependencies into a single jar: target/pedestal-standalone.jar
+$LEIN_HOME/bin/lein uberjar
+# -server is much faster
+# 'lein run' passes '-client -XX:+TieredCompilation -XX:TieredStopAtLevel=1' which make it starts fast, but runs slow
+$JAVA_HOME/bin/java -server -jar target/pedestal-standalone.jar &
```
hamiltont: After modifying the above, it's actually better to just use `java` here instead of the specific path

hamiltont:
FYI, when I merged #1512 that's probably why your test failed with `Setup pedestal: setup.sh: line 14: /usr/lib/jvm/java-1.7.0-openjdk-amd64/bin/java: No such file or directory`. Just rebase this PR onto master (or merge in master, whatever you prefer), and it should get past that error

frameworks/Clojure/pedestal/install.sh
msmith-techempower: I would add that he, as the first submitter since #1512 should update `leiningen` and have the install script 1) create `leiningen.installed`, 2) have that mimic the `java7.intalled` and export `LEIN_HOME` there, and 3) `source $IROOT/leiningen.installed` in his `setup.sh`

frameworks/Clojure/pedestal/setup.sh
msmith-techempower: Agreed - let's enforce `$IROOT/[software].installed` files to export all their variables and include any regularly used binaries in the path.

zane-techempower:
I just made the changes. lein should be accessible for all frameworks just like java. In issue #1515 I note that I will be adding the update test to the remaining clojure frameworks. I will also update lein/java to the new way of doing things.

PR #1516, opened 2015-04-10
Added MongoDB client to PHP install scripts. Fixes #1401

### LINKED ISSUE #1401: PHP's Phalcon Mongodb - Verification Failed
This test failed preview 3 of Round 10 and it also fails locally.

* [php-phalcon-mongodb](https://github.com/TechEmpower/TFB-Round-10/tree/master/peak/linux/results-2015-03-06-peak-preview3/latest/logs/php-phalcon-mongodb) (single query, multiple queries, and fortunes):

  ```bash
Accessing URL http://10.0.3.2:8080/mongodb/queries?queries=501 
Response: 
"HTTP/1.1 500 Internal Server Error
Server: nginx/1.4.1
Date: Sat, 07 Mar 2015 16:31:32 GMT
Content-Type: text/html
Transfer-Encoding: chunked
Connection: keep-alive
X-Powered-By: PHP/5.5.17

"
   FAIL for http://10.0.3.2:8080/mongodb/queries?queries=2
     Empty Response
   WARN for http://10.0.3.2:8080/mongodb/queries?queries=0
     Empty Response
   WARN for http://10.0.3.2:8080/mongodb/queries?queries=
     Empty response (this will be a failure in future rounds, please fix)
   WARN for http://10.0.3.2:8080/mongodb/queries?queries=501
     Empty Response
```

### PR DESCRIPTION
**Pull request by ssmith-techempower**
The MongoDB PHP extension wasn't installed. Fixes #1401. This involves changes to PHP's install scripts, so if there's an existing PHP installation on your test box it'll have to be wiped and rebuilt.

### DIFF
--- a/config/php.ini
+++ b/config/php.ini
@@ -870,6 +870,7 @@ zend_extension=opcache.so
 extension=redis.so
 extension=phalcon.so
 extension=yaf.so
+extension=mongo.so
 ;extension=php_bz2.dll
 ;extension=php_curl.dll
 ;extension=php_fileinfo.dll
--- a/toolset/setup/linux/languages/php.sh
+++ b/toolset/setup/linux/languages/php.sh
@@ -49,6 +49,8 @@ printf "\n" | $IROOT/php-${VERSION}/bin/pecl -q install -f redis
 # yaf.so
 printf "\n" | $IROOT/php-${VERSION}/bin/pecl -q install -f yaf
 
+printf "\n" | $IROOT/php-${VERSION}/bin/pecl -q install -f mongo
+
 # phalcon.so
 #   The configure seems broken, does not respect prefix. If you 
 #   update the value of PATH then it finds the prefix from `which php`
--- a/frameworks/PHP/php-phalcon/app/views/layouts/mongobench.volt
+++ b/frameworks/PHP/php-phalcon/app/views/layouts/mongobench.volt
@@ -1,0 +1,1 @@
+<!DOCTYPE html><html><head><title>Fortunes</title></head><body>{{ content() }}</body></html>

### CODE REVIEW

File: config/php.ini [hunk 1]
```diff
@@ -870,6 +870,7 @@ zend_extension=opcache.so
 extension=redis.so
 extension=phalcon.so
 extension=yaf.so
+extension=mongo.so
```
msmith-techempower: This line... feels important.

toolset/setup/linux/languages/php.sh
```diff
@@ -49,6 +49,8 @@ printf "\n" | $IROOT/php-${VERSION}/bin/pecl -q install -f redis
 # yaf.so
 printf "\n" | $IROOT/php-${VERSION}/bin/pecl -q install -f yaf
```
msmith-techempower: I know you didn't add this, but isn't `yaf` a framework for PHP? Why is it installed in the base `php.sh`?

ssmith-techempower: It is, but I think a couple of the PHP frameworks are built on top of it. It can probably be pulled out into its own separate install, but I don't know which frameworks that would affect.

msmith-techempower: Eh, don't bother in this pr.

LadyMozzarella:
Fantastic! Thank you!!

toolset/setup/linux/languages/php.sh
hamiltont: I think it's here because it requires `yaf.so` in the PHP config e.g. turning on PHP would result in errors if yaf wasn't installed. To pull it into a separate file you'd need to modify the php config from the yaf installer

Issue #1425, opened 2015-03-17
Javascript: Add Koa


ebramanti-techempower:
Koa.js is a smaller, more expressive Node framework from the team who made Express. Koa uses generators to avoid Node's callback hell and make error-handling much easier. I'm curious about how this will affect performance compared to Express.js.

Here's some extremely basic differences between Koa & Express, with an extremely small performance comparison included in the MD: https://github.com/koajs/koa/blob/master/docs/koa-vs-express.md

ebramanti-techempower:
I've completed the test server for this framework. However, I need to do some more work before PR. Koa required a bump of Node to a newer version, and this is creating some problems for some of the other Node frameworks (particularly ones that use SQL). I'm taking a look into it, and you can check out my progress [here](https://github.com/ebramanti-techempower/FrameworkBenchmarks/tree/koa).

hamiltont:
@ebramanti-techempower FWIW, I *think* the best thing here would be if you could use something like `nvm` to enable each framework to choose the nodejs version it uses. See #995

ebramanti-techempower:
@hamiltont I can work on this, but should we not be updating nodejs test to latest version of Node anyways? The optimizations/bug fixes introduced in Node over the past two versions should be working with the current Node framework tests. 

Just wondering, because I know using `nvm` will save me a ton of time but that shouldn't stop me from making our current tests compatible with the latest version of Node.

hamiltont:
Updating nodejs is a great idea! If possible, let's do both - update nodejs to the latest, and use nvm to isolate the environment of each nodejs-based framework

ebramanti-techempower:
@hamiltont So I'm working on setting up `nvm`, but every time I try to source the necessary .sh file I'm getting weird errors. I'm not sure whether it has to do with our own scripts, or `nvm`. However, since I'm getting error 2, I'm assuming it's something related to `nvm`.

What's weird is the command works in `install.sh` for installing the necessary binary, but when I try to source `nvm.sh` in `setup.sh` it completely does not work. Here is my error output for reference:

    INFO:root:Running setup module start (cwd=/home/vagrant/FrameworkBenchmarks/frameworks/JavaScript/nodejs)
    INFO:root:To run nodejs manually, copy/paste this:
          export FWROOT=/home/vagrant/FrameworkBenchmarks && \
          export TROOT=/home/vagrant/FrameworkBenchmarks/frameworks/JavaScript/nodejs && \
[...]

Issue #995, opened 2014-08-11
Create isolated environment for each framework


hamiltont:
As mentioned in #966, most languages with package systems also have tools to isolate environments (e.g. environment A has packages foo-v1,bar-v1.2,zlib-v0.6 while environment B has foo-v3,bar-v4.4,zlib-v0.6). We should use this! The first step is to separate all the framework-specific requirements out into their respective directories. @methane is making huge strides for python over in #984, @jberger basically finished this for perl with #986. This leaves languages like scala, php, etc, many of which already have some of their dependencies declared inside each framework folder. 

The next step is to choose an environment isolation framework for each language. In #966 we discuss this step for Perl. @jberger prefers Carton, but also mentions local::lib and some other stuff I don't know (perlbrew, plenv). Once we settle on an environment manager for that language, we can apply it to all the tests. 

~~Personally I think we should delay this until #448 is completed as it will become much cleaner to implement with a language-based directory structure (e.g. stuff like `frameworks/php/...` and `frameworks/perl/...`)~~ Done

msmith-techempower:
This is now the overarching design of the tests, and while probably still not perfect it is essentially in the master branch.

PR #1518, opened 2015-04-10
Koa [#1425] and Node versioning (nvm)

ebramanti-techempower:
I've implemented the Koa framework as discussed in #1425. However, this brought forth a bigger problem: A lack of versioning for Node.js (depending on framework). Koa has a requirement of `>=v0.11` (needs ES6 Harmony features) for Node.js, however incrementing Node manually broke all the other frameworks.

The most ideal thing is to update each and every framework to work with the current latest version of Node. With `nvm` now used in the Node.js frameworks, this provides a way for quick Node versioning. Creating isolated environments for each framework is something that has been discussed in #995, and this is an important step forward for good versioning amongst the Node frameworks.

I'm going to continue work on getting these frameworks updated to `0.12.x`. In the meantime, I wanted to push the updated versioning so people can continue working on new frameworks (believe @zane-techempower is about to start work on Sails.js).


ebramanti-techempower:
Need to do some more cleanup.

frameworks/JavaScript/hapi/app.js
```diff
@@ -8,8 +8,8 @@ var cluster = require('cluster'),
 	Hapi = require('hapi'),
 	mongoose = require('mongoose'),
 	async = require('async'),
-	conn = mongoose.connect('mongodb://localhost/hello_world'),
-	connMap = { user: 'benchmarkdbuser', password: 'benchmarkdbpass', database: 'hello_world', host: 'localhost', charset: 'utf8' };
+	conn = mongoose.connect('mongodb://127.0.0.1/hello_world'),
+	connMap = { user: 'benchmarkdbuser', password: 'benchmarkdbpass', database: 'hello_world', host: '127.0.0.1', charset: 'utf8' };
```
LadyMozzarella: Think you accidentally committed the 127.0.0.1 change. Switch back to localhost.

frameworks/JavaScript/koa/bash_profile.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+export NODE_HOME=${IROOT}/node-v0.12.0-linux-x64
```
LadyMozzarella: We've removed the need for bash_profile.sh from TFB. If you could rebase and move NODE_HOME to setup that would be helpful.

frameworks/JavaScript/nodejs/hello.js
```diff
@@ -19,9 +19,9 @@ var http = require('http')
   , url = require('url')
   , async = require('async')
   , mongoose = require('mongoose')
-  , conn = mongoose.connect('mongodb://localhost/hello_world')
+  , conn = mongoose.connect('mongodb://127.0.0.1/hello_world')
```
LadyMozzarella: This file looks like `localhost` was replaced with `127.0.0.1` accidentally as well.

frameworks/JavaScript/koa/setup.sh
```diff
@@ -0,0 +1,34 @@
...
+# ${NODE_HOME}/bin/node --harmony app.js &
+
+# !DO NOT REMOVE!
+#
+# It takes `node app` a few seconds to turn on and 
+# then fork. If you remove this sleep, the parent shell 
+# executing this script will be terminated before the 
+# application has time to awaken and be forked, and 
+# express will fail to be started
+sleep 5
```
LadyMozzarella: I believe we're already sleeping 60 seconds before starting each framework. Is this necessary?

frameworks/JavaScript/koa/setup.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# run app
+npm install
+node --harmony app &
+
+# export NODE_HOME=${IROOT}/node-v0.12.0-linux-x64
+# export PATH=$PATH:$NODE_HOME/bin
+
+# ${NODE_HOME}/bin/npm install
+# ${NODE_HOME}/bin/node --harmony app.js &
```
LadyMozzarella: Are all of these left commented out for a reason?

frameworks/JavaScript/koa/setup.sh
msmith-techempower: Pinging @hamiltont
I DO believe this is necessary, but I don't fully recall why.

hamiltont: Now that #1505 is merged, this is no longer necessary. Explanation: 

The critical change was switching from `subprocess.check_call` to `subprocess.Popen` - the former will return when the subprocess is finished, while the latter will not. Without this sleep command, `setup.sh` would occasionally exit before the forking (e.g. `node --harmony app &`) was complete, python's subprocess would take care of cleaning up process resources, and therefore the app would never start. Now that we've switched to a non-terminating method of calling setup.sh from python, there is no need to sleep before returning.

frameworks/JavaScript/koa/setup.sh
ebramanti-techempower: Ah, I missed a spot! Good catch!

frameworks/JavaScript/koa/setup.sh
ebramanti-techempower: I'll get this update done for all the Node frameworks, seems that this is duplicated.

hamiltont: Thanks!

ebramanti-techempower:
@msmith-techempower Do you want me to switch the `fw_depends` to `nvm` to be like `rvm.sh`, or is it OK how it is? Think we were talking about that over [here](https://github.com/TechEmpower/FrameworkBenchmarks/issues/1425#issuecomment-91645618).

ebramanti-techempower:
I've updated the pull request with the requested changes!

ebramanti-techempower:
So it looks like Express is failing on Travis, but this seems to be a module problem (not a problem with the Node version). Here is what I see in the log:

    Setup express: npm ERR! Error: No compatible version found: brace-expansion@'^1.0.0'

I'll fix this issue, as well as update the other Node frameworks, after this PR is merged to `master`.

ebramanti-techempower:
Last commit should fix the Express issue.

hamiltont:
> Last commit should fix the Express issue

Awesome, let's see what it says. For what it's worth, express on Travis-CI seems to succeed about 50% of the time. It's one of the ones that is not deterministic for some reason. My (total noob to npm) guess is that the dependencies are specified without a fixed version number e.g. brace-expansion>1.0 is ok vs a fixed e.g. brace-expansion 1.0.3.5a is required. I've seen something similar with python frameworks use a `requiremetns.txt` file without first issuing a `pip freeze` command to hardcode the requirement numbers

frameworks/JavaScript/nodejs/hello.js
```diff
@@ -29,12 +29,12 @@ if (!windows) {
     
     Mapper.connect(connMap, {verbose: false, strict: false});
     var World = Mapper.map("World", "id", "randomNumber")
-    libmysql.connectSync('localhost', 'benchmarkdbuser', 'benchmarkdbpass', 'hello_world');
+    libmysql.connectSync('127.0.0.1', 'benchmarkdbuser', 'benchmarkdbpass', 'hello_world');
```
LadyMozzarella: The `127.0.0.1`s snuck in again.

ebramanti-techempower:
@hamiltont So the test passed for Express, I believe it is actually an issue with npm being out of date. What I now do is run `npm install -g npm` before running `npm install`, and this seems to have fixed the issue. Tests are all green, and I'm going to implement a fix for that line @LadyMozzarella caught.

hamiltont:
> seems to have fixed the issue

Great, would love to see express start greenlighting continuously

ebramanti-techempower:
@hamiltont Now my real work begins: updating Node tests to v0.12.x. I'll post this as an issue just so people know I'm working on it!

Issue #999, opened 2014-08-11
Add expert contacts to test README's


hamiltont:
For a lot of the minor "improve framework Foo" pull requests that tweak a framework slightly, e..g add final to the variables, techempower can't know if this is actually an improvement or not without consulting an expert. We should ask experts that show up on our project somewhere if they are willing to allow us to occasionally ping them for advice. If they agree, we can add their github username to the README for that project and then ping them if we need help looking at a pull request

LadyMozzarella:
Closing. I have added suggestions to add experts in the Ruby and Python Readmes (that don't already have experts listed) as examples for other contributors. Also, the [documentation now has some basic outline ideas for the Readmes](http://frameworkbenchmarks.readthedocs.org/en/latest/Development/Readme-Formats/), which include suggestions to add expert contacts.

Issue #977, opened 2014-08-08
Proposal: Push discussion from issues/pull requests into test README files


hamiltont:
There are a number of comment threads for some situations, like the aspnet thread management stuff.  We should potentially start to encourage people to address these tricky issues in their own readme files once the relevant discussion has reached a consensus. This might help to prevent further questions on tricky situations that have already been answered, allowing more time for threads discussing if the current solution is the right one

Issue #958, opened 2014-08-06
Store links to interesting discussion somewhere in readme


hamiltont:
All the data available here is worthy of a *lot* of discussion. There are a lot of good posts already, perhaps we should start keeping a list of discussions somewhere in a README. This would have really helped me get a grip on some of the more subtle points. 

Here's a start: 
http://tiamat.tsotech.com/unfair-comparisons
http://www.reddit.com/domain/techempower.com/ (perhaps pick a few great threads?)
http://theholyjava.wordpress.com/2013/04/01/the-value-and-perils-of-performance-benchmarks-in-the-wake-of-techempowers-web-framework-benchmark/
https://news.ycombinator.com/item?id=5454775
https://news.ycombinator.com/item?id=7680242
http://www.techempower.com/blog/2014/05/01/framework-benchmarks-round-9/
http://spray.io/blog/2013-05-24-benchmarking-spray/

etc. 

Thoughts?

Issue #877, opened 2014-06-22
Nancy: What is Nancy.Hosting.Event2.dll? + Add discussion summary to README


phillip-haydon:
What is this host? 

https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/nancy/lib/Nancy.Hosting.Event2.dll

The benchmarks for any .NET stuff are pretty terrible so I'm trying to figure out what this junk is and see if we can get some better benchmarks in. 

Is there any source code? 

Also why does this exist? https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/nancy/src/Global.asax.cs

PR #1523, opened 2015-04-13
Add Information and Reorganize Ruby Readmes

LadyMozzarella:
I am wanting to add more information to READMEs for a few languages, so that other contributors have some references/examples for what could be useful to add to a Readme. All this does is refactor and add information to the Ruby language Readme and all of the Readmes for the Ruby frameworks.


hamiltont:
Linking in #999, #958, #977, #877 - feel free to close those as you see fit

Issue #1536, opened 2015-04-17
Remove ringojs-convenient


msmith-techempower:
It does not appear to have ever worked, it is extremely stale at this point, and it looks like the main file `ringo-main.js` does not actually attempt to implement anything.

Should we remove this since `ringojs` works?

grob:
what makes you think ringojs-convenient never worked? ringojs-conv is part of the benchmark since round 7 (same as ringojs), works and passed every round of the benchmark since up to the latest preview of round 10. neither is it (resp. the modules it uses) _extremely stale_ nor does `ringo-main.js` _implement anything_, so i don't see any reason to remove it (if something in ringojs-conv doesn't work i'd be happy to help or fix it).

msmith-techempower:
Must be my misunderstanding, then. I will review, thanks for the quick response.

msmith-techempower:
I am trying to convert all tests away from `install.sh` scripts and instead relying more on their `setup.sh` files in conjunction with `fw_depends`.

The specific error I'm seeing when trying to run ringojs-conv is the following:

```
INFO:root:Sleeping 15 seconds to ensure framework is ready
Server ringojs-convenient: Wrapped java.lang.ClassNotFoundException: com.mysql.jdbc.Driver
Server ringojs-convenient:      at ringo-sqlstore/lib/sqlstore/connection/pool.js:50 (anonymous)
Server ringojs-convenient:      at ringo-sqlstore/lib/sqlstore/connection/pool.js:186 (anonymous)
Server ringojs-convenient:      at ringo-sqlstore/lib/sqlstore/store.js:23 (anonymous)
Server ringojs-convenient:      at ringo-sqlstore/lib/sqlstore/store.js:51 (anonymous)
Server ringojs-convenient:      at /home/techempower/FrameworkBenchmarks/frameworks/JavaScript/ringojs-convenient/app/models.js:17
Server ringojs-convenient:      at /home/techempower/FrameworkBenchmarks/frameworks/JavaScript/ringojs-convenient/app/views.js:4
Server ringojs-convenient:      at /home/techempower/FrameworkBenchmarks/frameworks/JavaScript/ringojs-convenient/ringo-main.js:5
Server ringojs-convenient:
```

Any help would be great.

grob:
[...]

PR #1541, opened 2015-04-19
fixed setup of ringojs-convenient and upgraded to ringojs 0.11

grob:
fixes a setup issue with ringojs-convenient (see #1536)


msmith-techempower:
I have this all sorted out on `remove-installed-scripts` already, but I guess I didn't push it yet. Thanks

PR #1543, opened 2015-04-20
Php lumen

knewmanTE:
Adds tests for the new PHP framework [Lumen](http://lumen.laravel.com/).


hamiltont:
Another nice PR, thanks

knewmanTE:
Not sure why it's still failing. All the tests pass locally. My guess is that it's something in the nginx config, since that should be the only place where there are differences between local and Travis, but everything appears to be in order.

knewmanTE:
Moving `index.php` from `lumen/public/` to `lumen/` gives passing tests for json, db, queries, and updates. Fortunes can't seem to find the template file. If we can figure out what the template issue is, this will likely resolve the similar issue that's coming up in #1590.

mcocciaTE:
I got all the tests to pass locally but I had to mess with the vendor file that composer downloads to do so I am going to add to the setup script switching out the two files that need fixing. 

As for the template issue I had to hard code the path in because the $path variable which should be the path was always just the empty string. The template was being found by findInPaths in FileFindViewer.php except because the $paths array only had a single empty string value in it the check to see if the path + file combo was valid was always failing.

Let's see if travis likes what I did

mcocciaTE:
Lumen passed but somehow Moonshine failed to even start. So LGTM?

msmith-techempower:
EXACTLY!

PR #1544, opened 2015-04-20
Fix updates

knewmanTE:
It turns out some of my frameworks weren't properly updating the database when their update requests. This has been fixed.


hamiltont:
o.O wowo. Glad you caught that

knewmanTE:
Looks like the fails came from `Perl/plack` and `Perl/web-simple` (neither of which were touched with this Pull Request).

hamiltont:
LGTM, thanks!

PR #1547, opened 2015-04-21
update to tapestry 5.3.8 and java8 (5.4-beta-31 exists)

based2:
https://git-wip-us.apache.org/repos/asf?p=tapestry-5.git

### DIFF
--- a/frameworks/Java/tapestry/hello/build.gradle
+++ b/frameworks/Java/tapestry/hello/build.gradle
@@ -5,8 +5,8 @@ apply plugin: "java"
 apply plugin: "jetty"
 apply plugin: "idea"
 
-sourceCompatibility = "1.5"
-targetCompatibility = "1.5"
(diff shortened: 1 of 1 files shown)


hamiltont:
@based2 Can you also update [this file](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/tapestry/install.sh#L3) to depend on `java8` instead of `java7` if tapestry now relies on java 8 being installed? 

I'm guessing #1548 requires the same change based on the PR description, so just change that file in one PR, wait for travis to verify and us to merge, then the second PR can be checked against the updated dependency

hamiltont:
@msmith-techempower @knewmanTE if someone over there has a minute, these two PRs are ready-to-merge except for a single dependency declaration

msmith-techempower:
Okay, I see what's happening here; closing in favor of #1684

PR #1548, opened 2015-04-21
update to tapestry 5.3.8 and java8 (5.4-beta-31 exists)

based2:


hamiltont:
blocking on changes requested in #1547

zane-techempower:
Looks like this is close, but not quite there. Seems like there is a disagreement between jdk version and target release, in this case it seems like the target is 1.8 i.e. Java8. However the process is using Java7 to try to compile the code.

See: http://stackoverflow.com/questions/28291691/fatal-error-compiling-invalid-target-release-1-8-help-1

msmith-techempower:
Will require `install.sh` to be updated to `fw_depends java8` from java7.

mcocciaTE:
I can get everything to compile when I change source $IROOT/java7.installed it java8 in the setup.sh but it causes all of the tests to give a generic server error back. I'm looking more into why.

mcocciaTE:
I believe the version of maven we use has an issue with java8. If we updated maven to 3.1.1+ it should work.

mcocciaTE:
Updating maven did not seem to work. Poking around more.

based2:
https://maven.apache.org/docs/3.3.3/release-notes.html
https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12316922&version=12332054

mcocciaTE:
So maven ended up working but redis was not. #1684 is my fixes which get everything happy.

PR #1550, opened 2015-04-21
Disable logging in Laravel

bratao:
We get an extra performance if we disable error logging


hamiltont:
> disable error logging

Not sure this is a good idea. Most current frameworks leave the error logging enabled - there's almost zero penalty (for most implementations) if there is nothing to send to the error log, and it's incredibly useful to see what is going wrong if the framework *does* start dumping non-200 responses

bratao:
@hamiltont , That make sense. I just applied some optimization that I used in a real website. Usually bots try some non-existing resources and this was slowing down my system.

PR #1551, opened 2015-04-22
Python TurboGears Template Assignment

knewmanTE:
fixes a bug where TurboGears Fortune template was being initialized multiple times.


knewmanTE:
Turbogears is passing fine. Seems to be another case of other tests randomly failing.

hamiltont:
For the push my travis code couldn't figure out what directories you had modified so it re-tested every framework (which will continue to 100% fail until every framework greenlights in travis :P )

PR #1552, opened 2015-04-22
Python Klein framework

knewmanTE:
https://github.com/twisted/klein


frameworks/Python/klein/app.py
```diff
@@ -0,0 +1,123 @@
...
+	fortunes = db_session.query(Fortune).all()
+	fortunes.append(Fortune(id=0, message="Additional fortune added at request time."))
+	fortunes.sort(key=attrgetter("message"))
+	for f in fortunes:
+		f.message = bleach.clean(f.message)
+	template = env.get_template("fortunes.html")
+	return template.render(fortunes=fortunes)
+
+if __name__ == "__main__":
+    app.run("localhost", 8080)
```
hamiltont: `0.0.0.0` may be better than `localhost` here, in the event of multiple NICs

hamiltont:
Very cool, it's going to be good to include something based on twisted

frameworks/Python/klein/app.py
knewmanTE: I changed `localhost` to `0.0.0.0`, bit it seems to be failing now.

hamiltont:
Thanks!

PR #1557, opened 2015-04-22
Scala http4s

mfirry:
For now it only provides the json test

### DIFF
--- a/frameworks/Scala/http4s/.gitignore
+++ b/frameworks/Scala/http4s/.gitignore
@@ -1,0 +1,2 @@
+target
+projet/target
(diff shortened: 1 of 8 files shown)


hamiltont:
Please take a quick look at the [guide for adding frameworks](https://frameworkbenchmarks.readthedocs.org/en/latest/Development/Add-Frameworks-or-Tests/) - this PR needs to have an `install.sh`, `setup.sh`, and `benchmark_config.json` file added

hamiltont:
Nice, looking better. Feel free to reopen when you're ready. Perhaps expand the Readme a bit to link to the parent project - here's a [template of how other frameworks structure the readme](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Scala/lift-stateless/README.md). 

If you're willing to be pinged via your github username if there is a proposed modification that we cannot evaluate, drop your github handle in the readme under something like a maintainers section so we know where to find you.

PR #1561, opened 2015-04-22
Scala http4s

mfirry:
For additional info on the framework: http://http4s.org


hamiltont:
Oh - moving comment here: 

Nice, looking better. Feel free to reopen when you're ready. Perhaps expand the Readme a bit to link to the parent project - here's a template of how other frameworks structure the readme.

If you're willing to be pinged via your github username if there is a proposed modification that we cannot evaluate, drop your github handle in the readme under something like a maintainers section so we know where to find you.

mfirry:
Great. Will add what's missing. 
One more thing: I just put the execution commands on the readme, is that enough?

hamiltont:
> I just put the execution commands on the readme, is that enough?

Definitely include a link to the parent project to 100% identify what the target is. Looks like that's https://github.com/http4s/http4s ? Other than that, your readme looks fine to me

mfirry:
How about now?

mfirry:
Any hints on how I can fix this CI issue?

hamiltont:
Readme looks good! To overcome the exception, add keys for `orm` and `database_os` to your `benchmark_config.json`

hamiltont:
@mfirry Doh! I forgot to tell you to add a line to the `.travis.yml` so that http4s would actually be picked up by travis-CI and tested Can you please add that in?

mfirry:
At last :) I'm gonna fix the `Required response header missing` soon

hamiltont:
LGTM! PS - You are not required to set `charset` as part of your content-type header, and doing so means you are sending a few more bytes than are required across the network with each request. If http4s is a top contender then the network will become your main bottleneck and these extra bytes will limit your top requests/sec performance. 

I'm happy to merge this now, or you can add one more commit to drop the charset declaration, let me know which way you wanna go! (This is what those 'WARN' results are trying to tell you, we need a better message as the current output is misleading - your headers are valid, just too verbose!)

mfirry:
There you go. Working on the `Server` header. 
Will probably push that next week though. 
Also working on other Scala goodies (mainly [akka-http](http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0-RC1/scala/http/index.html) and [tumblr colossus](http://tumblr.github.io/colossus/)).
After that I'd love to add the db tests, but that'll take me a bit more.
Thanks for your time!

hamiltont:
LGTM - While I'd love to have db tests implemented at some point in the future, I'm very glad to at least have the simpler tests setup and working, thanks!

mfirry:
One last question: `http4s` works with both [blaze](https://github.com/http4s/blaze) and jetty. I can provide both implementation and I think it'll be interesting to see both performances. Do I file another PR? I'd basically provide `htt4s-blaze` and `http4s-jetty` both inside the `http4s` folder. Is it ok? When is round 11 due?

hamiltont:
>  I'd basically provide htt4s-blaze and http4s-jetty both inside the http4s folder. Is it ok?

Perfectly fine. We call that two "configurations" of the same "framework". Update the `benchmark_config.json`, leaving one of the configurations as "default" (so that when someone says `--test http4s` we know which configuration to test by default). You'll want to use a different value for the `setup` key so that it has a different setup file. If all the files for blaze/jetty are the same and the only difference is how you turn it on, then your two setup files can coexist at the current http4s root folder. If they require an entirely separate codebase, feel free to create subfolders e.g. `http4s/blaze/<your blaze code here>`

> round 11

@bhauer is shooting for some time in june IIRC

mfirry:
Ok. Check this out and let me know if i can PR again https://github.com/mfirry/FrameworkBenchmarks/tree/master/frameworks/Scala/http4s

Issue #1534, opened 2015-04-16
Header verification throwing unnecessary warnings and exceptions


zane-techempower:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/fcd51040a1ddf06046d69ed16c39c90c825e533d/toolset/benchmark/test_types/query_type.py#L105-L107

Some frameworks use a `Content-Type: application/json; charset=utf-8` header, combining the charset and content type. This will throw a warn as it doesn't match `Content-type: application/json` exactly.

msmith-techempower:
This is a correct warning - we want to enforce equal footing for all tests, and if TestX is sending `Content-Type: application/json; charset=utf-8` instead of `Content-Type: application/json`, then they are sending extra bytes, which will affect their throughput negatively.

This is a warning because some frameworks might not be able to control this, but it should be noted that this **CAN** affect the results.

hamiltont:
@msmith-techempower I'd vote to restate the warning though - instead of the current "Missing required headers", seems we could instead say "Returning X bytes more than required for header 'Content-Type', this may affect your test performance" - thoughts? 

To me, the more user-friendly we can make the verification warnings the more framework devs can do their work withou having to consult the TFB team for guidance

msmith-techempower:
@hamiltont Agreed - the wording could use a change.

zane-techempower:
I think that section of the test can be improved. Another helpful feature would be to state specifically which header is missing rather than the general statement + warn that the code I specified is currently doing.

Letting devs know that the warn is for doing something that is likely reducing performace is definitely a user-friendly move.

hamiltont:
Agreed on both - let's reopen this as an Enhancement request

hamiltont:
Same concern with `Content-Type: text/plain; charset=UTF-8` and friends. header is correct but longer than required

hamiltont:
Also, here's one more stack trace for verifying headers. Different problem, same big of relevant code

```
Server play2-scala-activate: [WARN] [04/25/2015 09:12:31.033] [New I/O worker #1] [EventStream(akka://play)] [akka.event-handlers] config is deprecated, use [akka.loggers]
  Response (trimmed to 40 bytes): "{"id":1792,"randomNumber":7486}"
WARNING:root:Verifying test db for play2-scala-activate caused an exception: 'NoneType' object has no attribute 'lower'
   FAIL for http://127.0.0.1:9000
     Caused Exception in TFB
[...]

PR #1566, opened 2015-04-23
Scala play2 - added plaintext test

andreaferretti:
Scala play2 - added plaintext test


hamiltont:
oh, this needs manual testing as it took >50 minutes

hamiltont:
LGTM. Note that it appears to fully pass but is currently affected by a bug in our verification code #1534. Once that bug is cleared up this should go from red to green without any further changes to the scala play2 codebase

PR #1569, opened 2015-04-24
Updating vert.x to the latest release 2.1.5

victorbriz:
Maintainer of vert.x seems to like to have vert.x back in FrameworkBenchmarks, maybe for Round 11?
https://groups.google.com/forum/#!msg/vertx/kism9YHbpTo/zSwWhs952dwJ
It is also important for us, developers who is using vert.x to have it in FrameworkBenchmarks.

### DIFF
--- a/frameworks/Java/vertx/README.md
+++ b/frameworks/Java/vertx/README.md
@@ -12,7 +12,7 @@ This is the vertx portion of a [benchmarking test suite](../) comparing a variet
 ## Versions
 
 * [Java OpenJDK 1.7.0_09](http://openjdk.java.net/)
-* [vertx 1.3.1](http://vertx.io/)
+* [vertx 2.1.5](http://vertx.io/)
 
 
 ## Test URLs
--- a/frameworks/Java/vertx/WebServer.java
+++ b/frameworks/Java/vertx/WebServer.java
@@ -17,10 +17,33 @@ import java.util.concurrent.ThreadLocalRandom;
 
 public class WebServer extends Verticle implements Handler<HttpServerRequest> {
 
-  private Buffer helloWorldBuffer = new Buffer("Hello, World!");
-  private String helloWorldContentLength = String.valueOf(helloWorldBuffer.length());
-  private DateFormat DATE_FORMAT = new SimpleDateFormat("EEE, dd MMM yyyyy HH:mm:ss z");
+  private final Buffer helloWorldBuffer = new Buffer("Hello, World!");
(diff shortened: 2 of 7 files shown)


hamiltont:
Looks great! 

While we are here, can I ask you to move all of the code from `toolset/setup/linux/frameworks/vertx.sh` to `frameworks/Java/vertx/install.sh`, underneath the line for `fw_depends`

For various reasons, we are trying to be really restrictive about what goes under `toolset/`, so this change would move the installation process for vert.x into the specific install.sh file for vert.x. A copy/paste should be all that's required, with the final install.sh looking like the code below and vertx.sh being deleted

```
#!/bin/bash

fw_depends java7 

RETCODE=$(fw_exists ${IROOT}/vert.x-2.1.5.installed)
[ ! "$RETCODE" == 0 ] || { return 0; }

fw_get http://dl.bintray.com/vertx/downloads/vert.x-2.1.5.tar.gz?direct=true -O vert.x-2.1.5.tar.gz
fw_untar vert.x-2.1.5.tar.gz

touch ${IROOT}/vert.x-2.1.5.installed
```

hamiltont:
By the way, take a look at the verification results from Travis-CI. Everything passes, but there are a few places you might be shooting your performance in the foot a bit with errors like this: 

```
  Response (trimmed to 40 bytes): "[{"id":8909.0,"randomNumber":3396.0},{"i"
   WARN for http://127.0.0.1:8080/queries?queries=2
     Response key 'id' contains extra quotations or decimal points.
             This may negatively affect performance during benchmarking
```

It's telling you that `8909.0` could be `8909` (e.g. saving two bytes) and still be considered a valid response

victorbriz:
Done and done.

hamiltont:
thanks!

purplefox:
I would not like to see any results for Vert.x published until we have had time to do a proper submission based on Vert.x 3.

I discussed this with @bhauer some time back, so please don't publish this.

purplefox:
BTW @victorbriz - I don't mean to rain on your parade - but if you're interested in helping get benchmarks together for Vert.x 3 I think that would be a better use of time.

I can help you optimise too. wdyt?

martin-g:
See #1697

bhauer:
Vert.x will be marked as "Removed at request of maintainer - outdated version" in the final for Round 11.  It may appear in previews, but we will continue to mark it as such in the final results for each round until it's updated and @purplefox signs off.

PR #1571, opened 2015-04-24
Run optimize for laravel

bratao:


hamiltont:
For the uninitiate, what's the difference between composer's `--optimize-autoloader` and `php artisan optimize --force`?

bratao:
Internally it will call Composer with the optimize flag, but in addition it will create bootstrap/compiled.php , that is a 'pre-compiled' file with common classes.

hamiltont:
good to know

hamiltont:
@bratao Would this be useful for the other PHP-based tests too? If yes, how do you feel about sending me a PR for that? :-)

PR #1563, opened 2015-04-22
Update nodejs to v0.10.38 and fix warnings

mhart:
Fixes #1520 and all other warnings regarding things like the `Server` header.

Also updates all module dependencies to their latest versions and removes some duplicate code.

Issue #1546, opened 2015-04-21
Javascript: add io.js to Node tests


ebramanti-techempower:
We've been getting a bunch of requests in #1537 to add io.js tests to the Javascript frameworks to see if there is any performance difference. As soon as #1537 is closed, I'll begin work on this.

branning:
The decision to merge io.js and node.js and form the Node Foundation elminates the need to do this.

Sinewyk:
How can we help if we want the next benchmark to run on node v4 (being the LTS) and node v5 (stable) ?

asilvas:
+1 for using latest stable. This benchmark has nothing to do with LTS releases, so should be more than reasonable to use latest release.

nbrady-techempower:
We are up to v5, should be v6 soon!

Sinewyk:
I still see 0.12 here https://github.com/TechEmpower/FrameworkBenchmarks/blob/bfc1f5088798d561e71e95d23d0467b1974886c7/toolset/setup/linux/languages/nodejs.sh#L8.

You can probably go straight up to 6.2.2, it's going in LTS in October.

nbrady-techempower:
@Sinewyk the current node version for round 14 is in the round-14 branch. Feel free to submit a pull request to the round-14 branch for further upgrades!

Issue #1537, opened 2015-04-17
Nodejs: Update all tests to v0.12.x


ebramanti-techempower:
Now that my addition of `nvm` for versioning has been merged into master, I want to begin work on updating the Node tests to the most recent version 0.12.x (even Koa is only running on 0.11.x, which I just implemented). The MySQL driver, [node-mysql-libmysqlclient](https://github.com/Sannis/node-mysql-libmysqlclient), might need to be updated as well. It has very particular compatibility requirements, so the tests might have to be switched over to a new driver (might pertain/fix #1519 & #1520, but not exactly sure yet).

Either way, wanted to open this issue since all of the Node frameworks will be audited/updated.

ebramanti-techempower:
Also, I'll work on updating the warns since the Node frameworks have not been updated with the necessary response headers.

mhart:
Would be great to add io.js to the mix too – it's a more modern, up-to-date fork of Node.js and should be pretty much a drop-in replacement

https://iojs.org

ebramanti-techempower:
@mhart I've been keeping up with io.js, it will be much easier to figure out a solution for this now that we are employing `nvm`. For the moment, these tests **need** to be updated before we consider adding this as an option.

mhart:
Funny you say that – just playing with it on my machine now :+1:

ebramanti-techempower:
@mhart Sweet! What frameworks are you using it with? Right now I'm updating Hapi.js from 1.2 to 8.4, however since io.js is a fork of Node this shouldn't affect it working in io.

rayshan:
May be good to test io.js with this patch or wait for it to land:
https://github.com/iojs/io.js/pull/933

ebramanti-techempower:
I can open an issue to add io.js to the JS framework benchmarks! Thanks for your input!

mhart:
I updated the nodejs benchmark for Node.js v0.10 at least, here: #1563 

The basic JSON/text tests work fine under v0.12 and io.js, and are certainly more stable (not sure how much faster) – but the mysql dependency won't build at all under those versions, and I'm not familiar enough with that code to figure out an alternative.

ebramanti-techempower:
This is done! Onwards to io.js support!

PR #1572, opened 2015-04-24
Node.js Mega-Update: Version/framework/dep bump, ORM change, added functionality

ebramanti-techempower:
This update is a continuation of the `nvm` support I added in my last PR (#1518) and to address issue #1537. The changes are listed below:

* **Bump all Node frameworks to v0.12.x of Node**: All frameworks now run on the latest version of Node from time of writing (0.12.2).
* **Update all Node frameworks to latest**: Hapi.js was found to be running v1.2, and the current version was 8.4. Express was an entire version behind as well. This required modification of the code to meet the new standards of the current versions.
* **Update all Node dependencies to latest**: I updated all `npm` packages to their latest versions. This was mostly straightforward, except for the ORM. Which leads to the NEXT bullet...
* **Switch to a new ORM ([Sequelize](https://github.com/sequelize/sequelize))**: This is one of the biggest changes in this PR. Before, we were using Mapper and [mysql-libmysqlclient](https://github.com/Sannis/node-mysql-libmysqlclient). Neither have had recent updates, both have conflicts with Windows, and mysql-libmysqlclient does not even support 0.12.x. Therefore, I switched all tests using a MySQL ORM over to Sequelize.
* **Added to stock Node.js tests**: I added a couple of `/update` routes for the ORMs in the nodejs test In general, `nodejs` needed a HUGE cleanup anyways. I integrated one of @mhart's changes from #1563 as well, and I'd like to thank him for his work and suggestions.
* **Add missing route to Express/Hapi**: No `/plaintext` in either Hapi or Express.

Once this is merged into `master`, I'll work on addressing issue #1546. These need to be integrated, since the ORM change should pave the way for good io.js support.


ebramanti-techempower:
Fixed a test that was unable to connect to mongo, but I'm going to add a couple more cleanup commits before I merge this.

msmith-techempower:
@ebramanti-techempower @hamiltont This change seems to require some magic from `nvm` that is incompatible with the new "remove install.sh" branch.

Basically, it seems like it relies on executing `nvm.sh` which is executed via the user's `.bashrc` file, but this causes problems when we start going deeper in each shell. Specifically, if we rely on that in my branch, it simply fails to successfully load that script.

msmith-techempower:
Additionally, with `nodejs` and `io.js` being merged (at some point) and nothing is actually USING `io.js`, I'm inclined to roll back to the singular `node.js` build (that worked on my branch).

PR #1574, opened 2015-04-24
Fix relative path name in create_database.py

mcdonc:
The relative path name of create-postgres.sql was for an earlier layout of the repository.


hamiltont:
Good catch! Could you use `os.environ['FWROOT'] + /config/create-postgres.sql` instead of the relative path?

mcdonc:
That should do it, thanks for the tip.

PR #1575, opened 2015-04-24
Speed up plaintext view rendering

mcdonc:
Shave some function calls off the time required to do plaintext rendering.  I don't think this makes the view "less realistic", it's just a faster way to do the same thing.


hamiltont:
Thanks for the pyramid updates @mcdonc!

mcdonc:
Thanks for the merges!

hamiltont:
:thumbsup:

PR #1577, opened 2015-04-25
io.js support for Node frameworks

ebramanti-techempower:
This is a continuation from Node Mega-Update (#1572) to add io.js support into the Node.js frameworks (#1546). 

In each Node.js test, we now have two setup scripts: `setup_node.sh` and `setup_iojs.sh`, which allows for the benchmark configuration to set up each accordingly. I added to each benchmark configuration all of the normal benchmarks with an `-iojs` on the end. Therefore, every benchmark run in Node.js is now also run in io.js. This will allow us in the next round to see which frameworks run faster on which platform.

On a side note, it's cool not to have to use the `--harmony` flag to run Koa, since io.js has so many neat ES6 features already added.


frameworks/JavaScript/express/benchmark_config.json
```diff
@@ -60,6 +60,66 @@
...
+      "classification": "Micro",
+      "database": "MySQL",
+      "framework": "express",
+      "language": "JavaScript",
+      "orm": "Full",
+      "platform": "iojs",
+      "webserver": "None",
+      "os": "Linux",
+      "database_os": "Linux",
+      "display_name": "express",
```
hamiltont: Should we set these as "Express on io.js" to distinguish them?

frameworks/JavaScript/express/benchmark_config.json
```diff
@@ -60,6 +60,66 @@
...
+      "notes": "",
+      "versus": "node"
```
hamiltont: Perhaps set the versus to blank until we have a vanilla io.js implementation?

frameworks/JavaScript/express/benchmark_config.json
msmith-techempower: No, we render the platform on the results website already - let's leave this as "express".

frameworks/JavaScript/express/benchmark_config.json
msmith-techempower: Agreed

ebramanti-techempower: Implementing now.

ebramanti-techempower:
I've updated the benchmark configurations. In my last Travis build, I was only running into an issue with the Node.js test. I updated one of the queries that I believe has been responsible for this issue. Other than that, all tests were green in the last run, so I guess we just have to wait on that change.

ebramanti-techempower:
@hamiltont Can you re-run my 2 failing jobs on my last build? MongoDB is giving the tests trouble again.

hamiltont:
reran, same issue. Mongo is so touchy, and we have to jump thru some hoops to get the version we want on travis-ci. Don't have time to dig into this deeper at the moment - can anyone over there take point on figuring out mongo's issue?

jadengore:
We definitely should hold off on this with all the nodejs/iojs convergence news.

brendanashworth:
FWIW, io.js releases will continue for the foreseeable future until both repositories can be merged and agree on all the differences (there are many). If it isn't too big a deal to add it and later remove it, I don't see a reason why it shouldn't be added for now.

jadengore:
@brendanashworth Cool, this PR is ready to go, it's just Mongo on Travis giving issues. Maybe ping @hamiltont one more time to see if these will run?

hamiltont:
@ebramanti-techempower @msmith-techempower This looks mergeable to me. I agree with @brendanashworth that any convergence will take a while (and it's easy for us to remove this anyway). 

If you guys are on the same page let's merge it in

If anyone is interested, [here](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/run-ci.py#L341) is where I configure mongo in travis ci. It's basic stuff, but I'm not the most experienced with mongodb. Maybe someone can figure out if there is a way to avoid all our mongo issues. I'll add some more details in mike's issue

msmith-techempower:
Can `io.js` be installed without `nvm`? My branch which removes `install.sh` files does not seem to work with `nvm` and I have not yet figured out why.

martin-g:
Since io.js and node.js will be merged soon (http://www.infoq.com/news/2015/05/nodejs-iojs) is there a need to add support for both now?

msmith-techempower:
That's what I asked in the other thread. I suspect that this will be a slow-going merge, but if they are going to converge to a single instance anyway, I'm inclined to roll back the `nvm` work in favor of vanilla `node.js` (which will have `io.js` merged in some meaningful way down the line).

jadengore:
@msmith-techempower Could it have something to do with `fw_depends nvm` being in the [`install.sh`](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/JavaScript/express/install.sh) script? Is that automatically run on your branch, or does this `fw_depends` need to be moved? Could be that `nvm` never installs.

msmith-techempower:
@jadengore `install.sh` has been removed, and `fw_depends` has been updated to be run from `setup.sh`. Specifically, `nvm` is installed, but fails to run correctly because of its reliance on being injected into the user's `.bashrc`.

msmith-techempower:
I'm inclined to close this without merging since `io.js` is being merged into `nodejs`.

PR #1579, opened 2015-04-25
Fix setup script calling bugs

hamiltont:
I suspect these bugs were causing #1566 to fail in Travis-CI


hamiltont:
Compared to the last full build on master I could find, this is only an improvement. Also allowed scala-play2 to be verified successfully in travis instead of erroring out.

PR #1076, opened 2014-09-01
Added PHP Framework CCF 2 

mario-deluna:
This PR adds the benchmarks for CCF2.

https://github.com/ClanCats/Framework

I hope everything is correct.

PR #1590, opened 2015-04-28
Php ccf2

knewmanTE:
Brings @mario-deluna 's work on [ClanCats Framework](http://www.clancats.io/)  (see #1076) up to speed with the changes we've made to TFB's structure. Also adds in the missing /plaintext and /updates tests.


mario-deluna:
Hi, thanks a lot that you took the time updating this. The travis tests seem to fail, i will also  take a look on this.

knewmanTE:
I think this might be related to the issues going on with [Lumen](https://github.com/TechEmpower/FrameworkBenchmarks/pull/1543). Both are PHP frameworks where the index file isn't in the root directory. They are both passing locally but can't locate any of the files once nginx has started on Travis. I'm experimenting on Lumen with what happens when I move the index file to the root directory.

mcocciaTE:
I can't seem to get this to pass at all locally. I also can't find an error log or see any errors in the out file. All of the tests keep returning a 404 page for me. I'm looking into why it might not be working for me.

mcocciaTE:
I found that in the setup "sed -i 's|root .*/FrameworkBenchmarks|root '"${TROOT}"'|g' deploy/nginx.conf" will always add onto the root so running the test twice in a row without checking out the nginx.conf file will result in all 404 errors. The reason it was failing for me was because it would have clancats/clancats/ even on the first run which isn't the right path.

I'm not sure how to change the command to make it not keep replacing after the first run but I believe my change will get travis to pass and will work for the benchmarks. We can just include in the setup to replace the file before every run but I'm not sure which solution will be the better one

frameworks/PHP/clancats/setup.sh
```diff
@@ -0,0 +1,11 @@
+export PHP_HOME=${IROOT}/php-5.5.17
+export COMPOSER_HOME=${IROOT}/php-composer
+export PHP_FPM=${PHP_HOME}/sbin/php-fpm
+export NGINX_HOME=${IROOT}/nginx
+
+sed -i 's|localhost|'"${DBHOST}"'|g' index.php
+sed -i 's|root .*/FrameworkBenchmarks|root '"${TROOT}"'|g' deploy/nginx.conf
```
msmith-techempower: A better fix for this would probably be: `sed -i 's|root /home/ubuntu/FrameworkBenchmarks/;|root '"${TROOT}"'|g' index.php

This way, the value present in your `nginx.conf` is matched directly (no wildcard) and replaced the first time while the second time you run the script, it simply won't find that direct match and sed will replace nothing on this line.

mcocciaTE: For some reason the /; was causing it never to match but other than that I made the change.

msmith-techempower:
What is the status here? It looks like the latest travis build worked for `clancats`; is this ready to be merged?

mcocciaTE:
Should be good to merge.

PR #1593, opened 2015-05-03
Updated to Play 2.3.8, updated dependencies/plugins and changed to use Java 8

donovanmuller:


richdougherty:
This change looks good to me. Thanks @donovanmuller!

hamiltont:
another good one guys, thanks!

hamiltont:
ps - hopefully we can avoid adding more to the play2  folder - it's starting to time out on travis-ci (it only gets 50 minutes of wall time, and during their busy times that definitely doesn't mean 50 minutes cpu)

PR #1595, opened 2015-05-04
Update framework version and support more tests

jamming:
Update framework version and support more tests (with different backends and different databases).


frameworks/Java/sabina/jetty-mongodb.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+./setup.sh -Dsabina.backend=jetty -Dsabina.benchmark.repository=mongodb
```
hamiltont: ping @msmith-techempower - FYI, you'll notice that the `&` is no longer required with the new way we call the setup files. Cool, huh? :-)

frameworks/Java/sabina/src/main/java/sabina/benchmark/MongoDbRepository.java
```diff
@@ -0,0 +1,122 @@
...
+        final String DATABASE = settings.getProperty ("mongodb.database");
+        final String WORLD = settings.getProperty ("mongodb.world.collection");
+        final String FORTUNE = settings.getProperty ("mongodb.fortune.collection");
+
+        MongoClient mongoClient = new MongoClient (HOST, PORT);
+        MongoDatabase db = mongoClient.getDatabase (DATABASE);
+        worldCollection = db.getCollection (WORLD);
+        fortuneCollection = db.getCollection (FORTUNE);
+
+        loadData ();
```
hamiltont: @jamming please remove this call and the associated `loadData` function + `FORTUNES` data. We already setup the databases for you.

frameworks/Java/sabina/src/test/java/sabina/benchmark/ApplicationTest.java
```diff
@@ -39,13 +38,13 @@
...
 
     private static final String ENDPOINT = "http://localhost:5050";
     private static final Gson GSON = new Gson ();
 
     @BeforeClass public static void setup () {
-        main (null);
+        Application.main (null);
     }
 
     @BeforeClass public void warm_up () throws IOException {
```
hamiltont: We also perform an initial warmup before we actually benchmark your application, this is probably redundant :-)

hamiltont:
looks good. Requested some minor changes, then we can merge

jamming:
Hi, I just made the updates you proposed. Thanks for the feedback... and kudos for the great work!

jamming:
I am not sure which is the correct place to ask this, so I'll try here (as in the issues there is no 'question' label):

In my fork's Travis build I'm getting this error:

    +timeout 15s sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 7F0CEB10

Do you know how to fix this?

Thanks again!

hamiltont:
> Do you know how to fix this?

Sadly, there is basically no way. You just have to keep restarting the build until that line passes. Travis-CI is known to have occasional network issues

frameworks/Java/sabina/src/main/java/sabina/benchmark/MongoDbRepository.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.*;
+import java.util.concurrent.ThreadLocalRandom;
+
+import com.mongodb.*;
+import com.mongodb.client.MongoCollection;
+import com.mongodb.client.MongoDatabase;
+import org.bson.Document;
+
+final class MongoDbRepository implements Repository {
+    private static final String [] FORTUNES = {
```
hamiltont: Please delete this variable too

frameworks/Java/sabina/src/main/java/sabina/benchmark/MongoDbRepository.java
```diff
@@ -0,0 +1,106 @@
...
+        return worlds;
+    }
+
+    private World findWorld (int id) {
+        return createWorld (worldCollection.find(eq ("_id", (double)id)).first ());
+    }
+
+    private World createWorld (Document world) {
+        try {
+            return new World (world.get ("_id", Double.class).intValue (), world.get
```
hamiltont: Just double-checking - this appears to be creating a `new World` object on every request. Is this what you intended?

jamming: This is ok... I create a World instance when a World is fetched from DB. But as fas as I see, creates an instance per findWorld call (which is the same as 'queries' supplied in the test).

jamming:
I won't be able to make any update in the next two weeks, I would highly appreciate if this PR could make it for the 10th round... I can't wait to check the results!

hamiltont:
lgtm, nice pr

PR #1597, opened 2015-05-05
vibed

nin-jin:
Language: D
Framework: Vibe.D
DataBase: MongoDB

Actions:
1. JSON Serialization
2. Single Database Query
3. Multiple Database Queries
6. Plaintext


frameworks/D/vibed/install.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+sudo wget http://master.dl.sourceforge.net/project/d-apt/files/d-apt.list -O /etc/apt/sources.list.d/d-apt.list
```
hamiltont: You'll have to do this without `sudo` - we can't accept further PRs that utilize sudo in their install.sh scripts (we are busy now trying to remove all instances of this)

hamiltont:
Cool, new language. Please update `.travis.yml` to include the line for this new framework

nin-jin:
```
INSTALL: sudo -u testrunner -E -H bash -c "source $FWROOT/toolset/setup/linux/bash_functions.sh && source /home/travis/build/TechEmpower/FrameworkBenchmarks/frameworks/D/vibed/install.sh" (cwd=$FWROOT//installs)
/etc/apt/sources.list.d/d-apt.list: Permission denied
E: Could not open lock file /var/lib/apt/lists/lock - open (13: Permission denied)
E: Unable to lock directory /var/lib/apt/lists/
E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
/home/travis/build/TechEmpower/FrameworkBenchmarks/frameworks/D/vibed/install.sh: line 6: dub: command not found
```

nin-jin:
How can i install this packages without sudo?
http://d-apt.sourceforge.net/

nin-jin:
```
Accessing URL http://127.0.0.1:8080/db:
  Response (trimmed to 40 bytes): "{"randomNumber":6658,"_id":8855,"id":885"
   PASS for http://127.0.0.1:8080/db
Accessing URL http://127.0.0.1:8080/queries?queries=:
  Response (trimmed to 40 bytes): "404 - Not Found
Not Found
```

It's impossible...

```
	router.get("/db", &db);
	router.get("/queries", &queries);
```

nin-jin:
```
Accessing URL http://127.0.0.1:8080/db:
  Response (trimmed to 40 bytes): ""
   FAIL for http://127.0.0.1:8080/db
Accessing URL http://127.0.0.1:8080/queries?queries=:
  Response (trimmed to 40 bytes): ""
Accessing URL http://127.0.0.1:8080/queries?queries=:
  Response (trimmed to 40 bytes): ""
Accessing URL http://127.0.0.1:8080/queries?queries=:
  Response (trimmed to 40 bytes): ""
Accessing URL http://127.0.0.1:8080/queries?queries=:
  Response (trimmed to 40 bytes): ""
```
FTW .-.

hamiltont:
> How can i install this packages without sudo?

Youll need to download and compile the source. Starter instructions [appear to be here](http://wiki.dlang.org/Building_DMD)

> It's impossible

Correct! You've run into bug #1600 - the urls being printed are incorrect. [Here](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/benchmark/test_types/query_type.py#L28) is the urls we are hitting, basically add suffix'es like `2`, `foo`, `30`, etc to your base URL

nin-jin:
> Correct!
Server was not started (https://github.com/rejectedsoftware/vibe.d/issues/956). But tests prints "empty response". I am working on that..

> Youll need to download and compile the source. Starter instructions appear to be here
This is too dificult dor me. Can you intergate docker? http://www.docker.com/

hamiltont:
> you intergate docker

No, not at this time. I think the best thing would be for you to get this working with the use of sudo, and then we can see if someone else is willing to migrate the working-with-sudo code to working-without-sudo.

methane:
http://dlang.org/download.html
There is zip distribution for Ubuntu / Debian.
Isn't it enough?

nin-jin:
This is so difficult for me..

msmith-techempower:
We're here to help - essentially, you need to answer the following questions and we'll see if we can help guide:

1. Upon what software does this test rely?
2. Does that software have binary distributions?

nin-jin:
http://dlang.org/download.html - compiler
http://code.dlang.org/download - package manager
http://vibed.org/download - framework

But i can not resolve this issue to start server: https://github.com/rejectedsoftware/vibe.d/issues/956#issuecomment-99937162

msmith-techempower:
I'm not sure what is required to get it working either, but at least I can try and help on my end. I will try and get it working over here and let you know what I find.

msmith-techempower:
Okay, I ran `vibed` on a branch and it failed: https://gist.github.com/msmith-techempower/b46c0bcd907987c17802

It seems like `dlang` and `dub` are successfully installed, but there is an error setting up `vibed`:

```
Setup vibed: object.Exception@src/rt/minfo.d(162): Aborting: Cycle detected between modules with ctors/dtors:
Setup vibed: vibe.http.server -> vibe.http.dist -> vibe.http.server
```

I'm not sure how much control over this you have, but this appears to be the error hanging this PR up.

etcimon:
The error can be resolved with this change: https://github.com/etcimon/FrameworkBenchmarks/commit/d8a2fd3f67abf84b34490804943c880e6703ba40

It might require a little more memory to build.

Note that the line `sudo apt-get install -qq -y g++ gcc-multilib xdg-utils` shouldn't be necessary.

etcimon:
Also, since this is for benchmarks, you should also consider building with `dub build --build=release` to enable compiler optimizations ;)

There's a chance that the LDC compiler (LLVM-based) will build faster code than DMD, but that would require another setup process.

etcimon:
I would also recommend to add `VibeManualMemoryManagement` to the versions list here: https://github.com/etcimon/FrameworkBenchmarks/blob/03b0f6ccaf7253f11a526f373fdc2f524f17ccf0/frameworks/D/vibed/dub.json#L9

Performance sensitive applications can go to greater lengths to avoid the GC entirely, but this is a good start.

msmith-techempower:
Let's just get vibed working at all before we start worrying about optimizations.

I made that change, but still see the following error:
```
Setup vibed: Linking...
vibed: setup.sh and framework processes have terminated
Setup vibed: object.Exception@src/rt/minfo.d(162): Aborting: Cycle detected between modules with ctors/dtors:
```

etcimon:
Yes, I think we could debug this further with some verbose. Can you pipe the script to stdout some way with `dub --verbose` or `--vverbose` flag ?

msmith-techempower:
So, `dub --verbose build` ?

etcimon:
`dub build --verbose`

msmith-techempower:
Same results: https://gist.github.com/msmith-techempower/5e7a91c0c4dd7d0341c7

msmith-techempower:
Here's the very verbose output: https://gist.github.com/msmith-techempower/8bffeebf8ca962c375f0

etcimon:
If this is on the same machine as the singleFile build, you can use  `--force` flag to refresh the vibed binary

etcimon:
`dub build --force`

msmith-techempower:
I think that may have done the trick, but mongodb is being annoyingly difficult right now... I'll keep hacking away at it for now.

etcimon:
Ok I'll start to support this benchmark script, just point out the problems to me and I'll get to it. But not now, I have to head out =)

msmith-techempower:
Not a problem - it feels like this MIGHT be the last problem, and if it is, I'll just merge it myself.

etcimon:
Yes, don't forget to optimize it

msmith-techempower:
Psssh, nuts to that. I'll get it working at all and you can open a pull request to optimize it.

msmith-techempower:
Closing in favor of #1606 (which I tested and works)

nin-jin:
Cool! When we can see results?

PR #1605, opened 2015-05-14
Remove install scripts

msmith-techempower:
This pull request removes and improves the installation and dependency management supported by the benchmark suite. Prior to this merge, each framework required an `install.sh` file to specify how to install the software required by the framework.

This pull request removes the need for the `install.sh` files and moves the functionality of `fw_depends` into each test's `setup.sh` file.

Additionally, the setup files on which `fw_depends` relies (so, `java7.sh` for example) are updated to use the model first introduced by `mono.sh` wherein `$IROOT/[software].installed` is checked to determine whether the software is installed and then `source`d to set the environment.

As an example, `java7.installed` is as follows:
```bash
export JAVA_HOME=/usr/lib/jvm/java-1.7.0-openjdk-amd64
export PATH=/usr/lib/jvm/java-1.7.0-openjdk-amd64:$PATH
```

So, when a test specifies `fw_depends java7`, the toolset will now check that `$IROOT/java7.installed` exists, and `source` it (thus exporting the environment variables `JAVA_HOME` and `PATH`) before running the rest of the `setup.sh` file for that test. If the file did not exist, then `$FWROOT/toolset/setup/linux/languages/java7.sh` would be executed (and that would install java7, create, and then `source` this "installed" file).


frameworks/Python/bottle/setup.sh
```diff
@@ -1,7 +1,7 @@
...
-export PY2_PIP=$PY2_ROOT/bin/pip
-export PY2_GUNICORN=$PY2_ROOT/bin/gunicorn
 
-$PY2_GUNICORN app:app -c gunicorn_conf.py &
\ No newline at end of file
+fw_depends python2
+
+$PY2_ROOT/bin/pip install --install-option="--prefix=${PY2_ROOT}" -r $TROOT/requirements.txt
+
+gunicorn app:app -c gunicorn_conf.py &
```
methane: `$PY2_ROOT/bin/gunicorn` ?

msmith-techempower: Not necessary anymore. Unfortunately, the line about it still requires `$PY2_ROOT/bin/pip` because our software suite is written in Python and requires `pip` prior to running, so there is a name collision on the path. However, `gunicorn` is NOT on the path prior to running a benchmark, and `fw_depends python2` tell it to put `$PY2_ROOT/bin` onto the path, so `gunicorn` is sufficient.

methane: Wow! great improvements!

msmith-techempower: Thanks! I'm super happy about 1) removing the need of `install.sh` and 2) improving the implementation of `fw_depends` to make it SUPER easy for contributors!

hamiltont:
Just triggered a full verification on the current state of the master branch [here](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/63119298). That should be comparable to the last travis verification for this PR, so we can check that there are no regressions

msmith-techempower:
Going to postpone merging this until after Round 11.

msmith-techempower:
As an updated, we probably cannot avoid merging this before round 11 since we cannot get a full run to complete anymore due to having too many environment variables/functions defined as described in #1699.

This branch changes the way in which environment variables are loaded to be ONLY the current test's variables INSIDE the testrunner's new shell which should completely side-step issue #1699

msmith-techempower:
Unfortunately, Travis seems to continue failing for Postgresql tests, so it is unclear whether this is ready to merge or not.

msmith-techempower:
At this point, I'm thinking that we will probably merge this in for round 11 since it resolves #1699 and #1700 

The Travis issues seem to be travis-issues, not issues with the suite.

PR #1607, opened 2015-05-15
Jetty 9.3 Stripped Test

gregw:
Updated Jetty Servlet test to jetty-9.2
Added Jetty Stripped test at jetty-9.3

### DIFF
--- a/frameworks/Java/jetty/README.md
+++ b/frameworks/Java/jetty/README.md
@@ -1,0 +1,15 @@
+# Jetty-9 Benchmarking Test
+
+This is the Jetty portion of a [benchmarking test suite](../) comparing a variety of web development platforms.
+
+### JSON Encoding Test
+* [JSON test source](src/main/java/hello/HelloWebServer.java)
+
+## Versions
+Jetty-9.1.0 (http://eclipse.org/jetty)
+
+## Test URLs
+
+### JSON Encoding Test
+
+    http://localhost:8080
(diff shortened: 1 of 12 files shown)


frameworks/Java/jetty/README.md
```diff
@@ -0,0 +1,15 @@
+# Jetty-9 Benchmarking Test
+
+This is the Jetty portion of a [benchmarking test suite](../) comparing a variety of web development platforms.
+
+### JSON Encoding Test
+* [JSON test source](src/main/java/hello/HelloWebServer.java)
+
+## Versions
+Jetty-9.1.0 (http://eclipse.org/jetty)
```
martin-g: 9.3.0

gregw:
Updated with fixed version numbers in README.md files for jetty and jetty-servlet

hamiltont:
@gregw Are jetty and jetty-servlet two different frameworks, or are they just one framework that's run using two different configurations? E.g. would it be possible to have both of the source codes underneath `frameworks/Java/jetty`? If so, that would be preferred - each additional row in travis-CI adds non-trivial overhead to our verification runtime

gregw:
@hamiltont The jetty-servlet uses jetty-9.2 and jetty uses jetty-9.3, which are actually significantly different internally. We are also running them in slightly different modes. Our intention was to run them both in the next round so we can compare and then drop jetty-servlet from subsequent rounds and would do any other diversity of configuration under one framework.

However, if it is preferable to merge them now, happy to do so.  Can you point me at some other example java frameworks that run two different versions as different configurations?  Note that they cannot share a classpath.

gregw:
Pushed an update to update to 9.3.0.RC1

PR #1610, opened 2015-05-16
Updated scruffy versions

sksamuel:


hamiltont:
If you have trouble debugging the compile errors through travis-ci, you can easily spin up a local development [virtual machine](https://frameworkbenchmarks.readthedocs.org/en/latest/Development/Installation-Guide/#vagrant-development-environment)

sksamuel:
Thanks for the tip @hamiltont luckily I know what this error is :)

sksamuel:
@hamiltont should be good to go now :)

mcocciaTE:
Everything looks good to me!

frameworks/Scala/scruffy/src/test/scala/BenchmarkVerify.scala
```diff
@@ -11,7 +11,7 @@ object BenchmarkVerify extends App {
 
   val sw = Stopwatch()
 
-  val futures = for ( k <- 1 to 10000 ) yield {
+  val futures = for ( k <- 1 to 30000 ) yield {
```
msmith-techempower: What is this I don't even...

sksamuel: Just creating 30,000 deferred requests to verify the endpoints work as expected when a change is made. It's code in the src/test source folder so isn't include in the JAR that is built when the build script is invoked.

msmith-techempower: So, this code is not executed during an actual benchmark run?

sksamuel: Correct. It's not even included in the package that is built when you build the jar. It's just a smoke test I run when I upgrade the version of Scruffy used by the tests. It's typical in Java/Scala to keep the test code in a source folder called src/test and the main code in src/main.

msmith-techempower:
The let that which must be done be done!

PR #1612, opened 2015-05-18
Move root password set to reliable location

greg-hellings:
When set in the previous way, the root password was not being
reliably set for me in Ubuntu 14.04. This technique moves the
password reset into the configuation tools that are a part of
MySQL itself.


hamiltont:
thanks!

PR #1623, opened 2015-05-25
Add Crystal

will:
Only the JSON and Plain Text tests, but it's a start.


msmith-techempower:
Is there no framework being exercised? What is the reason for the directory being called "stdlib" instead of "crystal-raw" or similar?

will:
> Is there no framework being exercised?

Correct

> What is the reason for the directory being called "stdlib" instead of "crystal-raw" or similar?

I wasn't aware of the naming conventions. I can change it to crystal-raw :)

msmith-techempower:
Great, `crystal-raw` sounds good. Also, after you do that, I think we can safely merge it in since [Travis-CI](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/64016022) has checked it.

will:
Updated, thanks!

Question for future tests: Were I to include a [postgres driver](github.com/will/crystal-pg) for the database tests, would that still fit under -raw?

msmith-techempower:
Yeah, if you plan on including the database tests, I would suggest adding them to the `raw` one you have here.

LGTM!

Vikaton:
Someone is working on a framework on Crystal called Moonshine

Issue #1594, opened 2015-05-04
Crystal: add Moonshine


Vikaton:
Could you guys maybe add Crystal and the Moonshine web framework?

https://gist.github.com/Vikaton/3102c16ee14ef3d76244

https://github.com/manastech/crystal

https://github.com/dhruvrajvanshi/Moonshine

Need any help with it also?

hamiltont:
> Need any help with it also?

That would be great! [here](https://frameworkbenchmarks.readthedocs.org/en/latest/Development/Add-Frameworks-or-Tests/) is the how-to for adding a new framework if it helps

zane-techempower:
Interested in this one, will start building out today

Vikaton:
Yeah I've been meaning to to to do this but I have lots of School work to do, This will be interesting!

will:
I took a stab at this in #1623

Vikaton:
When will round 11 experimental be available?

msmith-techempower:
Soon™

PR #1629, opened 2015-05-28
Add Crystal-Moonshine and improve Crystal installation

zane-techempower:
_resolves #1594, original PR posted yesterday but was branched incorrectly, this is much cleaner._

Add the Moonshine framework in the Crystal language. All tests implemented, database tests are done on Redis so far, I couldn't get the main Crystal MySQL driver to work. I may add Postgres tests soon.

Crystal-raw was also recently added, this pull request modifies the way that Crystal is installed: no sudo, install specific version. I have amended Crystal-raw in the merge wrt the changes I made.


zane-techempower:
Redis is not starting up/ being connected to in these travis fails. Will look into.

zane-techempower:
All of the databases were not starting up and not being tested to be active in travis. I went through the toolset files but could not definitively figure out why this framework was being flagged as not needed a database.

It may be the case that the flag happened because the `default` test in `benchmark_config.json` only listed `/json` and `/plaintext` tests, with the database-required (redis) routes listed separately under a `"redis"` test. However nodejs has this exact configuration i.e. default has `/json` and `/plaintext` and the rest of the db-required tests were in other tests.

For the fix I just merged the two test specs together, I believe this will work because there is only one test config to look at and it for-sure has db-required tests in it. Still doesn't explain why nodejs plays just fine with Travis and moonshine didn't when they had similar configurations.

msmith-techempower:
Travis doesn't seem to be aware of redis... for some reason. @hamiltont Any idea on this?

That said, I pulled your branch on a clean VM and tested - all passed; LGTM!

PR #1630, opened 2015-05-28
 some tuning and add setting on prerequisites.sh for real-time priority

stefanocasazza:


frameworks/C++/ULib/benchmark_config.json
```diff
@@ -20,6 +19,24 @@
       "notes": "",
       "versus": ""
     },
+    "json": {
```
msmith-techempower: Why add an entire new test for `json` when it was already included in the default?

toolset/setup/linux/prerequisites.sh
```diff
@@ -62,6 +62,8 @@ RETCODE=$(fw_exists ~/.bash_profile.bak)
 }
 
 sudo sh -c "echo '*               -    nofile          65535' >> /etc/security/limits.conf"
+sudo sh -c "echo '*            hard    rtprio             99' >> /etc/security/limits.conf"
```
msmith-techempower: Would prefer this be done in a separate pull request with a VERY detailed explanation as to why it is needed and what it does.

frameworks/C++/ULib/benchmark_config.json
stefanocasazza: Hi,

in this way I am able to differentiate the execution context (plaintext has a
greater need for resources than json) so I can tuning better...

Greetings

2015-05-28 19:48 GMT+02:00 Mike Smith <notifications@github.com>:

toolset/setup/linux/prerequisites.sh
stefanocasazza: Hi,

I would use the *FIFO* scheduling policy, I found it can give an advantage
to my server application (as in my test environment).
I need for this to use the linux support for fine-grained capabilities
<http://linux.die.net/man/7/capabilities> in Linux kernel, which allows
granting privileges to a process to do things like, for example, opening
raw sockets or raising thread priority without granting the process root
privileges.

I do this in install.sh with:
setcap cap_sys_nice,cap_sys_resource+eip $ULIB_ROOT/bin/userver_tcp

It is also necessary to change the bash limit (real-time priority) for the
test user.
In the context of execution of benchmark:
sudo -u testrunner -E -H stdbuf -o0 -e0 bash -ex setup_plaintext.sh

it is not possible change the ulimit setting real-time priority unless
I use this
command in setup.sh that I find very ugly:
sudo /bin/bash -c "ulimit -r 99 && exec su testrunner -p -c
\"$IROOT/ULib/bin/userver_tcp -c $IROOT/ULib/benchmark.cfg\"" &

so to avoid this I added the two line in /etc/security/limits.conf
so others may use this possibility also

Greetings

2015-05-28 19:51 GMT+02:00 Mike Smith <notifications@github.com>:

msmith-techempower: This sounds completely reasonable, but I am WAY out of my depth here. Pinging @hamiltont and @bhauer to try and get a concensus.

msmith-techempower:
No one is responding to this >_< I will have to test this on a vanilla environment before merging.

msmith-techempower:
Since I have gotten no feedback and I tested this locally and it worked, I'm inclined to merge. LGTM!

stefanocasazza:
Hi,

thank you very much!

Greetings

2015-06-03 19:24 GMT+02:00 Mike Smith <notifications@github.com>:

PR #1631, opened 2015-05-28
"New" framework to have symfony2 run on hhvm

mcocciaTE:


msmith-techempower:
Hmmm, did you have to change any of the files themselves (PHP) to support running on `hhvm`?

If not, I would suggest simply creating a new entry in the existing `symfony2`'s `benchmark_config.json` and having that setup file use `hhvm`.

mcocciaTE:
I did not. My thought on copying everything over was because I think there are a lot of hhvm specific changes that can be made to improve it's performance on hhvm that might not want to be made to the original one. That would be the only reason to do it how I did though.

msmith-techempower:
Okay, let's close this PR and see if you can open one which only adds a new entry to the `symfony2` `benchmark_config.json`

PR #1633, opened 2015-05-28
Add Phoenix ORM tests and add elixir-cowboy (new)

waiteb3:
Adds the Ecto ORM for phoenix and the a elixir using cowboy similar to the erlang one.

Also, switched to dpkg -x since compiling from source takes forever.


msmith-techempower:
Seems to be returning empty responses based on travis's results.

waiteb3:
Looks like cowboy-elixir passed now.

msmith-techempower:
True, but it looks like you broke the Erlang tests.

toolset/setup/linux/languages/erlang.sh
```diff
@@ -3,16 +3,12 @@
...
-	./configure --prefix=$IROOT/erlang --without-termcap
-	make
-	make install
-)
+fw_get http://packages.erlang-solutions.com/site/esl/esl-erlang/FLAVOUR_1_esl/esl-erlang_${VERSION}~ubuntu~${RELEASE}_${ARCH}.deb
+dpkg -x esl-erlang_${VERSION}~ubuntu~${RELEASE}_${ARCH}.deb $IROOT/erlang
+$IROOT/erlang/usr/lib/erlang/Install -minimal $IROOT/erlang/usr/lib/erlang
 
-echo "export PATH=$IROOT/erlang/bin:$PATH" >> $IROOT/erlang.installed
+echo "export PATH=\$IROOT/erlang/usr/bin:\$PATH" >> $IROOT/erlang.installed
```
msmith-techempower: Is this the location of `rebar`?

toolset/setup/linux/systools/rebar.sh
```diff
@@ -1,15 +0,0 @@
...
-fw_depends erlang
-source $IROOT/erlang.installed
-
-fw_get https://github.com/rebar/rebar/archive/2.5.1.tar.gz -O rebar-2.5.1.tar.gz
-fw_untar rebar-2.5.1.tar.gz
-mv rebar-2.5.1 rebar
-cd rebar
-./bootstrap
-
-echo "export REBAR_HOME=$IROOT/rebar" >> $IROOT/rebar.installed
```
msmith-techempower: ... because you removed `rebar` as a dependency here. That's fine if it is included in `erlang`, but I just want to make sure.

waiteb3: Yes, the packaged install comes with rebar now. Not sure why the other tests are breaking.

msmith-techempower:
Travis again... after round 11 we will have to figure out all these travis bugs.

This worked for me locally; LGTM!

hamiltont:
@msmith-techempower note for future - when you see `dpkg -x`, we should promote usage of [this function](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/bash_functions.sh#L45) instead. This is because doing this dpkg trick requires a few special things on file ownership that may not be true inside the vagrant environment (where the filesystem they expose is not actually posix compliant), so I want to be able to modify that one function in the future and have the changes flow into all usages

msmith-techempower:
@hamiltont Okay, will do.

waiteb3:
The issue with using the provided apt function is that the default Erlang installs are too out of date and there's no canonical download for Elixir yet. To be able to use that function would require
```bash
wget http://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb
sudo dpkg -i erlang-solutions_1.0_all.deb
```
which breaks the sudo-less install rule and annoyingly assumes that it's installed /usr/bin (or elsewhere, I forget).

msmith-techempower:
Wat? For one, you aren't at your desk right now, so it's hard to take you seriously.

Secondly, this is why i opted for using `dpkg-deb` with `dlang`. It allows you to specify a root directory and it will treat it as `/` for the installation; really, all you need to do is ensure your `PATH` is set up correctly.

@hamiltont Upon looking into this further, I found that I'm using `dpkg-deb` and not `dpkg` for `dlang`. I cannot explain why, except that I remember trying `dpkg` and it didn't work.

PR #1635, opened 2015-05-29
Improve JS frameworks

zane-techempower:
_Addresses  #1586_

Tasks:
* ~~[nodejs] Improve routing, no longer using harmful switches~~
* ~~[nodejs] Refactor and organize code better~~
* ~~[nodejs] Add fortunes test to the 4 database configurations~~
* ~~[nodejs] Add redis tests~~
* ~~refactor handlers out of main script~~

Will address in another PR:
* Add fortunes testing to other frameworks
* Add redis tests to other frameworks

Making PR early, this is a work-in progress but should be passing every step of the way. Work has been done on plain nodejs so far.


zane-techempower:
Will have this PR just address nodejs improvement. If fortunes and redis support are added to other JS frameworks, then those will be in their own PR's.

frameworks/JavaScript/nodejs/create-server.js
```diff
@@ -0,0 +1,84 @@
...
+
+module.exports = http.createServer(function (req, res) {
+  var values = url.parse(req.url, true);
+  var route = values.pathname;
+
+  // Basic routes, no db required
+  if (route === '/json') {
+    return h.responses.jsonSerialization(req, res);
+  } else if (route === '/plaintext') {
+    return h.responses.plaintext(req, res);
```
naugtur: It shouldn't affect performance directly, but if you want to choose a function based on a string in JS, you might consider doing something like this:
```javascript
var handlers={
  '/json': h.responses.jsonSerialization,
  '/plaintext': h.responses.plaintext,
...
};

return handlers[route](req,res);
```

frameworks/JavaScript/nodejs/create-server.js
```diff
@@ -0,0 +1,84 @@
...
+var MongooseHandler = require('./handlers/mongoose');
+
+// Sequelize is a popular Node/SQL driver
+var SequelizeHandler = require('./handlers/sequelize');
+
+// Node's redis package uses the C bindings of the hiredis library
+var HiredisHandler = require('./handlers/redis');
+
+module.exports = http.createServer(function (req, res) {
+  var values = url.parse(req.url, true);
```
naugtur: `url` module is pretty big and complicated, not my first choice. You could try `parseurl` https://github.com/pillarjs/parseurl (used in express, don't judge by github stars) or even a simple regexp, as you don't use the vast output generated by url.parse

frameworks/JavaScript/nodejs/create-server.js
```diff
@@ -0,0 +1,84 @@
...
+  } else if (route === '/mysql/fortunes') {
+    return MySQLRawHandler.Fortunes(req, res);
+  } else if (route === '/hiredis/db') {
+    return HiredisHandler.SingleQuery(req, res);
+  } else if (route === '/hiredis/fortunes') {
+    return HiredisHandler.Fortunes(req, res);
+  }
+
+  else {
+    var queries = isNaN(values.query.queries) ? 1 : parseInt(values.query.queries, 10);
```
naugtur:
```javascript
var queries = ~~(values.query.queries) || 1;
```
looks like magic, but converting numbers with `~~` and defaulting with `||` are pretty common in JS

frameworks/JavaScript/nodejs/handlers/mongodb-raw.js
```diff
@@ -0,0 +1,84 @@
...
+    world.value._id = undefined; // remove _id from query response
+    callback(err, world.value);
+  });
+}
+
+module.exports = {
+
+  SingleQuery: function (req, res) {
+    mongodbRandomWorld(function (err, result) {
+      if (err) { throw err; }
```
naugtur: the callback is called asynchronously, throwing an error in such callback will not get caught anywhere and it might kill the app instance and force restart in cluster.
If you want to have asynchronous error handling, you need to switch to promises (I recommend bluebird for that) or generators (node12 have them)

But in this case you don't need to do anything other than just returning a http500 if `err` is truthy

frameworks/JavaScript/nodejs/app.js
```diff
@@ -0,0 +1,16 @@
...
+var numCPUs = require('os').cpus().length;
+
+if (cluster.isMaster) {
+  // Fork workers.
+  for (var i = 0; i < numCPUs; i++) {
+    cluster.fork();
+  }
+
+  cluster.on('exit', function (worker, code, signal) {
+    console.log('worker ' + worker.pid + ' died');
```
naugtur: Not a single worker should die during the benchmark. If this is called, it means the app is missing error handling somewhere. I recommend shutting everything down if this callback is called and stopping the test.

frameworks/JavaScript/nodejs/create-server.js
zane-techempower: Done, this should be a slight, slight improvement on performance by using a hash to route rather than if-else.

frameworks/JavaScript/nodejs/create-server.js
zane-techempower: Done, nice module to know about.

frameworks/JavaScript/nodejs/create-server.js
zane-techempower: This is new to me, but I'm down with it, added.

frameworks/JavaScript/nodejs/handlers/mongodb-raw.js
zane-techempower: I believe that the errors would get thrown at some point. However I have added error throwing to be present at this level as it is with some of the other database configurations. Should be "extra safe" now.

frameworks/JavaScript/nodejs/app.js
zane-techempower: I've only seen this happen when an `err` is thrown in a database test. Otherwise this should never go off, it's more of a dev-helper as that log will be printed at the end of the stack trace that details why e.g. ERR_CONNECTION_REFUSED, the worker/thread died.

frameworks/JavaScript/nodejs/create-server.js
```diff
@@ -0,0 +1,79 @@
...
+
+    '/sequelize/db':       SequelizeHandler.SingleQuery,
+    '/sequelize/fortunes': SequelizeHandler.Fortunes,
+
+    '/mysql/db':           MySQLRawHandler.SingleQuery,
+    '/mysql/fortunes':     MySQLRawHandler.Fortunes,
+
+    '/hiredis/db':         HiredisHandler.SingleQuery,
+    '/hiredis/fortunes':   HiredisHandler.Fortunes
+  }
```
ngryman: This hash is built for each request, you should move this at the module level.

frameworks/JavaScript/nodejs/create-server.js
```diff
@@ -0,0 +1,79 @@
...
+
+      '/sequelize/queries': SequelizeHandler.MultipleQueries,
+      '/sequelize/updates': SequelizeHandler.Updates,
+
+      '/mysql/queries':     MySQLRawHandler.MultipleQueries,
+      '/mysql/updates':     MySQLRawHandler.Updates,
+
+      '/hiredis/queries':   HiredisHandler.MultipleQueries,
+      '/hiredis/updates':   HiredisHandler.Updates
+    }
```
ngryman: Same here.

frameworks/JavaScript/nodejs/handlers/mongoose.js
```diff
@@ -0,0 +1,101 @@
...
+
+var Worlds = connection.model('World', WorldSchema);
+var Fortunes = connection.model('Fortune', FortuneSchema);
+
+function mongooseRandomWorld(callback) {
+  Worlds.findOne({
+    id: h.randomTfbNumber()
+  }).exec(function (err, world) {
+    callback(err, world);
+  });
```
ngryman: Can be simplified by `exec(callback)`.

frameworks/JavaScript/nodejs/handlers/mongoose.js
```diff
@@ -0,0 +1,101 @@
...
+}
+
+function mongooseGetAllFortunes(callback) {
+  Fortunes.find({}).exec(function (err, fortunes) {
+    callback(err, fortunes);
+  });
```
ngryman: Can be simplified by `exec(callback)`.

zane-techempower:
@ngryman thanks for that second round of looking, made the adjustments.

ngryman:
@zane-techempower You're welcome. Your refactoring is really good. It makes the code way more readable with probably better performances.

 - [ ] For aesthetic purpose I would recommend to remove mixed `else` and `return`. You should choose one way (branches or early returns) and stick with it.

 - [ ] As mentioned by @naugtur, `throw`s are weird as nobody will ever catch them. At the end of the refactoring, if you just want to exit the process, then `process.exit(1)` would be better because it's explicit and quicker.

 - [ ] You should bump your dependencies as it seems some are outdated. This can have a bad performance impact. You can use [npm-check-updates] or something similar, that does it for you automatically.

- [ ] Not directly related, but you should add [iojs] to your benchmark

- [ ] To conclude, if you are interested in, you can dive more in performance optimizations and take a look on how JS VM work and optimize / deoptimize / inline code. Here is a good starter: https://github.com/petkaantonov/bluebird/wiki/Optimization-killers.

[iojs]: https://github.com/nodejs/io.js
[npm-check-updates]: https://www.npmjs.com/package/npm-check-updates

naugtur:
@ngryman I don't think it's worthwile to add io.js to benchmarks, as io.js already decided to merge with Node Foundation, and it's just a matter of waiting for another node version that results from merge.

ngryman:
@naugtur Hm I think it can be interesting for a benchmark. They already compare different VMs.

`node` stable uses v8 `3.28`, `iojs` uses v8 `4.2`. This is a huge difference when you consider performances. Without mentioning `iojs` moves really faster than `node`.

I think both `iojs` and `node` will live in parallel for a while as they both afdress different needs: `node` is for stability, `iojs` is for state-of-the-art.

naugtur:
@ngryman I'm not speculating here. It's a fact. There will be no new versions of io.js and all work on io.js will be merged into node. You seem to have missed some great news recently ;)
Next node will have it all (including new major version number, so no 0.xx as far as I remember). No point setting up a different benchmark now.

msmith-techempower:
How are we feeling about this? Travis has signed off; is it ready to merge?

naugtur:
I'd add `process.exit(1)` in `exit` event handler for cluster to destroy the whole test if a worker fails (cause it should not, ever) and it's fine.

ngryman:
@naugtur Indeed! I totally missed that... This is good news!
So yes, I guess forgot about this advice. Even if I would have been curious to see the difference between both runtimes now and the future *merged nodejs*.

zane-techempower:
io.js was in the works along with improved node versioning. For this round it looks like the node versioning via `nvm` will stay. Based on what @naugtur is saying, waiting till next round will be nice as we may have a much-improved node at our disposal.

Right now I will:
* add in the `process.exit(1)` call to the exit handler
* get rid of the `throws` and get any error case to hit the `exit` handler
* consistency between `else` and `return`
* Ill do a node check update, but if it "breaks everything" and will take a while to correct, then package updates will wait till next time.

ngryman:
@msmith-techempower That seems pretty cool :)

zane-techempower:
Added the `process.exit(1)` calls, caught errors as cleanly as possible e.g. tried not to catch the same error twice in a row which is just un-runnable code. I poked around and from what I've seen, a worker will exit, which gets caught in the master 'exit' handler which then shuts down everything with it's own process.exit(1) call.

npm-check-updates was awesome. Most deps didn't need to be touched, however Sequelize shifted to promises for its 3.0 update so I made changes for that, looks very clean now.

Ready for merge if all is well. This will make it into round 11 where hopefully we see a speed-up :)

zane-techempower:
I saw that the build was failing for `/json` and `/plaintext` because travis thinks there doesn't need to be a database active. However I had the tests exiting if there were any database connectivity issues so the server would not be operational for those tests (Redis seemed to be the main culprit).

I have since changed the code to do nothing if there are database connectivity issues. This will be fine as if there is a true failure, then the actual endpoints will error out. Also updated the readme to reflect new software + version upgrades.

If for some reason this fails again, just re-run locally and see that `/json` and `/plaintext` are indeed working (all db-related routes are fine).

msmith-techempower:
Travis error was unrelated to javascript tests; LGTM!

PR #1636, opened 2015-05-31
Add Rust

steveklabnik:
Hey there! Now that Rust is 1.0, I figured adding some Rust stuff to the benchmarks would be cool :)

I haven't gotten it all working, but, in the interest of 'release early, release often', I wanted to check that my setup is vaguely on track. VirtualBox isn't working on my computer right now, so I don't have any error messages or anything, but was interested in early feedback anyway.

Thoughts?

### DIFF
--- a/frameworks/Rust/hyper/setup.sh
+++ b/frameworks/Rust/hyper/setup.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-sed -i 's|tcp(.*:3306)|tcp('"${DBHOST}"':3306)|g' src/main.rs
+# Something like this will need to be done once we add database tests
+# sed -i 's|tcp(.*:3306)|tcp('"${DBHOST}"':3306)|g' src/main.rs
 
 CARGO = $IROOT/rust/bin/cargo
 
--- a/frameworks/Rust/hyper/setup.sh
+++ b/frameworks/Rust/hyper/setup.sh
@@ -3,7 +3,7 @@
 # Something like this will need to be done once we add database tests
 # sed -i 's|tcp(.*:3306)|tcp('"${DBHOST}"':3306)|g' src/main.rs
 
-CARGO = $IROOT/rust/bin/cargo
+CARGO=$IROOT/rust/bin/cargo
 
 $CARGO build --release
 
--- a/frameworks/Rust/hyper/setup.sh
+++ b/frameworks/Rust/hyper/setup.sh
@@ -2,7 +2,7 @@
 
 sed -i 's|tcp(.*:3306)|tcp('"${DBHOST}"':3306)|g' src/main.rs
 
-set CARGO = $IROOT/rust/bin/cargo
+CARGO = $IROOT/rust/bin/cargo
 
 $CARGO build --release
 
--- a/frameworks/Rust/hyper/setup.sh
+++ b/frameworks/Rust/hyper/setup.sh
@@ -2,6 +2,8 @@
 
 sed -i 's|tcp(.*:3306)|tcp('"${DBHOST}"':3306)|g' src/main.rs
 
-cargo build --release
+set CARGO = $IROOT/rust/bin/cargo
 
-cargo run --release &
+$CARGO build --release
+
+$CARGO run --release &
--- a/frameworks/Rust/hyper/setup.sh
+++ b/frameworks/Rust/hyper/setup.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+sed -i 's|tcp(.*:3306)|tcp('"${DBHOST}"':3306)|g' src/main.rs
+
+cargo build --release
+
+cargo run --release &
--- a/frameworks/Rust/hyper/setup.sh
+++ b/frameworks/Rust/hyper/setup.sh
@@ -1,10 +1,7 @@
 #!/bin/bash
 
-# Something like this will need to be done once we add database tests
-# sed -i 's|tcp(.*:3306)|tcp('"${DBHOST}"':3306)|g' src/main.rs
+source $IROOT/rust.installed
 
-CARGO=$IROOT/rust/bin/cargo
-
-$CARGO build --release
-
-$CARGO run --release &
+cargo clean
+cargo build --release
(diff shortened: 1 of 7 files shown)


torhve:
There's an old PR here that might be revived: https://github.com/TechEmpower/FrameworkBenchmarks/pull/1101

frameworks/Rust/hyper/setup.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+sed -i 's|tcp(.*:3306)|tcp('"${DBHOST}"':3306)|g' src/main.rs
+
+cargo build --release
```
msmith-techempower: Cargo would not be on your path at this point - try pointing to it directly `$IROOT/rust/bin/cargo` or whatever the sandboxed path will be.

steveklabnik:
Status update: I haven't gotten VirtualBox to work, but will be updating this PR tomorrow.

frameworks/Rust/hyper/setup.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+sed -i 's|tcp(.*:3306)|tcp('"${DBHOST}"':3306)|g' src/main.rs
+
+set CARGO = $IROOT/rust/bin/cargo
```
untitaker: Remove `set` here.

steveklabnik: yeah i'm bad at bash, sigh

steveklabnik:
So, I _still_ cannot get virtualbox to work on my computer. But, I've added an implementation of the `plaintext` test.

Can anyone tell if this actually runs or not yet? :/

At least once this setup is done, it will work for all future Rust projects...

steveklabnik:
Oh, I need to do the `sed` properly to set the port...

steveklabnik:
... in the future, anyway. i just left a note for now, since I didn't implement the database test

steveklabnik:
@msmith-techempower i think this is ready for a re-review

msmith-techempower:
Currently being reviewed by Travis-CI.

steveklabnik:
Ahh interesting, so I saw the failures in other builds, but didn't realize that they were builds per language. That's super neat. I'll pay close attention to the specific build, thank you! Still getting used to navigating this project. :)

frameworks/Rust/hyper/src/main.rs
```diff
@@ -0,0 +1,28 @@
...
+    Server::http(|req: Request, mut res: Response| {
+        match (req.method, req.uri) {
+            (hyper::Get, RequestUri::AbsolutePath(ref path)) if path == "/plaintext" => {
+                res.headers_mut().set(ContentLength(HELLO_WORLD.len() as u64));
+                res.headers_mut().set(ContentType::plaintext());
+                res.headers_mut().set(header::Server("Hyper".to_owned()));
+
+                res.send(HELLO_WORLD).unwrap();
+
+                StatusCode::Ok
```
seanmonstar: this Ok is dangling

frameworks/Rust/hyper/src/main.rs
```diff
@@ -0,0 +1,28 @@
...
+            }
+            (hyper::Get, _) => StatusCode::NotFound,
```
seanmonstar: likewise, these statuscodes aren't being assigned to anything. you could do `*res.status_mut() = NotFound`.

frameworks/Rust/hyper/src/main.rs
```diff
@@ -0,0 +1,28 @@
...
+
+fn main() {
+    Server::http(|req: Request, mut res: Response| {
+        match (req.method, req.uri) {
+            (hyper::Get, RequestUri::AbsolutePath(ref path)) if path == "/plaintext" => {
+                res.headers_mut().set(ContentLength(HELLO_WORLD.len() as u64));
+                res.headers_mut().set(ContentType::plaintext());
+                res.headers_mut().set(header::Server("Hyper".to_owned()));
+
+                res.send(HELLO_WORLD).unwrap();
```
seanmonstar: `send()` will add the Content-Length for you.

frameworks/Rust/hyper/src/main.rs
steveklabnik: I would get a type error when it's not included

frameworks/Rust/hyper/src/main.rs
steveklabnik: Ah! So, this may be a side effect of my stupidity with `curl` the other day: I wasn't getting the header, but that's probably because I was accidentally sending `HEAD` requests

frameworks/Rust/hyper/src/main.rs
steveklabnik: same as above. the `match` is an expression, so these are the return types of the closure

frameworks/Rust/hyper/src/main.rs
steveklabnik: .... but now I don't. cool. :)

steveklabnik:
@geal and I paired, and this should be working now, so let's just wait for travis! :confetti_ball:

msmith-techempower:
You should probably rebase off of master and take a look at #1638 to get Hyper working; it's a pretty good starting point.

steveklabnik:
Will do in the next day or two :)

steveklabnik:
rebased and updated. #1638 had a lot of goodies :)

msmith-techempower:
Hmm, it claims that `cargo` could not be found, but an audit of your install and setup shows that you are following the correct order. I'll try and pull this to test manually and let you know.

steveklabnik:
Yeah, strange. I followed the other PR here, so if it works, this should work.

msmith-techempower:
I have no idea why Travis doesn't like this - I tested it locally and it seems to work.

msmith-techempower:
Oh I see, there was an issue handshaking (SSL) when trying to get the binary. Oh well, this is working locally; LGTM!

steveklabnik:
:confetti_ball:

PR #1638, opened 2015-06-02
Add Rust with Iron and Nickel

waiteb3:
Currently just the JSON and Plaintext tests only. They both use hyper as their HTTP lib, so they should be close in performance.

### DIFF
--- a/toolset/setup/linux/languages/rust.sh
+++ b/toolset/setup/linux/languages/rust.sh
@@ -10,5 +10,5 @@ fw_untar rust-1.0.0-x86_64-unknown-linux-gnu.tar.gz
 	./install.sh --prefix=$IROOT/rust
 )
 
-echo "export LD_LIBRARY_PATH=\${LD_LIBRARY_PATH}:\$IROOT/rust/lib" >> $IROOT/rust.installed
-echo "export PATH=\$IROOT/rust/bin:$PATH" >> $IROOT/rust.installed
+echo -e "export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$IROOT/rust/lib" > $IROOT/rust.installed
+echo -e "export PATH=$IROOT/rust/bin:$PATH" >> $IROOT/rust.installed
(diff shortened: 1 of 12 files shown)


toolset/setup/linux/languages/rust.sh
```diff
@@ -0,0 +1,14 @@
...
+[ ! "$RETCODE" == 0 ] || { . $IROOT/rust.installed; return 0; }
+
+fw_get https://static.rust-lang.org/dist/rust-1.0.0-x86_64-unknown-linux-gnu.tar.gz
+fw_untar rust-1.0.0-x86_64-unknown-linux-gnu.tar.gz
+(
+	cd rust-1.0.0-x86_64-unknown-linux-gnu
+	./install.sh --prefix=$IROOT/rust
+)
+
+echo "export LD_LIBRARY_PATH=\${LD_LIBRARY_PATH}:\$IROOT/rust/lib" >> $IROOT/rust.installed
```
msmith-techempower: Should be a single `>` here - always create/override the file on the first instance.

toolset/setup/linux/languages/rust.sh
```diff
@@ -0,0 +1,14 @@
...
+echo "export PATH=\$IROOT/rust/bin:$PATH" >> $IROOT/rust.installed
```
msmith-techempower: I think you have your echo statements a bit messed up. I would expect:

`echo -e "export LD_LIBRARY_PATH=${IROOT}/rust/lib:\$LD_LIBRARY_PATH" > $IROOT/rust.installed
echo -e "export PATH=${IROOT}/rust/bin:\$PATH" >> $IROOT/rust.installed`

The `-e` flag allows escaping the dollar sign, and otherwise it doesn't matter if $IROOT gets expanded at this point, so save a char.

steveklabnik:
see also https://github.com/TechEmpower/FrameworkBenchmarks/pull/1636

msmith-techempower:
Tested this locally and it works (not sure what is up with Travis of late); LGTM!

steveklabnik:
:confetti_ball:

PR #1639, opened 2015-06-02
Added hhvm test for Yii2

mcocciaTE:


msmith-techempower:
The one I care about passed. LGTM!

PR #1641, opened 2015-06-03
Added hhvm test to Slim

mcocciaTE:


msmith-techempower:
Tested this locally and it works - LGTM!

PR #1642, opened 2015-06-03
forgot to change setup file to hhvm

mcocciaTE:
oops


msmith-techempower:
Y U DO DIS!? Restested and it still works - LGTM!

PR #1643, opened 2015-06-03
Missed setup file change for raw test

mcocciaTE:


msmith-techempower:
Retested and it looks good; LGTM!

PR #1644, opened 2015-06-03
copy paste fail on slim setup sourceroot

mcocciaTE:


msmith-techempower:
Passing again... but I'm going to leave this open for 15m to ensure you don't find more minor fixes.

PR #1646, opened 2015-06-03
Fixes gin test

manucorporat:
1. Context.Fail() was deprecated, using c.AbortWithError() instead.
2. Gin by default is set in debug mode. This mode reloads the HTML templates in each request, so you don't have to restart the server when doing changes in the html templates.

### DIFF
--- a/frameworks/Go/gin/hello.go
+++ b/frameworks/Go/gin/hello.go
@@ -72,7 +72,8 @@ func db(c *gin.Context) {
 	var world World
 	err := worldStatement.QueryRow(rand.Intn(worldRowCount)+1).Scan(&world.Id, &world.RandomNumber)
 	if err != nil {
-		c.Fail(500, err)
+		c.AbortWithError(500, err)
+		return
 	}
 	c.JSON(200, &world)
 }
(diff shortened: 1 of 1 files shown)


msmith-techempower:
The travis failure was unrelated to gin; LGTM!

manucorporat:
@msmith-techempower great! btw I noticed some warnings about missing response headers in the travis log, should I care about?

msmith-techempower:
At some point, that will be bumped from a warning to an error, so if you have the time then it would be great to fix those up.

manucorporat:
@msmith-techempower oh! I see. I did not wrote the code.

Some observations:
1. Missing the "Server" header.
2. Content-type
```
9.The response content type must be set to application/json.
18.The response content type must be set to text/html.
```
 that does not sound right. For example, Gin automatically sets the charset in order to avoid very common issues between unicode and browsers.

Does `application/json; charset=utf-8` trigger a warning?

msmith-techempower:
I'm not sure where you found numbers 9 and 10, but that is probably a copypasta error.

`application/json; charset=utf-8` should NOT trigger a warning... but it might.

manucorporat:
I just looked into the code: https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/benchmark/test_types/json_type.py#L48

I think it would pass

hamiltont:
> application/json; charset=utf-8 should NOT trigger a warning... but it might

I actually think it does trigger a warning currently. We need to revamp the header checking at some point to be be more maintainable - the current approach is kind of cryptic to read and still tosses out false positives like this

PR #1647, opened 2015-06-04
Updated Haywire script to initialize properly to match latest in the Haywire repository.

kellabyte:


msmith-techempower:
Ooof, it tested everything. Oh well, Haywire passed; LGTM!

PR #1649, opened 2015-06-05
Improve hapi framework

zane-techempower:
This PR covers:
* ~~Improve hapi's organization w.r.t. handlers for each database set up~~
* ~~Update dependencies to newest version~~
* ~~Sequelize (mysql, postgres) uses Promises now~~
* ~~Add Postgres and Redis tests~~
* ~~[Promisified](https://github.com/petkaantonov/bluebird/blob/master/API.md#promisepromisifyallobject-target--object-options---object) the redis library so now that uses a Promise approach rather than async~~
* ~~Promisify mongodb handler~~
* Improve start up time -- server takes about a minute between when the master cluster forks and all the workers are up and ready. _I was able to do a slight optimization that dropped the time by a second, but I was not able to determine the cause of the long delay._
* ~~Add band-aid fix of adding `sleep 30` to the `setup.sh` to give more time.~~


zane-techempower:
Took a look at travis. The fails are non-hapi related.

`Crystal/moonshine`, `Rust/iron`, and `Rust/akka-http` are all failing because they do not define what Travis considers to be valid test cases:
* `Crystal/moonshine` only defines a redis test, this is being fixed in #1653 
* The two Rust frameworks have the empty string as their database. I will make a fix right now to set this to `"None"`.

Hopefully those two changes will clean up Travis for working tests such as this. Please test manually for confidence and merge.

msmith-techempower:
Man... Travis is letting me down of late - this worked locally; LGTM!

PR #1650, opened 2015-06-05
Set rust frameworks' database property to None type to see if that broke travis CI

waiteb3:
Travis CI broke on Rust, which may be caused by using "" rather than "None" as the database config.

### DIFF
--- a/frameworks/Rust/iron/benchmark_config.json
+++ b/frameworks/Rust/iron/benchmark_config.json
@@ -8,7 +8,7 @@
       "port": 8080,
       "approach": "Realistic",
       "classification": "Framework",
-      "database": "",
+      "database": "None",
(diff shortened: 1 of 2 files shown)


zane-techempower:
Looks like I made the same fix in #1654 with the addition of making the frameworks `"versus": "rust"` as well. However I see that the tests are still failing due to lack of `cargo`:

    Setup iron: + cargo build --release
    Setup iron: setup.sh: line 6: cargo: command not found

Lack of `cargo` also appears to be the latest development in #1636 so I will leave it to the rust crew to settle this.

msmith-techempower:
Looks good, but `cargo` missing is part of the travis installation issue (SSL handshake is failing for some reason); LGTM!

PR #1652, opened 2015-06-08
Improve sailjs, add/improve DB support

zane-techempower:
Refactored MySQL database tests to make better use of Promises e.g. removed all callbacks as they are improper with the promise way of doing things.

Added Postgres and Redis database tests.

Everything is configured well so fails should only come up if a database isn't available in travis.


zane-techempower:
Seems like sails shut down in the middle of the postgres updates part of the test. I ran the tests manually with curl and was not able to replicate a database shutdown on any of the `/updates` requests.

I believe that this fail caused the redis tests to not run. I have made a small change to give travis another go as this is unexpected.

mcocciaTE:
It worked for me locally. Must have been a travis hiccup.

msmith-techempower:
No idea what's up with travis lately, but this worked for me locally; LGTM!

PR #1653, opened 2015-06-08
Add Redis as valid database

zane-techempower:
Redis was not listed as a supported database in `run-ci.py`. While making the change I also removed `sqlite` as that is not a valid database for our tests anymore.

Finally I made a trivial change in `Crystal/moonshine`. This particular test _only_ requires a redis database. This causes failures because when redis wasn't on the list of databases, it appeared as if this framework didn't require a valid database. Travis aborts in this case.

    INFO:run-ci:Found 1 usable tests (1 valid for linux, 0 valid for linux and {mysql,postgres,mongodb,cassandra,elasticsearch,sqlite,none}) in directory '$FWROOT/frameworks/Crystal/moonshine'

    CRITICAL:run-ci:Found no test that is possible to run in Travis-CI! Aborting!

So I made a change in `Crystal/moonshine` to ensure the tests will re-run on the directory (they should regardless).

Should see all tests passing, including `Crystal/moonshine` which is the litmus test.

PR #1654, opened 2015-06-08
Update Rust benchmark configs to get through Travis

zane-techempower:
As [commented](https://github.com/TechEmpower/FrameworkBenchmarks/pull/1649#issuecomment-110096958) in #1649, there were some tests that were failing to make it through travis due to their benchmark configs.

Adding Redis has already been put into a PR #1653 . These tests were failing because they specified the empty string as the database rather than `"None"`. I have updated the configs to have None specified as well as to be compared against other `rust` frameworks.

### DIFF
--- a/frameworks/Rust/iron/benchmark_config.json
+++ b/frameworks/Rust/iron/benchmark_config.json
@@ -8,17 +8,17 @@
       "port": 8080,
       "approach": "Realistic",
       "classification": "Framework",
-      "database": "",
+      "database": "None",
       "framework": "iron",
       "language": "rust",
-      "orm": "",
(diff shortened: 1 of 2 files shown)


msmith-techempower:
Loving these noticed things.

@zane-techempower Maybe an improvement would be to have the suite report and error out when there are incorrect values in `benchmark_config.json` for a given test. This would, assuming every one of these already is correct, would force contributors to fix it before submitting a pr.

zane-techempower:
@msmith-techempower I am looking into this now

msmith-techempower:
Closed in favor of #1658

PR #1658, opened 2015-06-09
Sanitize benchmark_config fields, improve helpful errors

zane-techempower:
Following discussion in #1654 I took a look into adding checks to the `benchmark_config` parser to prevent incorrect field values rather than just catching when required fields were not present at all.

Changes:
* Clearly list which field(s) are absent from the benchmark config
* Add checking for test urls (asserts non-empty and begins with "/"), suggests a default url e.g. `/mysql/queries?queries=`
* Add checking for database type (Includes edge cases "sqlite", "sqlserver" (used by aspnet), and "None")
* Add checking for language (asserts non-empty and suggests the frameworks language if empty)
* Add checking for approach, must be `["realistic" or "stripped"]`
* Add checking for classification, must be `["fullstack" or "micro" or "platform"]`
* Add checking for webserver (asserts non-empty)
* Add checking for orm, must be `["full" or "micro" or "raw"]`
* Add checking for os and database_os, must be `["linux" or "windows"]`, suggests `"Linux"`

In order for no errors to pop up, every framework in TFB must have a compliant `benchmark_config`. This is because `benchmark_config`'s are parsed under the hood when the tests are run i.e. when you run the tests on the `nodejs` framework, the `benchmark_config`'s of other, non-related frameworks are also parsed and thus, must also be correct. To address this I made the required changes in the benchmark configs of `Elixir/phoenix`, `Java/servlet-dsl`, `JavaScript/sailsjs`, `Rust/iron`, and `Rust/nickel`.

Assigned self as I believe there may be more work, or a correction or two to make.


zane-techempower:
This morning I took a look at the fails and for the most part they are not related to the changes I made here:
* `Duda` -- install was broken, fixed and posted PR: #1660 
* `onion` -- ok on my end
* `aspnet` -- ok on my end
* `nancy-libevent2` -- completely broken, many exceptions going around. I took a look around, but I lack the experience to track down the issue in a timely way.
* `servicestack` -- broken, responds with xml on every request...
* `cpoll_cpp` -- mysql randomly doesn't start for the test, ok on my end
* `wt` -- ok on my end, takes forever to be ready though
* `moonshine` -- I needed to have my changes to `run_ci.py` from #1653 , travis failed to run because it didn't recognize redis as a db. I have merged those changes into here as they are closely related to my toolset changes.

Really want to see moonshine work on travis on this latest build.

msmith-techempower:
Nice work! LGTM!

hamiltont:
Great start. By the way, there is redundancy in this code for checking the benchmark_config keys. The pattern is basically this (in pseudocode)

```
for key in keys: 
   val = config[key].lowercase()
   if val is "" or val not in list_of_allowed_values:
      print "invalid %key value %value"
      print "acceptable values: %list_of_allowed_values"
```

Given that, you could probably just construct a simple data structure that is the allowed schema for the benchmark_config, like so: 

```
schema = {
  'database': ['mysql', 'redis', 'postgres'],
  'os': ['linux', 'windows'],
  'classification': ['fullstack', 'micro']
}
```

This would be much more maintainable than enumerating all possible keys - the one data structure at the top shows what is "allowed", and the little bit of code just checks that the config conforms. For special keys (like database with this "supported" vs "edge" concept) you could have a final check at the bottom just for that key.

If you like this idea, we could also have more advanced representations of "schema" written into the code. For example, including documentation into each item by making it be a list of tuples instead of a list of strings would allow you to dump out guidance documentation when someone makes a mistake: 
```
schema = {
  'database': ['mysql', 'redis', 'postgres'],
  'os': ['linux', 'windows'],
  'classification': [('fullstack', 'A fullstack classification means the framework doex X,Y,and Z'),
                           ('micro', 'A micro classification means the framework does A,B,C things')]
}
```

zane-techempower:
I like the idea! I was already working with suggestions for some of the fields, but I like the idea of spitting out the list of options with some help text as you pointed out. I will submit a "round-2" pr later today.

PR #1662, opened 2015-06-09
Fixing host bind

sksamuel:
Scruffy was binding to localhost, when it should bind to all network interfaces.
This PR addresses an error in preview 11.


msmith-techempower:
Nice, LGTM!

PR #1667, opened 2015-06-10
update express to use latest jade

alubbe:
the old version of jade had a debug flag that was still enabled by default in production which this newer version disabled, so the performance increases quite significantly.


msmith-techempower:
Looks good; LGTM!

PR #1669, opened 2015-06-10
scala/http4s: server now bound to 0.0.0.0

mfirry:
server address caused error. now bound to 0.0.0.0


msmith-techempower:
Looks good; LGTM!

PR #1672, opened 2015-06-10
Simplified JSON serialization with Jackson.

nmihajlovski:
Fixed misuse of the slow internal JSON mapper, by using plain Jackson.


msmith-techempower:
Travis agreed - LGTM!

PR #1673, opened 2015-06-10
Enable MySQL logging during verify mode

zane-techempower:
These changes allow MySQL to log during verify mode. One script creates a temporary log file that mysql can access, and restarts MySQL to log to that file. Once the test is complete the log is moved into the results directory for the test, the same one with the respective `err.txt` and `out.txt` of the test:

    $ ls -la results/ec2/latest/logs/sailsjs
    total 372
    drwxrwxrwx 1 vagrant vagrant   4096 Jun 10 21:21 .
    drwxrwxrwx 1 vagrant vagrant   4096 Jun 10 21:17 ..
    -rwxrwxrwx 1 vagrant vagrant    687 Jun 10 21:21 err.txt
    -rwxrwxrwx 1 vagrant vagrant 265576 Jun 10 21:21 mysql.log
    -rwxrwxrwx 1 vagrant vagrant  99660 Jun 10 21:21 out.txt

This change will allow a closer audit to be made on the database activities of a framework tests.

This change works if multiple tests are run at the same time, the results will be organized nicely in the same way as `err.txt` and `out.txt`.

Tasks:
- [x] Do initial work, get MySQL to log to a specific location then move that log to a test-specific place at the end of test
- [x] First round of feedback
- [ ] Ensure that the logic for doing the logging is `in_verify_mode && !in_travis_ci`

### DIFF
--- a/config/move-temp-mysql-log.sh
+++ b/config/move-temp-mysql-log.sh
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+cd "$(dirname "$0")"
+
+FROM=/tmp/tfb-temporary-mysql.log
+FINAL_DESTINATION=$1
+TEST_NAME=$2
+
+if [ -f $FROM ]
+then
+    cp /tmp/tfb-temporary-mysql.log $FINAL_DESTINATION
+    echo "Successfully moved MySQL temporary log file for $TEST_NAME" 
+else
+	echo "Expected MySQL log file at $FROM, but found none, no files copied"
+	return 1
+fi
+
+return 0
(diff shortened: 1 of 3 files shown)


msmith-techempower:
I like it. Let's see how this works...

mcocciaTE:
I have tested a few of the tests that failed and they so far have all worked locally for me.

msmith-techempower:
Let's postpone this until after round 11

hamiltont:
Great concept!! I'd expect major breakage on travis-ci that's subtle - in general it totally dies in weird ways when you swap out their config file (that's one big reason why I don't use `databases.sh` inside travis), so I'd say you might want to say "if verify and not in travis-ci"

msmith-techempower:
Agreed - let's ensure this doesn't happen in travis-ci.

msmith-techempower:
The underlying issue has been resolved in master, so I do not think this additional logging is necessary any longer.

Issue #1665, opened 2015-06-10
Incorrect use of Hibernate Session


denkab:
Hibernate session is not thread-safe, unlike JDBC Connection:

```
Server grizzly-jersey: java.util.concurrent.ExecutionException: java.lang.NullPointerException
Server grizzly-jersey: 	at java.util.concurrent.FutureTask.report(FutureTask.java:122)
Server grizzly-jersey: 	at java.util.concurrent.FutureTask.get(FutureTask.java:188)
Server grizzly-jersey: 	at hello.DbResource.db(DbResource.java:61)
Server grizzly-jersey: 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
Server grizzly-jersey: 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
Server grizzly-jersey: 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
Server grizzly-jersey: 	at java.lang.reflect.Method.invoke(Method.java:606)
Server grizzly-jersey: 	at com.sun.jersey.spi.container.JavaMethodInvokerFactory$1.invoke(JavaMethodInvokerFactory.java:60)
Server grizzly-jersey: 	at com.sun.jersey.server.impl.model.method.dispatch.AbstractResourceMethodDispatchProvider$ObjectOutInvoker._dispatch(AbstractResourceMethodDispatchProvider.java:258)
Server grizzly-jersey: 	at com.sun.jersey.server.impl.model.method.dispatch.ResourceJavaMethodDispatcher.dispatch(ResourceJavaMethodDispatcher.java:75)
Server grizzly-jersey: 	at com.sun.jersey.server.impl.uri.rules.HttpMethodRule.accept(HttpMethodRule.java:302)
Server grizzly-jersey: 	at com.sun.jersey.server.impl.uri.rules.ResourceClassRule.accept(ResourceClassRule.java:108)
Server grizzly-jersey: 	at com.sun.jersey.server.impl.uri.rules.RightHandPathRule.accept(RightHandPathRule.java:147)
Server grizzly-jersey: 	at com.sun.jersey.server.impl.uri.rules.RootResourceClassesRule.accept(RootResourceClassesRule.java:84)
[...]

Issue #1663, opened 2015-06-10
NPE in Grizzly when no parameters


denkab:
Round 11 preview 1:

Server grizzly-jersey: java.lang.NumberFormatException: null
Server grizzly-jersey: 	at java.lang.Integer.parseInt(Integer.java:454)
Server grizzly-jersey: 	at java.lang.Integer.parseInt(Integer.java:527)
Server grizzly-jersey: 	at hello.DbResource.getQueries(DbResource.java:70)

denkab:
I broke it, so I'll be fixing it.

PR #1674, opened 2015-06-11
Fix parameter parsing and Hibernate Session use

denkab:


denkab:
#1663 and #1665 are addressed here

mcocciaTE:
If you put 
'''
final Session session = sessionFactory.openSession();
session.setDefaultReadOnly(true);
'''
back into the main body of the function the query test will pass instead of giving a warning and fortune will pass instead of timing out and failing.

denkab:
@mcocciaTE , can you please elaborate why? I honestly don't understand, as fortune code was not changed and is not using the code path that is being worked on.

PR #1675, opened 2015-06-11
Clarify what happens when you pass `--test`

pbowyer:
I didn't understand how to test one framework and not install everything. While the docs gave the correct commands, this wasn't clear to me what was happening.

I have clarified it based on my converstaion with TE-msmith on IRC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,10 +43,10 @@ required.
 
         $ vagrant ssh
 
-5. Run a test.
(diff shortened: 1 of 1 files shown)


msmith-techempower:
Need to replace the `README.md` with ["look at readthedocs"](http://frameworkbenchmarks.readthedocs.org/en/latest/)

PR #1677, opened 2015-06-11
Finch added

mfirry:
I've added code for [finch](https://github.com/finagle/finch).


mfirry:
Can't really get the error from travis

msmith-techempower:
Finch passed; LGTM!

PR #1678, opened 2015-06-12
bump express' jade dep to 1.11.0

alubbe:
improves string escaping performance


mcocciaTE:
Worked locally for me. Also lots of different frameworks on travis failed so this might need a rerun on travis?

Issue #1661, opened 2015-06-09
Databases unstable on Travis


zane-techempower:
When verifying PR's we are getting a lot of framework fails simply because databases are not available to connect to. Here are some connection results from a recent Travis run. The other frameworks have all the databases up for the most part. On a previous run I had a test failing when only mysql failed to start.

```
--------------------------------------------------------------------------------
  Running Test: wt
--------------------------------------------------------------------------------
database connection test results:
mysql: is GO!
mongodb: is _NO_ GO!: ERROR: [Errno 111] Connection refused
redis: is GO!
postgresql: is _NO_ GO!: ERROR: [Errno 111] Connection refused
cassandra: is _NO_ GO!: ERROR: [Errno 111] Connection refused
elasticsearch: is _NO_ GO!: ERROR: [Errno 111] Connection refused
```

```
--------------------------------------------------------------------------------
  Running Test: wt-postgres
--------------------------------------------------------------------------------
database connection test results:
mysql: is GO!
mongodb: is _NO_ GO!: ERROR: [Errno 111] Connection refused
redis: is GO!
postgresql: is _NO_ GO!: ERROR: [Errno 111] Connection refused
cassandra: is GO!
elasticsearch: is _NO_ GO!: ERROR: [Errno 111] Connection refused
```

Very common for cassandra to fail
```
--------------------------------------------------------------------------------
  Running Test: gemini-mysql
--------------------------------------------------------------------------------
database connection test results:
mysql: is GO!
mongodb: is GO!
redis: is GO!
postgresql: is GO!
cassandra: is _NO_ GO!: ERROR: [Errno 111] Connection refused
elasticsearch: is GO!
```

In job https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/66095360 all databases have started up, however when the tests start, the server reads `null`s from redis. From my experience this means that Redis is down or not available. May suggest instability during test runtime.

Same thing happens in https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/66095480 where is said to be GO but then all the redis tests fail in a way that suggests the database was inoperable.

Two main issues:
* Databases sometimes do not start for the tests
* Redis specifically may be unavailable even when it is reported to be operational when Travis tests it at the start of each run.

zane-techempower:
This is related to #1604

hamiltont:
[...]

PR #1679, opened 2015-06-12
Travis has not been configured properly for redis

hamiltont:
See my comment at https://github.com/TechEmpower/FrameworkBenchmarks/issues/1661#issuecomment-111341509

### DIFF
--- a/toolset/run-ci.py
+++ b/toolset/run-ci.py
@@ -34,7 +34,8 @@ class CIRunnner:
   Only verifies the first test in each directory 
   '''
 
-  SUPPORTED_DATABASES = "mysql postgres mongodb cassandra elasticsearch sqlite redis none".split()
+  # Note: This list should not be updated without also updating the run_travis_setup function
+  SUPPORTED_DATABASES = "mysql postgres mongodb cassandra elasticsearch sqlite none".split()
   
   def __init__(self, mode, testdir=None):
     '''


zane-techempower:
I took a look around w.r.t. getting Redis working on travis. It seems that we could use `Redis 2.8.4` on travis fairly easily: http://docs.travis-ci.com/user/database-setup/#Redis
This is the version of `redis-server` that gets installed with `apt-get` so I'm assuming that that's what Travis will do if the option is activated. 

What sort of blockers are there to upgrading TFB to Redis 2.8.4? I wouldn't mind undertaking adding Redis, but I suspect there are issues beyond my knowledge.

hamiltont:
> I took a look around w.r.t. getting Redis working on travis. It seems that we could use Redis 2.8.4 on travis fairly easily: http://docs.travis-ci.com/user/database-setup/#Redis

Actually it's not too hard. Looking [here](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/database.sh#L278), you *don't* want to overwrite the configuration they use on travis, but you *do* want to run the bash script to populate the database. Remember - for travis we don't care about performance tuning, we just want it to turn on and have the correct data when the verification runs. 

> What sort of blockers are there to upgrading TFB to Redis 2.8.4?

From my understanding, the upgrade *might* break some of our configuration options because they occasionally change the valid options (and redis refuses to start with a bad config, IIRC). You would also probably have to modify [inside here](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/database.sh#L248) a little bit to keep our minimal database support on ubuntu 12.04.

zane-techempower:
Sounds good to me, thanks for the tips. I will put this on my todo list.

msmith-techempower:
I BELIEVE this was resolved with our latest Travis-CI updates to master.

@ssmith-techempower Please confirm

blee-techempower:
"Travis has not been configured properly for redis." which is right.

For this reason, Crystal/Moonshine, hapi, nodejs and other frameworks which use redis fail the Travis test. 

So, I edited " .travis.yml " like this: https://github.com/TechEmpower/FrameworkBenchmarks/commit/9e61a671fb0cff32c2c5748a9498286f8bf182bd

After this change, the frameworks which fail travis test due to redis pass travis test.

kbrock:
@blee-techempower this is great. thanks so much

PR #1680, opened 2015-06-12
try to fix using of FIFO scheduling policy

stefanocasazza:
I don't understand why I am not able to use FIFO scheduling policy, as happen by looking at travis stderr:
Server ULib: userver_tcp: WARNING: 04/06/15 07:41:46 (pid 31152) Cannot set posix realtime scheduling policy
I hope to have more information with this change...

Greetings


stefanocasazza:
I give up, now travis is broken etc..

Issue #1511, opened 2015-04-09
Add testing to TFB toolset


hamiltont:
Not sure what kind of testing or what library, so any suggestions welcome. Main goal is to improve reliability of TFB toolset, secondary goal would be to help modularlize the toolset in a manner that better enables reuse of the codebase by third-party programs. 

Some python testing frameworks (unsorted, feel free to share more or advocate for any of these): 
- [hypothesis](https://hypothesis.readthedocs.org/en/latest/)

LadyMozzarella:
For Reference: [Python Testing Tools Taxonomy](https://wiki.python.org/moin/PythonTestingToolsTaxonomy)

Looks like there are several options for us and I don't know anything about unit testing in Python. Based on quickly skimming the internet, it appears that [pytest](https://pytest.org/latest/contents.html) is a popular choice.

hamiltont:
[pytest](http://pytest.org/latest/), [unittest](https://docs.python.org/2/library/unittest.html)

nbrady-techempower:
Moved this to the toolset roadmap https://github.com/TechEmpower/FrameworkBenchmarks/projects/1

PR #1681, opened 2015-06-12
Benchmark config warnings 2

zane-techempower:
_Changes prompted by conversation in the [previous pull request](https://github.com/TechEmpower/FrameworkBenchmarks/pull/1658#issuecomment-111344061)_

Changes:
- [x] Improve the Python of the area I was working on, make it conform more closely with PEP8, clean up comments, improve variable names in some cases
- [x] Use a schema dict to define required fields and help messages in the event that they are absent or non-conforming (main motivation for changes)
- [x] Break `parse_config` into smaller functions
- [ ] Improve quality of help messages (not fully done, some are still `"..."`, others can be improved with some feedback)

To test:
1.  Pick a couple `benchmark_config.json` files to use, different languages with more than one test type is ideal.
2.  The validation is done on the whole project from what I've seen, so if the `benchmark_config.json` for `Java/gemini` is broken, you will still get errors if you are running `toolset/run-tests.py --mode verify --test sailsjs`. This is nice and can save some typing.
3.  The test urls `["json_url", "db_url", "query_url", "fortune_url", "update_url", "plaintext_url"]` must begin with a `/` character, some nice test cases are `"."`, `""`, `"db"`, `"queries/"`. See that an exception and suggested url is generated. Omitting these fields is not a test case as they are never required.
4.  The `["language", "webserver", "platform"]` fields are required, but free-entry, so they must be any string that is non-empty. So cases like `"JavaSScript"` and `"Pyton"` will work. Test will empty string to ensure that error and help message come up.
5.1.  The `["classification", "database", "approach", "orm", "os", "database_os"]` fields are required but only accept a set of options. In several `benchmark_configs` of choice, set these keys to values that are clearly not apart of set e.g.:
    * `"classification": "quite superb"`
    * `"database": "HoldsStuffDB"`
    * `"orm": ""`
    * `"os": "Yosemite"`
See that a list of possible values plus short descriptors for each are presented.
5.2.  These fields are not case-sensitive, try some acceptable, un-acceptably cased values and see that they work e.g.:
    * `"database": "PoSTGRES"`
    * `"os": "LiNuX"`
6.  Finally take a look at the `schema` dict in `framework_test.py` (look through files changed) and examine it. I have done my best with the help comments so far but I plan for another commit of improvements before this gets merged.

Starting this PR now to let travis have at it over the weeked. Best-case scenario is I only need to complete the help strings.

### DIFF
--- a/toolset/benchmark/framework_test.py
+++ b/toolset/benchmark/framework_test.py
@@ -839,165 +839,148 @@ class FrameworkTest:
 # End FrameworkTest
 ############################################################
 
-##########################################################################################
-# Static methods
-##########################################################################################
 
-##############################################################
-# parse_config(config, directory, benchmarker)
-# parses a config file and returns a list of FrameworkTest
-# objects based on that config file.
-##############################################################
+# Static methods
+
+def test_order(type_name):
+  """
+  This sort ordering is set up specifically to return the length
+  of the test name. There were SO many problems involved with
+  'plaintext' being run first (rather, just not last) that we
+  needed to ensure that it was run last for every framework.
+  """
+  return len(type_name)
+
+def validate_test(test_name, test_keys, directory):
+  """
+  Validate benchmark config values for this test based on a schema
+  """
+  # Ensure that each FrameworkTest has a framework property, inheriting from top-level if not
+  if not test_keys['framework']:
+    test_keys['framework'] = config['framework']
+
+  recommended_lang = directory.split('/')[-2]
+  windows_url = "https://github.com/TechEmpower/FrameworkBenchmarks/milestones/Windows%%20Compatibility"
+  schema = {
+    'language': {
+      'help': ('language', 'The language of the framework used, suggestion: %s' % recommended_lang)
+    },
+    'webserver': {
+      'help': ('webserver', 'Name of the webserver also referred to as the "front-end server"')
+    },
+    'classification': {
+      'allowed': [
+        ('Fullstack', '...'),
+        ('Micro', '...'),
+        ('Platform', '...')
+      ]
+    },
+    'database': {
+      'allowed': [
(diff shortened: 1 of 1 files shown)


hamiltont:
Excellent!!! 

> Break parse_config into smaller functions

It might even be worth sending in a pr version 3 on top of this one that pulls this bit into an independent file and adds some simple testing against that file! See #1511 - parsing the configs is a nice easy place to start adding some unit tests to the codebase

zane-techempower:
@hamiltont are you suggesting that we take the config sanitizing function and move that to a suite where other unit tests will go?

Or to write it so that `benchmark_config.json` gets broken up into test-specific configs e.g. `node-default-test-config.json`, `node-mongodb-raw-test-config.json` and so on. And then test from that perspective?

What sort of TFB test cases do you have in mind? That may help me understand the next set of changes.

hamiltont:
> I get it now, the idea is to make the TFB toolset itself testable

Yup, right on!

> How do you want unit tests to work? I think that separate command

Sounds good to me. I'd say to do a small bit of searching for a decent python testing framework like pytest or unittest, and just use whatever approach they have. We can always modify later if needed. 

By the way - the really cool part is this: Once you have some unit tests running, you can add a "special" TESTDIR entry to our `.travis.yml`, like `TESTDIR=run_toolset_tests`, then modify `run-ci.py` to pick up on that special entry and run the actual testing framework.....and boom! goes the dyno-mite....we finally have some automated unit testing running **against our toolset**.

zane-techempower:
Woo, I am all-in on this testing biz.

This PR still needs help messages to be finalized. The start of unit test layout will likely come in #1685 for those verification methods.

hamiltont:
Amazing work, I'm loving the code changes!

PR #1684, opened 2015-06-16
Update Tapestry

mcocciaTE:


mcocciaTE:
Woops wanted to test on my own fork first but this is to fix/replace #1548 if passes I believe #1547 can be merged as well and #1548 can be closed

toolset/setup/linux/webservers/resinJava8.sh
```diff
@@ -0,0 +1,24 @@
...
+make
+make install
+
+mv conf/resin.properties conf/resin.properties.orig
+cat $FWROOT/config/resin.properties > conf/resin.properties
+
+mv conf/resin.xml conf/resin.xml.orig
+cat $FWROOT/config/resin.xml > conf/resin.xml
+
+touch ${IROOT}/resin-$RVER.installed
```
zane-techempower: A little nit-pick: `resin-java8` would conform a bit better to install file conventions.

Convention is also to have the install name match the `.installed` name. There may be issues since `resinJava8` != `resin-4.0.41`. Not catastrophic, but this will cause the toolset to always install `resinJava8` even if it is already installed.

Recommendation: use `resin-java8` and touch `resin-java8.installed` on this commented line.

mcocciaTE: done :)

frameworks/Java/tapestry/install.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-fw_depends java7 resin maven
+fw_depends java8 resin-java8 maven
```
msmith-techempower: Wat

toolset/setup/linux/webservers/resin-java8.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+RVER=4.0.41
+
+RETCODE=$(fw_exists ${IROOT}/resin-java8-$RVER.installed)
+[ ! "$RETCODE" == 0 ] || { return 0; }
+
+fw_depends java8
+sudo cp -r $JAVA_HOME/include $JAVA_HOME/jre/bin/
```
msmith-techempower: Oh god, no. Let's not do this anymore... this version thing is officially untenable. I see no difference between this file you have added and [this file](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/webservers/resin.sh) aside from depending on Java8. To me, that reads "resin will work with java8 or java7"

Question - if you were to leave resin alone and still have `source $IROOT/java8.installed` in the setup file, will it run?

frameworks/Java/tapestry/install.sh
hamiltont: Agreed. If we have to lock down the java version prior to installing resin, it would be better to just remove the line `fw_depends java7` from `resin.sh` and rely on the caller explicitly setting the java version desired via `fw_depends java7 resin` or `fw_depends java8 resin`, instead of creating `resin-java8.sh`. Perhaps we could replace `fw_depends java7` with a check that either `java7.installed` or `java8.installed` exists and fail if that is not true just as one sanity check

msmith-techempower: We discussed this offline and apparently resin needs to be compiled with the same java runtime that is used to invoke it. So, this `resin-java8` must exist unless we move EVERYTHING to java8 explicitly. Very annoying.

mcocciaTE:
looks like a lot of unrelated tests failed in travis.

blee-techempower:
Localtest failed. ( http://sprunge.us/fUNM )
Please resolve the issue.

frameworks/Java/tapestry/setup.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 # load java environment variables
-source $IROOT/java7.installed
+source $IROOT/java8.installed
 export RESIN_HOME=${IROOT}/resin-4.0.41
```
msmith-techempower: These two lines need to be replaced by `fw_depends java8 resin-java8`

frameworks/Java/tapestry/install.sh
msmith-techempower: This file needs to be removed completely.

toolset/setup/linux/webservers/resin-java8.sh
```diff
@@ -0,0 +1,24 @@
...
+make
+make install
+
+mv conf/resin.properties conf/resin.properties.orig
+cat $FWROOT/config/resin.properties > conf/resin.properties
+
+mv conf/resin.xml conf/resin.xml.orig
+cat $FWROOT/config/resin.xml > conf/resin.xml
+
+touch ${IROOT}/resin-java8.installed
```
msmith-techempower: This file needs to be updated to better match the latest `toolset/setup/linux/webservers/resin.sh`'s dependency structure.

msmith-techempower:
@ssmith-techempower See if you can make a branch off of this PR and get it updated to work.

msmith-techempower:
Unrelated: does anyone know why this PR is called "Based2 patch 2"?

msmith-techempower:
Closing in favor of #1816

Issue #1568, opened 2015-04-23
Replace curl with python requests library


hamiltont:
Or something similar.  See http://docs.python-requests.org/en/latest/

Avoids having to have curl installed on host, which will be a barrier to entry for windows

zane-techempower:
I'm using requests on a side project. It's great! Don't know of any better options.

GMLudo:
Personally, I use aiohttp.client for concurrency support with Requests-like API. If you don't need to launch several HTTP requests in the same time, requests is enough.

zane-techempower:
That's really good to know. From what I know, these requests are just used to assert that responses look at certain way i.e. unit test them. The load-testing requests are done by wrk.

GMLudo:
@zane-techempower: Then, Requests it's enough.
Unless you want to speed-up the unit-tests and launch concurrently all requests at the same time for one framework, but it must be benchmarked.
If you had a lot of URLs to validate at the same time, `aiohttp.client` is clearly a winner compare to Requests, I've made a CouchDB crawler and I've benchmarked that.
But, for only 5 requests, especially if the HTTP client is executed on the same machine as the HTTP server, Requests should be quicker.
If you're interested in that I'm explore that, or you want some others tips, don't hesitate to contact me, it's my main day job to optimize Python applications.

msmith-techempower:
We aren't really interested in speeding up the verification process, which is where `curl` is used, and doing the requests in parallel doesn't really get us much more than that.

I suspect that `requests` (or similar) will be a fine replacement.

PR #1685, opened 2015-06-16
Replace curl usage with requests

### LINKED ISSUE #1600: Incorrect 'accessing url' output
Here is the bug - note that the Accessing URL field never changes, but somehow the response appears to be different!

```
Accessing URL http://127.0.0.1:8080/index.php/bench/db/:
  Response (trimmed to 40 bytes): "[{"id":778,"randomNumber":5795},{"id":30"
Accessing URL http://127.0.0.1:8080/index.php/bench/db/:
  Response (trimmed to 40 bytes): "{"id":5754,"randomNumber":1388}"
Accessing URL http://127.0.0.1:8080/index.php/bench/db/:
  Response (trimmed to 40 bytes): "{"id":8660,"randomNumber":4689}"
Accessing URL http://127.0.0.1:8080/index.php/bench/db/:
  Response (trimmed to 40 bytes): "[{"id":1995,"randomNumber":2759},{"id":3"
   WARN for http://127.0.0.1:8080/index.php/bench/db/0
     Top-level JSON is an object, not an array
   WARN for http://127.0.0.1:8080/index.php/bench/db/foo
     Top-level JSON is an object, not an array
```

The problem is that we [print "Accessing URL" in _curl](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/benchmark/test_types/framework_test_type.py#L67), but the [calls to _curl do not include the full URL](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/benchmark/test_types/query_type.py#L27), only the calls to `_curl_body`. This is a mistake - _curl needs the full URL passed to it, just in case the returned headers are different

### LINKED ISSUE #1625: Warn for incorrect field counts on response JSON
Noticed when trying to get `vibed` working; example response for `DB` test:

`[{"randomNumber":3634,"_id":6208,"id":6208}]`

While this technically passes our rules, it should be warned because the additional field is going to impact the results (more data sent means fewer possible responses).

zane-techempower:
_Early PR for a chunk of in-progress work_

These changes are centered around addressing #1568 which aims to use the native, and excellent Python [requests](http://docs.python-requests.org/en/latest/) package which prevents the need of using `curl` through Python. Notably this will improve Windows compatibility by removing the curl requirement. These changes touch the framework test types and their super class `framework_test_type` (which previously had the `_curl` and `_curl_response_body` methods). I have taken the liberty to improve the formatting of the files that I touch to be more closely in line with PEP8 in the interest of a cleaner TFB toolset.

- May fix #1600, I am checking that url's are correctly formed, and I am working in that area of files
- Fixes #1625 
- Will help address #1511 by breaking verification functionality into smaller and easier-to-test chunks. I have been keeping unit testing in mind as I've been working through this.
- Addresses #1534 by giving better warning messages. I have addressed concerns about [frameworks including charset in `Content-Type` field](https://github.com/TechEmpower/FrameworkBenchmarks/issues/1534#issuecomment-95766652) by adding an explicit check for the case. This makes the toolset `WARN`-city for certain frameworks that always append the charset.

- [x] Replace `_curl` and `_curl_response_body` with a similar method that utilizes the `requests` module
- [x] Go through test type files and:
    - Change their header checks for `requests`
    - Change their body checks for `requests`
    - Refactor checking logic to be more informative (mostly for `WARN`'s) and helpful to developers trying to add a framework
    - PEP 8 them, 4 space indentation, formatting, break into smaller functions
- [x] Refactor repeated/shared code
- [x] Refactor verification functions to be more grouped together
- [x] All test types inherit from `FrameworkTestType` directly (previously queries and updates were inheriting from `DBTestType` in order to share validation functions)
- [ ] Implement unit tests for the verification of headers and response bodies

This is Windows Compatibility / Round 12 work

### DIFF
--- a/frameworks/Crystal/moonshine/server.cr
+++ b/frameworks/Crystal/moonshine/server.cr
@@ -14,9 +14,9 @@ app = App.new
 
 class CONTENT
   UTF8 = "; charset=UTF-8"
-  JSON = "application/json" + UTF8
+  JSON = "application/json" #+ UTF8
   PLAIN = "text/plain"
-  HTML = "text/html" + UTF8
+  HTML = "text/html" #+ UTF8
 end
 
 ID_MAXIMUM = 10_000
--- a/toolset/benchmark/framework_test.py
+++ b/toolset/benchmark/framework_test.py
@@ -839,165 +839,148 @@ class FrameworkTest:
 # End FrameworkTest
 ############################################################
 
-##########################################################################################
-# Static methods
-##########################################################################################
 
-##############################################################
-# parse_config(config, directory, benchmarker)
-# parses a config file and returns a list of FrameworkTest
-# objects based on that config file.
-##############################################################
+# Static methods
+
+def test_order(type_name):
+  """
+  This sort ordering is set up specifically to return the length
+  of the test name. There were SO many problems involved with
+  'plaintext' being run first (rather, just not last) that we
+  needed to ensure that it was run last for every framework.
+  """
+  return len(type_name)
+
+def validate_test(test_name, test_keys, directory):
+  """
+  Validate benchmark config values for this test based on a schema
+  """
+  # Ensure that each FrameworkTest has a framework property, inheriting from top-level if not
+  if not test_keys['framework']:
+    test_keys['framework'] = config['framework']
+
+  recommended_lang = directory.split('/')[-2]
+  windows_url = "https://github.com/TechEmpower/FrameworkBenchmarks/milestones/Windows%%20Compatibility"
+  schema = {
+    'language': {
+      'help': ('language', 'The language of the framework used, suggestion: %s' % recommended_lang)
+    },
+    'webserver': {
+      'help': ('webserver', 'Name of the webserver also referred to as the "front-end server"')
+    },
+    'classification': {
+      'allowed': [
+        ('Fullstack', '...'),
+        ('Micro', '...'),
+        ('Platform', '...')
+      ]
+    },
+    'database': {
+      'allowed': [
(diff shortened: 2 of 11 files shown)


hamiltont:
!!! Very cool !!!

My only nitpick - why don't we use `verifyBody` instead of `verifyObject` for the method name - it seems more precise to me as the return value is not always a JSON object e.g. plaintext

BTW, I was thinking that we could add a few "common" verify* methods to framework_test_type to avoid code duplication, such as methods like `verify_response_is_json`, `verify_json_has_top_level_keys([key1,key2])`, etc. Thoughts?

zane-techempower:
> `verifyBody` instead of `verifyObject`

Sounds good to me

> BTW, I was thinking that we could add a few "common" verify* methods to framework_test_type to avoid code duplication, such as methods like `verify_response_is_json`, `verify_json_has_top_level_keys([key1,key2])`, etc. Thoughts?

Currently stuff isn't very DRY, I've just been working through the changes on each file. I will do a nice refactor job at the end to pull out the common verify methods like you mentioned. This will also make unit testing as easy as possible. Good suggestions and certainly where I want to go with this.

zane-techempower:
Starting tomorrow I will not be able to commit time to this task, or other tfb tasks due to being assigned a full-time project. Aside from lack of unit tests, I am happy with the current state of this PR. I have cleaned up the `test_types` directory and moved much of it's functionality to `verifications.py` where it is more detailed (better warning and failure messages) and easier to understand (code formatting/shorter functions/DRYer).

Hopefully I will have time to at least start/set up the testing suite for these changes before round 12. I will accept PR's though!

PR #1687, opened 2015-06-17
Add Dropwizard Postgres, Upgrade to DW 0.8.1

yunspace:
- Added postgres tests
- Added README
- Use Maven profiles to build different shaded jars for different tests
- Split DAO into interface and implementation.
- Removed unnecessary Optionals since we are converting them to Null anyway.
- Use int instead of long for ids

https://github.com/dropwizard/dropwizard/issues/1010


yunspace:
- Upgraded Dropwizard version to 0.8.1 which uses Jersey2. Local tests show it's 3 times faster than 0.7.1.
- Upgraded MongoJack to latest 2.3.0
- Fixed Update test from using a single Transaction for all updates to using individual Transactions for each read/write.

current_benchmark.txt
```diff
@@ -0,0 +1 @@
+dropwizard
```
msmith-techempower: Please remove this file. In fact, this file SHOULD be in `.gitignore`, but I will make that change elsewhere.

yunspace:
Oops, how did that sneak in there? Removed. Thanks for pointing this out.

msmith-techempower:
Travis is happy, I am happy... LGTM!

PR #1688, opened 2015-06-17
Mangooio

svenkubiak:
I've added a new framework for benchmark testing: mangoo I/O.

As I am to late for round 11, is there any rough schedule for round 12?

Cheers,
Sven


frameworks/Java/mangooio/src/main/resources/ehcache.xml
```diff
@@ -0,0 +1,13 @@
+<ehcache xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../config/ehcache.xsd" updateCheck="false">
```
martin-g: Is this used by the test?
AFAIK caching is not allowed.

svenkubiak: No, a Cache is not used at all. But the Framework has an configuration for ehCache which creates an instance of a Cache provider. I could remove the file, but ehCache would still autoconfigure.

mcocciaTE:
I believe you still need to add mangooio to the ./travis.yml if you could add it in alphabetical order that would be awesome.

msmith-techempower:
Travis agrees; LGTM!

PR #1691, opened 2015-06-19
Remove wget

msmith-techempower:
`wget` was failing to work on a number of https urls for older-than-ubuntu14 systems because it would attempt to use SSLv3 as the handshake, get turned down, and fail.

After attempting a number of options to fix `wget`, I fell back on using `curl` instead. It was just easier.


frameworks/PHP/php-laravel/setup_hhvm.sh
```diff
@@ -15,5 +15,5 @@ sed -i 's|/home/vagrant/FrameworkBenchmarks/installs/nginx/|'"${IROOT}"'/nginx/|
 
 export PATH="${PHP_HOME}/bin:${PHP_HOME}/sbin:$PATH"
 
-hhvm --config $TROOT/deploy/config.hdf --user $(whoami) -m daemon
+hhvm -m daemon --config $TROOT/deploy/config.hdf --user $(whoami)
```
msmith-techempower: Just so we have some explanation on this SOMEWHERE: this was an annoying fix to `hhvm` simply ignoring input flags because of an incorrect order and STILL starting up but binding to completely different ports as a completely different user.

When `-m daemon` was at the end, `hhvm` would bind to 9000 as `www-user` regardless of the other flags, but moving it to the first argument (hint: not an argument, apparently) fixes this.

SO ANNOYING!!!

hamiltont:
LGTM!

frameworks/PHP/php-laravel/setup_hhvm.sh
hamiltont: How about a note on the PHP readme about this

PR #1692, opened 2015-06-21
go: GOGC=800

methane:
Default GOGC value (100) is bit conservative.
For high throughput servers, 400~800 is better.


hamiltont:
LGTM

PR #1703, opened 2015-06-30
Wicket: respond with error code when exception is caught

dashorst:
The tests currently fail silently for the Wicket project. This commit fixes the silent failure by returning a proper error code (500) and the exception message. This will allow the Wicket team to figure out why the tests fail.


martin-g:
Thanks, Martijn!

For less-noisy diff use `?w=1` in the URL for ignoring the whitespace, i.e. https://github.com/TechEmpower/FrameworkBenchmarks/pull/1703/files?diff=unified&w=1

msmith-techempower:
Looks good; LGTM!

PR #1711, opened 2015-07-07
klein, turbogears: Fix updating DB

### LINKED ISSUE #1708: Klein (and others) MUST NOT update the Fortune table
I have had to update the mysql "Fortune" table to be readonly for `benchmarkdbuser` because SOMEONE was writing to it (:cough: @knewmanTE :cough:).

Fix Klein to not attempt to write to the Fortune table.

methane:
* Use autoescape instead of bleach.clear
* Close db_session for each request

Fixes #1708

### DIFF
--- a/frameworks/Python/klein/app.py
+++ b/frameworks/Python/klein/app.py
@@ -1,23 +1,19 @@
 # -*- coding: utf-8 -*-
 
-import os
-import sys
-import json
-
-import bleach
-
-from random import randint
 from functools import partial
+import json
 from operator import attrgetter
-
-from klein import Klein, run, route
+import os
+from random import randint
+import sys
 
 from jinja2 import Environment, PackageLoader
-
-from sqlalchemy.ext.declarative import declarative_base
+from klein import Klein, run, route
 from sqlalchemy import create_engine, Column
-from sqlalchemy.types import String, Integer, Unicode
+from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import sessionmaker
+from sqlalchemy.types import String, Integer, Unicode
+
 
 if sys.version_info[0] == 3:
     xrange = range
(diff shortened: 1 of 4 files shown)


methane:
My travis result is here: https://travis-ci.org/methane/FrameworkBenchmarks/builds/69886540

msmith-techempower:
Needs to be tested against #1605 
Also, doesn't seem to resolve the call to update the fortune table.

frameworks/Python/klein/app.py
```diff
@@ -67,57 +64,72 @@ def getQueryNum(queryString):
...
 
 @app.route("/fortune")
+@close_session
 def fortune(request):
-	request.setHeader("Content-Type", "text/html; charset=UTF-8")
-	fortunes = db_session.query(Fortune).all()
-	fortunes.append(Fortune(id=0, message="Additional fortune added at request time."))
-	fortunes.sort(key=attrgetter("message"))
-	for f in fortunes:
-		f.message = bleach.clean(f.message)
```
msmith-techempower: Oh, maybe it does address that here. You're saying that `bleach.clean(f.message)` might actually cause a db write? Weird.

methane: Yes. SQLAlchemy is "Unit of Work" ORM.
When you modify object which is managed by session and commit the session, all modified objects are updated.

There are no `db_session.commit()` in this request. But since this request doesn't remove session, the session alive
after this request.
When calling "/update" test, modified objects are actually written to DB.

Transaction over HTTP request is nightmare.  So, closing session for each request is required.

methane:
> Needs to be tested against #1605

#1605 seems don't touch klein and turbogears.
What I should do?

msmith-techempower:
Ah, good point - I think this is safe to merge as-is and I'll just merge it into #1605

msmith-techempower:
Confirmed that both klein and turbogears now successfully run AND don't update the mysql database. Thanks for the help @methane

PR #1714, opened 2015-07-11
Trying out the Travis CI beta setup

joshk:
Sorry for using this PR to test out the new setup we are working on, but I thought it would be a great test repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 sudo: 9000
+
 language: python
+
 python:
   - "2.7"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: required
+sudo: 9000
 language: python
 python:
   - "2.7"


joshk:
Ok, looks like our test image isn't ready yet, sorry for the noise.

hamiltont:
Ha ha, glad we could help out! We've noted a few times that we push travis-ci to the limits, happy that was a useful feature for someone ;-)

joshk:
Hey @hamiltont 

This new setup is still highly beta, but will allow you and your team to do some really awesome things on it. 

More to come soon!

joshk:
@hamiltont I was wondering, I see that each language has multiple sets of tests, eg. 

```
TESTDIR=C/lwan
TESTDIR=C/duda
TESTDIR=C/haywire
TESTDIR=C/onion
```

Are you able to combine these into one? 

Each VM we start, because you request sudo, takes between 40-60 seconds, but yet your tests only take 15 seconds, so we spend more time starting a VM then actually testing the code.

If you could merge these into one then it would be a great help to us.

joshk:
The new setup seems to work for you now, and is faster, and I think once we reduce the matrix size and combine some jobs together, like mentioned above, it will be even faster as we won't be doing a git clone each time.

hamiltont:
We've actually had direct teleconference with some people from travis a few times to discuss these things. The TLDR - it's not that simple ;-)

Each TESTDIR must run fully (and each will take up to 50 minutes) if something in our `toolset` directory changes. This happens quite a bit actually, you can easily notice which travis builds take 1 hour and which take ~20 hours - the long ones were when the underlying toolset changed. I programmed some fast-fail logic by examining the git log and deciding if that TESTDIR could be quickly aborted to save travis resources in those cases, but we can't just combine TESTDIRs or we will consistently hit the time limit when they need to actually run. IIRC in our telecon we tried out an increase in the time limit per job and even that didn't help much - unless our limit per job was multiple hours, at which point we might be able to say `TESTDIR=Java` (e.g. one TESTDIR per language). We are actually setup to do that (because I tried it), but one-testdir-per-language requires quite large job time limits. 

The best feature (for us reducing our usage) would be the ability to have one "matrix-building" job that runs first and decides what needs to be run. I'm subscribed to some issues on the travis repos that reference this desire, but it's probably a niche need. 

Sadly, until the container setup allows sudo, we can't utilize that for our testing. As you can imagine, it's hard to support a few hundred web frameworks without allowing sudo usage ;-) Once that can be done in your container infrastructure though I'm excited to switch - it is indeed a shame to launch a bunch of VMs, then have them check the git logs and abort themselves.

joshk:
Hi @hamiltont 

Thanks for the summary on what you and Dan have talked about.

This new setup we are working on has better boot times (compared to our legacy setup), allows for sudo, and has more cores and memory. I would still be tempted to see what a full test run is like, and what it would look like if we merged some of the jobs together, for example sake on the new setup.

This new setup will also make it easier for us to increase your job limit as it will be using a cloud provider.

There is another option available to you, it won't be perfect, and we may need to add some features, but we have a requests API where you can post a custom Travis config in a json payload. This would mean you could control what is run yourself.

If you would like to know more, ping me at support@travis-ci.com and we can arrange a follow up call to talk more.

Josh

methane:
@joshk Today Travis-CI announce docker support.  While we don't use docker for now, we use Ubuntu 14.04 for benchmarking environment.
Can we use following to use Ubuntu 14.04?

```yaml
sudo: required
services:
    - docker
```

Or should we merge this PR?

msmith-techempower:
I have no idea whether we should merge this or not. @hamiltont if you have a moment, can you weigh in?

msmith-techempower:
After messing with travis a lot more lately, I am inclined to close this PR.

PR #1715, opened 2015-07-11
Http4s updates

fiadliel:
Fix plaintext, json tests

Add db, queries, updates, fortunes tests
 - uses doobie/postgresql for database queries
 - uses HikariCP for connection pooling
 - uses argonaut-shapeless for class-JSON conversion
 - uses Twirl for HTML templating

Add support for DBHOST environment variable


blee-techempower:
Passed localtest.

```
 Verification Summary

| Test: http4s
|       fortune     : PASS
|       plaintext   : PASS
|       db          : PASS
|       update      : PASS
|       json        : PASS
|       query       : PASS
```

hamiltont:
@blee-techempower Just as a sanity check, are you running each different local verification in a new virtual machine? We've had substantial issues in the past with this, because one framework might install some package or change some setting that will then affect any later framework verifications. The easy thing here is to setup a brand new VM, do a snapshot, and then always rollback the snapshot before doing a local verification. If you're not doing that then the PASS/FAIL results can't be trusted

msmith-techempower:
@hamiltont He is running each different local verification on the same virtual machine (just like I do); although, you are right and I probably SHOULD snapshot my VM after initial setup and roll back after each test.

hamiltont:
Highly recommend you do ;-) It's super simple to get started with the snapshots and will save you from the world of pain that happens when frameworks decide to conflict (not to mention the slow creep of tech debt that happens if broken PRs are accidentally accepted)

msmith-techempower:
Your current employer has you using terms like "tech debt"? Such words make me sad.

I agree with the overall sentiment, though. Ensuring a clean VM on every run locally makes sense. I probably still won't do it unless I know there are potentially conflicting issues.

Additionally, with the shift to (at least attempting) having every software be installed in `$IROOT`, I DO practice deleting that folder between tests, and usually that and a quick spot-check of the `fw_depends` of the test is enough to ensure a clean environment from test to test.

PR #1720, opened 2015-08-07
Include Comsat Servlet framework support

circlespainter:
Could you also help me with some questions? https://groups.google.com/forum/#!topic/framework-benchmarks/Gujdipuua7Q


circlespainter:
It actually runs fine in my Vagrant env (tried with the latest `ubuntu/trusty64` box but also previously) but it fails on CI; it looks like the `java8` dependency can't be setup for some reason:

```
'ERROR: /toolset/setup/linux/languages/java8.sh: Command '\''sudo update-alternatives --set java $J7_HOME/jre/bin/java'\'' exited with status 2 (dependency=java8) (cwd=$FWROOT/installs)'
Setup comsat-webactors-servlet-tomcat: ERROR: /toolset/setup/linux/languages/java8.sh: Command 'sudo update-alternatives --set java $J7_HOME/jre/bin/java' exited with status 2 (dependency=java8) (cwd=$FWROOT/installs)
```

Do you see anything in the framework's build source that could cause this (apart from the dependency to java8 - which is there only because it's more performant, although the framework support java7 too)?

blee-techempower:
Passed localtest. 
```
  Verification Summary

| Test: comsat-servlet
|       plaintext   : PASS
|       json        : PASS

```

PR #1722, opened 2015-08-10
Tornado/Postgres better utilization of IOLoop

Dith3r:
Current code don't utilize asynchronous features. Proposal is to fire queries without waiting for single result.


blee-techempower:
Passed localtest.
```
 Verification Summary

| Test: tornado
|       plaintext   : PASS
|       query       : PASS
|       json        : PASS
|       db          : PASS
```

Issue #1521, opened 2015-04-13
Php yaf: Separate install from PHP


LadyMozzarella:
As mentioned in PR #1516, [yaf is installed inside of the PHP install file](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/languages/php.sh#L49). Move this to a separate install file. Keep in mind that other frameworks may depend on yaf.

```bash
# yaf.so
 printf "\n" | $IROOT/php-${VERSION}/bin/pecl -q install -f yaf
```

msmith-techempower:
I will probably be addressing this in my PR that removes `install.sh` files.

LadyMozzarella:
Okay. Will assign to you then.

PR #1734, opened 2015-08-30
update PHP Fat-Free Framework

ikkez:
fixed an issue found in the logs and updated to v3.5.0.


blee-techempower:
All test faild.

msmith-techempower:
Ah, the folder names all got changed for round 11.

Basically, I would suggest closing this pull request and opening a new one off of master.

ikkez:
Alright. I've updated the branch. It should be able to merge now.

ikkez:
hm. Is this CI check related to this Framework test? What's the problem?

blee-techempower:
Localtest failed. Here is log. ( http://sprunge.us/SOQY )
Please resolve this issue, so this branch can be merged.

ikkez:
Well actually I have no idea why it's not working. It worked in Round 10 before. What was changed? 
I've adjusted a path in the nginx config now. hopefully that was the issue. 
In the log you showed me, I noticed this notice:
```
Setup fat-free: [20-Oct-2015 15:29:49] NOTICE: PHP message: PHP Warning:  PHP Startup: Unable to load dynamic library '/home/blee/FrameworkBenchmarks/installs/php-5.5.17/lib/php/extensions/no-debug-non-zts-20121212/yaf.so' - /home/blee/FrameworkBenchmarks/installs/php-5.5.17/lib/php/extensions/no-debug-non-zts-20121212/yaf.so: cannot open shared object file: No such file or directory in Unknown on line 0
```
is that relevant? It does not seem to belong to our framework.

hamiltont:
@msmith-techempower Is there a readme we can point to that helps explain all the reasons why we don't like PRs with the source code for the entire framework included? 

@ikkez Some context: all other frameworks (besides the 5-10 that have yet been updated, incl. fatfree) do not put their entire framework source code directly in the TFB repo, that is downloaded by the setup.sh script. Is this something you could add to this PR?

msmith-techempower:
@hamiltont Not to my knowledge... but there isn't much of that anyway.

hamiltont:
not anymore ;-)

ikkez:
@hamiltont really? As far as I can see others have included the source files as well, and fat-free is one of those with a relative small code base (~150kb, hand-picked for this tests). I can remove the changelog file to save additional 30kb if you want. I am currently more interested in getting the test-suite running instead of setup optimizations. I guess I'm lost with both tasks.

Rayne:
My Vagrant box returns the following result with https://github.com/F3Community/FrameworkBenchmarks/commit/4c034236ff0559678fd20c2931a4af0d9a9b8fef:

```
toolset/run-tests.py --install server --mode verify --test fat-free

[…]

  Verification Summary

| Test: fat-free
|       plaintext   : WARN
|       query       : PASS
|       json        : PASS
|       db          : WARN
|       update      : PASS
```

The Travis CI runner instead fails because of `./toolset/run-ci.py prereq "$TESTDIR"` terminating with exit code `1` ([run-tests.py#L176-L186](https://github.com/TechEmpower/FrameworkBenchmarks/blob/339022960c4df60d86c393a27998966672dbab4e/toolset/run-tests.py#L176-L186)).

I haven't found a solution in the documentation or a similar problem in Google Groups. Do you have an idea how to fix this problem?

Also I would like to reset our `master` branch to create a new pull request (as suggested by @msmith-techempower) on a dedicated branch (probably [this branch](https://github.com/Rayne/FrameworkBenchmarks/tree/fat-free-framework-setup)) after resolving the mentioned problems.

ikkez:
> Also I would like to reset our master branch... as suggested by msmith-techempower

@Rayne I did that already and force-pushed it into the current master that is used in this pull-request. These should be no merge conflicts anymore.

Rayne:
`Java/sabina` has the same problem.

* PR https://github.com/TechEmpower/FrameworkBenchmarks/pull/1761
* [Travis CI](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/91083838)

@ikkez You are right that there aren't merge conflicts, but our `master` branch isn't deployable and diverges from TechEmpower's `master` branch.

> There's only one rule: anything in the master branch is always deployable.
>
> Because of this, it's extremely important that your new branch is created off of master when working on a feature or a fix.
>
> Source: https://guides.github.com/introduction/flow/index.html

msmith-techempower:
@ikkez @Rayne We have recently made a number of fixes for Travis-CI on our master; can you please rebase and push your branch so that Travis-CI will automatically test for us?

ikkez:
it's rebased now.

msmith-techempower:
Awesome!

ikkez:
Can I find the complete travis ci error logs somewhere? I'm looking for the full response of thit test: https://travis-ci.org/F3Community/FrameworkBenchmarks/jobs/96214489#L1701

ikkez:
Fat-Free uses a strict error_reporting setting by default. Because Php yaf seems to be installed globally as a php module #1521 and randomly fails to load, our error handler is triggered and shouts things like this:
> Server fat-free: 2015/12/11 15:04:49 [error] 28739#0: *2 FastCGI sent in stderr: "PHP message: Fatal error: PHP Startup: Unable to load dynamic library '/home/vagrant/FrameworkBenchmarks/installs/php-5.5.17/lib/php/extensions/no-debug-non-zts-20121212/yaf.so' - /home/vagrant/FrameworkBenchmarks/installs/php-5.5.17/lib/php/extensions/no-debug-non-zts-20121212/yaf.so: cannot open shared object file: No such file or directory" while reading response header from upstream, client: 127.0.0.1, server: localhost, request: "GET /db-orm HTTP/1.1", upstream: "fastcgi://127.0.0.1:9001", host: "127.0.0.1:8080"

Because of this, your test verification randomly fails... **sometimes**. Of course we don't need yaf, and we could probably adjust the error reporting level to get around this, but I don't see that our framework should clean up install requirements or ignore errors from others participants.

msmith-techempower:
Hmm, I'm not convinced that `yaf` is the problem.

@ssmith-techempower Could you pull this and test locally; at least this will output error logs that can be attached to the task to better understand the issue.

ikkez:
Fat-Free scans for already occurred system errors on its initialization. We already have a similar issue here https://github.com/bcosca/fatfree/issues/851 which reveals a server issue before F3 starts. A way to clear system errors is to use the `error_clear_last();` function, but it's only available with PHP7.

nbrady-techempower:
@ikkez we've added php7 to the toolset

msmith-techempower:
Additionally, @nbrady-techempower has done quite a bit of work getting the php installation into a more sensible state; please rebase off our latest master.

ikkez:
rebased and adjusted. Verification tests passed, but for any reason it stopped [somewhere]() without an error :-/

frameworks/PHP/fat-free/benchmark_config.json
```diff
@@ -43,6 +43,49 @@
...
+      "platform": "PHP-FPM",
+      "webserver": "nginx",
+      "os": "Linux",
+      "database_os": "Linux",
+      "display_name": "fat-free",
+      "notes": "",
+      "versus": "php"
+    },
+    "php5-raw": {
+      "setup_file": "setup",
```
msmith-techempower: This should be `"setup_file": "setup_php5"`

frameworks/PHP/fat-free/benchmark_config.json
```diff
@@ -43,6 +43,49 @@
       "display_name": "fat-free",
       "notes": "",
       "versus": "php"
+    },
+    "php5": {
+      "setup_file": "setup",
```
msmith-techempower: This should be `"setup_file": "setup_php5"`

ikkez: oh yes, indeed. thanks.

msmith-techempower:
I'm not quite sure what happened on Travis; it got into the `fat-free-php5` test, successfully installed php5 and nginx, then just stopped doing work entirely so the job timed out.

I see the line `nginx is installed!` in the Travis log, so it definitely finished its `fw_depends` calls, but the next call should have been `sed`, and it never got to that... curious.

ikkez:
hm, should I just try to drop the php5 test for now, to see if it works, or will it be out of comparison with other frameworks, which are not using php7 so far?

msmith-techempower:
Hmm, according to [the log](https://s3.amazonaws.com/archive.travis-ci.org/jobs/103807578/log.txt), it seems that there is still this timeout issue.

I'm not really sure what is going on here, but it's hanging up in Travis-CI.

msmith-techempower:
Seems that there is STILL a hang-up in Travis and I am not sure why.

@ssmith-techempower Can you test all the `fat-free` variants locally and report back?

ssmith-techempower:
It's timing out my local install as well. Gets to "`nginx is installed!`" and then hangs.

msmith-techempower:
Thanks - it's odd that of all the PHP tests, this is the only one hanging in this way.

frameworks/PHP/fat-free/setup.sh
```diff
@@ -1,12 +1,20 @@
...
-sed -i 's|.*/FrameworkBenchmarks/php-fatfree|'"${TROOT}"'|g' deploy/php
-sed -i 's|Directory .*/FrameworkBenchmarks/php-fatfree|Directory '"${TROOT}"'|g' deploy/php
-sed -i 's|root .*/FrameworkBenchmarks/php-fatfree|root '"${TROOT}"'|g' deploy/nginx.conf
+sed -i 's|root .*/FrameworkBenchmarks/fat-free|root '"${TROOT}"'|g' deploy/nginx.conf
 sed -i 's|/usr/local/nginx/|'"${IROOT}"'/nginx/|g' deploy/nginx.conf
 
+
+F3DIR="$TROOT/src"
+
+[[ ! -e "$F3DIR" ]] || rm -r "$F3DIR"
```
ssmith-techempower: The `rm` command here needs the `f` flag to force the `rm` to happen without bringing up a prompt, which was hanging Travis and my local install due to prompting about write-protected files.

ikkez:
looks good here https://travis-ci.org/F3Community/FrameworkBenchmarks/jobs/116216563
thank you all for the support :)

ssmith-techempower:
Looks good on my local install, and I trust your Travis job means it would succeed for us. Merging in.

PR #1742, opened 2015-09-22
Clean zend warnings on db calls

greg-hellings:
Adds in data type-casting to zend in order to clean up test warnings and return proper integer types instead of string types.


greg-hellings:
Travis failed this test because there is warning text in the output during Travis that is not present in other modes of execution. I'll see if I can't clean up that warning.

blee-techempower:
Passed localtest.
```
 Verification Summary

| Test: zend
|       query       : PASS
|       json        : PASS
|       db          : PASS
```

PR #1743, opened 2015-09-23
Clean up warnings for zend1

greg-hellings:
Adds model layer to zend1 that includes type casing integers in order to clean up warnings during verification.


blee-techempower:
Passed all localtests.

PR #1746, opened 2015-10-03
upgrade version of crystal to 0.8.0

kbrock:
add postgres support too

### DIFF
--- a/frameworks/Crystal/moonshine/.gitignore
+++ b/frameworks/Crystal/moonshine/.gitignore
@@ -2,3 +2,4 @@
 .deps.lock
 libs/
 .crystal
+.shards
--- a/frameworks/Crystal/moonshine/Projectfile
+++ b/frameworks/Crystal/moonshine/Projectfile
@@ -1,4 +1,0 @@
-deps do
-  github "dhruvrajvanshi/Moonshine", name: "moonshine"
-  github "stefanwille/crystal-redis"
-end
--- a/frameworks/Crystal/moonshine/benchmark_config.json
+++ b/frameworks/Crystal/moonshine/benchmark_config.json
@@ -2,12 +2,12 @@
   "framework": "moonshine",
   "tests": [{
     "default": {
-      "setup_file": "setup",
+      "setup_file": "setup-redis",
       "json_url": "/json",
-      "db_url": "/redis/db",
-      "query_url": "/redis/queries?queries=",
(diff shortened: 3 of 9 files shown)


kbrock:
Hmm...This is not working.

Unfortunately, This is returning valid values for my local redis instance, but no values are coming back here.

kbrock:
@zane-techempower was your thought that 1 program would serve all 3 database types?

Also, can you spot what I am doing wrong with redis?
Locally, on my mac, I populate my redis database with the tech empower scripts. And the values are working. Unfortunately, redis seems to just return null in the environment.

Thanks

zane-techempower:
> @zane-techempower was your thought that 1 program would serve all 3 database types?

This is flexible, endpoints of a certain db can be separated by their route, e.g. `/redis/json`, `/postgres/json`. Or a whole separate server can be made. The way you have it currently is fine—your `benchmark-config.json` looks good, which is the important part as the toolset uses that to setup and test the endpoints.

> Ideas why redis would be failing? It looks like it is not getting configured correctly. (I used the techempower script to populate my redis)

I believe you when you say that redis is working just fine locally, as we are aware of [redis having issues on travis.ci](https://github.com/TechEmpower/FrameworkBenchmarks/pull/1679). On travis, redis is set up in its own way, different to our other test environments, which is leading to erroneously failing tests. Fixing up the redis bug on travis is on my todo-list but I have not had time amongst my other responsibilities. Travis does great work for us, and this redis issue is a small unfortunate thing.

Two things can happen now:
1) Someone fixes up the Redis on travis bug, removing those erroneously failing tests. Here's the [issue](https://github.com/TechEmpower/FrameworkBenchmarks/issues/1661), and the [pr with some more discussion](https://github.com/TechEmpower/FrameworkBenchmarks/pull/1679).
2) Someone, likely @msmith-techempower will run your PR locally, confirm it works, and manually merge it. This is what we did to include redis-using frameworks for round 11. This will not happen until work for round 12 opens—which will happen shortly. We are currently on the verge of putting out round 11 data, and are not actively working to merge new contributions at the moment.

Thanks for the work! Crystal is a favorite of mine :)

blee-techempower:
Pass all localtests. Looks good.

```
 Verification Summary

| Test: moonshine
|       fortune     : PASS
|       plaintext   : PASS
|       db          : PASS
|       update      : PASS
|       json        : PASS
|       query       : PASS
```

PR #1747, opened 2015-10-04
fix warning: historical needs default

kbrock:
builds are complaining that historical does not have a `"default"` test

```
WARNING:root:Framework historical does not define a default test in benchmark_config.json
```


blee-techempower:
Pass all localtest.

PR #1748, opened 2015-10-04
fix warnings for sabina

kbrock:
This this is what the benchmark json should look like for sabina.

```
WARNING:root:Framework sabina does not define a default test in benchmark_config.json
```
I noticed the warning and it looked different from others.

I may be way off on this one.


blee-techempower:
Passed all tests.

PR #1751, opened 2015-10-10
Fix broken download and update phoenix to 1.0

waiteb3:


kbrock:
nice. passing in travis to

msmith-techempower:
Well well well, the prodigal contributor has returned! You should set up an avatar @waiteb3 

Round 11 is imminent - I will test this out and get it merged shortly after the release.

thousandsofthem:
I think some improvements needed:

* https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Elixir/phoenix/config/prod.exs#L9
Should use pool size 256 (as other frameworks do, affects speed alot, like ~10x slowdown due to smaller pool size), postgres adapter (as reported in metadata)
```
adapter: Ecto.Adapters.Postgres,
...
pool_size: 256
```

* router should pipe through api instead of browser (all the interaction goes through json so it makes sense, also it doing way too much job that other frameworks skip)
https://github.com/TechEmpower/FrameworkBenchmarks/pull/1751/files#diff-8cb5241f16d26b7178d94f665b759f29R17

* `plug :scrub_params` is useless and could be removed:
https://github.com/TechEmpower/FrameworkBenchmarks/pull/1751/files#diff-92a2750f03cbfd63d2c3b27eeeef442aR6

waiteb3:
@thousandsofthem Since this isn't merged yet I've added your suggestions in this PR.

Also, since Ecto's pool_size seems to be undocumented outside of source code and release notes, I'd ask for you (or someone invested in elixir) to fix that.

thousandsofthem:
Thanks!
`pool_size` is indeed undocumented :/ thanks for the note
Still, phoenix adds it as a default config option in newer versions:
https://github.com/phoenixframework/phoenix/blob/master/installer/lib/phoenix_new.ex#L263
So i think it's not that bad 

P.S.I'm not related to phoenix/ecto/elixir dev team

blee-techempower:
Passed local test.

```
 Verification Summary

| Test: phoenix
|       fortune     : PASS
|       plaintext   : PASS
|       db          : PASS
|       update      : PASS
|       json        : PASS
|       query       : PASS
```

PR #1754, opened 2015-10-12
finalize vertx benchmark for version 3.1.0

jebbstewart:
Updated vertx test suite for latest version 3.1.0.  Only including plaintext and json tests.


blee-techempower:
Pulled and tested locally and both tests pass.

Can you please remove `install.sh` as that file is deprecated and no longer used? After that, LGTM.

jebbstewart:
install.sh was removed.  Thanks

PR #1767, opened 2015-11-19
Replace curl usage with requests

ssmith-techempower:
This is a new pull request for #1685. No changes have been made aside from fixing merge conflicts from bringing it up to the current master.


msmith-techempower:
Oh, we need to fix up the Travis run file before we can merge this. Also, we need to fix up the vagrant bootstrap.

https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/.travis.yml#L181
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/deployment/vagrant-common/bootstrap.sh#L125

I believe removing those two lines will fix the issue.

msmith-techempower:
I haven't any idea why asynchio and ulib failed on travis. I'm merging this.

PR #1772, opened 2015-11-24
avoid Content encoding warning with tests

stefanocasazza:
Hi,

this patch is to avoid Content encoding warning when we test ULib.

There is a problem with the name of the directory for the framework 'ULib' and the name used for travis test 'ulib' and also Ulib is disappearance from the result of Round 10 results...

Greeting
Stefano


nbrady-techempower:
Tested locally - no warnings for plaintext.

PR #1776, opened 2015-11-26
use snake_case for function names in iron

steveklabnik:
This is more idiomatic Rust.

### DIFF
--- a/frameworks/Rust/iron/src/main.rs
+++ b/frameworks/Rust/iron/src/main.rs
@@ -14,19 +14,19 @@ struct Message {
 
 fn main() {
     let mut router = Router::new();
-    router.get("/json", jsonHandler);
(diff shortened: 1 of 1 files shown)


steveklabnik:
This failure seems unrelated.

msmith-techempower:
@steveklabnik Looks like you need to rebase from our latest master for Travis to test this properly.

steveklabnik:
Will do.

msmith-techempower:
@steveklabnik Looks like there was an error: https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/96127834

steveklabnik:
Ah, this change is unrelated to my fix, actually. It's due to the dependencies resolving in an incompatible way.

I don't have time to fix this right now, so I'll just give it a close. In the future, others will end up with this error and need to fix it though.

PR #1781, opened 2015-11-28
Add sinatra-sequel benchmark

mwpastore:
A new test/benchmark of Sinatra on Ruby 2.2 using Sequel ORM instead of ActiveRecord. MRI and Puma only for now; I will start work on the others when this is merged. I deliberately did not squash the two commits so a reader can easily see the differences between the `sinatra` and `sinatra-sequel` benchmarks (without manual `diff`'ing). Thank you!


mwpastore:
This isn't stable. Working on a fix. Thanks.

PR #1782, opened 2015-11-29
Add sinatra-sequel benchmark

mwpastore:
I'd like to add a new test/benchmark of Sinatra on Ruby 2.2 using Sequel ORM instead of ActiveRecord. MRI and Puma only for now; I will start work on the others when this is merged. I deliberately did not squash the two commits so a reader can easily see the differences between the sinatra and sinatra-sequel benchmarks (without manually diff'ing corresponding files in different directories). Thank you!

Housekeeping: This replaces and obsoletes #1781.


nbrady-techempower:
fortune, plaintext, db, update, json, query: all tests pass, no warnings - @msmith-techempower

msmith-techempower:
Looks good to me.

@nbrady-techempower LGTM!

mwpastore:
Great, I have a sinatra-sequel-jruby-puma queued up so I'll push that soon. I need to rebase it first. Thanks all!

PR #1783, opened 2015-11-29
flask: Use PyMySQL on PyPy

### LINKED ISSUE #1779: PyPy should not use python-mysqldb
Hi,

I've done some benchmarks of mysql libraries with pypy 4.0.1:

Lib |	Version	| Queries per second
----|---------------|---------------------------
PyMySQL|0.6.7|13590
MySQL connector|2.1.3|10688
MySQLdb|1.2.5|5841

I've seen yours explanation for not trying different libs for databases access, but i think that this need to be reconsidered one more time.

methane:
fixes #1779


nbrady-techempower:
Pulled and tested locally - plaintext, db, update, json, and query all pass
with a warn on forture - LGTM

PR #1787, opened 2015-12-01
Add sinatra-sequel-puma-jruby benchmark 

mwpastore:
### sinatra-sequel on JRuby

* Make sinatra-sequel framework compatible with JRuby 9K
* Add runner, language setup, and benchmark config for sinatra-sequel-puma-jruby
* Bug Fix: sinatra-sequel framework was applying updates on /updates, but returning the original records (instead of the updated records)

PR #1788, opened 2015-12-01
Reworked RVM configuration for Ruby

mwpastore:
RVM [stable] typically only has one "[known](https://github.com/rvm/rvm/blob/stable/config/known)" minor version for each major version, e.g. Ruby 2.1.5 for 2.1 and 2.2.1 for 2.2, JRuby 1.7.19 for JRuby 1.7, etc. In light of this, the way the Ruby languages are configured in FrameworkBenchmarks today is somewhat obtuse. A lot of the tests are broken at install time because they depend on a version of Ruby that is not available in RVM [stable]! 

To address this problem, I've renamed and slightly modified the language setup scripts for Ruby to point to the current versions in RVM [stable] and establish environment variables for the test setup scripts to reference (after invoking `fw_depends`). This system seems to work well and will be much easier to keep up-to-date with RVM as they turn out new stable releases. (There is a little bit of a crisis at the moment because the maintainer of RVM is essentially AWOL and hasn't merged into the stable branch since March, but this will hopefully soon be resolved.)

I've also taken this opportunity to reorganize the way JRuby depends on Java. Each setup script for a JRuby-based test was requiring (via `fw_depends`) a jruby and a jdk. I moved the dependency on java out of the test setup script and into the jruby language setup script. I think this makes much more sense but I'm interested to know if the team agrees. Allowing different combinations of the same JRuby and Java versions makes the benchmarks less comparable and adds a complexity to FrameworkBenchmarks that should be the responsibility of the JRuby team (i.e. figuring out how JRuby should interoperate with Java). 

Finally, I noticed that Bundler was being used inconsistently and in some cases incorrectly by the various Ruby test setup scripts. I've modified them to consistently install Gems locally, four ways simultaneously, and added `.gitignore` rules to exclude `.bundler` and `vendor/bundle` per standard Ruby project practices. Please let me know if you have any questions or corrections.

Thank you!


mwpastore:
*sigh* My git-fu is weak. I will restructure this so it doesn't include my other commit that is a separate pull request (#1787). Sorry...

PR #1789, opened 2015-12-01
Reworked RVM configuration for Ruby

mwpastore:
RVM [stable] typically only has one "[known](https://github.com/rvm/rvm/blob/stable/config/known)" minor version for each major version, e.g. Ruby 2.1.5 for 2.1 and 2.2.1 for 2.2, JRuby 1.7.19 for JRuby 1.7, etc. In light of this, the way the Ruby languages are configured in FrameworkBenchmarks today is somewhat obtuse. A lot of the tests are broken at install time because they depend on a version of Ruby that is not available in RVM [stable]! 

To address this problem, I've renamed and slightly modified the language setup scripts for Ruby to point to the current versions in RVM [stable] and establish environment variables for the test setup scripts to reference (after invoking `fw_depends`). This system seems to work well and will be much easier to keep up-to-date with RVM as they turn out new stable releases. (There is a little bit of a crisis at the moment because the maintainer of RVM is essentially AWOL and hasn't merged into the stable branch since March, but this will hopefully soon be resolved.)

I've also taken this opportunity to reorganize the way JRuby depends on Java. Each setup script for a JRuby-based test was requiring (via `fw_depends`) a jruby and a jdk. I moved the dependency on java out of the test setup script and into the jruby language setup script. I think this makes much more sense but I'm interested to know if the team agrees. Allowing different combinations of the same JRuby and Java versions makes the benchmarks less comparable and adds a complexity to FrameworkBenchmarks that should be the responsibility of the JRuby team (i.e. figuring out how JRuby should interoperate with Java). 

Finally, I noticed that Bundler was being used inconsistently and in some cases incorrectly by the various Ruby test setup scripts. I've modified them to consistently install Gems locally, four ways simultaneously, and added `.gitignore` rules to exclude `.bundler` and `vendor/bundle` per standard Ruby project practices. Please let me know if you have any questions or corrections.

Thank you!

Housekeeping: This replaces and obsoletes PR #1788. (I swear I'll get the hang of this eventually...)


msmith-techempower:
Relies on #1794 being addressed; then needs to be rebased and retested with Travis-CI

msmith-techempower:
OKAY, @mwpastore we have fixed the `rvm` issues on Travis, but we need you to rebase your branch and push for a test build to get kicked off.

toolset/setup/linux/languages/jruby-1.7.sh
```diff
@@ -1,8 +1,11 @@
 #!/bin/bash
 
-fw_depends rvm
+fw_depends rvm java7
+
+# rvm stable [typically] only provides one version of jruby-1.7
+# update this when it changes
+JRUBY_VERSION="1.7.19"
```
msmith-techempower: This should be moved down past the `RETCODE` check, and it needs to be included in the `.installed` file so that subsequent runs have access to he variable in the standard way.

Github is not allowing me to comment on the line I want, but I would expect it to look like this:

`echo "export JRUBY_VERSION=${JRUBY_VERSION}" > $IROOT/jruby-$JRUBY_VERSION.installed`

This change would also have to be made for the other Ruby language files.

msmith-techempower:
In addition to my comments from above, Travis-CI reports that JRuby is not working at all for most of the tests; some cases nothing starts because of an error while in other cases it starts but responds with 500s.

I doubt that your changes are related to that particular problem, but I wanted to document it somewhere.

msmith-techempower:
I believe the `jruby` issues to be unrelated, so I think I am just going to merge this in.

toolset/setup/linux/languages/jruby-1.7.sh
mwpastore: It looks like you or someone else made these changes. Thank you!

mwpastore:
Thanks Mike!

PR #1800, opened 2015-12-15
Add ffead-cpp-2.0 to techempower benchmarks

sumeetchhetri:


toolset/setup/linux/frameworks/ffead-cpp.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+RETCODE=$(fw_exists ${IROOT}/ffead-cpp.installed)
+[ ! "$RETCODE" == 0 ] || { return 0; }
+
+sudo apt-get install -y uuid-dev unixodbc unixodbc-dev
```
msmith-techempower: Is it strictly necessary to use `apt-get` for these requirements?

If so, I recommend moving these to `toolset/setup/linux/prerequisites.sh`

Although, now that we have this dependency management stuff in place (via `fw_depends`), maybe it makes sense to do away with a lot of `prerequisites.sh` and move each into the specific needing dependency files, which would speed up travis by quite a bit. Hmmm, I guess this is fine for now.

PR #1801, opened 2015-12-15
fasthttp: code clarification

valyala:
This is just a cosmetic change aimed to improving code readability for fasthttp benchmark.


msmith-techempower:
LGTM!

PR #1802, opened 2015-12-15
Quick fix for vagrant

nbrady-techempower:
Updated bootstrap.sh 

Commented out syncing for now, added progressbar support, added necessary sudo commands for testrunner, added 'source toolset/setup/linux/prerequisites.sh'

### DIFF
--- a/deployment/vagrant-common/bootstrap.sh
+++ b/deployment/vagrant-common/bootstrap.sh
@@ -100,7 +100,7 @@ if [ ! -e "~/.firstboot" ]; then
   sudo apt-get update
   sudo apt-get install -y git
   sudo apt-get install -y python-pip
-  sudo pip install progressbar
+  source toolset/setup/linux/prerequisites.sh
 
   # Make project available
   # If they synced it to /FwBm, just expose it at ~/FwBm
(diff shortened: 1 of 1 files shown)


deployment/vagrant-common/bootstrap.sh
```diff
@@ -157,6 +168,7 @@ if [ ! -e "~/.firstboot" ]; then
   # Setup 
   echo "Installing $ROLE software"
   cd $FWROOT
+  source toolset/setup/linux/prerequisites.sh
```
msmith-techempower: This could probably move WAY up; maybe, we could replace the `sudo pip install progressbar` line way up there since it is called out again in this file. Less duplication.

msmith-techempower:
Yay; LGMT!!!

Notes: this branch breaks file-[something] between the host environment and the virtual environment, so after you `vagrant ssh`, changes made in the VM/Host will NOT be mirrored in the other. That means that if one is doing dev on the host, those changes will have to be gotten INTO the VM somehow, be it pushing to a branch and pulling those in the VM for testing, or just copy-pasting.

PR #1807, opened 2015-12-17
fixed toolset prerequisites command

nbrady-techempower:
fixed the toolset prerequisites command to include home directory


msmith-techempower:
Love it; fixed forever... LGTM!

PR #1808, opened 2015-12-17
Vagrant Fix - added apt-get updates, set global line endings

nbrady-techempower:
ensure git is up to date with proper security patches by first doing apt-get update
forced global config of linux style line endings
moved running the prerequisites until after the repo is cloned


nbrady-techempower:
Tested locally, vagrant working now.

PR #1810, opened 2015-12-17
Verification change for charset

ssmith-techempower:
Previously, we would flag responses that set a `charset` as warnings as this extra header data was unnecessary and may impact performance. After some internal discussions, it's been decided that the `charset` should be considered mandatory for `text/html`, which is the content type for the fortunes test. This PR changes the verification script to reflect this. Failure to include `charset` for `text/html` or including `charset` for other content types is flagged as a warning.

### DIFF
--- a/toolset/benchmark/test_types/verifications.py
+++ b/toolset/benchmark/test_types/verifications.py
@@ -66,19 +66,36 @@ def verify_headers(headers, url, should_be='json'):
              'No content encoding found, expected \"%s\"' % (
                  expected_type),
              url))
-    elif content_type.lower() == includes_charset:
-        problems.append(
-            ('warn',
-             ("Content encoding found \"%s\" where \"%s\" is acceptable.\n"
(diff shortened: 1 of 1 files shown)


nbrady-techempower:
````
   WARN for http://127.0.0.1:8080/plaintext

     Content encoding found "text/plain; charset=utf-8" where "text/plain" is acceptable.
````

@ssmith-techempower then this should not be a warning, correct?

msmith-techempower:
@nbrady-techempower I do not remember the specifics, but the comment in the code is what was decided upon: `# "text/html" requires charset to be set. The others do not`

ssmith-techempower:
After some internal discussion, we've decided that `charset` should be optional for `plaintext`, so we will consider both "`text/plain`" and "`text/plain; charset=utf-8`" as being correct.

msmith-techempower:
Hard to know who this would affect, but since most of the tests we expect to pass are, I say LGTM!

PR #1813, opened 2015-12-20
Blee

blee-techempower:
Testing Purpose. DO NOT MERGE


msmith-techempower:
Note to self - these changes need to be reverted.

frameworks/Crystal/moonshine/server-redis.cr
```diff
@@ -30,6 +29,11 @@ end
 private def randomWorld
   id = rand(1..ID_MAXIMUM)
   num = REDIS.get("world:" + id.to_s)
+
+  if num == Nil
+    num = 7777
+  end
```
msmith-techempower: This seems wrong to me - if you get a null pointer back from this, then there is something wrong with how we are getting the value for `id`, and we should definitely create special cases to deal with that.

frameworks/Crystal/moonshine/setup-postgres.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 
-fw_depends crystal
+curl http://dist.crystal-lang.org/apt/setup.sh | sudo bash
+
+sudo apt-get install crystal
```
msmith-techempower: These lines are 100% wrong - use the `fw_depends` structure in all setup files. If these are the correct steps (and I am not saying that they are), then they should be in [this file](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/languages/crystal.sh), not in setup files like this one.

msmith-techempower:
As a final note - please do not merge in pull requests you have made; I will merge them in after these types of review cycles.

frameworks/Crystal/moonshine/setup-postgres.sh
blee-techempower: I doubt that also. WIll keep fw_depnds structure thing.

frameworks/Crystal/moonshine/server-redis.cr
blee-techempower: Hi, Kain.

That code was my mock data to test crystal/moonshine on local machine.

I did not have redis on my local machine in that time, so just need to
pass redis test to see what will happen. (In reality, that if statement
will not happen)

Doing this I could figure out why crystal/moonshine fails. Redis problem
happens on travis causes the error; due to this all other frameworks
which fails redis test were resolved.

I just forgot to delete those line.

Now everything is correct and already merged.

No need to worry.

On 12/21/15 11:16 AM, Mike Smith wrote:
> In frameworks/Crystal/moonshine/server-redis.cr
> <https://github.com/TechEmpower/FrameworkBenchmarks/pull/1813#discussion_r48179437>:
> 
>> @@ -30,6 +29,11 @@ end
>>  private def randomWorld
>>    id = rand(1..ID_MAXIMUM)
>>    num = REDIS.get("world:" + id.to_s)
>> +
>> +  if num == Nil
>> +    num = 7777
>> +  end
> 
> This seems wrong to me - if you get a null pointer back from this, then
> there is something wrong with how we are getting the value for |id|, and
> we should definitely create special cases to deal with that.
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/TechEmpower/FrameworkBenchmarks/pull/1813/files#r48179437>.
> 

-- 
Best,
Byung Lee

frameworks/Crystal/moonshine/setup-postgres.sh
blee-techempower: Hi, Kain.

I did not know all the framework should need to use fw_depends structure
or whatever that function is.

What I hear from you guys was just fix problem.

I fixed this issue and now crystal/moonshine uses fw_depends function to
setup environments.

Everything is good now.

On 12/21/15 11:17 AM, Mike Smith wrote:
> In frameworks/Crystal/moonshine/setup-postgres.sh
> <https://github.com/TechEmpower/FrameworkBenchmarks/pull/1813#discussion_r48179586>:
> 
>> @@ -1,6 +1,8 @@
>>  #!/bin/bash
>>  
>> -fw_depends crystal
>> +curl http://dist.crystal-lang.org/apt/setup.sh | sudo bash
>> +
>> +sudo apt-get install crystal
> 
> These lines are 100% wrong - use the |fw_depends| structure in all setup
> files. If these are the correct steps (and I am not saying that they
> are), then they should be in this file
> <https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/languages/crystal.sh>,
> not in setup files like this one.
> 
> EDIT: Additionally, we are striving to moving away from |apt-get|, which
> is why this file
> <https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/languages/crystal.sh>
> does not use it.
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/TechEmpower/FrameworkBenchmarks/pull/1813/files#r48179586>.
> 

-- 
Best,
Byung Lee

PR #1815, opened 2015-12-22
Blee nancy modification.

blee-techempower:
Now, at least this framework is testable. While I was working on this problem, I could find some issues regarding scripting file for environment setup. 

This framework does not test plaintext, so I added.

I am pretty sure query test part is not correct. Need to make sure it later. 

For this reason I just like to keep this branch for a while.


blee-techempower:
This branch is dead due to master branch revert. Just leave it to refer what I did for now.

PR #1820, opened 2015-12-26
Create a postgres-based alternative to the mysql Go Fasthttp

silviucm:
This is a Postgres-based benchmark based on fasthttp

It uses Jack Christensen's pgx: https://github.com/jackc/pgx

At Aliaksandr's suggestion, I renamed fasthttp to fasthttp-mysql to better differentiate between the underlying databases


msmith-techempower:
Looks good; LGTM!

... but we have to do it from the command line.

PR #1822, opened 2015-12-26
Blee mono patch

blee-techempower:
I carefully tested CSharp related frameworks and I found that CSharp related environment setup script files such as Mono.sh have some problems. So I modified those bad ones. After modification, I was able to test CSharp/Nancy. To make sure everything, I also tested other CSharp related frameworks and my conclusion is that my change does not make side effcet. For this reason I am going to pull request this branch first and later I will open a pull request my nancy update.


toolset/setup/linux/languages/mono.sh
```diff
@@ -16,8 +16,7 @@ echo "deb http://jenkins.mono-project.com/repo/debian sid main" | sudo tee /etc/
 sudo apt-get update
 
 # Find the most recent snapshot
-#SNAPSHOT=$(apt-cache search 'mono-snapshot-.*-assemblies' | cut -d'-' -f3 | tail -1)
-SNAPSHOT="20150202010831"
+SNAPSHOT=$(apt-cache search 'mono-snapshot-.*-assemblies' | cut -d'-' -f3 | tail -1)
```
msmith-techempower: This line was purposefully changed to NOT find the most recent because the builds of mono would break day to day. We found the most recent (at the time) build that worked and used it directly. I am weary of returning to this code that finds the latest because of this same reason.

toolset/setup/linux/languages/mono.sh
```diff
@@ -36,17 +35,20 @@ source $IROOT/mono.installing
...
+# Download and extract debs.
 fw_apt_to_iroot mono-snapshot-$SNAPSHOT
 fw_apt_to_iroot mono-snapshot-$SNAPSHOT-assemblies mono-snapshot-$SNAPSHOT
 
+# Require sudo for some commands due to permission problem.
+# Other than that, this script will be stuck and will fail test.
+#
 # Simplify paths
-mv $MONO_HOME/opt/mono-*/* $MONO_HOME
+sudo mv $MONO_HOME/opt/mono-*/* $MONO_HOME
```
msmith-techempower: Should not require sudo as the move is from `$IROOT/mono-snapshot-$SNAPSHOT/opt/mono-*/*` to `$IROOT/mono-snapshot-$SNAPSHOT`. Neither of these places require any special privileges for the testrunner user.

toolset/setup/linux/languages/mono.sh
```diff
@@ -36,17 +35,20 @@ source $IROOT/mono.installing
...
 file $MONO_HOME/bin/* | grep "POSIX shell script" | awk -F: '{print $1}' | xargs sed -i "s|/opt/mono-$SNAPSHOT|$MONO_HOME|g"
-sed -i "s|/opt/mono-$SNAPSHOT|$MONO_HOME|g" $MONO_HOME/lib/pkgconfig/*.pc $MONO_HOME/etc/mono/config
+sudo sed -i "s|/opt/mono-$SNAPSHOT|$MONO_HOME|g" $MONO_HOME/lib/pkgconfig/*.pc $MONO_HOME/etc/mono/config
```
msmith-techempower: Should not require sudo.

toolset/setup/linux/languages/mono.sh
```diff
@@ -36,17 +35,20 @@ source $IROOT/mono.installing
...
 
 echo "mozroots --import --sync" >> $IROOT/mono.installing
 
-mv $IROOT/mono.installing $IROOT/mono.installed
+sudo mv $IROOT/mono.installing $IROOT/mono.installed
```
msmith-techempower: Should not require sudo.

toolset/setup/linux/languages/xsp.sh
```diff
@@ -15,7 +15,7 @@ git checkout e272a2c006211b6b03be2ef5bbb9e3f8fefd0768
 # build
 ./autogen.sh --prefix=$MONO_HOME --disable-docs
 make
-make install
+sudo make install
```
msmith-techempower: Should not require sudo if properly configured to install into $IROOT.

toolset/setup/linux/bash_functions.sh
```diff
@@ -42,16 +42,18 @@ fw_unzip() {
...
+#   Without using sudo,
+#   Does not download dependant packages.
+#   (This is one of reasons why Mono did not run correctly.)
 #
 # Example: fw_apt_to_iroot <package> [<directory>]
 fw_apt_to_iroot() {
   DIR=${2:-$1}
   echo "Downloading $1 to $IROOT"
-  apt-get download $1
+  sudo apt-get download $1
```
msmith-techempower: Should not require sudo.

toolset/setup/linux/bash_functions.sh
```diff
@@ -42,16 +42,18 @@ fw_unzip() {
...
   echo "Extracting $1 to $DIR"
-  dpkg-deb -x $1*.deb "$IROOT/$DIR" && rm $1*.deb
+  sudo dpkg-deb -x $1*.deb "$IROOT/$DIR" && sudo rm $1*.deb
```
msmith-techempower: Should not require sudo.

msmith-techempower:
The only real change I can see here that **may** fix the issues mono is having is using a more updated snapshot rather than the static one I mentioned in comments.

Other than that, nothing here has changed as none of those `sudo` additions should be required.

Mono has been notoriously fickle with regard to installation in the past because sometimes the server simply doesn't respond when we are trying to download it. It is unreliable and difficult to test, but the mono installation **should** be working as is without these changes.

PR #1824, opened 2015-12-28
Crystal/Moonshine update

blee-techempower:
Ready to be merged. 

I solved the problems that @msmith-techempower pointed on branch blee.

Now, this version uses "fw_depends structure".
Also, I deleted my debug purpose code for local testing. (Actually the code does not do anything. It was just mock data to do local test without redis.)

Let me know if there is other issues.


msmith-techempower:
@blee-techempower What is the state of this branch?

PR #1825, opened 2015-12-29
JavaScript/Express update.

blee-techempower:
Now, it passes build test.

Ready to be merged.


nbrady-techempower:
@blee-techempower we shouldn't use wildcards for versions.
 2.1.2 works as the latest npm release for mongodb.

blee-techempower:
Oops! I suppose delete that line and I forgot.

PR #1827, opened 2015-12-29
add support for Test type 6: Plaintext

jochenberger:


nbrady-techempower:
@jochenberger Would you mind checking out the conflicts? Also, a note that we are moving to warnings for text/plain responses that don't carry the charset="utf-8" parameter, and removing the warning for JSON responses that do not.

jochenberger:
I updated the PR.
> and removing the warning for JSON responses that do not.

That's supposed to be "that do", right?

nbrady-techempower:
@jochenberger Sorry if that was unclear. As it stands, we are warning if your JSON response is missing the charset=utf-8 parameter. This should not be the case, as there are no parameters for JSON application types. We're removing the warnings for JSON responses that do not have charset=utf-8.

nbrady-techempower:
Pulled and tested locally as well as passing travis. Looks good! Thanks for your contributions @jochenberger

PR #1836, opened 2015-12-30
fixed create-redis.sh

nbrady-techempower:
Without this hash bang travis was not recognizing the proper commands


blee-techempower:
This thing is already verified, so I merge it in.

Issue #1500, opened 2015-04-07
Support additional linux distributions


hamiltont:
As discussed on IRC (and briefly in #401) there is a desire to see additional distributions beyond Ubuntu 14.04 (and our very minimal support for 12.04 such as [database.sh](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/database.sh)). 
Currently we are focusing on adding Windows support, but this issue should help us clarify and plan for expanding to multiple linux distros. 

A key point is that we do not want to allow sudo in any of the setup scripts. This is critical for ensuring that tests do not interfere with each other. 

Another key point is that travis-CI only runs Ubuntu 12.04 and cannot help us test scripts for other platforms. 

The current linux setup system uses individual installer scripts, such as [this one](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/languages/dart.sh), and generally each script does one of these: 
- OK - Download and unzip appropriate binaries
- BAD - Use `sudo apt-get` to install system wide
- OK - Download source code and compile. Install in local prefix
- OK - Use `apt-get download` and `dpkg -x` to download/unzip

Currently a number of these scripts are fairly rigid (e.g. see #1499). 

The question is how to best update the existing system to support multiple operating systems. 

Personally, my preference is to use distro-specific scripts (e.g. `java.debian.sh`, `java.ubuntu12.sh`), with each script using the distro-specific package manager (`yum`,`apt-get`, etc) as much as possible to automate finding the correct binary and downloading it, and then manually extract the binary into a prefix folder. This is what I propose in #1479 and what is currently implemented for [mono.sh](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/languages/mono.sh) using this function in bash_functions.sh:

```
# Download *.deb file and install into IROOT without using sudo
# Does not download dependant packages
#
# Example: fw_apt_to_iroot <package> [<directory>]
fw_apt_to_iroot() {
  DIR=${2:-$1}
  echo "Downloading $1 to $IROOT"
  apt-get download $1
  echo "Extracting $1 to $DIR"
  dpkg-deb -x $1*.deb "$IROOT/$DIR" && rm $1*.deb
}
```

This replaced a download+compile based script that took 20+ minutes with a download+extract approach that takes <5 minutes. 

I'm not sure if something like puppet or chef can help us manage scripts for different platforms.

[...]

PR #1839, opened 2015-12-30
Update bash_functions.sh

blee-techempower:
I tested it several times. This change does not make side effects. 

Without using sudo, permission problem occurs which causes script being stuck. In detail, prompt will take away the process, and our script does not report this issue, so for user it looks like there is no progress.

### DIFF
--- a/toolset/setup/linux/bash_functions.sh
+++ b/toolset/setup/linux/bash_functions.sh
@@ -42,16 +42,18 @@ fw_unzip() {
   rm -f "$@"
 }
 
-# Download *.deb file and install into IROOT without using sudo
-# Does not download dependant packages
-#
+# Download *.deb file and install into IROOT 
+# Cautions:
+#   Without using sudo,
+#   Does not download dependant packages.
+#   script will be stuck and will not make progress. (e.g: CSharp/nancy)
 # Example: fw_apt_to_iroot <package> [<directory>]
 fw_apt_to_iroot() {
   DIR=${2:-$1}
   echo "Downloading $1 to $IROOT"
-  apt-get download $1
+  sudo apt-get download $1
   echo "Extracting $1 to $DIR"
-  dpkg-deb -x $1*.deb "$IROOT/$DIR" && rm $1*.deb
+  sudo dpkg-deb -x $1*.deb "$IROOT/$DIR" && sudo rm $1*.deb
 }
 
 # Was there an error for the current dependency?


blee-techempower:
Please merge this.(I intentionally separated this change from my original nancy update that passed travis-ci test before: https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/99140181 for separation of concerns.)   Travis failing is not caused by my changes. Its because of broken frameworks.

nbrady-techempower:
@blee-techempower Please see #1500. Can you address that?
Anyway, is mono the only framework that uses fw_apt_to_iroot()? I believe it is and we want to get away from apt_get anyway, maybe we should consider fixing the mono.sh instead and deprecating fw_apt_to_iroot

blee-techempower:
I think my change is valid. 

On #1500, hamiltont stated that " OK - Use apt-get download and dpkg -x to download/unzip "

My change does not make side effect. It just allow proper permission. 

I am not sure only CShap frameworks uses  fw_apt_to_iroot(). Since it is in bash_functions.sh, there is strong possibility that other frameworks also utilize that function.

Well I already tested about this several time. My conclusion is my change does not make side effect. Its good to go. After this change and some other stuff, CSharp/Nancy and CSharp/evhttp-sharp passed travis ci and local test before.

nbrady-techempower:
A quick grep reveals that only mono.sh utilizes fw_apt_to_iroot() so I will merge this and we should consider updating the mono.sh script at some point.

blee-techempower:
@nbrady-techempower : Thanks so much, man.

This problem was tricky because OS process was taken by terminal prompt and our scripting system does not report anything regarding this issue. Therefore, it seemed like our scripting stuck at one point and does not make progress forever. Looked like a "starvation" and "deadlock".

But after I allow super permission, the problem disappeared magically.

msmith-techempower:
This sounds like a red herring.

I mentioned in another issue that mono is notoriously finicky and sometimes just fails to download. It **probably** does not have anything to do with using `sudo`, but if this works as-is, I do not see any harm in doing it this way (though is **should** be unnecessary).

blee-techempower:
If you are going to take out SUDO from bash_funcitions.sh, the script
file will not work properly. Process will be taken by bash prompt, so
script stuck forever. This can be observed with Nancy. Try it.


On 1/4/16 8:51 AM, Mike Smith wrote:
> This sounds like a red herring.
> 
> I mentioned in another issue that mono is notoriously finicky and
> sometimes just fails to download. It *probably* does not have anything
> to do with using |sudo|, but if this works as-is, I do not see any harm
> in doing it this way (though is *should* be unnecessary).
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/TechEmpower/FrameworkBenchmarks/pull/1839#issuecomment-168731576>.
> 

-- 
Best,
Byung Lee

Issue #1838, opened 2015-12-30
Travis configuration and broken frameworks


nbrady-techempower:
Currently, when editing a configuration or toolset file, travis will run through every framework and run a complete test. Broken frameworks eat up a ton of time and cause travis to fail. I suggest commenting out the frameworks in travis.yml that we know aren't working, so that if a complete test needs to be run it will save time and not cause contributers to panic, thinking they may have broken something. When the framework is fixed, it can be uncommented. 

Example:
````yml
    - "TESTDIR=Erlang/chicagoboss"
   # (Framework Not Working) - "TESTDIR=Erlang/cowboy"
    - "TESTDIR=Erlang/elli"
    - "TESTDIR=Go/beego"
    - "TESTDIR=Go/falcore"
    - "TESTDIR=Go/fasthttp"
````

If this is approved, I will go through and update.

**Comment by nbrady-techempower** [comment]
Note: Of course, it's preferable to fix either the framework if it's broken, the travis-ci configuration for that framework if it needs it, or remove the framework if it's no longer being supported in lieu of just commenting out frameworks and forgetting about them. The intention here is not to sweep things under the rug, but at the same time I think it's counter-intuitive (OHH That's what the ci in travis-ci stands for!) for travis to ever be failing on master, as master should probably always be ready (pull requests not withstanding) to run the next round of benchmarks.

That being said, it seems there may be fixes for the Scala issues in the PR's, the javascript frameworks fixed, ulib framework passing and the redis config patched. Also, I think this move could put us back on track for the workflow suggestion @LadyMozzarella made in #1437

**Comment by blee-techempower** [comment]
It will save tons of time, so I will disable all the broken frameworks from now.

**Comment by msmith-techempower** [comment]
I like this suggestion, but I would also like those comments in `.travis.yml` to have a link to an open issue describing the problem. Example:

```
    - "TESTDIR=Erlang/chicagoboss"
   # "TESTDIR=Erlang/cowboy" (Framework Not Working: https://github.com/TechEmpower/FrameworkBenchmarks/issues/1838)
    - "TESTDIR=Erlang/elli"
    - "TESTDIR=Go/beego"
    - "TESTDIR=Go/falcore"
    - "TESTDIR=Go/fasthttp"
```

msmith-techempower:
[...]

PR #1845, opened 2015-12-31
Update Mono: Fix environment setup file issues based on mono and xsp manual 

blee-techempower:
This change is already verified before : ( https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/99149368 )

Just separated this from previous branch is for "Separation of Concern"

Here is brief note:
- Mono needs to have the most recent snapshot.
- Some statements required super permission.
- Disabled all broken frameworks to save some time; also it does not confuse a person who is going to review this pull request because of travis test error due to malfunction frameworks.. ( #1838 )


nbrady-techempower:
@blee-techempower I wanted to wait for approval/consensus/discussion on #1838 before committing it. I also wanted to add comments to the frameworks that just need one test fixed, stuff like that. Can you separate out the travis.yml file from this PR?

blee-techempower:
Note:
Point1 : Getting errors because of frameworks which are not related with CSharp does not make sense. 

I only changed "Mono and XSP" which are for C#. Modification of "mono.sh" and "xsp.sh" cannot affect non-CSharp frameworks because there is no relationship. Only CSharp related frameworks utilize those two.

Point2: Everytime I repeat this test, I get different results. 

I tested same thing many times. I get different result each time.
https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/99632570
https://travis-ci.org/blee-techempower/FrameworkBenchmarks/builds/99624921
https://travis-ci.org/blee-techempower/FrameworkBenchmarks/builds/99639452

To verify my point, I intentionally opened a one more pull request : https://github.com/TechEmpower/FrameworkBenchmarks/pull/1846

Just check history of above request. Only change is that I added dummy file to run Travis-CI one more time. But it gives different result.

Conclusion:

 - Nothing wrong with my update.
  Again, my change is already verified before : ( https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/99149368 )

- There must be a some other issues that causes this. ( Getting different result every time is abnormal. )

blee-techempower:
I tested this update one more time and Travis-ci gives me different result again.

https://travis-ci.org/blee-techempower/FrameworkBenchmarks/builds/100051237

Again, this update only affect frameworks which use CSharp but test failed with "Go".

zapov:
I don't think thats a good way to install Mono.
In other PR I changed mono setup to install it from latest official version through standard apt-get: https://github.com/zapov/FrameworkBenchmarks/blob/master/toolset/setup/linux/languages/mono.sh
I don't know of anyone who installs Mono from snapshots in such a way. And that snapshot is pretty old anyway.
Java is installed the same way though apt-get, what's the rationale for installing Mono in such a strange way?

msmith-techempower:
@zapov The short answer is because we have been trying to move away from `apt-get` for a long time now.

The longer answer is because we need to be able to keep our environment sandboxed such that at any point, we can say definitively that we have removed all software for a full test run. `apt-get` (and other such management suites) makes this difficult as it would necessitate an "uninstall" script for every install script.

Currently, Java is installed via `apt-get` because we haven't had to time to figure out a different way of installing it. Mono was originally installed via `apt-get`, but someone found a different way to install it that does this proper sandboxing that we need, and for a while it worked fine. Now, it is showing that we need to have a more permanent solution, though I am not sure that giving up to `apt-get` again is that solution.

toolset/setup/linux/languages/mono.sh
```diff
@@ -16,8 +16,7 @@ echo "deb http://jenkins.mono-project.com/repo/debian sid main" | sudo tee /etc/
 sudo apt-get update
 
 # Find the most recent snapshot
-#SNAPSHOT=$(apt-cache search 'mono-snapshot-.*-assemblies' | cut -d'-' -f3 | tail -1)
-SNAPSHOT="20150202010831"
+SNAPSHOT=$(apt-cache search 'mono-snapshot-.*-assemblies' | cut -d'-' -f3 | tail -1)
```
msmith-techempower: Stated in another thread, but I would prefer to NOT using rolling snapshots like this; we often have to rerun tests weeks apart and there is no logging at any moment describing which version is being used.

I am fine updating our static snapshot number to a more recent version, but we cannot use this "use latest snapshot" implementation.

toolset/setup/linux/languages/mono.sh
```diff
@@ -40,13 +39,13 @@ rm -rf $MONO_HOME && mkdir -p $MONO_HOME
...
-sed -i "s|/opt/mono-$SNAPSHOT|$MONO_HOME|g" $MONO_HOME/lib/pkgconfig/*.pc $MONO_HOME/etc/mono/config
 
+# Simplify paths
+sudo mv $MONO_HOME/opt/mono-*/* $MONO_HOME
+file $MONO_HOME/bin/* | grep "POSIX shell script" | awk -F: '{print $1}' | xargs sudo sed -i "s|/opt/mono-$SNAPSHOT|$MONO_HOME|g"
+sudo sed -i "s|/opt/mono-$SNAPSHOT|$MONO_HOME|g" $MONO_HOME/lib/pkgconfig/*.pc $MONO_HOME/etc/mono/config
 echo "mozroots --import --sync" >> $IROOT/mono.installing
 
-mv $IROOT/mono.installing $IROOT/mono.installed
+sudo mv $IROOT/mono.installing $IROOT/mono.installed
```
msmith-techempower: Also stated in another task: I do not believe that any of these `sudo`s are needed.

msmith-techempower:
Also, it looks like all the CSharp tests were removed from `.travis.yml`, so they were not tested.

zapov:
I think a better way to make sure you have a clean system is by keeping a ghost image of the system, not by trying to find alternatives for apt-get.
Anyway, I think the best way to solve this (at least for now) is to allow apt-get on the official version (and not some random snapshot version) and look into installation issue later when C# frameworks are not second class citizens anymore ;)

BTW, when I tried fixing it, latest snapshots from Mono didn't work either, that's why I used the official stable one.

msmith-techempower:
Well, I'd like to get a pointer to that stable one that works so that we can link to it directly instead of guessing whether we had the same version of mono from round 12 to round 13.

blee-techempower:
@zapov @msmith-techempower 
Well, since you guys do not like my mono setup file, I would fix that mono setup file. I welcome any suggestion.  But that xsp.sh file requires sudo for sure. Without sudo , "make install" statement  did not do its job.

At least my full fix makes mono works. 
This is what I got from my local machine and it also passed travis-ci test before.
```
 Verification Summary
| Test: nancy-mono
|       query       : WARN
|       json        : WARN
|       db          : WARN
```

msmith-techempower:
@blee-techempower Okay, let's get a definitive pull request open to handle the mono and xsp fixes.

zapov:
Im reapating myself but I dont see a reason  to have different way of handling Java and C#. You can adress that issue for both at a later time. Java version changed between runs too.
Therefore I still think you should install Mono the official way (like I did)

msmith-techempower:
I tried to explain above, but I will include a shorter version of why we moved away from using `apt-get` as it pertains to mono.

A long time ago we relied on `apt-get` to install mono, but there were many version updates which seemed to actually break implementations with no discernible announcement. So, we would run several previews for a round (six, if my memory serves) which would work just fine with mono, but then I would kick off the official run and mono would have been updated, so all the C# tests would fail.

This became obnoxious, so I locked the version in the manner we see today, and that was working fine for a long time. Now, it seems we are facing a new problem in that that version is not up to date.

I still prefer our method as it gives us pin-point versions and we are trying our best to move all of the installations away from `apt-get` for different reasons.

zapov:
Thanks. Thats much better explanation. 
Mono is much more stable now than few years ago... but if that was the issue we can use fixed deb version plus apt get.
Although thats much more distribution specific.
As I said snapshots from official server did not work for some reason

PR #1846, opened 2016-01-01
Update mono test

blee-techempower:
I observed some weird fact about travis-ci. 

To verify the this, I am going to test it on TFB repersitory.


msmith-techempower:
This would not fix anything as you have commented out all the CSharp tests, neither `mono` or `xsp` will ever attempt to be installed.

blee-techempower:
@msmith-techempower  This is not full fix update. Purpose of this open request is for the separation of concern. While I was working on this, I just obsereved weird thing, so i just like to test it. So I put it here.

PR #1847, opened 2016-01-03
Upgrade python3 to 3.5.1

GMLudo:
Hi,

This PR is to upgrade python3 to 3.5.1.

Have a nice day.

### DIFF
--- a/frameworks/Python/asyncio/aiohttp.web/setup.sh
+++ b/frameworks/Python/asyncio/aiohttp.web/setup.sh
@@ -4,7 +4,7 @@ fw_depends python3
 
 sed -i 's|host: 127.0.0.1|host: '${DBHOST}'|g' aiohttp.web/etc/hello/main/main.yaml
 
-$IROOT/py3/bin/pip install --install-option="--prefix=${IROOT}/py3" -r $TROOT/requirements.txt
+pip3 install --install-option="--prefix=${PY3_ROOT}" -r $TROOT/requirements.txt
 
 cd $TROOT/aiohttp.web
 api_hour -ac hello:Container &
--- a/frameworks/Python/asyncio/aiohttp.web/setup.sh
+++ b/frameworks/Python/asyncio/aiohttp.web/setup.sh
@@ -4,7 +4,7 @@ fw_depends python3
 
 sed -i 's|host: 127.0.0.1|host: '${DBHOST}'|g' aiohttp.web/etc/hello/main/main.yaml
 
-pip install --install-option="--prefix=${PY3_ROOT}" -r $TROOT/requirements.txt
+$IROOT/py3/bin/pip install --install-option="--prefix=${IROOT}/py3" -r $TROOT/requirements.txt
 
 cd $TROOT/aiohttp.web
 api_hour -ac hello:Container &
--- a/frameworks/Python/asyncio/yocto_http/setup.sh
+++ b/frameworks/Python/asyncio/yocto_http/setup.sh
@@ -2,7 +2,7 @@
 
 fw_depends python3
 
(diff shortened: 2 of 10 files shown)


blee-techempower:
Need verification!

I checked travis-ci report and some frameworks such as cherrypy and bottom which passed travis-ci test before failed travis-ci test. (e.g: cherrypy-py3, bottom-py3) Since all of these fails are related with python3, need to verify this update.

Compare these two:
Previous one : https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/97529668 
Current one : https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/100016660

GMLudo:
@blee-techempower: Sorry, I hoped that this change didn't break something.

The error is that python3 doesn't find theses libraries: `ImportError: No module named site`: https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/100016789#L1423
I'm trying to avoid to upgrade setuptools and pip to see if it's better.

Be my guest if somebody has an idea.

methane:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Python/asyncio/aiohttp.web/setup.sh#L7

How about specify pip by absolute path?

If pip for Python 3.4 (installed by system) called, it searches `site` module from `$PYTHONHOME/lib/python3.4`.

methane:
Or `hash -r` after changing PATH?

GMLudo:
@methane: Thanks for your suggestions, it makes sense.

> How about specify pip by absolute path?

If I understand correctly the workflow, it's this file that manipulates PATH to point the right pip: https://github.com/Eyepea/FrameworkBenchmarks/blob/python_3.5.1/toolset/setup/linux/languages/python3.sh#L23

To be honest with you, I've forked my setup.sh file from: https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Python/flask/setup_py3.sh

It has the same issue with the new Python 3.5: https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/100016801#L2250

Maybe something changes with Python 3.5, I could point the exact pip executable in the script.

> Or hash -r after changing PATH?

What this command should change something ? or you mean bash -r ?

methane:
bash doesn't lookup command every time.
It remembers where the command exists in hash.
So I suspect that bash find `pip` from the hash, not from the PATH.

To investigate this issue is relating to the hash, add `type pip` before `pip`.
It shows where `pip` come from.

msmith-techempower:
@methane We used to require the absolute path for `pip`, but I **believe** I resolved that with my meganuke branch. Like you said, it is easy to verify via `type pip`, but I suspect that it is finding the correct one.

GMLudo:
@msmith-techempower I've made a new commit to test all hypothesis in this issue.
We'll see the results, it's running on travis-ci.

GMLudo:
Thank you a lot @methane and @msmith-techempower :-)

Even if the build isn't yet finished, I think I've found the issue with your suggestions:
In fact, pip points to `/usr/local/bin/pip`: https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/100134628#L2187
Only pip3 seems to be present now, it's why Django tests are still passing with my PR.

This change was present for Python 3.4, I don't know why it continues to work for us: https://docs.python.org/3/whatsnew/3.4.html#pep-453-explicit-bootstrapping-of-pip-in-python-installations

Now, with this commit, it should be good: https://github.com/Eyepea/FrameworkBenchmarks/commit/f8c127a412098cfe858aeb765601c55ff0832239

I'll push a new PR when the build will be done: https://travis-ci.org/Eyepea/FrameworkBenchmarks/builds/100228904

Have an happy new year.

PR #1849, opened 2016-01-04
Added Erlang/misultin framework

nbrady-techempower:
I will be adding the 3 frameworks from #938 over time, as I fix them to meet the correct structure. This is the misultin framework. Passes all tests locally for plaintext and json. It seems there are issues with db connections for all the erlang frameworks so I will look into that as well, while updating these.


nbrady-techempower:
Curious as to why this ran the gambit on travis tests while the mochiweb one did not. However, both erlang frameworks pass travis.

msmith-techempower:
misultin tests passed; LGTM!

PR #1851, opened 2016-01-04
Update CSharp/Nancy

blee-techempower:
- At least nancy-mono runs now.
- Added plaintext route test.
- Separate query test from db test.
- Temporarily remove libevent2 test. (Currently, this one does not work.)

### DIFF
--- a/toolset/setup/linux/languages/mono.sh
+++ b/toolset/setup/linux/languages/mono.sh
@@ -16,8 +16,7 @@ echo "deb http://jenkins.mono-project.com/repo/debian sid main" | sudo tee /etc/
 sudo apt-get update
 
 # Find the most recent snapshot
-#SNAPSHOT=$(apt-cache search 'mono-snapshot-.*-assemblies' | cut -d'-' -f3 | tail -1)
-SNAPSHOT="20150202010831"
+SNAPSHOT=$(apt-cache search 'mono-snapshot-.*-assemblies' | cut -d'-' -f3 | tail -1)
 
 # save environment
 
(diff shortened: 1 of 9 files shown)


toolset/setup/linux/languages/mono.sh
```diff
@@ -16,8 +16,7 @@ echo "deb http://jenkins.mono-project.com/repo/debian sid main" | sudo tee /etc/
 sudo apt-get update
 
 # Find the most recent snapshot
-#SNAPSHOT=$(apt-cache search 'mono-snapshot-.*-assemblies' | cut -d'-' -f3 | tail -1)
-SNAPSHOT="20150202010831"
+SNAPSHOT=$(apt-cache search 'mono-snapshot-.*-assemblies' | cut -d'-' -f3 | tail -1)
```
msmith-techempower: Please run this command on your local and simply use the value returned statically like before. It will probably be something like "2015100512512".

Add a comment explaining that we are doing this purposefully.

blee-techempower:
@msmith-techempower 
Oh well, this is failing because of the "sudo" issue you pointed yesterday.

Here is why I got error.
```
$MONO_HOME'\'' exited with status 1 (dependency=mono) (cwd=$FWROOT/installs)'
Setup nancy-mono: ERROR: /toolset/setup/linux/languages/mono.sh: Command 'mv $MONO_HOME/opt/mono-*/* $MONO_HOME' exited with status 1 (dependency=mono) (cwd=$FWROOT/installs)
Setup nancy-mono: Setup has terminated
```

This problem will be solved after I give sudo permission to that statement. 
So "SUDO" is required!!!

See the result of my original experimental result:
- https://github.com/TechEmpower/FrameworkBenchmarks/pull/1823
- https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/98843634

So, I need to revert this branch as my original.

toolset/setup/linux/languages/mono.sh
```diff
@@ -17,7 +17,13 @@ sudo apt-get update
...
 # Find the most recent snapshot
 #SNAPSHOT=$(apt-cache search 'mono-snapshot-.*-assemblies' | cut -d'-' -f3 | tail -1)
-SNAPSHOT="20150202010831"
+
+# REMARK:
+# Rollback this after I execute above command manually due to "msmith-techempower"'s request.
+# According to him "apt-get to install mono is not stable", so keep this way. 
+# If you see mono fail, please doubt this SNAPSHOT and execute above command manually, then
+# copy and paste the value to SNAPSHOT variable just like below.
+SNAPSHOT="2016.01.04+14.28.05"
```
msmith-techempower: Perfect comment here.

msmith-techempower:
Nancy checks passed; LGTM!

PR #1852, opened 2016-01-05
Fix Python/Flask

blee-techempower:
- now, build test does not fail.
- update version of some dependencies.
- Did local test. Passed every single tests.

In addition, I also tested this on my personal account and it passes all test. Here is result:
https://travis-ci.org/aznnobless/FrameworkBenchmarks/builds/100255428


blee-techempower:
Travis-ci passed. Ready to be merged.

PR #1854, opened 2016-01-05
Added php7 to toolset and updated nginx

nbrady-techempower:
This pr is to add php7 to the toolset. I've added this as a separate language dependency and will work with all the php frameworks to get them all running with php7. Some are not yet available to work with php7, which is why the need to have a separate setup file. The php7 setup file is still a work in progress, as I'm trying to fix configuration warnings that appeared in the original setup file. I don't like having to include all the shared object files because they're all in the php.ini config. So it might be a better option to create separate php configs and copy them over as needed - but I'd love feedback on that.

I've updated the PHP/php php-raw framework to use php7 as proof that the toolset currently works.

This pr also updates nginx from 1.4.1 to 1.9.9 which has serious performance implications. I've tested most of the frameworks that rely on nginx - there seem to be no breaking changes.


toolset/setup/linux/languages/php7.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
+RETCODE=$(fw_exists ${IROOT}/php.installed)
```
msmith-techempower: This will be problematic on our production system where php and php7 need to coexist. Specifically, this should probably be `php7.installed` throughout.

toolset/setup/linux/languages/php7.sh
```diff
@@ -0,0 +1,65 @@
...
+[ ! "$RETCODE" == 0 ] || { \
+  echo "Moving PHP config files into place";
+  sudo cp $FWROOT/config/php.ini /usr/local/lib/php.ini
+  sudo cp $FWROOT/config/php-fpm.conf /usr/local/lib/php-fpm.conf
```
msmith-techempower: However, once we see that php needs to be globally installed, we find the real problem; going back and forth from php5 to php7 could very well break based on this config files having different layouts.

Maybe I'm being pessimistic and this will just work.

toolset/setup/linux/webservers/nginx.sh
```diff
@@ -7,9 +7,9 @@ RETCODE=$(fw_exists ${IROOT}/nginx.installed)
...
 
-fw_get -O http://nginx.org/download/nginx-1.4.1.tar.gz
-fw_untar nginx-1.4.1.tar.gz
-cd nginx-1.4.1
+fw_get -O http://nginx.org/download/nginx-1.9.9.tar.gz
+fw_untar nginx-1.9.9.tar.gz
+cd nginx-1.9.9
 
 # There is no --quiet flag that I could find...
 echo "Configuring nginx..."
```
msmith-techempower: Would have preferred the nginx update in a separate PR for future reference; just leave as-is for this one.

nbrady-techempower:
Ah, right. Should have done that for nginx. 

And I missed the php.install. That makes sense. Still learning pieces about how tfb works as a whole. Didn't think to test another php framework after but now I realize it wouldn't work. I will remove this and continue work on it.

nbrady-techempower:
Changes to php7 toolset seem to allow me to go back and forth between different versions, both from a fresh install, and from a previously installed state.

nbrady-techempower:
@msmith-techempower this passes all relevant tests both locally and on travis, including the frameworks that were passing that depend on nginx, as no breaking changes are introduced. php7 vs php5 tests for php, hhvm (though probably irrelevant as the framework probably only depends on php for composer), slim and laravel.

I am looking into making changes for the php.ini files to be included with each individual framework, though since that is a rather large update to all the php frameworks it should probably be done after Rnd-12.

msmith-techempower:
Looks good; LGTM!

PR #1856, opened 2016-01-06
Fix Java/wildfly-ee7 travis-ci build fail issue

blee-techempower:
Fix build fail issue. 
The error occurs because of Java 7. 
Java 7 trying to use IPv6 to connect, but this framework uses IPv4.  
So telling java to use IPv4 is developer's role. 
After I explicitly indicate this using java option, problem disappeared magically.

Sounds simple, but this kind of error is super hard to debug. So keep in mind.


msmith-techempower:
Amazing catch; LGTM!

PR #1860, opened 2016-01-07
Fix Elixir/phoenix travis-ci test failing issue

blee-techempower:
Fix this framework failing travis-ci test.

### DIFF
--- a/frameworks/Elixir/phoenix/setup.sh
+++ b/frameworks/Elixir/phoenix/setup.sh
@@ -2,6 +2,21 @@
 
 fw_depends elixir
 
+# 
+# Caution: "line 20 has flaw."
+#
+#   Below sed command will cause another travis-ci test failing 
+#   if a person is executed local test. I think that line is original author's mistake.
+#   The logic is correct, but below command will make permanent change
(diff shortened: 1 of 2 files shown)


frameworks/Elixir/phoenix/setup.sh
```diff
@@ -2,6 +2,21 @@
...
+#   to 'config/proud.exs'. For this reason, if a tester of this framework is 
+#   going to execute this on local machine, then push this to remote
+#   it will fail travis-ci test.
+#
+#   For now, I keep this original code and leave this comment. 
+#   I will fix this problem later with better solution.
+#
+#   blee@techempower.com
+#
 sed -i 's|localhost|'${DBHOST}'|g' config/prod.exs
```
msmith-techempower: This looks correct to me; however, you are right that contributors need to be careful when submitting pull requests not to include files that were changed by these `sed` commands. The comment is probably overkill since we end up auditing code (like I am doing now) before accepting pull requests and i think that your fix to `config/prod.exs` is sufficient on its own. I will make this change and merge it in.

PR #1861, opened 2016-01-07
Fix Erlang/cowboy travis-ci test fail issue

blee-techempower:
Fix Travis-CI test failing issue. 
- Indicated 'dangerous' code to prevent problems that can be occurred by contributors or testers.
- Shrink size of db pool which causes problem on the travis-ci environment. Minimizing database pool size was a perfect solution. After shrinking the size of db pool, travis test stop failing.

### DIFF
--- a/frameworks/Erlang/cowboy/src/hello_world_app.erl
+++ b/frameworks/Erlang/cowboy/src/hello_world_app.erl
@@ -10,10 +10,16 @@
 
 %% API.
 
+%% NOTE: 
+%%   If size of db testpool is too big (e.g: 5000), 
+%%   it will fail travis ci test. So I shrink this to 256.
+%%   blee@techempower.com
+
+
 start(_Type, _Args) ->
         crypto:start(),
         application:start(emysql),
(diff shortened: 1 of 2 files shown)


frameworks/Erlang/cowboy/src/hello_world_app.erl
```diff
@@ -24,7 +30,7 @@ start(_Type, _Args) ->
       {"/query", query_handler, []}
 		]}
 	]),
-	{ok, _} = cowboy:start_http(http, 5000, [{port, 8080}], [
```
msmith-techempower: Presumably, these values were chosen for a specific reason. My worry here is that by lowering the number of connections available in the pool to get it working in Travis-CI, we may inadvertently degrade the performance of the actual benchmark when running on a machine that can handle that many connections.

msmith-techempower:
After discussing it over, we think 256 pool size (for db and connections) is probably okay; if we see degraded performance, we will revisit this... LGTM!

PR #1864, opened 2016-01-07
Updated WRK git hash to version 4

ssmith-techempower:
This should up the WRK version to 4.

### DIFF
--- a/toolset/setup/linux/client.sh
+++ b/toolset/setup/linux/client.sh
@@ -23,7 +23,7 @@ sudo sh -c "echo '*               -    nofile          65535' >> /etc/security/l
 
 git clone https://github.com/wg/wrk.git
 cd wrk
-git checkout 205a1960c8b8de5f500bb143863ae293456b7add
+git checkout a20969192f7bc095ad3ad3e238a9a581e64ae4f9
 make
 sudo cp wrk /usr/local/bin
 cd ~
--- a/toolset/setup/linux/client.sh
+++ b/toolset/setup/linux/client.sh
@@ -21,9 +21,9 @@ sudo sh -c "echo '*               -    nofile          65535' >> /etc/security/l
 # wrk
 ##############################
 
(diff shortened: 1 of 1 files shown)


toolset/setup/linux/client.sh
```diff
@@ -23,7 +23,7 @@ sudo sh -c "echo '*               -    nofile          65535' >> /etc/security/l
 
 git clone https://github.com/wg/wrk.git
 cd wrk
-git checkout 205a1960c8b8de5f500bb143863ae293456b7add
+git checkout a20969192f7bc095ad3ad3e238a9a581e64ae4f9
```
msmith-techempower: I think we would all prefer to use the zipped release instead of a git hash moving forward - any way we could change this to use 4.0.1?

https://github.com/wg/wrk/archive/4.0.1.zip

toolset/setup/linux/client.sh
```diff
@@ -21,9 +21,9 @@ sudo sh -c "echo '*               -    nofile          65535' >> /etc/security/l
 # wrk
 ##############################
 
-git clone https://github.com/wg/wrk.git
-cd wrk
-git checkout 205a1960c8b8de5f500bb143863ae293456b7add
+curl -sL -o wrk-4.0.1.tar.gz https://github.com/wg/wrk/archive/4.0.1.tar.gz
+tar xzf wrk-4.0.1.tar.gz
```
msmith-techempower: Change these lines to the following:
```bash
fw_get -o wrk-4.0.1.tar.gz https://github.com/wg/wrk/archive/4.0.1.tar.gz
fw_untar wrk-4.0.1.tar.gz
```

msmith-techempower:
We need to test this directly since Travis-CI does not use `wrk` and cannot determine whether the output matches the version we are currently using.

msmith-techempower:
I finally got around to testing this, and it looks good; LGTM!

PR #1866, opened 2016-01-08
Fix and update Erlang/elli

blee-techempower:
- No more build fail on travis-ci test.
- Reduced db pool size. Assigning large db pool size is a main reason of failing travis-ci test.
- Add plaintext route.
- Disable multi query test temporarily. Even this framework has multi query test route, the logic is not correct. For this reason, I just disable it for now. This can be added later by Erlang expert.


msmith-techempower:
Looks good; LGTM!

PR #1867, opened 2016-01-08
cache the content of the local Maven repository between builds

jochenberger:


.travis.yml
```diff
@@ -211,3 +211,7 @@ before_script:
 script:
   # Pick one test in this directory and verify
   - time ./toolset/run-ci.py verify "$TESTDIR"
+
+cache:
+  directories:
+  - $HOME/.m2/repository
```
msmith-techempower: This is interesting, but not the correct directory. `mvn` (and others) are run by the testrunner user, so it needs to be `/home/testrunner/.m2`, I believe.

msmith-techempower:
Well, Travis ran, so I suppose it is not broken; though, it will be hard to tell if this is really going to help us at all. 

Additionally, `testrunner` does not exist prior to the boot of the Travis-CI environment, so I'm not sure whether that will cause issues from run to run.

I guess we can always revert this if it breaks something we notice later; LGTM!

PR #1868, opened 2016-01-08
Fix rvm.sh

blee-techempower:
While I was investigating why most Ruby related framework are broken,
I notice that one problem at 'rvm.sh'. It does not solve current problem,
but it seems that RETCODE variable would have nonexisting value, so I change
it to have right value.

### DIFF
--- a/toolset/setup/linux/languages/rvm.sh
+++ b/toolset/setup/linux/languages/rvm.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-RETCODE=$(fw_exists rvm.installed)
+RETCODE=$(fw_exists ${IROOT}/rvm.installed)
 [ ! "$RETCODE" == 0 ] || { \
   # Assume single-user installation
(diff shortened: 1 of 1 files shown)


toolset/setup/linux/languages/rvm.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-RETCODE=$(fw_exists rvm.installed)
+RETCODE=$(fw_exists ${IROOT}/rvm.installed)
```
msmith-techempower: Good catch; this was likely just me being bone-headed.

blee-techempower: Thanks. I like to figure out what causes Ruby frameworks failing, but I am assigned to another project, so I just report this fact here.

msmith-techempower:
`rbx` is still broken for all the Ruby tests, but that is unrelated to this pull request; otherwise, it looks good, so LGTM!

PR #1869, opened 2016-01-08
PHP/silex fix

nbrady-techempower:
Silex was failing because of an outdated composer dependency which wasn't updating properly because of the 767 vendor files that shouldn't be in the repo


nbrady-techempower:
@msmith-techempower passes all relevant tests

msmith-techempower:
Looks good; LGTM!

PR #1870, opened 2016-01-08
Fix Go/falcore : logic error at multi query test

blee-techempower:


frameworks/Go/falcore/src/framework_benchmarks/falcore.go
```diff
@@ -182,13 +182,21 @@ var dbFilter = falcore.NewRequestFilter(func(req *falcore.Request) *http.Respons
...
 
-		if n <= 1 {
-			return dbFilter.FilterRequest(req)
+		// In the case of nStr is non-number string, n will be 0. So need to change n to 1.
+		if n == 0 {
+			n = 1
+		}
+
+		// In the case of nStr is number and its value is higher than 500, change n to 500
+		if n > 500 {
```
msmith-techempower: Hard tabs are hard sads.

blee-techempower: @msmith-techempower 
Hahaha.. sorry Mike. I blame VirtualBox. Sometimes I am not able to access VM using ssh( I still don't know a reason why. its weird), so I just did my work using vim on the VM console without text editor. Let me fix this problem.

msmith-techempower: It's not necessary; I just noticed the difference. I prefer spaces (like you used) to hard tabs, but the original files had hard tabs so w/e.

msmith-techempower:
Looks good; LGTM!

frameworks/Go/falcore/src/framework_benchmarks/falcore.go
kostya-sh: FYI, almost all Go code is formatted using gofmt that uses tabs for indentation. Some people do not like tabs but all Go developers like gofmt ;)

PR #1872, opened 2016-01-10
go-raw: use prepared statements

kostya-sh:
My benchmarks show:
- the same results for /db and /queries
- /update becomes slightly faster and scales better

### DIFF
--- a/frameworks/Go/go-raw/src/hello/hello.go
+++ b/frameworks/Go/go-raw/src/hello/hello.go
@@ -34,18 +34,7 @@ type Fortune struct {
 
 // Databases
 const (
-	// Go 1.4's sql.DB has scalability problem when using (explicitly reused) prepared statement.
-	// https://github.com/golang/go/issues/9484
-	//
-	// Using db.Query() instead of stmt.Query() avoid the issue.
-	// But it makes 3 round trips per query: prepare, execute and close.
-	// `interpolateParams=true` enables client side parameter interpolation.
-	// It reduces round trips without prepared statement.
-	//
-	// Before Go 1.5 is released, we can see real power of Go with this benchmark.
-	// After Go 1.5 is released, we can see prepared statement vs interpolation by comparing
-	// this and another lightweight Go framework.
-	connectionString   = "benchmarkdbuser:benchmarkdbpass@tcp(localhost:3306)/hello_world?interpolateParams=true"
+	connectionString   = "benchmarkdbuser:benchmarkdbpass@tcp(localhost:3306)/hello_world"
 	worldSelect        = "SELECT id, randomNumber FROM World WHERE id = ?"
 	worldUpdate        = "UPDATE World SET randomNumber = ? WHERE id = ?"
 	fortuneSelect      = "SELECT id, message FROM Fortune;"
@@ -61,6 +50,9 @@ var (
 
 	// Database
 	db *sql.DB
+
+	worldSelectStmt *sql.Stmt
+	worldUpdateStmt *sql.Stmt
 
 	helloWorldBytes = []byte(helloWorldString)
 )
@@ -83,6 +75,15 @@ func main() {
 		log.Fatalf("Error opening database: %v", err)
 	}
 	db.SetMaxIdleConns(maxConnectionCount)
+
+	worldSelectStmt, err = db.Prepare(worldSelect)
+	if err != nil {
+		log.Fatalf("Error preparing %s statment: %v", worldSelect, err)
+	}
+	worldUpdateStmt, err = db.Prepare(worldUpdate)
+	if err != nil {
+		log.Fatalf("Error preparing %s statment: %v", worldUpdate, err)
+	}
 
 	http.HandleFunc("/db", dbHandler)
 	http.HandleFunc("/queries", queriesHandler)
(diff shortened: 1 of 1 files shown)


frameworks/Go/go-raw/src/hello/hello.go
```diff
@@ -34,18 +34,7 @@ type Fortune struct {
...
-	// https://github.com/golang/go/issues/9484
-	//
-	// Using db.Query() instead of stmt.Query() avoid the issue.
-	// But it makes 3 round trips per query: prepare, execute and close.
-	// `interpolateParams=true` enables client side parameter interpolation.
-	// It reduces round trips without prepared statement.
-	//
-	// Before Go 1.5 is released, we can see real power of Go with this benchmark.
-	// After Go 1.5 is released, we can see prepared statement vs interpolation by comparing
-	// this and another lightweight Go framework.
```
methane: Have you read this comment?
All other go frameworks uses prepared statement.
If go-raw uses prepared statement, I can't see performance difference between prepared statement and
client side interpolation.

kostya-sh: I've read this comment however I'd like to ask how important this comparison is given that:

- there are other differences between frameworks, so it is impossible to explain the difference in numbers by a single factor (prepared vs unprepared statements)
- scalability of prepared statements has been fixed in Go 1.5
- a version with prepared statements has the same or better performance with Go 1.5 and mysql
- using prepared statements is more secure and generally should be prefered in real production usage

methane: > there are other differences between frameworks, so it is impossible to explain the difference in numbers by a single factor (prepared vs unprepared statements)

There are many "micro" frameworks in Go.  Performance difference between them are only URL routing.
In case of "multiple queries" benchmark, difference other than DB access is small enough to ignore.

> scalability of prepared statements has been fixed in Go 1.5

Yes. I'm who fixed it.
Have you *really* read the comment?
"After Go 1.5 is released, we can see prepared statement vs interpolation by comparing
 this and another lightweight Go framework."

> a version with prepared statements has the same or better performance with Go 1.5 and mysql

I want to see it on "Framework Benchmarks result" page.

> using prepared statements is more secure and generally should be prefered in real production usage

I implemented "interpolateParams" secure and I use it on real production.
There are some pros in prepared statement:

* Some ORM doesn't support reusing prepared statement.  In such case, prepared statement requires three
roundtrips (prepare, exec, close).
* "SHOW FULL PROCESSLIST" is handy tool for DB admins.  Prepared statement hides real SQL from PROCESSLIST.

methane:
Could you add "go-raw-prepared" test?
It shouldn't have "json" and "plaintext" tests.

methane:
I've did it.
Could you review #1873?

kostya-sh:
@methane, I like your approach with a separate test. I've left a few comments for  #1873, please review.

A few other comments/questions:

- Should a version with prepared statement be a default one (go-raw)? interpolateParams is something that is only supported by the mysql driver. If in the future more databases and/or drivers are added I think it would be more logical to have entries like `go-raw-mysql`, `go-raw-mysql-interpolated`, `go-raw-postgres`, etc.
- If some ORMs do not support re-using prepared statements then it should be shown in the benchmarks numbers for these ORMs. This will help developers understand limitations of these ORMs.
- I didn't imply that your implementation of interpolatedParams is insecure. However the server side interpolation with prepared statements has been much more thoroughly tested and easier to trust.

Basically I agree with you that the comparison is interesting and useful. However I believe that go-raw version of the benchmark should use common best practices. And I believe that using prepared statements is one of such best practices that should be used by default unless there are good reasons not to use it (such as simplifying life for DB admins or not being able to reuse prepared statements due to user requirements).

Also you might be interested in the benchmarks numbers that I've got using 3 m4.4xlarge EC2 instances (mysql, wrk and the app running on separate hosts):

    /update?queries=1               
                                    
    original:                       
    Threads	Conns	Reqs/s          
    1	1	349.31                    
    2	2	781.37                    
    4	4	1382.53                   
    8	8	2844.12                   
    16	16	5438.69                  
    32	32	8416.48                  
    64	64	11799.48                 
    128	128	12629.00                
                                    
    prepared statements:            
                                    
    Threads	Conns	Reqs/s          
    1	1	372.30                    
    2	2	762.52                    
    4	4	1427.83                   
    8	8	2748.05                   
    16	16	5652.36                  
    32	32	10481.72                 
    64	64	16208.63                 
    128	128	23571.69                
                                    
                                    
    /update?queries=20              
                                    
    original:                       
    Threads	Conns	Reqs/s          
    1	1	22.57                     
    2	2	46.26                     
    4	4	81.95                     
    8	8	160.32                    
    16	16	294.96                   
    32	32	468.16                   
    64	64	602.36                   
    128	128	630.06                  
                                    
    prepared statements:            
    Threads	Conns	Reqs/s          
    1	1	22.17                     
    2	2	47.33                     
    4	4	83.07                     
    8	8	166.52                    
    16	16	313.52                   
    32	32	592.91                   
    64	64	918.15                   
    128	128	1243.36                 

Finally thank you for your contributions and making Go better! ;)

kostya-sh:
Closing this PR in favor of #1873

PR #1873, opened 2016-01-11
Add go-prepared test

methane:


frameworks/Go/go-raw/src/hello/hello.go
```diff
@@ -48,11 +48,17 @@ const (
 	connectionString   = "benchmarkdbuser:benchmarkdbpass@tcp(localhost:3306)/hello_world?interpolateParams=true"
```
kostya-sh: Does `interpolateParams=true` affect prepared statement version? I assume that no

frameworks/Go/go-raw/src/hello/hello.go
```diff
@@ -77,17 +106,15 @@ func main() {
...
+	http.HandleFunc("/json", jsonHandler)
 	http.HandleFunc("/db", dbHandler)
+	http.HandleFunc("/dbPrepared", dbPreparedHandler)
 	http.HandleFunc("/queries", queriesHandler)
-	http.HandleFunc("/json", jsonHandler)
+	http.HandleFunc("/queriesPrepared", queriesPreparedHandler)
 	http.HandleFunc("/fortune", fortuneHandler)
+	http.HandleFunc("/fortunePrepared", fortunePreparedHandler)
 	http.HandleFunc("/update", updateHandler)
 	http.HandleFunc("/plaintext", plaintextHandler)
```
kostya-sh: `/updatePrepared` is missing.

frameworks/Go/go-raw/src/hello/hello.go
```diff
@@ -194,15 +256,23 @@ func fortuneHandler(w http.ResponseWriter, r *http.Request) {
...
+	if err := tmpl.Execute(w, fortunes); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
 	}
-	rows.Close()
+}
+
+func fortunePreparedHandler(w http.ResponseWriter, r *http.Request) {
+	rows, err := fortuneSelectPrepared.Query()
+	if err != nil {
+		log.Fatalf("Error preparing statement: %v", err)
```
kostya-sh: This error message is a bit confusing. How about 'Error executing statement'?

frameworks/Go/go-raw/src/hello/hello.go
kostya-sh: It would be good to update the comment to mention that there are two benchmarks now.

frameworks/Go/go-raw/src/hello/hello.go
```diff
@@ -187,6 +225,30 @@ func queriesHandler(w http.ResponseWriter, r *http.Request) {
...
+	if nStr := r.URL.Query().Get("queries"); len(nStr) > 0 {
+		n, _ = strconv.Atoi(nStr)
+	}
+
+	if n < 1 {
+		n = 1
+	} else if n > 500 {
+		n = 500
+	}
+
```
kostya-sh: This code (from line 229) is duplicated 4 times now. May be extract it in a separate method? (`getQueriesParam(r *http.Request)`)

methane:
fixed

frameworks/Go/go-raw/benchmark_config.json
```diff
@@ -24,6 +24,27 @@
       "notes": "",
       "versus": "go"
     },
+    "prepared": {
```
kostya-sh: should be "interpolate"

kostya-sh:
One minor comment, otherwise looks good

methane:
good catch!

msmith-techempower:
All tests are passing; LGTM!

Unrelated question: should `plaintext` and `json` be removed from `go-prefork` as well? It seems like these are only testing database stuff, so why include those two tests, right?

PR #1874, opened 2016-01-11
Upgraded wicket to 7.1 and other deps

dashorst:
This PR upgrades wicket to 7.1 and upgrades other dependencies to their latest counterparts as well.
It also removes the Eclipse meta-data, which gets in the way when you use Eclipse 4.4 or newer to
import the project as a Maven project, which is the preferred way of managing your projects when they
are based on a Maven POM.

### DIFF
--- a/frameworks/Java/wicket/src/main/java/hellowicket/WicketApplication.java
+++ b/frameworks/Java/wicket/src/main/java/hellowicket/WicketApplication.java
@@ -4,7 +4,7 @@ import javax.naming.InitialContext;
 import javax.sql.DataSource;
 
 import org.apache.wicket.protocol.http.WebApplication;
-import org.apache.wicket.settings.IRequestCycleSettings;
+import org.apache.wicket.settings.RequestCycleSettings;
 
 import com.zaxxer.hikari.HikariDataSource;
 
@@ -42,7 +42,7 @@ public class WicketApplication extends WebApplication
 
 		// disable response caching to be more close to other
 		// test applications' behavior
-		IRequestCycleSettings requestCycleSettings = getRequestCycleSettings();
+		RequestCycleSettings requestCycleSettings = getRequestCycleSettings();
 		requestCycleSettings.setBufferResponse(false);
 
 		// set UTF-8 for /fortunes test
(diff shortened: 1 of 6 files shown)


frameworks/Java/wicket/src/main/java/hellowicket/fortune/FortunePage.java
```diff
@@ -16,48 +16,42 @@
...
-  }
+public class FortunePage extends WebPage {
+	private static final long serialVersionUID = 1L;
+
+	public FortunePage() throws Exception {
+		List<Fortune> fortunes = new ArrayList<>(10000);
+
+		DataSource dataSource = WicketApplication.get().getDataSource();
+		try ( //
+				Connection connection = dataSource.getConnection();
```
martin-g: It seems the connection and the statement are not closed now.

msmith-techempower:
Travis reports SQLExceptions on all of the database tests.

frameworks/Java/wicket/src/main/java/hellowicket/WicketApplication.java
```diff
@@ -85,17 +85,17 @@ private DataSource newDataSource()
 				InitialContext jndiContext = new InitialContext();
 				dataSource = (DataSource) jndiContext.lookup("java:comp/env/jdbc/hello_world");
 			}
-			else
+			else try(HikariDataSource ds = new HikariDataSource();)
```
martin-g: This close the data source right after the setup

frameworks/Java/wicket/src/main/java/hellowicket/fortune/FortunePage.java
dashorst: No: this is the try-with resources statement that ignores exceptions upon close, so all resources claimed in the try() block are auto-closed. See https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

msmith-techempower: I agree that the connection, prepared statement, and result set are closed via the try-with-resources, but does the DataSource not need to be closed as well (as it is outside the try)?

martin-g: @dashorst I know try-with-resources, just the diff is awful due to the change of the whitespaces and I used `?w=1` in the url to ignore them, and this seems to confused me even more.

@msmith-techempower No the data source should not be closed after usage. It is being closed when the application stops, i.e. on Servlet's `Filter#onDestroy()`. Aren't there any logs from Resin that contain the exceptions which lead to error code 500 ?

msmith-techempower: There are, but we cannot get them out of the Travis-CI environment because they spin-up and tear-down after each test run.

martin-g: Locally the exception is:
```
java.sql.SQLException: HikariDataSource HikariDataSource (null) has been closed.
	at com.zaxxer.hikari.HikariDataSource.getConnection(HikariDataSource.java:79)
	at hellowicket.HelloDbResponse.getDataFromDatabase(HelloDbResponse.java:69)
	at hellowicket.HelloDbResponse.newResourceResponse(HelloDbResponse.java:45)
	at org.apache.wicket.request.resource.AbstractResource.respond(AbstractResource.java:629)
	at org.apache.wicket.request.handler.resource.ResourceRequestHandler.respond(ResourceRequestHandler.java:105)
```

@dashorst Please read https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html and remove it from `hellowicket.WicketApplication#newDataSource()` for Hikari ;-)
There were few PRs in the last days for using Hikari instead of Resin's DataSource. I think Wicket should do this change too.

martin-g: Actually it is already
```java
private boolean useResinDataSource()
{
	return false;
}
```

so we don't use Resin's DS for some time.
This explains why it fails at Travis too.

frameworks/Java/wicket/src/main/java/hellowicket/WicketApplication.java
dashorst: will remote the try with...

msmith-techempower:
All tests pass; LGTM!

dashorst:
:+1: Merge, merge, MERGE! :octocat: :beers:

PR #1875, opened 2016-01-11
Perl/web-simple - Fixed Failing tests

nbrady-techempower:
Fixed failing query tests and fixed all warnings.
````
  Verification Summary

| Test: web-simple-raw
|       query       : PASS
|       db          : PASS
````

### DIFF
--- a/frameworks/Perl/web-simple/app.pl
+++ b/frameworks/Perl/web-simple/app.pl
@@ -3,7 +3,7 @@ use Web::Simple;
 use JSON::XS;
 use DBI;
 
-my $dsn = "dbi:mysql:database=hello_world;host=localhost";
+my $dsn = "dbi:mysql:database=hello_world;host=127.0.0.1";
 my $dbh = DBI->connect( $dsn, 'benchmarkdbuser', 'benchmarkdbpass', { RaiseError => 1 });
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: frameworks/Perl/web-simple/app.pl [hunk 1]
```diff
@@ -3,7 +3,7 @@
 use JSON::XS;
 use DBI;
 
-my $dsn = "dbi:mysql:database=hello_world;host=localhost";
+my $dsn = "dbi:mysql:database=hello_world;host=127.0.0.1";
```
msmith-techempower: This change seems unnecessary, but if it **is** necessary, then [this line](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Perl/web-simple/setup.sh#L3) will have to be updated accordingly.

nbrady-techempower: That must have been a change made by a sed operation, as I didn't make it.

msmith-techempower: Righto - let's clean this line up, then.

nbrady-techempower: All set

msmith-techempower:
All tests pass; LGTM!

PR #1876, opened 2016-01-11
Compojure, http-kit: Switch from c3p0 to HikariCP db pool

k-r-g:
HikariCP is significantly faster in my testing (and the HikariCP published benchmarks), while still being suitable for production use.


msmith-techempower:
Looks good; LGTM!

PR #1878, opened 2016-01-12
Grizzly-Jersey: Update deps, use HikariCP instead of c3p0 db pool

k-r-g:
HikariCP is significantly faster than c3p0 in published benchmarks and my crude testing.

I also updated all the Grizzly dependencies to the latest versions.

@trautonen Since you authored the original Grizzly db and fortunes tests, I'd like to confirm with you: Is this change reasonable?


msmith-techempower:
@k-r-g `Unsupported major/minor version: 52.0` [at around here](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/101730454#L3038)

trautonen:
Mustache.java requires Java 8:
```
As of release 0.9.0 mustache.java is now Java 8 only. For Java 6/7 support use 0.8.x.
```

So lets just update to Java 9 like @msmith-techempower suggests. This uses Jersey 1, but the latest versions should be Java 8 compatible.

Other than that, this looks good to go.

k-r-g:
@trautonen Sounds good. When #1882 is merged, I'll fix this up. Thanks!

msmith-techempower:
@k-r-g Rebase off master please!

msmith-techempower:
All tests pass; LGTM!

PR #1879, opened 2016-01-12
go-raw: simplify deployment

kostya-sh:
This change shouldn't affect performance numbers but it makes it much easier to run go-raw applications without TFB framework (e.g. `DBHOST=dbhost go run hello.go`)

See individual commit messages for more details.


msmith-techempower:
Travis tests look good; LGTM!

PR #1881, opened 2016-01-12
Undertow jersey x2

cmsimike:
Two different undertow jersey implementations using two different connection pools


msmith-techempower:
Looks good; LGTM!

PR #1883, opened 2016-01-12
Added the --clean results option to the toolset

nbrady-techempower:
@msmith-techempower What do you think of this? I set it up so I could easily add `--clean installs` and `--clean all` as well, but right now `toolset/run-tests.py --clean results` will remove the results directory


toolset/benchmark/benchmarker.py
```diff
@@ -963,6 +964,8 @@ def __init__(self, args):
 
     # setup results and latest_results directories 
     self.result_directory = os.path.join("results", self.name)
+    if args['clean'] == 'results':
+        shutil.rmtree(self.result_directory)
```
msmith-techempower: I'm fine with adding `--clean`.

Broader question: does it make sense for `--clean` to also nuke the installs directory?

nbrady-techempower:
@msmith-techempower i think in most cases, while you're testing the framework, you just want to remove the results dir because reinstalling can add a ton of time, that's why i suggested the 3 different options for `--clean [results, installs, all]`

msmith-techempower:
I see the comment now.

After thinking about it, I think we should just add two flags: `--clean` and `--clean-all`. Your current implementation of `--clean` is great, and `--clean-all` would also include the installs directory.

nbrady-techempower:
@msmith-techempower is there some travis config that needs to be done to recognize additional arguments? this works locally but bombs on travis

nbrady-techempower:
@msmith-techempower this should be good to go now. I added that quick fix for vagrant users, and was sure to do a check if the directory existed first, just in case it caused any errors

msmith-techempower:
Looks good; LGTM!

PR #1886, opened 2016-01-13
Luminus: Add routes to pass benchmark verification tests

k-r-g:
The verify script tries /queries/ and /updates/ with no count param, and
the Luminus test implementation needs to treat this the same as if a 1 
was passed in.

Also use ring.util.response to set the Content-Type header for plaintext.


msmith-techempower:
Looks good; LGTM!

PR #1887, opened 2016-01-14
Lua language fixes

nbrady-techempower:
Fixes an issue where postgres doesn't start on vagrant because it's requesting too much memory


blee-techempower:
Nice catch!

nbrady-techempower:
@msmith-techempower lua tests pass!

msmith-techempower:
Looks good; LGTM!

PR #1888, opened 2016-01-14
ULib: some tuning plus add mongodb test

stefanocasazza:
Hi,

I point out again that no longer appear the results of ULib in the [round 10](https://www.techempower.com/benchmarks/#section=data-r10&hw=peak&test=plaintext)

Greetings
Stefano


toolset/setup/linux/frameworks/ulib.sh
```diff
@@ -28,6 +27,17 @@ if [ $? -ne 0 ]; then
    sudo apt-get install -y libcap2-bin
 fi
 
+# We need to install mongo-c-driver (we don't have a ubuntu package)
+RETCODE=$(fw_exists ${IROOT}/mongo-c-driver.installed)
+if [ "$RETCODE" != 0 ]; then
```
msmith-techempower: This `RETCODE` business is unnecessary since you have the sentinel check above (`ulib.installed`). If the script gets here, then ulib is not installed and therefore the mongo-c-driver is not installed.

toolset/setup/linux/frameworks/ulib.sh
```diff
@@ -28,6 +27,17 @@ if [ $? -ne 0 ]; then
...
+  wget https://github.com/mongodb/mongo-c-driver/releases/download/1.1.10/mongo-c-driver-1.1.10.tar.gz
+  tar -xzf mongo-c-driver-1.1.10.tar.gz
```
msmith-techempower: Please change these to `fw_get` and `fw_untar` respectively.

toolset/setup/linux/frameworks/ulib.sh
```diff
@@ -28,6 +27,17 @@ if [ $? -ne 0 ]; then
...
+  cd mongo-c-driver-1.1.10/
+  ./configure --prefix=$IROOT --libdir=$IROOT
+  make && sudo make install
```
msmith-techempower: Probably do not need `sudo` here.

toolset/setup/linux/frameworks/ulib.sh
stefanocasazza: I noticed that other frameworks (ffead-cpp) must do the same installation of mongo-c-driver,
I thought that this would spare duplication of work...

msmith-techempower: Understood; that's fine then.

msmith-techempower:
[Travis](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/102634269) says that many of the ULib tests failed to start/run.

stefanocasazza:
Hi,

this happens to be by a bit of time and I have no idea why, I believe it is
related to the Travis environment. I don't see any problem in an
local environment.

Stefano
​

stefanocasazza:
Hi,
I do not understand the meaning of the label attached ' PR: Please update!'.
what should I do... ? @msmith-techempower

Stefano
​

msmith-techempower:
I would say create a completely clean test environment, clone the repository, set it up, then merge your pull request in and test ulib. This is essentially what Travis-CI does every time it runs, so if the logs are not enough information, you will have to do the process manually to try and reproduce the issue.

stefanocasazza:
In my opinion the issue is not relative to the pull request, it happen also
now with the current status of framework, there are also plenty of other
frameworks that have the same failure on travis...

2016-01-29 16:18 GMT+01:00 Mike Smith <notifications@github.com>:

toolset/setup/linux/frameworks/ulib.sh
```diff
@@ -28,6 +27,17 @@ if [ $? -ne 0 ]; then
    sudo apt-get install -y libcap2-bin
 fi
 
+# We need to install mongo-c-driver (we don't have a ubuntu package)
+#RETCODE=$(fw_exists ${IROOT}/mongo-c-driver.installed)
+#if [ "$RETCODE" != 0 ]; then
+  fw_get -O https://github.com/mongodb/mongo-c-driver/releases/download/1.1.10/mongo-c-driver-1.1.10.tar.gz
```
msmith-techempower: So, this is causing issues. Apparently, github is responding with a head suggesting the filesize is [something] but the actual size of the file is [something else], which causes `curl` to return a status 18.

stefanocasazza: Hi,

I added the fix but Travis remain broken, it can someone please pull and test locally so it can be verified my opinion of correctness for the pull request...? @msmith-techempower

Greetings
Stefano

ssmith-techempower:
I've recently attempted to run these tests locally, and they seem to all be working correctly minus the MongoDB tests, which are returning 404s. I'll be looking into ulib on Travis and see if we can't figure out what the problem is there.

stefanocasazza:
Ok,
thanks...

2016-02-01 23:09 GMT+01:00 ssmith-techempower <notifications@github.com>:

toolset/setup/linux/frameworks/ulib.sh
```diff
@@ -22,11 +21,22 @@ fi
...
 
-if [ $? -ne 0 ]; then
+#if [ $? -ne 0 ]; then
    sudo apt-get install -y libcap2-bin
-fi
+#fi
+
+# We need to install mongo-c-driver (we don't have a ubuntu package)
+#RETCODE=$(fw_exists ${IROOT}/mongo-c-driver.installed)
+#if [ "$RETCODE" != 0 ]; then
```
msmith-techempower: These commented-out checks will need to be reinstated before we merge in.

msmith-techempower:
We have also elected to disallow Redis tests for the time being, so please remove that from your `benchmark_config.json`

msmith-techempower:
Additionally, seems that all your database connection attempts are [failing](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/114781643#L4208).

Issue #1894, opened 2016-01-19
aleph, compojure, and http-kit benchmarks are incorrect


yogthos:
These benchmarks appear to have been created from the same template and use `def` to define queries, e.g:

```clojure
(def single-query-test-raw
  "Test 2: Single database query (raw)"
  (-> 1
      (run-queries-raw)
      (first)
      (ring-resp/response)))
```

This runs the code at compile time and binds the resulting value to the variable. This code isn't testing the performance at runtime.

k-r-g:
Ouch, good catch. I'll create a PR to fix these.

msmith-techempower:
Resolved with #1896

PR #1896, opened 2016-01-20
aleph/compojure/http-kit: Use defn instead of def for handler functions

k-r-g:
This is intended to address #1894 

They were mistakenly using def and using a value set at compile time. This
contradicts the intention of the benchmark test.

### DIFF
--- a/frameworks/Clojure/aleph/hello/src/hello/handler.clj
+++ b/frameworks/Clojure/aleph/hello/src/hello/handler.clj
@@ -7,21 +7,23 @@
     [clj-tuple :as t])
   (:gen-class))
 
-(def plaintext-response
+(defn plaintext-response
+  []
   (t/hash-map
     :status 200
     :headers (t/hash-map "content-type" "text/plain; charset=utf-8")
     :body (bs/to-byte-array "Hello, World!")))
 
-(def json-response
+(defn json-response
+  []
   (t/hash-map
     :status 200
     :headers (t/hash-map "content-type" "application/json")))
 
 (defn handler [{:keys [uri] :as req}]
   (cond
-    (= "/plaintext" uri) plaintext-response
-    (= "/json" uri) (assoc json-response
+    (= "/plaintext" uri) (plaintext-response)
+    (= "/json" uri) (assoc (json-response)
                       :body (json/encode (t/hash-map :message "Hello, World!")))
     :else {:status 404}))
 
--- a/frameworks/Clojure/aleph/hello/src/hello/handler.clj
+++ b/frameworks/Clojure/aleph/hello/src/hello/handler.clj
@@ -7,23 +7,21 @@
     [clj-tuple :as t])
   (:gen-class))
 
-(defn plaintext-response
-  []
+(def plaintext-response
   (t/hash-map
     :status 200
     :headers (t/hash-map "content-type" "text/plain; charset=utf-8")
     :body (bs/to-byte-array "Hello, World!")))
 
-(defn json-response
-  []
+(def json-response
   (t/hash-map
     :status 200
     :headers (t/hash-map "content-type" "application/json")))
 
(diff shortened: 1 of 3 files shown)


k-r-g:
@ztellman and @shenfeng : Are these changes to aleph and http-kit reasonable? I think with these changes we comply with the test requirements.

ztellman:
I don't really see the changes here as comparable to the one described in #1894.  The code example shown there fails to actually make a query, which clearly contravenes the intent of the benchmark.

Conversely, all that's being done here is pre-allocating some data structures, which seems very similar in spirit to, for instance: https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/netty/src/main/java/hello/HelloServerHandler.java#L37-L44.  Certainly this is an optimization I've used in actual web applications for certain fast paths.

If there's some final authority on what's appropriate here, please call them in, but I don't agree that this is an obvious change to make.

k-r-g:
@ztellman For the JSON serialization test I think I agree with you. 

https://www.techempower.com/benchmarks/#section=code
"1. For each request, an object mapping the key message to Hello, World! must be instantiated."

Your original implementation complies with this because when the /json request comes in you create the new map and merge it with the compile-time-generated headers. I will undo my change to the aleph json test.

However, I'm not sure I agree with you on the plaintext test.

https://www.techempower.com/benchmarks/#section=code
"4. This test is not intended to exercise the allocation of memory or instantiation of objects. Therefore it is acceptable but not required to re-use a single buffer for the response text (Hello, World). However, the response must be fully composed from this and its headers within the scope of each request and it is not acceptable to store the entire payload of the response, headers inclusive, as a pre-rendered buffer."

I believe your original implementation does "store the entire payload of the response, headers inclusive, as a pre-rendered buffer." Do you disagree?

ztellman:
"Buffer" implies encoded bytes to me, not a data structure (Clojure map) which will be encoded to a different data structure (Netty `HttpResponse`), which will then be encoded to bytes.  The use of the verb "render" seems to back this up.

To be honest, I think this change will have negligible performance impact, and if anyone can show me a precedent for doing it this way, I'm fine with it.  I just don't think this is actually in the spirit of the benchmark.

bhauer:
Hi @ztellman.  I'd like to invite @yogthos to weigh in here as well since he brought this to our attention in #1894.

For my part, I will confirm that my intent with the words "pre-rendered buffer" was a byte array or String that fully represented the response, inclusive of some or all HTTP response headers and payload.

I feel that the implementation in Aleph wherein a singleton HashMap is constructed with three attributes (status, a response header, and the payload) is within the spirit of the test.  By way of illustration, although I am not aware of a Java framework that operates this way, I can conceive of a Java framework that has a "Response" data structure that contains similar attributes (status, zero or more headers, and the payload).  I would then expect the implementation to construct a singleton representing that response and for the plaintext handler method to be something like:

    return plainTextResponse;

Which at first glance would look alarmingly simplistic.  But the "heavy lifting" of composing that Response object into a String or byte array/buffer inclusive of headers (such as `Date`) remains the work of the framework, which is the spirit of the test.

k-r-g:
@ztellman I have reverted all my changes to aleph.

@yogthos I have reverted my changes to the plaintext test for compojure and http-kit. I left the json handler as defn so that we continue to exercise the json serializer.

I think aleph, compojure, and http-kit are now all in agreement on this issue. Have I missed anything? @yogthos did I address your concerns from #1894 ?

yogthos:
Yeah, I think this makes sense to me and addresses the original issue.

bhauer:
I have updated the Plaintext test requirements to be more clear about the intent of avoiding a pre-rendered buffer for the entirety (or unnaturally large subset) of the response.

https://www.techempower.com/benchmarks/#section=code

msmith-techempower:
All tests passed; LGTM!

PR #1899, opened 2016-01-25
Clojure tests: Update to Clojure 1.8.0

k-r-g:
Also update other dependencies to be current.


msmith-techempower:
Looks good; LGTM!

PR #1903, opened 2016-01-28
fasthttp-postgresql: small /update optimization

valyala:
This change improves /update performance a bit and makes /update request handler more clear.


msmith-techempower:
All tests pass; LGTM!

PR #1904, opened 2016-01-28
Yesod to use stack

ssmith-techempower:
This PR updates the Yesod (Haskell) framework to be built using Stack, a build tool built on top of the usual Haskell build tools with an emphasis on reproducible builds. This should both fix the currently broken Yesod tests as well as hopefully make the Yesod tests more resistant to future changes of Hackage packages.

### DIFF
--- a/frameworks/Haskell/yesod/bench/bench.cabal
+++ b/frameworks/Haskell/yesod/bench/bench.cabal
@@ -18,17 +18,17 @@ executable         bench
                 EmptyDataDecls
                 CPP
 
-    build-depends: base                          >= 4.7        && < 5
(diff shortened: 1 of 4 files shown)


msmith-techempower:
Getting pretty close; `yesod` failed to startup, but `yesod-mongodb-raw` passed all tests.

msmith-techempower:
`Codec.Compression.Zlib: compressed data stream format error (invalid stored block lengths)` was the error, apparently.

msmith-techempower:
Actually, the first build worked fine. I'm going to chalk it up to "internet is hard"; LGTM!

PR #1906, opened 2016-01-29
ninja/compojure/http-kit: Fix setup.sh to set DBHOST for HikariCP

k-r-g:
I had neglected to fix setup.sh to properly replace the database host IP with DBHOST.


msmith-techempower:
Tests pass; LGTM!

PR #1907, opened 2016-01-29
Fixed some Java frameworks that were still depending on 'java8'

ssmith-techempower:
A couple of Java frameworks were still depending on `java8` instead of just `java`. This PR fixes those frameworks.


msmith-techempower:
All tests pass; LGTM!

PR #1909, opened 2016-02-02
update to wildfly 10 Final

codylerum:


ssmith-techempower:
Runs on Travis and works for me locally, looks good to me.

PR #1911, opened 2016-02-05
OfficeFloor

sagenschneider:
Requires the:
 - java 8 fix
 - PR to use latest maven version

PR #1912, opened 2016-02-05
Using latest maven version

sagenschneider:


sagenschneider:
Note the $IROOT/maven.installed file will need to be deleted for the later maven version to be installed.

msmith-techempower:
I am not sure why this is necessary; #1911 passed all tests without this being merged in.

sagenschneider:
Yes, I realised I had put in a fix within the OfficeFloor plug-ins to avoid this issue (flag that turns of the functionality, as was not necessary for this test).  Hence now working.

I will leave it to you, as to whether a later version of Maven is worth the risk of introducing.

msmith-techempower:
Thanks for the explanation - I am in favor of leaving this until it is necessary.

PR #1913, opened 2016-02-08
Fixed setup script for Luminus

ssmith-techempower:
Luminus's setup script had a mistake in the `sed` command which caused the DB host to not be updated. Since the DB host defaults to localhost this error only manifested on the Peak servers and not on Travis or when run locally with everything on the same machine.


msmith-techempower:
Tests passed; LGTM!

PR #1919, opened 2016-02-12
migrate old postgres code to new async driver and added mysql to the …

pmlopes:
…list of implementations

Signed-off-by: Paulo Lopes <paulo@mlopes.net>


ssmith-techempower:
Looks good to me, merging in.

PR #1921, opened 2016-02-13
Upgrade Ur/Web version

achlipala:


ssmith-techempower:
Not sure why Travis is having trouble with this, but looks good to me.

Issue #1152, opened 2014-10-29
Java tests - java.util.Random usage may lead to non-optimal performance in heavy concurency conditions


zloster:
[Java documentation](http://docs.oracle.com/javase/7/docs/api/java/util/Random.html) states the following:
>Instances of java.util.Random are threadsafe. However, the concurrent use of the same java.util.Random instance across threads may encounter contention and consequent poor performance. Consider instead using ThreadLocalRandom in multithreaded designs. 

[Here](http://java-performance.info/java-util-random-java-util-concurrent-threadlocalrandom-multithreaded-environments) you can find a nice article with comparision IMO.

And the files where java.util.Random is used :
```shell
:~/source/FrameworkBenchmarks$ grep -r "import java.util.Random;" ./
./frameworks/Java/grizzly-jersey/src/main/java/hello/DbResource.java:import java.util.Random;
./frameworks/Java/play1siena/app/controllers/Application.java:import java.util.Random;
./frameworks/Java/ninja-standalone/src/main/java/controllers/SetupController.java:import java.util.Random;
./frameworks/Java/ninja-standalone/src/main/java/controllers/HelloDbController.java:import java.util.Random;
./frameworks/Java/activeweb/src/main/java/app/controllers/DatabaseController.java:import java.util.Random;
./frameworks/Java/spring/src/main/java/com/techempower/spring/web/WorldDatabaseController.java:import java.util.Random;
./frameworks/Java/dropwizard/src/main/java/com/example/helloworld/resources/WorldResource.java:import java.util.Random;
./frameworks/Java/restexpress/src/main/java/hello/controller/MysqlController.java:import java.util.Random;
./frameworks/Java/restexpress/src/main/java/hello/controller/MongodbController.java:import java.util.Random;
./frameworks/Java/play1/app/controllers/Application.java:import java.util.Random;
./frameworks/Java/vertx/WebServer.java:import java.util.Random;
./frameworks/Java/play2-java/play2-java-ebean/app/controllers/Application.java:import java.util.Random;
./frameworks/Java/play2-java/play2-java-jpa/app/controllers/Application.java:import java.util.Random;
./frameworks/Java/spark/src/main/java/hello/web/SparkApplication.java:import java.util.Random;
./frameworks/Java/dropwizard-mongodb/src/main/java/com/example/helloworld/resources/WorldResource.java:import java.util.Random;
:~/source/FrameworkBenchmarks$

dgryski:
[...]

PR #1935, opened 2016-02-20
Changed the range of the id

zloster:
Resolves the last bit of #1152.
Now the id value will be from 1 to 10000 inclusive. The 'nextInt(int maxValue)' returns results in the range [0, maxValue) i.e. maxValue is excluded.


ssmith-techempower:
Looks good to me, merging in.

PR #1939, opened 2016-02-24
Update rust

steveklabnik:
I tried to keep these in small commits in case you didn't want any of these changes.

This PR:

1. Updates Rust from 1.0.0 to 1.6.0. It would be nice if we could have a "latest stable", but that would require infrastructure work I can't do right now.
2. Update the various frameworks and their versions. There was also breaking changes to one of the libraries, so the code was updated.
3. Various style/formatting fixes. They are now in line with the standard Rust style, and are formatted according to `rustfmt`, the automatic formatting tool.
4. A small fix to use 'cargo clean' rather than 'rm' to remove cached builds.

### DIFF
--- a/frameworks/Rust/hyper/Cargo.toml
+++ b/frameworks/Rust/hyper/Cargo.toml
@@ -4,5 +4,4 @@ version = "0.1.0"
 authors = ["Steve Klabnik <steve@steveklabnik.com>"]
 
 [dependencies]
-
-hyper="0.5.2"
+hyper="0.7.2"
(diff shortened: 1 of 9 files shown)


steveklabnik:
Hm, I don't know why unrelated tests are failing here. Any ideas?

ssmith-techempower:
Changes to files in `toolset` trigger full test runs. Normally only the changed frameworks are run, so broken frameworks that aren't run are flagged as passing.

steveklabnik:
Gotcha. Thank you!

PR #1941, opened 2016-02-25
Update Finagle benchmark

vkostyukov:
Cleanup and updated Finagle benchmark so it now includes JSON serialization (`/json`) and plain text endpoints (`/plaintext`).


Rydgel:
@vkostyukov Aren't you supposed to serialize the json at each hit? You are serializing once and outputing it like static binary data.

> 1. For each request, an object mapping the key message to Hello, World! must be instantiated.
> 2. […]
> 3. A JSON serializer must be used to convert the object to JSON.

vkostyukov:
Oh you're right, @Rydgel! I guess I just misunderstood that while porting the implementation from a different benchmark. Fixed it now. Thanks!

Rydgel:
Lookin' good now!

ssmith-techempower:
Looks good to me, merging in.

PR #1946, opened 2016-03-01
Move Json writer out side of the action method

AmirKarimi:
The writer would be instantiated on every request when it's inside the action.


ssmith-techempower:
Looks good to me, merging in.

PR #1948, opened 2016-03-01
Rust updates

steveklabnik:
Various updates for the Rust programs; they previously had several warnings, and with these changes, now all show `PASS`.


ssmith-techempower:
Looks good to me, merging in.

PR #1952, opened 2016-03-06
akka-http version up and fix code

sxend:
upgrade akka-http 2.4.2
fix http server code


ssmith-techempower:
This looks good to me, but since a part of this PR is removing the `charset` from the content type headers where it isn't necessary I wanted to mention that `charset` isn't needed for the plaintext test either.

PR #1953, opened 2016-03-07
Use the same version of the MySQL connector 5.1.38

### LINKED ISSUE #1932: Every "Java" frameworks should use the same version of the MySQL connector
To be fair I think every Java/Scala frameworks should use the same version of the MySQL connector.
Latest version is 5.1.38, but some frameworks still use 5.1.16 version.

Mogztter:
Resolves #1932


msmith-techempower:
All the tests pass in Travis (except for two known issues); LGTM!

PR #1955, opened 2016-03-07
Disabled tests that use Redis

ssmith-techempower:
We've decided to remove the Redis tests (see previous discussion [here](https://groups.google.com/forum/#!msg/framework-benchmarks/NIOs_lzc-X8/GbSBRXt8OeEJ)). This PR disables the test configurations that currently use it. There may still be some I've missed.


GMLudo:
If your next setup doesn't have Redis server anymore, API-Hour won't start, because, to be easy to maintain, I've made an all-in-one server for all tests.
Is it necessary to disable Redis connection in the daemon ?

ssmith-techempower:
For the first round of changes we'll just be disabling the test configurations, not removing Redis itself. We're not currently planning on removing it to avoid accidentally breaking existing frameworks/tests. You can assume Redis server will still be installed/available at least for the next round (tests using Redis will be disabled, though).

PR #1958, opened 2016-03-09
updated Luminus benchmark to latest Luminus version

yogthos:
This change updates the benchmark to use the latest template. In addition, the benchmark provides an optimized version of the routes that should demonstrate better performance characteristics.

The previous version of the benchmark used default middleware functions, such as sessions, for the benchmark routes. Since this functionality is not necessary, it's been removed from the benchmark to illustrate best case performance.


k-r-g:
Nice! I'm happy to see this use the latest template and get optimized routing.

A couple notes:
1) Were your changes to README.md intentional? No big deal either way.
2) setup.sh has "fw_depends java7 resin leiningen" but should have "java" instead of "java7". (java just means Java 1.8 now, for all the frameworks)

yogthos:
Perfect, I fixed the `setup.sh` and I cleaned up the readme notes as those were out of date as well.

yogthos:
@k-r-g I seem to be having a problem setting the database connection:

    Server luminus: Caused by: com.zaxxer.hikari.pool.HikariPool$PoolInitializationException: Exception during pool initialization: HikariPool-0 - Connection is not available, request timed out after 30005ms.

As far as I can tell though, the script is doing the right thing, and I'm not able to replicate the issue locally. Is there anything that jumps out at you here.

The database URL is defined [here](https://github.com/yogthos/FrameworkBenchmarks/blob/master/frameworks/Clojure/luminus/hello/env/prod/resources/config.edn#L3)

It's being updated [here](https://github.com/yogthos/FrameworkBenchmarks/blob/master/frameworks/Clojure/luminus/setup.sh#L6), and when I try that locally it works fine

    sed -i 's|127.0.0.1|'"${DBHOST}"'|g' hello/env/prod/resources/config.edn

So, I'm not quite sure why it fails on Travis, I could try hardcoding the db url in the code, but figured this way would be cleaner.

ssmith-techempower:
I've been looking into this PR trying to see what the problem is. Unfortunately, it fails in my local instance for a separate reason. The `HikariPool` seems to start up successfully (everything seems to be starting up successfully), but the connection is refused when the endpoints are hit. Is there anything about your environment that might be different?

Also, just as a sanity check, there's a `config.edn` for both `prod` and `dev`. The `prod` one is the one used when the app starts up, correct? The `sed` command to update the DB host works as far as my testing shows.

yogthos:
Hmm, the server starts up on port 3000, so you should be able to hit the endpoints there. In terms of the config, the `dev` one will be used when the app runs in development mode, but the `prod` one is used when it starts up from the compiled jar.

ssmith-techempower:
Alright, I've gotten it to spit out some exceptions and it looks like it's failing to authenticate with the database. Are you setting the username/password when you connect to Postgres?

yogthos:
Ah that would do it, the connection url doesn't have the credentials in it:

    "jdbc:postgresql://127.0.0.1:5432/hello_world?jdbcCompliantTruncation=false&elideSetAutoCommits=true&useLocalSessionState=true&cachePrepStmts=true&cacheCallableStmts=true&alwaysSendSetIsolation=false&prepStmtCacheSize=4096&cacheServerConfiguration=true&prepStmtCacheSqlLimit=2048&zeroDateTimeBehavior=convertToNull&traceProtocol=false&useUnbufferedInput=false&useReadAheadInput=false&maintainTimeStats=false&useServerPrepStmts&cacheRSMetadata=true"

I'll update that and thing should be back to noraml

msmith-techempower:
Seems to be errors thrown when trying to connect to the database.

yogthos:
@msmith-techempower I think I tracked down the problems, hopefully travis will like the build this time around :)

valyala:
FYI, I think [this travis test](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/116159320) may be canceled, since it is overridden by [the latest test](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/116160018). This should speed up [pending tests' execution](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/pull_requests).

msmith-techempower:
`Server luminus: org.postgresql.util.PSQLException: ERROR: column "randomNumber" of relation "world" does not exist`

I believe that PostgreSQL `toLower`s all of its schema/table names.

yogthos:
@msmith-techempower yeah that's exactly what happened, the key got renamed after roundtripping. Should be fixed in the latest commit

msmith-techempower:
[Same error](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/116898391#L2442):

`Server luminus: org.postgresql.util.PSQLException: ERROR: column "randomNumber" of relation "world" does not exist`

yogthos:
Hmm interesting, I managed to reproduce this locally with lower case as well

`ERROR: column "randomnumber" of relation "world" does not exist Position: 18`

msmith-techempower:
It has been a while for me, but maybe PostgreSQL is case-sensitive? That is, it should be `World` and not `world`. Similarly, it should be `randomNumber` and not `randomnumber`?

yogthos:
Ah, it does appear that Postgres is case sensitive. My local table is:

```sql
CREATE TABLE world
(
  id numeric,
  "randomNumber" numeric
)
```

so I get the error for "randomnumber", however it looks like the table in tests has all lowercase for "randomnumber", hence why tests pass for me locally, but not on travis.

msmith-techempower:
I believe the schema name should also be capitalized. See [the PostgreSQL setup script](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/config/create-postgres.sql).

yogthos:
ah ok, let me take a quick look, I definitely misnamed something in my schema

yogthos:
ok so running your migration script will create tables called "World" and "world" the "World" table will have the "randomNumber" column, while the "world" table will have "randomnumber". Since I was using "world" as the table name that's where the problem was coming from.

yogthos:
actually, looking at the tables, it looks like columns end up with lower case regardless, when I run

```
DROP TABLE IF EXISTS "World";
CREATE TABLE  "World" (
  id integer NOT NULL,
  randomNumber integer NOT NULL default 0,
  PRIMARY KEY  (id)
);
```

I see this in my db:

```
CREATE TABLE "World"
(
  id integer NOT NULL,
  randomnumber integer NOT NULL DEFAULT 0,
  CONSTRAINT "World_pkey" PRIMARY KEY (id )
)
```

yogthos:
So, this is what I've got right now:

```sql
-- :name get-worlds :? :*
SELECT * FROM "World"
WHERE id IN (:v*:ids)

--:name get-all-fortunes :? :*
-- select all records from the fortune table
SELECT * FROM "Fortune"

-- :name update-world! :! :1
-- update an existing world record
UPDATE "World"
SET "randomnumber" = :randomNumber
WHERE id = :id
```

It appears to work against my local db that's initialized with your script. Hopefully, this should work same on travis now.

msmith-techempower:
Hoping intensifies!

yogthos:
I've definitely a learned things about Postgres schemas here :)

ssmith-techempower:
Updates is working on my local instance with the latest. I'm feeling hopeful for this latest Travis run.

yogthos:
this is exciting! :)

msmith-techempower:

yogthos:
:+1: and thanks for the help :)

PR #1959, opened 2016-03-09
Remove keepalive from PHP frameworks using fastcgi

gjuric:
A few years ago I have added a keepalive option to nginx configuration for PHP frameworks and since then almost all of them started getting a lot of failed request.

I finally decided to fix it, and judging by the tests I did on the Symfony2 framework it looks like it works now.

### DIFF
--- a/frameworks/PHP/clancats/deploy/nginx.conf
+++ b/frameworks/PHP/clancats/deploy/nginx.conf
@@ -17,7 +17,6 @@ http {
 
 	upstream fastcgi_backend {
 		server 127.0.0.1:9001;
-		keepalive 32;
 	}
 
 	server {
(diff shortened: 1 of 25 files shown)


methane:
Can't PHP use unix domain socket?
When disabling keepalive, unix socket is very faster than TCP.

gjuric:
Yes it can, but that needs some more work and testing. 

I propose that we leave that for another pull request and merge this anyway because it affects drastically the benchmark results (a lot of requests were failing before this).

msmith-techempower:
Looks like most of the PHP frameworks are passing in Travis (and I'm not going to put the onus of fixing the others on you) - LGTM!

PR #1962, opened 2016-03-11
Add dependency to java8 to startup script

nkasvosve:
The missing dependency on java8 is why this framework failed to start in round 12

### DIFF
--- a/frameworks/Java/beyondj/setup.sh
+++ b/frameworks/Java/beyondj/setup.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-fw_depends java8
+fw_depends java
 
 WORKING_DIR=beyondj-launcher/deploy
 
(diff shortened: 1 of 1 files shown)


k-r-g:
@nkasvosve Try making it "fw_depends java" instead of "fw_depends java8". We've made it so all Java is now Java 8, and there is only a "java" dependency to define.

nkasvosve:
Hmm, the round 12 run for the framework failed because Java 7 was on the system, and not the required Java 8 ...
 ________________________________
 

    On Friday, March 11, 2016 4:58 PM, Keith Gustafson <notifications@github.com> wrote:
 

 @nkasvosve Try making it "fw_depends java" instead of "fw_depends java8". We've made it so all Java is now Java 8, and there is only a "java" dependency to define.—

msmith-techempower:
All tests passed in Travis; LGTM!

PR #1963, opened 2016-03-11
bee fix

astaxie:
works on beego latest version


ssmith-techempower:
Looks good to me, merging in.

PR #1968, opened 2016-03-14
fasthttp: switch from html/template to github.com/valyala/quicktemplate

valyala:
This should improve `/fortune` performance.


valyala:
@msmith-techempower , could you cancel [this travis test](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/116303684), since it became stale after [this test](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/116305574) has been queued after the latest PR update.

valyala:
It looks like unrelated tests [are failing](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/116305574) again, while all tests inside `frameworks/Go` pass (except revel* tests, which are broken due to defunct dependency on code.google.com, which is unrelated to this PR).

methane:
LGTM.

One thing I'm concerned is we can't compare go-raw and fasthttp.
Someone should update go-raw to use same template.

valyala:
> One thing I'm concerned is we can't compare go-raw and fasthttp.
Someone should update go-raw to use same template.

`go-raw` is based on standard go packages - [net/http for server](https://golang.org/pkg/net/http/) and [html/template for templates](https://golang.org/pkg/html/template/). So I don't think it should be switched to [quicktemplate](https://github.com/valyala/quicktemplate).

methane:
quicktemplate is not a part of fasthttp, is it?
What people think when they see benchmark result of fasthttp and go-raw?

I think `raw` means "Not using framework.  Use HTTP server directly.", not "Use only standard library."
When comparing `go-raw` and `fasthttp[-raw]`, it should shows performance difference between http servers.

methane:
Uh, I'm sorry.  I had not know about **you** are the author of fasthttp **and** quicktemplate.
I agree that "Standard library stack" vs "@valyala's stack" is good comparison.

Additionally, `-raw` suffix is used for `not using ORM.`
go-std is better name for go-raw, maybe.

valyala:
> go-std is better name for go-raw, maybe.

Agreed. Will prepare the corresponding PR soon...

valyala:
See the PR #1977 , which renames `go-raw*` to `go-std*`

valyala:
@msmith-techempower , could you cancel [this travis test](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/116585164), since I updated this PR and [new test](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/116696106) has been queued?

msmith-techempower:
Done

bhauer:
@methane @valyala This comment thread identifies some potential inconsistency in the terminology we have used and I'd like to discuss the following with you.  This may make sense as a separate issue.

@methane is correct that the "-raw" suffix usually has been applied to test implementations that do not include an ORM for database connectivity.

While I have tried to use the word "plain" to describe an implementation that uses the platform's standard libraries, I would not be surprised if I and others have sometimes used the word "raw" in this context as well.  I expect I've said the words "raw PHP" for example, just because it rolls off the tongue more easily than "plain PHP."

But it has been my intent that at least verbally and in written communication, the plain implementations be referred to as, for example:

* Plain PHP
* Plain Servlet
* Plain Go (or just "Go")

(Amusing aside: we have a framework included that is named "Plain.")

I am agreeable with the adjective "Standard" you have used here in place of plain.  Does anyone else have an opinion on this?  Again, this may be better raised as a separate issue.

One final point is that these plain/standard implementations are *most commonly* implemented with the platform's standard libraries.  However, where the platform is missing a component, the guidance would be to use the most common (or a very common) library for the component in question.  For example, on the Plain Servlet implementation [we use Jackson as the JSON serializer](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/servlet/src/main/java/hello/Common.java#L5) because Jackson is the de facto standard JSON serializer on Java.

ssmith-techempower:
All tests pass for me locally as well as on Travis, merging in.

PR #1964, opened 2016-03-12
go/revel: Fix revel installation

methane:

PR #1969, opened 2016-03-15
Switch from go1.5 to go1.6

valyala:
Go 1.6 is out a month ago, so it would be great to run Go tests with it.

### DIFF
--- a/toolset/setup/linux/languages/go.sh
+++ b/toolset/setup/linux/languages/go.sh
@@ -5,7 +5,7 @@ RETCODE=$(fw_exists ${IROOT}/go.installed)
   source $IROOT/go.installed
   return 0; }
 
-VERSION=1.5.2
+VERSION=1.6
 GOROOT=$IROOT/go
 
 fw_get -O https://storage.googleapis.com/golang/go$VERSION.linux-amd64.tar.gz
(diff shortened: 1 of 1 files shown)


kostya-sh:
A while ago I ran go TFB benchmarks using Go 1.5 and 1.6. One thing that I found is that the garbage collector in 1.6 is more accurate at keeping heap size based on `GOGC` value. GC in 1.5 generally tends to overcommit memory. As result the application compiled with 1.6 uses less memory but performs worse in throughput based benchmark like TFB.

I suggest to raise GOGC to 1000 (from 800) to account for more accurate GC algorithm in 1.6:

    echo "export GOGC=1000" >> $IROOT/go.installed

methane:
+1 for @kostya-sh

valyala:
@kostya-sh , done.

valyala:
It looks like Go/revel* [failed to build with go 1.6](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/116053178), since it depends on deprecated code from code.google.com, which has been disabled in go1.6:

> package code.google.com/p/go.net/websocket: unable to detect version control system for code.google.com/ path

@msmith-techempower , let's leave revel tests broken until @robfig removes deprecated code.google.com dependency from [github.com/robfig/revel/](https://github.com/robfig/revel).

methane:
@valyala I've noticed it and fixed in #1964.  Please ignore it.

methane:
Uh...
I haven't noticed about revel-jet and revel-qbs.
I'll fix it.

msmith-techempower:
I am content with this and Travis tests mostly pass; LGTM!

PR #1972, opened 2016-03-15
ULib: substitute for #1888

stefanocasazza:


msmith-techempower:
ULib passes all Travis tests with these changes; LGTM!

PR #1977, opened 2016-03-17
Go: renamed go-raw* to go-std* according to the discussion at #1968

valyala:


msmith-techempower:
This is just a rename; LGTM!

PR #1978, opened 2016-03-17
updating colossus to the latest version

DanSimon:
Upgrading to the latest colossus release, switching to json4s, cleaning up the benchmark code.


ssmith-techempower:
Travis errors are unrelated, colossus passes in it and my local install. Looks good to me, merging in.

PR #1981, opened 2016-03-18
fix for the query test

yogthos:
I noticed that I was doing a range select with random ids, but that can result in duplicate ids and returns less than the required number of results. Also removed unnecessary rename-keys called from the fortunes result.


msmith-techempower:
This did not get a Travis run... for reasons unknown.

ssmith-techempower:
Don't know why this wasn't picked up by Travis, but everything passes on my local instance, so I'm merging it in.

msmith-techempower:
Doesn't seem to matter, the merge into master kicked off a Travis run that tested it and it passed... so GWGTM (glad we got this merged).

yogthos:
excellent

Issue #1947, opened 2016-03-01
"SET synchronous_commit TO OFF" should not be allowed (fasthttp-postgresql)


kostya-sh:
fasthttp-postgresql came first in Data Updates benchmark in Round 12. When browsing its code I found the following snippet:

		// Disable synchronous commit for the current db connection
		// as a performance optimization.
		// See http://www.postgresql.org/docs/current/static/runtime-config-wal.html
		// for details.
		if _, err := conn.Exec("SET synchronous_commit TO OFF"); err != nil {
			log.Fatalf("Error when disabling synchronous commit")
		}

I am not sure how much this affected the benchmark results however I beleive such code should not be allowed. As result of disabling synchronous commits fasthttp-postgresql behaves differently from other benchmarks, in particular data loss is possible in some cases.

zloster:
About transactions and data loss: the rules for [`Database updates` test](http://frameworkbenchmarks.readthedocs.org/en/latest/Project-Information/Framework-Tests/#database-updates) doesn't _mandate_ the use of `transactions`. Check **Rule 10**.
IMO this "tweak" for the transactions doesn't break the rules in general - it gives `fasthttp-postgresql` advantage over the other frameworks that are using PostgreSQL and not using this tweak.

However I've got curious about how MySQL and PostgreSQL are configured. And here it is what I've found (it is _always_ possible that I've missed something):

1. [MySQL config](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/config/my.cnf#L67) enables [_defacto_ asynchronous transaction log write to the disk](https://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_flush_log_at_trx_commit). With the current setting it **is** possible to loose data from latest committed transactions before crash.
2. [PostgreSQL config](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/config/postgresql.conf#L166-L170) seems to use the _production safe_ [defaults](http://www.postgresql.org/docs/9.4/static/runtime-config-wal.html). With the current setting It is **not** possible to loose data from the latest committed transactions before crash.

So `fasthttp-postgresql` just sets PostgreSQL and MySQL behaviour (regarding the transaction log flushing) on equal ground.

[...]

PR #1982, opened 2016-03-18
fasthttp-postgresql: cleanup: removed 'synchronous_commit=off' from db setup

valyala:
This trick became obsolete after a48ab7fa73e46de6294553d1d81566789d682415


msmith-techempower:
This SHOULD be good to merge in because #1947 is done, right?

valyala:
> This SHOULD be good to merge in because #1947 is done, right?

@msmith-techempower , exactly!

msmith-techempower:
All tests have passed; LGTM!

PR #1984, opened 2016-03-18
"Fixed" C# frameworks

zapov:
Install XSP through apt-get

Tooling fails if "default" targets "Windows"
Renamed config names as a workaround

### DIFF
--- a/frameworks/CSharp/aspnet/benchmark_config.json
+++ b/frameworks/CSharp/aspnet/benchmark_config.json
@@ -1,7 +1,7 @@
 {
   "framework": "aspnet",
   "tests": [{
-    "default": {
+    "windows": {
       "setup_file": "setup_iis",
       "json_url": "/json/default",
       "plaintext_url": "/plaintext",
@@ -203,7 +203,7 @@
       "notes": "",
       "versus": ""
     },
-    "mono": {
+    "default": {
       "setup_file": "setup_nginx",
       "json_url": "/json/default",
       "plaintext_url": "/plaintext",
--- a/toolset/setup/linux/languages/xsp.sh
+++ b/toolset/setup/linux/languages/xsp.sh
@@ -7,18 +7,6 @@ RETCODE=$(fw_exists ${IROOT}/xsp.installed)
   source $IROOT/xsp.installed
   return 0; }
 
-# get
-git clone git://github.com/mono/xsp
-cd xsp
-git checkout e272a2c006211b6b03be2ef5bbb9e3f8fefd0768
-
-# build
-./autogen.sh --prefix=$MONO_HOME --disable-docs
-make
-sudo make install
-
-# cleanup
-cd ..
-rm -rf xsp
+sudo apt-get install -y mono-fastcgi-server
 
 echo "" > $IROOT/xsp.installed
--- a/frameworks/CSharp/nancy/benchmark_config.json
+++ b/frameworks/CSharp/nancy/benchmark_config.json
@@ -1,7 +1,7 @@
 {
   "framework": "nancy",
   "tests": [{
-    "default": {
+    "windows": {
       "setup_file": "setup_iis",
       "plaintext_url":"/plaintext",
       "json_url": "/json",
@@ -22,7 +22,7 @@
       "notes": "",
       "versus": ""
     },
-    "mono": {
+    "default": {
       "setup_file": "setup_nginx",
       "plaintext_url":"/plaintext",
       "json_url": "/json",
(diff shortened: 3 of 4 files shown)


frameworks/CSharp/aspnet/benchmark_config.json
```diff
@@ -203,7 +203,7 @@
       "notes": "",
       "versus": ""
     },
-    "mono": {
+    "default": {
```
msmith-techempower: These changes should not be necessary. We already have the "os" key to denote on which operating system the test can run. The fact that the default test does not run on linux is of no concern.

toolset/setup/linux/languages/xsp.sh
```diff
@@ -7,18 +7,6 @@ RETCODE=$(fw_exists ${IROOT}/xsp.installed)
...
-
-# build
-./autogen.sh --prefix=$MONO_HOME --disable-docs
-make
-sudo make install
-
-# cleanup
-cd ..
-rm -rf xsp
+sudo apt-get install -y mono-fastcgi-server
```
msmith-techempower: I was under the impression that this was installed via [this line](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/languages/mono.sh#L20).

frameworks/CSharp/aspnet/benchmark_config.json
zapov: Without those changes running test in vagrant fails early

toolset/setup/linux/languages/xsp.sh
zapov: Yeah, me to. but it's not xsp, but fastcgi they are using

frameworks/CSharp/aspnet/benchmark_config.json
msmith-techempower: Why? You should be running `toolset/run-tests.py --mode verify --test aspnet-mono`, for example.

zapov: I tried running aspnet, not aspnet-mono, so that might be the issue. Still even that should not fail (in my opinion)

toolset/setup/linux/languages/xsp.sh
zapov: Basically, I was wrong. It's still needed.

msmith-techempower: Then this file should be removed entirely `mono-fastcgi-server` should be added to [this file](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/languages/mono.sh) for accuracy. This change is the worst case scenario of requiring tests to depend on `xsp` when they really need `mono-fastcgi-server`.

frameworks/CSharp/aspnet/benchmark_config.json
msmith-techempower: If there is a bug (e.g. getting a stack trace instead of a nicely printed error when trying to run a windows test on linux or vise versa), then that should be an issue that we resolve in the toolset. The naming convention from before your changes is, however, correct.

toolset/setup/linux/languages/xsp.sh
zapov: I don't currently have access to linux/mac computer so it's unlikely I'll be able to change anything.

I suggest you just pick the xsp.sh change since xsp stands for multiple things and not all frameworks need it (thats why it's ok to be a separate dependency.. .although I have no problems if you install it with vanilla Mono)

msmith-techempower: We can cherry pick the idea ourselves to make it easier, but I want to better understand this issue - is `xsp` synonymous with `mono-fastcgi-server`, or is the latter just something on which many `mono` tests depend? If so, what is `xsp` and does it need to be its own dependency?

zapov: This might not be 100% correct...

XSP is mono web server similar to ASP.NET
standalone XSP server is installed when you do a mono-complete installation

but XSP can run over Apache or Nginx
In that case you need those extract mono-fastcgi-server dependencies

so xsp.sh is poorly named, it should be named mono-fastcgi and only some of the C# frameworks depend on that.

msmith-techempower: Is there a compelling argument for **not** including `mono-fastcgi-server` in `mono.sh`?

zapov: I think it's fine to include it there

msmith-techempower:
Closing as this has been addressed by a direct [commit](https://github.com/TechEmpower/FrameworkBenchmarks/commit/ef82fbb42e54fe352e193f4005853ad36463083c)

zapov:
But you should probably remove xsp dependency from those frameworks

msmith-techempower:
That's a good idea >_<

PR #1985, opened 2016-03-19
Resolves #1949: Add command to link VM code and local code, which enables mirroring

### LINKED ISSUE #1949: Vagrant box does not mount the host FrameworkBenchmark directory
I was under the assumption that I can just fire vagrant up and have it test my host's current working directory of FrameworkBenchmark. Turns out that is not the case and in order for me to test out any changes I make, I have to commit and push (possible incorrect code) to a repo, then pull in the VM. 

It was said that this is the case because Vagant does not work in Windows, which is fine, but why are we making it harder for osx and linux devs because Windows and Vagrant won't play well together?

I suggest making it work for the OSes that work well with vagrant. We can re-evaluate Windows as well to see if the situation has solved itself (or default to something like samba for windows)

cjnething:
Mirroring between VM code and local code was not working; need to add this line to the script to enable mirroring

### DIFF
--- a/deployment/vagrant-common/bootstrap.sh
+++ b/deployment/vagrant-common/bootstrap.sh
@@ -123,6 +123,9 @@ if [ ! -e "~/.firstboot" ]; then
     git config --global core.autocrlf input
     git clone -b ${GH_BRANCH} https://github.com/${GH_REPO}.git $FWROOT
     source ~/FrameworkBenchmarks/toolset/setup/linux/prerequisites.sh
+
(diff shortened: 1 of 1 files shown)


msmith-techempower:
Since this breaks the line-endings on Windows, is there a way to have it only call that line if the host OS is NOT windows?

cjnething:
Adding this line of code will break line-endings for Windows users. It's not possible to test for Windows inside the virtual machine, so instead of adding this as a line of code, we will add this as a line in the documentation.

PR #1986, opened 2016-03-19
fix revel-jet and revel-qbs

methane:


msmith-techempower:
Tests look good in Travis; LGTM!

PR #1988, opened 2016-03-21
Add new benchmark: Go + PostgreSQL

kostya-sh:
The code mostly follows go-std-mysql.

prefork and interpolate tests from go-std-mysql were not ported to PostgreSQL to
keep things simple.


msmith-techempower:

PR #1990, opened 2016-03-22
Add ant dependency for gemini test

cjnething:
Running the Gemini test gave an error because we were missing the ant dependency; added dependency to the prerequisites file


msmith-techempower:
Amazing that this has been broken for this long; good catch. LGTM!

Issue #1655, opened 2015-06-08
Logging improperly sending all to out.txt


msmith-techempower:
Seems that all output is sent to `out.txt` and nothing really is sent to `err.txt`

hamiltont:
[That one is on me](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/benchmark/framework_test.py#L254)

It was actually intentional - I found that having two separate files was not very useful for debugging purposes when something went wrong, and when things go right there are only 2-3 lines in these files so mixing them was not a big deal. Also there is a code maintenance and runtime resource benefit to having only one thread watching the process for IO instead of two, but that's a fairly minor concern. My main thought was just that it was annoyingly hard to rebuild the full picture of what happened by stitching err.txt and out..txt back together in the correct order

msmith-techempower:
I am going to repurpose this task - I think we should do the following: 

1. Remove `err.txt` entirely; it is unused now
2. Add a `verification.txt` that only contains the verification stage output
3. Add a `[test type].txt` for each test type that only contains the output for that everything beginning with "Benchmarking [test type]" until that test is done.

This assists with a few different things, but mainly it makes testing and benchmarking easier for me >_<

cjnething:
Resolved in #1991

PR #1991, opened 2016-03-22
Create file for verification stage output and get rid of error file

cjnething:
- Addresses tasks 1 and 2 from Issue 1655
- Create file "verification.txt" for all verification stage output to be logged into
- Get rid of "err.txt" file and send all previous information that would have logged to that file to "out.txt" (which already existed)

### DIFF
--- a/toolset/benchmark/benchmarker.py
+++ b/toolset/benchmark/benchmarker.py
@@ -522,8 +522,7 @@ class Benchmarker:
       os.makedirs(logDir)
     except Exception:
       pass
-    with open(os.path.join(logDir, 'out.txt'), 'w') as out, \
-         open(os.path.join(logDir, 'verification.txt'), 'w') as verification:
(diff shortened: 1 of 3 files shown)


msmith-techempower:
This looks good to me; will merge in after round 13.

msmith-techempower:
@cjnething This is a good start! I would like to break out `verification.txt` into several folder and files: `verification/json.txt`, `verification/db.txt`, etc with each only encompassing that particular verification.

Additionally, since we can, let's break out the benchmarking blocks into their own folder and files:
`benchmark/json.txt`, `benchmark/db.txt` etc. The contents of these files should be the data under the `Benchmarking [testname]` block (these look like "BENCHMARKING JSON... Complete").

cjnething:
Updated the PR based on your comments

msmith-techempower:
Just tested this and I love it; LGTM!

cjnething:
Resolves #1655

PR #1992, opened 2016-03-23
Optimized the Undertow-Jersey tests

cmsimike:
- Changed the SessionFactory to be a Singleton rather than RequestScoped. SessionFactories are thread safe by nature.
- Cleaned up the code dealing with moving request headers from Undertow to Jersey
- General import cleanup


ssmith-techempower:
Looks good to me, merging in.

PR #1993, opened 2016-03-23
Delete ant from prerequisites and move to its own script file

cjnething:
- Delete ant from the prerequisites file
- Create separate file for ant script (format based off the maven script)
- Depend on the ant script in Gemini's start script

### DIFF
--- a/toolset/setup/linux/systools/ant.sh
+++ b/toolset/setup/linux/systools/ant.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+RETCODE=$(fw_exists ${IROOT}/ant.installed)
+[ ! "$RETCODE" == 0 ] || { \
+source $IROOT/ant.installed
+return 0; }
+
+sudo apt-get update
+sudo apt-get -y --force-yes install ant
+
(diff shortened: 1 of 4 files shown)


toolset/setup/linux/systools/ant.sh
```diff
@@ -0,0 +1,21 @@
...
+return 0; }
+
+sudo apt-get update
+sudo apt-get -y --force-yes install ant
+
+if [ -e /usr/bin/ant ]
+then
+    sudo rm -f /usr/bin/ant
+fi
+sudo ln -s /usr/share/ant/bin/ant /usr/bin/ant
```
msmith-techempower: Wait, so `apt-get install ant` does **not** put `ant` onto the `PATH`?

msmith-techempower: If so, I think maybe get rid of `/usr/bin/ant` entirely and just replace this line with `echo "export PATH=/usr/share/ant/bin:\$PATH" > $IROOT/ant.installed`

msmith-techempower:
Looks good and fixes some setup for round 13; LGTM!

PR #1994, opened 2016-03-24
netty update

lneves:


ssmith-techempower:
Looks good to me, merging in.

PR #1999, opened 2016-03-26
Go echo v2

methane:
Echo v2 breaks backward compatibility.  Go/echo test is broken.


methane:
Test passed on methane#5

msmith-techempower:
How does this differ from #2002 ?

methane:
@msmith-techempower I've just ported to echo v2.
#2002 did more things:

* Switch to PostgreSQL
* Add prefork mode
* fasthttp backend and stdlib's net/http backend.

Please use #2002.

PR #2003, opened 2016-03-28
Merge revel-*

### LINKED ISSUE #1998: Revel should be one directory
Currently, there are three directories (revel, revel-jet, and revel-qbs), but these should be one dir with separate test/code as necessary.

My understanding is that `revel` (the framework) is the same for both, but `jet` is used as the ORM for `revel-jet` whereas standard database stuff is done in Golang for `revel`. The same is true for `qbs`

methane:
fixes #1998


methane:
Test for revel was passed.
https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/118896897

msmith-techempower:
Love it.

Question: should `revel` be broken up into `revel` and `revel-raw` with just json/plaintext in the former and db/query/fortunes/updates in the latter? My gut says yes.

methane:
I did it.

cjnething:
Same for this one! (Please reopen and target Round 14) Thanks!

methane:
#2089 overrides this

PR #2008, opened 2016-03-28
Resolves #2005: Lower WRK timeout value to fix benchmarks issue

### LINKED ISSUE #2005: WRK timeout value is breaking benchmarks
Currently, the scripts in `toolset/benchmark/framework_test.py` ([example](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/benchmark/framework_test.py#L48)) are breaking in wrk4; change the timeout to a more realistic number (maybe 8?).

cjnething:
Change default duration from 15 seconds to 8 seconds


methane:
LGTM

PR #2013, opened 2016-03-29
Resolves #1704: Remove the name attribute with default ec2

### LINKED ISSUE #1704: results.json contains a "name" attribute that is always "ec2"
Remove the "name" attribute from the results.json *or* make it meaningful.  Presently, it always is set to "ec2", regardless of the test environment, making it meaningless or confusing.

cjnething:
- Remove results.json name attribute (default is "ec2" and is never renamed)
- Remove README references to results/ec2/latest/...; file structure is now results/latest/...
- Resolves Issue #1704


msmith-techempower:
I do not see an update to where the timestamp folder. Specifically, there was a `results/ec2/latest` dir, and this update seems to address that; however, there was also a `results/ec2/[timestamp]` dir where the raw data goes.

cjnething:
On line 966 in benchmarker.py, we create a "result_directory" that used to be /results/ec2/ and is now just /results/, and the other folders (latest, timestamp, etc.) are joined to result_directory (specifically on line 256 of benchmarker.py, "os.path.join(self.result_directory, self.timestamp)")

msmith-techempower:
Got it; LGTM!

PR #2014, opened 2016-03-29
Change package.json files to show actual names for each framework

cjnething:
Current package.json files only show default name; change to show name of framework


msmith-techempower:
Ha, though I do not think this affects anything; LGTM!

PR #2015, opened 2016-03-30
Fixed maven installation

dreis2211:
Hey,

while waiting on my latest Travis build I recognized that the currently running builds are failing in almost every Java related build.

I guess the root-cause for this was introduced some days ago when the installation of Ant and Maven were changed to don't use symlinks anymore ( #1993 ). 

```
Setup spring: +++ mvn -version
Setup spring: /home/travis/build/TechEmpower/FrameworkBenchmarks/toolset/setup/linux/systools/maven.sh: line 14: mvn: command not found
```

Since the symlinks are not setup anymore, accessing mvn -version won't work directly. The file needs to be sourced beforehand. 

**Note:** Ant seems to work despite this, which I guess is related to a previous installation of Ant somewhere else. Nevertheless, I changed the order for Ant as well.

Cheers,
Christoph

### DIFF
--- a/toolset/setup/linux/systools/ant.sh
+++ b/toolset/setup/linux/systools/ant.sh
@@ -10,6 +10,6 @@ sudo apt-get -y --force-yes install ant
 
 echo "export PATH=/usr/share/ant/bin:\$PATH" > $IROOT/ant.installed
 
-ant -version
-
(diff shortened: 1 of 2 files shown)


msmith-techempower:
Good catch; LGTM!

PR #2016, opened 2016-03-30
Make minor changes to fix local test run of nodejs

cjnething:
- Deleted hiredis from the dependencies and
  altered the comments accordingly:
  Hiredis can no longer be installed properly; however,
  redis is also no longer being tested, which makes hiredis
  unnecessary (see PR #1955). If redis test are reinstated, the code will
  still work with the redis default JavaScript parser, so
  hiredis is not necessary but would be nice to get working
  again.
- Updated sequelize to a newer version
- Updated the "main" option in package.json and in source_code
  to the actual main file, "app.js"; "hello.js" does not exist
- Updated node version to the latest in the 0.x series (released
  in March 2016)


msmith-techempower:
Fixes NodeJS tests in Travis; LGTM!

PR #2019, opened 2016-04-05
Issue #1546: Update Node version to 5.10.0 which includes io.js

cjnething:
Beginning with Node 4.0, Node and io.js have merged together so there is no need to check performance between Node and io.js. Updated Node version to 5.10.0, which is the latest stable.

Resolves Issue #1546


jadbox:
Perhaps upgrade this to Node v6 for post-Round 13? I've noticed a solid improvement to performance in the version bump.

PR #2051, opened 2016-05-17
Revenj update bugfix for Round 13.

zapov:
Due to usage of random Revenj can fail on highly concurrent updates.
Avoid failure by sorting data before update.


msmith-techempower:
Bugfix looks good; LGTM!

PR #2059, opened 2016-06-14
Go 1.6.2

methane:


nbrady-techempower:
Tested locally against Go frameworks and passes travis on all Go frameworks. LGTM!

PR #2060, opened 2016-06-15
Simplify Revenj setup.

zapov:
Add DSL Platform as language dependency.
Move Java compilation to Maven plugin.
Add missing Fortunes implementation.


nbrady-techempower:
This is good to go. The failing on travis was a keyserver issue with ubuntu. Pulled and tested and all works.

PR #2061, opened 2016-06-15
Updated ffead-cpp version (resolved threadlocal issue in ffead-cpp) + Added support for ffead-cpp to run within apache server

sumeetchhetri:


nbrady-techempower:
@sumeetchhetri this passes local tests but it was never added to the `.travis.yml` file. However, instead of doing that, it might make more since to group this within `/ffead-cpp` and add it as a second test in the `benchmark_config.json` and create a `setup_apache2.sh`.

sumeetchhetri:
Done

nbrady-techempower:
closing this for #2162

PR #2064, opened 2016-06-15
update akka-http

dborovikov:


nbrady-techempower:
Passed all local tests, merging for round-14!

PR #2067, opened 2016-06-15
Upgrade deps, tune perf, add undertow servlet and undertow webactors conf

circlespainter:


nbrady-techempower:
Passed all local tests, merging for round-14!

PR #2069, opened 2016-06-15
Update Node version to 5.10.0 which includes io.js

cjnething:
Originally #2019


nbrady-techempower:
@cjnething since this is targeting round-14 now did you want to give node 6.0.0 a try? this is the next LTS and I don't believe it has any breaking changes from node > 4.

nbrady-techempower:
This passes on all local tests for all relevant js frameworks that are currently working. We'll look at bumping to node LTS 6.0.0 next. Thanks @cjnething! :)

PR #2012, opened 2016-03-29
Fix http headers spring

dreis2211:
Hey,

when working on a different merge request I noticed that the Spring suite doesn't include the **Server** header and sends the encoding along with the **Content-Type** header. Both result in a warning in the builds for all requests.

WARN for http://127.0.0.1:8080/json
`Required response header missing: Server`

WARN for http://127.0.0.1:8080/json
`Content encoding found in "application/json;charset=UTF-8" where "application/json" is acceptable.`

I needed to upgrade to Spring-Boot 1.3.3 to set the server header in Undertow via `server.server-header`.

Cheers,
Christoph

PR #2071, opened 2016-06-15
Fix http headers spring

dreis2211:
Hey,

this is basically the duplicate of #2012 targeting the new branching model introduced for round 14 and beyond, but for completeness reasons here is the original content.....

....when working on a different merge request I noticed that the Spring suite doesn't include the **Server** header and sends the encoding along with the **Content-Type** header. Both result in a warning in the builds for all requests.

WARN for http://127.0.0.1:8080/json
`Required response header missing: Server`

WARN for http://127.0.0.1:8080/json
`Content encoding found in "application/json;charset=UTF-8" where "application/json" is acceptable.`

I needed to upgrade to Spring-Boot 1.3.5 to set the server header in Undertow via `server.server-header`.

Cheers,
Christoph

### DIFF
--- a/frameworks/Java/spring/pom.xml
+++ b/frameworks/Java/spring/pom.xml
@@ -8,7 +8,7 @@
     <parent>
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-starter-parent</artifactId>
-        <version>1.2.5.RELEASE</version>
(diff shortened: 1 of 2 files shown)


nbrady-techempower:
Travis agrees, merging! Thanks @dreis2211!

PR #2075, opened 2016-06-17
fix warning on query tests, json tests with different preforked process, elasticsearch code

stefanocasazza:
fix warning on query tests.
Also I added some json tests to see wich data result for different server preforked capacity in test environment.
Also I added elasticsearch code but not the test because I see that fail on travis cause I think of the method to test if the  elasticsearch server is running before to run the test.


nbrady-techempower:
Hi  @stefanocasazza, from now on when you do a pull request for a framework simply target the round-## branch. So for this, it should be targeting round-14, not master. Thanks!

stefanocasazza:
Hi,

I am not sure to understand, you wish that I close the actual pull request
and doing another targeting round-14 ?

Stefano

2016-06-17 17:32 GMT+02:00 Nate <notifications@github.com>:

nbrady-techempower:
@stefanocasazza Yeah, exactly. Sorry if I didn't make that clear on the first comment. From now on any framework updates should be targeting the a round-## branch instead of master.

PR #2076, opened 2016-06-17
fix warning on query tests, json tests with different preforked process, elasticsearch code

stefanocasazza:
fix [warning on query test](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/119842891)
Also I added some json tests to see wich data result for different server preforked capacity in test environment.
Also I added elasticsearch code but not the test because I see that fail on travis cause I think of the method to test if the elasticsearch server is running before to run the test.


stefanocasazza:
something goes wrong...

PR #2078, opened 2016-06-17
aspnetcore test automation

nathana1:
Adds plaintext and json tests for asp.net core with several variations (Windows/Linux, Mvc, WebListener/Kestrel).

cc:  @sajayantony @damianedwards @halter73


DamianEdwards:
This doesn't seem to include the actual code, rather it just clones from https://github.com/aspnet/benchmarks as part of the automation script. Don't we need to include the benchmark apps themselves (like all the frameworks here)?

frameworks/CSharp/aspnetcore/run-linux.sh
```diff
@@ -0,0 +1,20 @@
...
+fw_depends mono
+sudo apt-get install unzip libunwind8 -y
+
+if [ -d test ]; then
+  rm -rf test
+fi
+
+mkdir test
+cd test
+git clone https://github.com/aspnet/benchmarks.git
```
msmith-techempower: Like @DamianEdwards said, this should be the actual code that handles requests for the various benchmark endpoints and live in our repo.

nathana1:
I'll get something sent out on Monday.

nathana1:
hi @msmith-techempower, just curious if you have any info yet on the timing or the hardware for round 14.  For lots of cores we needed to make tweaks around processor groups and numa to utilize all of the cores so we'd like to make sure we're getting full utilization for your runs as well.  

Thanks.

msmith-techempower:
We are working on the hardware situation at the moment; the timing for round 14 is unknown at the moment.

We expect to run the suite on a few environments: Azure (with minimal cores) and hardware with 40 cores.

msmith-techempower:
This looks good to merge in to me, but there is some discussion about getting this in for round 13.

DamianEdwards:
@msmith-techempower thanks. Do let us know if there's anything we can do to help.

msmith-techempower:
@DamianEdwards We are going to get this merged in for round 13's preview run.

DamianEdwards:
Fantastic, looking forward to it!

madhugilla:
Would having ConfigureAwait as false help in increasing the perf ?

  return context.HttpContext.Response.Body.WriteAsync(_helloWorldPayload, 0, _helloWorldPayload.Length);

halter73:
@madhugilla ASP.NET Core doesn't create a SynchronizationContext , so there typically isn't any performance benefit to using `ConfigureAwait(false)`.

Additionally `return context.HttpContext.Response.Body.WriteAsync(_helloWorldPayload, 0, _helloWorldPayload.Length);` isn't being directly awaited. It's returning a `Task` from a method while `ConfigureAwait()` returns a `ConfiguredTaskAwaitable` which isn't a Task.

madhugilla:
@halter73  lesson learnt, thank you.

PR #2081, opened 2016-06-18
Fix Nim frameworks - Jester and Nawak

nbrady-techempower:
Nim's csources repo was pulling from current, which was causing breaking changes.
Also mongrel2's repo had relocated. Found and tied that down as well.


nbrady-techempower:
Also passing travis!

knewmanTE:
LGTM!

PR #2087, opened 2016-06-19
Spray Fixes

mfirry:
Spray fixed


nbrady-techempower:
@mfirry Out of curiosity, can the same fixes be applied to spray-es to fix that framework?

mfirry:
Haven't tried that. 
Actually I removed the `db` part of this one. It's going to take me a bit more to port that to newer versions of slick and such. 
I'll take a look at `spray-es` but, if I was you, I'll skip it and focus on `sql` tests.

nbrady-techempower:
@mfirry this doesn't seem to be passing. Getting the same errors in local as on travis for Spray.

mfirry:
Can't see the errors. Any hint?

mfirry:
Ok. Found the issue.

nbrady-techempower:
@mfirry that fixed it! thanks! will merge for round-14 as soon as travis agrees!

nbrady-techempower:
Great and travis agrees!

mfirry:
Awesome! You guys should seriously think about removing `spray-es` altogether.

msmith-techempower:
Why? What is `spray -es`; how does it compare to `spray` proper?

nbrady-techempower:
@mfirry why do you say that? i'm not familiar with the framework

mfirry:
It's not a different framework. It's [spray](http://spray.io) but uses [elasticsearch](https://www.elastic.co/products/elasticsearch) as a db.

msmith-techempower:
Oh... that's set up incorrectly, then.

Those two implementations should be consolidated to the same directory and the `-es` variant should be just another test permutation in the benchmark config file.

mfirry:
I didn't even know you guys were testing elasticsearch to begin with. I figured you only cared for sql stuff.

msmith-techempower:
I honestly don't recall supporting elasticsearch.

mfirry:
`git rm -r frameworks/Scala/spray-es/` then.

Keep doing this. I know it's a lot of work, but what you guys do is very cool and useful.

nbrady-techempower:
@mfirry we are considering creating a sort of graveyard repository post round-14 for frameworks/implementations that aren't currently working as an easy reference point for people that might want to take a crack at them. I don't know how @msmith-techempower feels, but if we can't get it working and included as another test implementation within /spray it will probably end up there.

mfirry:
Alright. For what it's worth, I'll fool around with other Scala frameworks (currently trying to upgrade the `unfiltered` benchmark). I am not going to work on elasticsearch though. 

Have a good one!

nbrady-techempower:
@mfirry Thank you! We will probably remove it then.

PR #2000, opened 2016-03-26
Merge go-std-* tests

methane:
fixes #1997

PR #2090, opened 2016-06-19
go-std: integrate go-std-* tests

methane:
overrides #2000


nbrady-techempower:
passes local and travis tests, merging for round-14!

PR #2094, opened 2016-06-20
ULib: fix warning on query tests, json tests with different number of…

stefanocasazza:
fix [warning on query test](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/119842891).
Also I added some json tests to see which data result for different server preforked capacity in test environment.
Also I added elasticsearch code but not the test because I see that failed on travis cause I think of the method to test if the elasticsearch server is running before to run the test.


nbrady-techempower:
@stefanocasazza Thank you for this! I'm currently testing on a local vagrant environment. I see that you added ElasticSearch to the database test sources. Were you planning on implementing that in this PR? Should this be left as a work in progress?

stefanocasazza:
Hi,

I do not  inserted it in the test to avoid criticism about the failure in
travis, but if you want I add it with another request...

Stefano

2016-06-23 20:41 GMT+02:00 Nate <notifications@github.com>:

nbrady-techempower:
@stefanocasazza ah, makes sense! merging for round-14!

PR #2099, opened 2016-06-20
update haywire setup to use a specific commit

knewmanTE:
Currently, haywire is pulling the latest from its git repo and the framework is failing to start when running tests. I've switched it to pull from the last commit before haywire was originally merged into TFB master.

@kellabyte the commit I'm pulling from is fairly old (June 4, 2015). It works, but I realize there may be a more recent commit that still works with the framework benchmark. If there's a newer commit you'd like to use instead, feel free to let us know or open a new pull request with the updated commit.


nbrady-techempower:
This got tripped up at the prereqs for some reason. This appears to be unrelated and happens "sometimes". Passes local tests, merging.

PR #2093, opened 2016-06-19
Lwan: Fix validation for "queries" request

lpereira:
For round 13, the validation for the "queries" requests were expecting a JSON response if the "queries" parameter was not passed.  Newer Lwan test fixes this by assuming it's 1 if that parameter is absent.

This should fix #2074.

PR #2102, opened 2016-06-20
Lwan: Fix validation for "queries" request

lpereira:
For round 13, the validation for the "queries" requests were expecting a JSON response if the "queries" parameter was not passed.  Newer Lwan test fixes this by assuming it's 1 if that parameter is absent.

This is the same as #2093, but targeting the round-14 branch as asked by @nbrady-techempower.


nbrady-techempower:
thanks @lpereira !

nbrady-techempower:
Passed relevant local and Travis tests. Merging for round-14!

PR #2104, opened 2016-06-20
netty update

nbrady-techempower:
Reference #2055


nbrady-techempower:
Passed local and travis! merging for round-14

PR #2049, opened 2016-05-14
Update ringojs and ringojs-convenient

grob:
* updated dependencies
* re-added tests in benchmark config of ringojs
* fixed an issue with fortunes test in ringojs-convenient
* removed explicit transaction usage in updates test of ringojs-convenient (ringo-sqlstore uses transactions by default)
* update to mysql-connector-java-5.1.39.jar

PR #2105, opened 2016-06-20
Ringojs convenient sqlstore 0.7

nbrady-techempower:
Reference #2049


nbrady-techempower:
Passing local and travis! merging for round-14

PR #2040, opened 2016-05-09
PHP optimitzations

joanhey:
Disabled opcache.validate_timestamps and removed php extra header.
Commented old apc config.

PR #2106, opened 2016-06-20
PHP optimizations

nbrady-techempower:
Reference #2040


nbrady-techempower:
This is passing local tests and travis for all php frameworks that weren't already failing

PR #2107, opened 2016-06-20
Fix instructions output

gmile:


nbrady-techempower:
Thanks !

nbrady-techempower:
Good to go! Thanks again!

Issue #2092, opened 2016-06-19
prerequisites.sh needs fw_exists loaded


gmile:
At the time `prerequisites.sh` is executed, `fw_exists` is not available (see the screenshot).

![image](https://cloud.githubusercontent.com/assets/207112/16179731/489a526c-3677-11e6-9f75-58fbe59982a0.png)

Above picture was taken when running vanilla `vagrant up`.

nbrady-techempower:
@gmile `prequisites.sh` doesn't rely on `fw_exists`. If it does exist, it will exit because the prerequisites would have already been installed.

nbrady-techempower:
oh I misunderstood, sorry. Good catch!

LadyMozzarella:
See #2111

Issue #2088, opened 2016-06-19
Ruby rbx required packages install that fails on `requirements_debian_libs_install clang libedit-dev`


Zloy:
```
Setup sinatra-puma-rbx: Checking requirements for ubuntu.
Setup sinatra-puma-rbx: Installing requirements for ubuntu.
Setup sinatra-puma-rbx: Updating system.........
Setup sinatra-puma-rbx: Installing required packages: clang, libedit-dev....
Setup sinatra-puma-rbx: Error running 'requirements_debian_libs_install clang libedit-dev',
Setup sinatra-puma-rbx: showing last 15 lines of /home/testrunner/.rvm/log/1466325320_rbx-2.4.1/package_install_clang_libedit-dev.log
Setup sinatra-puma-rbx: ++++ sudo -p '%p password required for '\''apt-get --no-install-recommends --yes install clang libedit-dev'\'': ' apt-get --no-install-recommends --yes install clang libedit-dev
Setup sinatra-puma-rbx: Reading package lists...
Setup sinatra-puma-rbx: Building dependency tree...
Setup sinatra-puma-rbx: Reading state information...
Setup sinatra-puma-rbx: Some packages could not be installed. This may mean that you have
Setup sinatra-puma-rbx: requested an impossible situation or if you are using the unstable
Setup sinatra-puma-rbx: distribution that some required packages have not yet been created
Setup sinatra-puma-rbx: or been moved out of Incoming.
Setup sinatra-puma-rbx: The following information may help to resolve the situation:
Setup sinatra-puma-rbx:
Setup sinatra-puma-rbx: The following packages have unmet dependencies:
Setup sinatra-puma-rbx:  clang : Depends: clang-3.4 (>= 3.4~rc3-1~) but it is not going to be installed
Setup sinatra-puma-rbx: E: Unable to correct problems, you have held broken packages.
Setup sinatra-puma-rbx: ++++ return 100
Setup sinatra-puma-rbx: ++++ return 100
Setup sinatra-puma-rbx: Requirements installation failed with status: 100.
Setup sinatra-puma-rbx: Setup has terminated
Setup sinatra-puma-rbx: Status: Poll: 100, Port 8080 bound: False, Time Left: 1:44:38.056451
Setup sinatra-puma-rbx: run_rbx_puma.sh process exited naturally with 100
ERROR: Problem starting sinatra-puma-rbx

```

Zloy:
[...]

PR #2111, opened 2016-06-21
rbx fix and added bash_functions to prereq

nbrady-techempower:
This pr forces `rbx-2.4` and `rbx-2.5` to use `clang-3.4`. `clang-3.5` is being installed by the prerequisites, though it looks as though only silicon uses it. Long term, all gcc/clang compilers should be moved out of prereqs and into systools. #2088 

This also adds the `bash_functions.sh` source to `prerequisites.sh` for use of `fw_exists`. #2092


knewmanTE:
LGTM!

frameworks/PHP/cygnite-php-framework/apps/views/bench/fortunes.view.php
```diff
@@ -1,6 +1,9 @@
 <!DOCTYPE html>
 <html>
-<head><title>Fortunes</title></head>
+<head>
+    <title>Fortunes</title>
+</head>
+
```
msmith-techempower: For future reference, we should all strive to not make changes outside of the PR's goal. I didn't see anything about `cygnite` in the original comment. Additionally, why add bytes when the purpose of the test is to show the performance of the platform? They add up over thousands of requests and are unnecessary.

toolset/setup/linux/languages/rbx-2.5.sh
```diff
@@ -12,6 +12,9 @@ RETCODE=$(fw_exists ${IROOT}/rbx-2.5.installed)
 # update this when it changes
 RBX_VERSION=2.5.2
 
+# This version of RBX needs clang 3.4
+sudo apt-get -yq --force-yes install clang-3.4
```
msmith-techempower: Ugh... more apt-get, more sudo.

@nbrady-techempower if you have time, can you quickly experiment to see if you can make `clang-3.4` into its own systool dependency and test out `fw_apt_to_iroot`? I would love to move away from as many of these apt-get as possible.

frameworks/PHP/cygnite-php-framework/apps/views/bench/fortunes.view.php
knewmanTE: @msmith-techempower Nate and I were talking about this yesterday. We think this was a discrepancy where a few minor commits had been merged into master and were in Nate's branch, but hadn't been pulled into round-14 yet. In the future we'll be sure to update round-14 with the latest from master to ensure that this is the case.

toolset/setup/linux/languages/rbx-2.5.sh
nbrady-techempower: @msmith-techempower Sure thing!

PR #2124, opened 2016-06-24
Fix configurations for broken PHP frameworks

### LINKED ISSUE #2110: PHP setup script not properly installing redis extension
Despite having  [this line](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/languages/php.sh#L47) in the php.sh install script, redis doesn't seem to be properly installed. `fw_depends php` outputs:
```
++ /home/knewman/FrameworkBenchmarks/installs/php-5.5.17/bin/pecl -q install -f redis
++ printf '\n'
warning: pecl/redis requires PHP (version >= 7.0.0, version <= 7.1.0, excluded versions: 6.0.0), installed version is 5.5.17
```
and afterwards, you can see that `installs/php-5.5.17/lib/php/extensions/no-debug-non-zts-20121212/` does not contain `redis.so`. If you comment out the `extension=redis.so` line in [/config/php.ini](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/config/php.ini#L870), micromvc passes.

This is breaking PHP/micromvc (see #2074) and might be related to why silex-orm (see #2109) and cakephp are breaking.

**Update**
It looks like, as of June 8th, 2016, the default pecl redis package requires php7:
https://pecl.php.net/package/redis

Specifying pecl/redis version 2.2.5 in `php.sh` appears to address most of these issues.

**Update 2**
See #2124 for fix.

knewmanTE:
A bunch of small  (but important!) PHP fixes that fix the following frameworks:

- cakephp
- clancats
- micromvc
- phalcon
- phalcon-micro
- pimf
- yaf 
- yii2

The root of a lot of these issues is that we are using PECL to download PHP extensions, but we weren't specifying the version number for a lot of them. Many of our PHP frameworks are built on PHP 5, but the most recent versions of extensions they rely on are only compatible with PHP 7.

I also refactored the yaf and phalcon install scripts, which were previously in `php.sh`, to be in their own file so we aren't installing phalcon every time we run a PHP test.

Resolves #2110


nbrady-techempower:
@knewmanTE the `sed` on `php.ini` did the trick??!

msmith-techempower:
That's some magic right there.

knewmanTE:
@nbrady-techempower yup! Just had to make sure I was updating the one in `PHP_HOME` and not in `FW_ROOT/config`. Setting the version of composer we were downloading also fixed yii2. So now clancats is our only breaking PHP framework. Looking into it now.

nbrady-techempower:
@knewmanTE I literally can't even right now.

PR #2125, opened 2016-06-24
Updated Fintrospect version

daviddenton:


nbrady-techempower:
passes travis and local tests; merging for round-14! thanks again @daviddenton

PR #2129, opened 2016-06-27
Fortune test in postgres causes pooling issues

nbrady-techempower:
Removed the fortune tests from the postgres test. It's failing due to connection pooling problems.


nbrady-techempower:
WIP: Looking into error on travis while working in local environment

PR #2130, opened 2016-06-28
Clearly define PHP versions

nbrady-techempower:
Makes sure that all PHP frameworks versions and "language" in the `benchmark_config` specify `PHP5` or `PHP7`. Renames `php.sh` in the toolset to `php5.sh` along with `php7.sh`.


knewmanTE:
@nbrady-techempower all PHP tests are passing, but would you rather hold off on merging this until you finish up the branch that updates all of the benchmark_configs?

PR #2136, opened 2016-07-02
Fixed fortune view : added content type and fortune url in benchmark config

sanjoydesk:
[v1.3.2] Fixed fortune view : added content type and fortune url in benchmark config


sanjoydesk:
Hi @nbrady-techempower 

I still don't understand why fortunes test is giving WARN. It was passing before. Can you please let me know if anything i am missing ?

Thanks!

nbrady-techempower:
@sanjoydesk fortunes warn when charset is not set to utf-8. It looks like you had it and then removed it.

sanjoydesk:
@nbrady-techempower yes i removed because build was failing when ever i was content type. Let me try again.

sanjoydesk:
@nbrady-techempower Can you see build failing if I add content type in the html ? Please let me know how to fix this warning ?

nbrady-techempower:
@sanjoydesk I will look at it today and see what I can find.

sanjoydesk:
@nbrady-techempower ok sure.. thanks!

nbrady-techempower:
@sanjoydesk So you don't want a meta tag that sets the content type as this isn't the real response header. This should be happening somewhere within your `function fortunesAction()` in your controller. Does cygnite have an additional parameter in it's `$this->render()` function to allow you to set a response?

See this discussion on meta tags vs response headers: http://stackoverflow.com/questions/9417024/response-header-vs-meta-tag

nbrady-techempower:
Simply adding:
```php
header('Content-Type: text/html; charset=utf-8');
```
above your `render` in your fortune and removing the meta tag fixes and passes.

sanjoydesk:
Thanks! @nbrady-techempower :+1:  It works!! Please merge the PR.

nbrady-techempower:
@sanjoydesk yes, this is good to go now. Now that the framework is in working order, please submit any future PRs against the round-14 branch. Thanks!

sanjoydesk:
Yes @nbrady-techempower .. Next version PR will be against round-14 branch. Thank you very much.. :)

PR #2141, opened 2016-07-05
Benchmark Config cleanup (WIP)

nbrady-techempower:
Cleaning up the meta data in benchmark configs


nbrady-techempower:
I believe @msmith-techempower has taken over work on this and has probably surpassed where I was here. Closing in favor of his work.

PR #2144, opened 2016-07-06
Revert phalcon to use php5

knewmanTE:
I originally thought Phalcon was running on PHP 7, but as it turns out, it was installing PHP 7 in `php7.sh` and then switching to PHP 5 in the separate `phalcon.sh` file.

I also noticed and fixed a couple minor discrepancies between the PHP 7 and PHP 5 install scripts. We weren't disabling yaf or phalcon, and the redis download was commented out (possibly because, the last time someone was working on `php7.sh`, there wasn't a redis version that was compatible with PHP 7).

### DIFF
--- a/frameworks/PHP/phalcon-micro/setup.sh
+++ b/frameworks/PHP/phalcon-micro/setup.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-fw_depends php7 phalcon nginx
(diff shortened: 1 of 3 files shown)


knewmanTE:
PHP tests are all passing in Travis. The errors are coming from other frameworks that are known to be currently breaking.

PR #2145, opened 2016-07-07
H2O: Add implementation

volyrique:
This implementation embeds the HTTP server in the application using libh2o instead of running the server separately. All test types have been implemented.


volyrique:
As far as I can tell, all test failures occur in unrelated frameworks.

On a related note, initially the build script was using wrk to generate the profiling data (to be consistent with the tests), but that failed on Travis, so I switched to curl.

frameworks/C/h2o_app/setup.sh
```diff
@@ -0,0 +1,57 @@
...
+		-DCMAKE_C_FLAGS="-march=native $1" "$TROOT"
+	make -j "$(nproc)"
+}
+
+run_curl()
+{
+	for ((i = 0; i < 10; i++)); do
+		curl "${H2O_APP_PROFILE_URL}/$1" > /dev/null 2>&1
+	done
+}
```
msmith-techempower: This looks like test stuff; please remove.

frameworks/C/h2o_app/setup.sh
```diff
@@ -0,0 +1,57 @@
...
+{
+	"$1/h2o_app" "$2" "$3" -a2 -f "${TROOT}/template/fortunes.mustache" -m2 \
+		-d "host=$DBHOST dbname=hello_world user=benchmarkdbuser password=benchmarkdbpass" &
+}
+
+generate_profile_data()
+{
+	run_h2o_app . -p$H2O_APP_PROFILE_PORT -t1
+	local -r H2O_APP_PROFILE_PID=$!
+	while ! curl ${H2O_APP_PROFILE_URL} > /dev/null 2>&1; do sleep 1; done
```
msmith-techempower: More testing? Remove.

toolset/setup/linux/systools/mustache-c.sh
```diff
@@ -0,0 +1,22 @@
...
+
+RETCODE=$(fw_exists "${IROOT}/mustache-c.installed")
+[ ! "$RETCODE" == 0 ] || { \
+  source "${IROOT}/mustache-c.installed"
+  return 0; }
+
+MUSTACHE_C_HOME="${IROOT}/mustache-c"
+BUILD_DIR="${MUSTACHE_C_HOME}_build"
+
+git clone 'https://github.com/x86-64/mustache-c.git' "$BUILD_DIR"
```
msmith-techempower: Please change this to use a release version instead of latest.

toolset/setup/linux/webservers/h2o.sh
```diff
@@ -0,0 +1,22 @@
...
+
+RETCODE=$(fw_exists "${IROOT}/h2o.installed")
+[ ! "$RETCODE" == 0 ] || { \
+  source "${IROOT}/h2o.installed"
+  return 0; }
+
+H2O_HOME="${IROOT}/h2o"
+BUILD_DIR="${H2O_HOME}_build"
+
+git clone 'https://github.com/h2o/h2o.git' "$BUILD_DIR"
```
msmith-techempower: Same here, please use a release not latest.

frameworks/C/h2o_app/setup.sh
volyrique: No, this is not testing - the application is built using profile-guided optimization, which requires generating the profiling data in the first place. I really don't see another way to do it with a statically compiled language such as C.

frameworks/C/h2o_app/setup.sh
volyrique: No, the idea is to wait for the application to start in a slightly more robust way than simply sleeping for a hardcoded amount of time.

toolset/setup/linux/webservers/h2o.sh
volyrique: There is a `git checkout` afterwards.

frameworks/C/h2o_app/setup.sh
msmith-techempower: Got it

toolset/setup/linux/systools/mustache-c.sh
volyrique: Done.

nbrady-techempower:
@volyrique Looks good and passes travis, will test in the morning on a local environment. One final nit though, it looks like the name of the framework is h2o, but you have h2o_app for the directory name and framework test name in the `benchmark_config`, can you please change this to h2o to be consistent with the rest of the frameworks.

volyrique:
Done.

nbrady-techempower:
@volyrique did you miss somewhere?
```bash
Setup h2o: /home/vagrant/FrameworkBenchmarks/installs/h2o/include/h2o.h: At top level:
Setup h2o: /home/vagrant/FrameworkBenchmarks/installs/h2o/include/h2o.h:106:5: warning: type of bit-field ‘proxy_should_drop’ is a GCC extension [-Wpedantic]
Setup h2o:      char proxy_should_drop : 1;
Setup h2o:      ^
Setup h2o: /home/vagrant/FrameworkBenchmarks/installs/h2o/include/h2o.h:107:5: warning: type of bit-field ‘is_init_header_special’ is a GCC extension [-Wpedantic]
Setup h2o:      char is_init_header_special : 1;
Setup h2o:      ^
Setup h2o: /home/vagrant/FrameworkBenchmarks/installs/h2o/include/h2o.h:108:5: warning: type of bit-field ‘http2_should_reject’ is a GCC extension [-Wpedantic]
Setup h2o:      char http2_should_reject : 1;
Setup h2o:      ^
Setup h2o: /home/vagrant/FrameworkBenchmarks/installs/h2o/include/h2o.h:109:5: warning: type of bit-field ‘copy_for_push_request’ is a GCC extension [-Wpedantic]
Setup h2o:      char copy_for_push_request : 1;
Setup h2o:      ^
Setup h2o: /home/vagrant/FrameworkBenchmarks/frameworks/C/h2o/src/utility.c:26:27: fatal error: yajl/yajl_gen.h: No such file or directory
Setup h2o:  #include <yajl/yajl_gen.h>
Setup h2o:                            ^
Setup h2o: compilation terminated.
Setup h2o: make[2]: *** [CMakeFiles/h2o_app.dir/src/fortune.c.o] Error 1
Setup h2o: make[2]: *** Waiting for unfinished jobs....
h2o: setup.sh and framework processes have terminated
Setup h2o: make[2]: *** [CMakeFiles/h2o_app.dir/src/utility.c.o] Error 1
Setup h2o: make[1]: *** [CMakeFiles/h2o_app.dir/all] Error 2
Setup h2o: make: *** [all] Error 2
```

volyrique:
Is this error from Travis? I can't see it. Otherwise, I added `libyajl-dev` to `toolset/setup/linux/prerequisites.sh` (and it's still there in the latest revision), so perhaps `apt-get` failed to install it for some reason?

knewmanTE:
@volyrique @nbrady-techempower it looks like that last instance of "h2o_app" is from CMakeLists, which I'm assuming is in reference to `${IROOT}/h2o_app` where the compiled app lives (to distinguish it from `${IROOT}/h2o` where the framework boilerplate lives).

Tested locally and this LGTM!

volyrique:
Yes, I kept the `h2o_app` name for the executable on purpose - `h2o` is the name of the standalone server, which a future permutation may use (I have been thinking of using that approach for static file serving, for example). I noticed that some other frameworks do exactly the same thing.

Anyway, I discovered an issue, and I will push another revision, but I wanted to see if you had encountered any other test failure, and to push any potential fixes together - no need to overload the Travis queue with several submissions.

nbrady-techempower:
@volyrique Why is `libyajl-dev` added to the `prereqs`? Can we do this in setup? Maybe put it in `setup/linux/systools` and get it from there?

volyrique:
I noticed that the same thing was done for some of the other dependencies (e.g. libpq and OpenSSL), and since the package manager provided YAJL, I decided to be consistent, and to get it in the same way (I was also trying not to write unnecessary scripts). It can certainly be done in both of the ways that you suggest, though my personal preference would be for `toolset/setup/linux/systools`; in this way YAJL wouldn't be rebuilt whenever the application would be recompiled, and IMHO that is the logical location.

nbrady-techempower:
@volyrique agreed. I'll be moving others to the same model.

volyrique:
Just to be clear - is the current approach acceptable, or do you insist on adding a build script for YAJL as well? As I said, I'd rather not include extra code.

nbrady-techempower:
@volyrique I think that removing it from prereqs is all that really matters for me here. If you can move it to systools and just fw_get it in the setup.sh file, would that be good for you? The plan is to move all the framework specific stuff out of prereqs.

volyrique:
Done.

nbrady-techempower:
@volyrique great stuff! thank you! all tests pass locally. will merge for round-14 after travis confirms

volyrique:
@nbrady-techempower I noticed several small things that I forgot to clean up (whitespace, comments, etc.). Would you mind if I push another revision (since you ran some tests locally) instead of opening another pull request?

nbrady-techempower:
@volyrique absolutely

volyrique:
@nbrady-techempower Thanks, the code is ready.

PR #2147, opened 2016-07-07
JSON Tests to Fail Gracefully Fixes #2146

### LINKED ISSUE #2146: JSON test doesn't gracefully fail
None of the other tests for a framework will run if you have a JSON test doesn't return exactly "Hello, World!" For example, if you change the return value to "Hello Wrld" or "Hello Woorld" you are given the following error:

```
Accessing URL http://127.0.0.1:8080/json:
  Response (trimmed to 40 bytes): "{"message":"Hello, Woorld!"}"
Process Test Runner (express):
  File "/usr/lib/python2.7/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/lib/python2.7/multiprocessing/process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "/home/knewman/FrameworkBenchmarks/toolset/benchmark/benchmarker.py", line 611, in __run_test
    passed_verify = test.verify_urls(verificationPath)
  File "/home/knewman/FrameworkBenchmarks/toolset/benchmark/framework_test.py", line 442, in verify_urls
    verify_type(test_type)
  File "/home/knewman/FrameworkBenchmarks/toolset/benchmark/framework_test.py", line 405, in verify_type
    test.failed = any(result == 'fail' for (result, reason, url) in results)
  File "/home/knewman/FrameworkBenchmarks/toolset/benchmark/framework_test.py", line 405, in <genexpr>
    test.failed = any(result == 'fail' for (result, reason, url) in results)
ValueError: need more than 2 values to unpack
verify: 100% |###########################################################################| Rough Time: 00:00:25
WARNING:root:Framework fasthttp does not define a default test in benchmark_config.json
  Verification Summary

| Test: express
|      NO RESULTS (Did framework launch?)
```

This only seems to happen with the JSON test. Making slight variations to the other tests will cause that particular endpoint to fail, but all other tests will continue on. If you're writing your tests properly, this shouldn't be much of an issue, but it would still be nice if the test failed gracefully and allowed the other tests to run.

nbrady-techempower:
When "hello, world!" is not matched, the `problem` returned was missing the url parameter.

### DIFF
--- a/toolset/benchmark/test_types/verifications.py
+++ b/toolset/benchmark/test_types/verifications.py
@@ -125,7 +125,7 @@ def verify_helloworld_object(json_object, url):
         message = json_object['message']
 
         if message != 'hello, world!':
-            return [('fail', "Expected message of 'hello, world!', got '%s'" % message)]
+            return [('fail', "Expected message of 'hello, world!', got '%s'" % message, url)]
         return problems
 
 


knewmanTE:
LGTM!

PR #2149, opened 2016-07-07
remove Java play1siena framework

knewmanTE:
The Java [play1siena framework](https://www.playframework.com/modules/siena) hasn't been updated since 2012 and breaks because of dependencies that are not compatible with Java 8. I think we should probably remove this test because:

- It is very outdated
- All of our other Java tests are running on Java 8
- We already have a working play1 test without siena


nbrady-techempower:
Agreed! This puts all the Java frameworks in the green category!

PR #2150, opened 2016-07-08
Specify perl extension version

knewmanTE:
Fixes Perl/dancer and Perl/plack which were randomly breaking when trying to pull DBD:mysql with a cpanfile. Moved their dependencies to their respective setup.sh files and also specified version numbers for the dependencies in Perl/kelp and Perl/web-simple as well (which weren't breaking, but were just pulling the latest version of all of their dependencies).


knewmanTE:
Perl tests all passing on Travis.

PR #2152, opened 2016-07-08
Go/Beego: interpolateParams=true

methane:
beego/orm doesn't cache prepared statement.
It is huge performance penalty when using real prepared statement.
mysql driver provides direver's side parameter interpolation.  Let's use it.


nbrady-techempower:
Failing test is unrelated, passes local tests as well! Merging for round-14. Thanks for all your work @methane !

PR #2153, opened 2016-07-08
Updated readme to make changing dir clearer

nbrady-techempower:
Updated readme to make it clearer to move into the FrameworkBenchmarks directory in the vagrant machine before running the first test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,9 +43,12 @@ required.
 
         $ vagrant ssh
 
-5. Run a test.
+5. Move into the FrameworkBenchmarks directory in the vm.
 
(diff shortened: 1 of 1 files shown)


ryanlath:
That didn't work for me.  But I figured out the password: "test"  So all's good.

nbrady-techempower:
@ryanlath If you could reproduce, and show the steps you took to get to the problem you're having, it's the only way I can see what the issue is since following the steps for me works.

ryanlath:
You can close this.  I updated the vagrant box and re-provisioned and that (or something else) seems to get me in.

PR #2158, opened 2016-07-11
WIP: Upgrade hapiJS

knewmanTE:
Upgrades hapiJS to its newest release (13.5.0). Also upgrades most of its dependencies to their newest releases as well.

WIP: While I'm at it, I'm upgrading the dependencies for all of the other JS frameworks to as high as they can go, so hold off on merging this until then.


nbrady-techempower:
Tested locally as well! merging for round-14!

PR #2163, opened 2016-07-12
Upgrade nodeJS to 6.3.0

knewmanTE:
This merge request does the following:

- upgrades nodeJS to 6.3.0 (the latest release)
- upgrade all npm dependencies for node projects to their latest versions (unless the latest version breaks the framework, which happened with some versions of sequelize)
- remove ~ and ^ versioning modifiers for npm dependencies in an effort to keep the node projects more reliably running
- Comments out the call to `$NODE_HOME/bin/npm install -g npm` in the nodeJS install script. This was causing the node install script to break for nodeJS v6, and I've tested the version of npm that comes packaged along with nodeJS v6 and it works fine for all of the frameworks


nbrady-techempower:
Excellent! still just the sails issue which should be fixed with #2164 - merging for round-14!

PR #2160, opened 2016-07-12
Fix mongo login and bump vert.x core version to latest stable 3.3.2

pmlopes:

PR #2165, opened 2016-07-13
Fix mongo login and bump vert.x core version to latest stable 3.3.2

pmlopes:
Same as: https://github.com/TechEmpower/FrameworkBenchmarks/pull/2160

But now on the right branch


nbrady-techempower:
Good to go, merging for round-14!

PR #2161, opened 2016-07-12
ffead-cpp-apache2 addition

nbrady-techempower:
This updates #2061 ffead update and apache test addition

PR #2173, opened 2016-07-27
Sc ffead cpp

sumeetchhetri:
Merged changes from (nbrady-techempower:ffead-cpp) https://github.com/TechEmpower/FrameworkBenchmarks/pull/2161 and fixed the failing ffead-cpp-apache2 test.


sumeetchhetri:
Finally the ffead-cpp and ffead-cpp-apache2 tests pass successfully, Can someone please check why the build still fails, the logs for the test run can be found at https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/148493112. @nbrady-techempower can you please look at this and get this merged as this PR has been open for a long time now :-)

toolset/benchmark/benchmarker.py
```diff
@@ -561,6 +561,7 @@ def exit_with_code(code):
             sudo service postgresql restart
             sudo service cassandra restart
             /opt/elasticsearch/elasticsearch restart
+   	    sudo /etc/init.d/apache2 stop
```
knewmanTE: Is there a reason why this is a stop and not a restart like the other services in this function?

sumeetchhetri: This line can be removed, it is not required anymore!

knewmanTE: Awesome. Could you push a fix removing the line? Other than that I'd say this is good to merge!

sumeetchhetri: Done

knewmanTE:
Thanks @sumeetchhetri! So it looks like what's causing Travis to mark the test as a fail is this line at the end:

[The command "time ./toolset/run-ci.py verify "$TESTDIR"" exited with 1.](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/149349876#L3806)

We're looking into why this is happening, so we're going to hold off on merging this in for a little bit. Sorry for the delay, we'll do our best to get it merged in as soon as we can!

One thing that comes to mind is the ``ffead-cpp-apache.sh`` file. Is there a way to get the app running without having to access _/var/www/_ or _/etc/apache2/_ ? You also shouldn't have to add a new user, testrunner already has admin privileges.

sumeetchhetri:
I tried that already, having the ffead-cpp web folder anywhere else does not pass the travis test, it passed for me locally though. I had to go through several combinations to make travis pass my tests, moving the web folder to /var/www was the important one after disabling the default port Listen 80 of apache.

msmith-techempower:
Travis borked; this looks good to me; LGTM!

PR #2176, opened 2016-08-01
Remove PHP/micromvc

### LINKED ISSUE #2174: Is it worth keeping PHP/micromvc?
What the title says. For the PHP framework [micromvc](https://github.com/Xeoncross/micromvc), I am seeing that:
- It is no longer maintained
- Hasn't been updated in 3-5 years
- Is only compatible with PHP 5
- Its [website ](http://micromvc.com/)is defunct. You can find a [cached version](https://web.archive.org/web/20110201124701/http://micromvc.com/) of the site on the Web Archive, but there isn't a ton of info there
- The framework doesn't have a full test suite (only JSON, DB, and Queries are implemented)

Given these factors, it might make sense to remove the framework, as I don't imagine its results are particularly meaningful to anyone who might want to build with a PHP framework or compare different PHP frameworks.

That being said, it did perform decently well against other PHP frameworks on Round 12 (5th/31 on JSON, 7th/30 on DB, 17th/27 on Queries), but I don't know if that is reason enough to keep it in our suite.

knewmanTE:
It is outdated and unmaintained. See #2174 for a more comprehensive explanation.

Resolves #2174


nbrady-techempower:
I believe the Aye's have it. Merging this in.

PR #2179, opened 2016-08-02
Update PHP/phalcon to v3

knewmanTE:
The phalcon team has released [v3.0.0 of phalcon php](https://blog.phalconphp.com/post/phalcon-3-0-0-released)! I've updated our phalcon test to use the latest version of the framework, which also makes it compatible with PHP 7.

---

<b>Update</b>
It appears that Phalcon 3 [doesn't currently support MongoDB](https://github.com/phalcon/cphalcon/issues/12074). I'll be keeping an eye on it and will try to update this PR once Phalcon 3 has been updated.

For future reference, I've been currently trying to use [this PHP7-compatible MongoDB library](http://mongodb.github.io/mongo-php-library/). The problem I'm seeing is that Phalcon's built-in MongoDB Collection API doesn't match up with above library, which so far as I can tell is the only way to get MongoDB working with PHP7 at the moment.

<b>Update 2</b>
On August 11th, the MongoClient task was added to the 3.1 milestone for Phalcon, so keep an eye out for that version release!


knewmanTE:
Looks like php-phalcon-mongodb is failing, probably because of the php7mongodb driver.

msmith-techempower:
@knewmanTE Where does this PR stand?

knewmanTE:
@msmith-techempower I haven't looked into it in a bit, but I don't see any recent updates to PhalconPHP that appear to address the MongoDB issue. If I have some down time I'll try to check in with this branch and see if I can get it working.

knewmanTE:
Possibly relevant: https://forum.phalconphp.com/discussion/14885/mongodb-support-in-phalcon-3-php-7

knewmanTE:
@msmith-techempower should be fixed now. Let's see what Travis thinks.

knewmanTE:
Phalcon is passing on Travis. I made a change to the phalcon setup file, so all of the other tests were run, which is why we're seeing the fails on Travis.

PR #2180, opened 2016-08-02
Mf/django plaintext

mwflickner:
Python Django plaintext test

### DIFF
--- a/frameworks/Python/django/benchmark_config.json
+++ b/frameworks/Python/django/benchmark_config.json
@@ -49,8 +49,6 @@
     },
     "postgresql": {
       "setup_file": "setup_pg",
-      "plaintext_url" : "/plaintext",
-      "json_url": "/json",
       "db_url": "/db",
       "query_url": "/dbs?queries=",
       "fortune_url": "/fortunes",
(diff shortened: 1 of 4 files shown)


frameworks/Python/django/benchmark_config.json
```diff
@@ -47,6 +49,7 @@
     },
     "postgresql": {
       "setup_file": "setup_pg",
+      "plaintext_url" : "/plaintext",
       "json_url": "/json",
```
methane: Should we have `/plaintext` and `/json`for both of MySQL and PostgreSQL?

methane:
other than that, lgtm.

frameworks/Python/django/benchmark_config.json
knewmanTE: @mwflickner since plaintext and json don't require any DB access, we shouldn't expect any different results on postgresql vs. mysql. can you remove them from the postgresql test since we already have them elsewhere in the django tests?

mwflickner: @knewmanTE just removed with latest commit

knewmanTE: Awesome. Will merge now!

PR #2182, opened 2016-08-03
Round 14 hexagon

jaguililla:
Add Hexagon framework tests


knewmanTE:
Looks good @jaguililla! Always exciting to add a new language to the testing suite!

Is it possible to remove ``gradle-wrapper.jar`` and then generate it or pull it from somewhere online in your ``setup.sh`` file? For the sake of cleanliness and keeping the TFB repo maintainable, we try to reduce how much boilerplate code we keep in each framework and generate as much of it locally on the spot as possible.

jaguililla:
I understand your concerns, I do not like to have binaries in the VCS either. Anyway, it is the recommended way of [dealing with Gradle](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:wrapper_generation). The JAR is a small bootstrap program (53K) that downloads the proper Gradle version. The alternative would be installing the binary and dealing with Gradle versions in the OS instead on the project. I would stick to the Gradle recommendations, what do you think?

knewmanTE:
@jaguililla that's fine by me then! Merging this into round-14!

jaguililla:
Thanks!

PR #2183, opened 2016-08-03
Mf/java/grizzly jersey/plaintext

mwflickner:
plaintext test passing for Java Grizzly Jersey


knewmanTE:
LGTM! @mwflickner, while you're at it, could you take a quick look and see if you can fix the Content Type for JSON, DB, and Query to PASS instead of WARN? Should be as simple as specifying type `application/json` with no `charset=utf-8`.

mwflickner:
@knewmanTE JSON, DB, and Query all passing now

PR #2186, opened 2016-08-04
Mf/java/undertow jersey c3p0/plaintext

mwflickner:
plaintext passing on undertow-jersey-c3p0

### DIFF
--- a/frameworks/Java/undertow-jersey-c3p0/src/main/java/hello/PlaintextResource.java
+++ b/frameworks/Java/undertow-jersey-c3p0/src/main/java/hello/PlaintextResource.java
@@ -1,0 +1,19 @@
+package hello;
+
+import javax.inject.*;
+import javax.ws.rs.*;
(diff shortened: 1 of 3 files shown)


frameworks/Java/undertow-jersey-c3p0/src/main/java/hello/PlaintextResource.java
```diff
@@ -0,0 +1,19 @@
+package hello;
+
+import javax.inject.*;
+import javax.ws.rs.*;
+import java.util.*;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
```
knewmanTE: Take this line out, since it doesn't seem like we're using it. Otherwise, looks good to me!

mwflickner: @knewmanTE removed!

PR #2187, opened 2016-08-04
Mf/java/undertow jersey hikaricp/plaintext

mwflickner:
Plaintext is now a PASS on Undertow Jersey Hikaricp


knewmanTE:
looks good to me!

PR #2189, opened 2016-08-07
Update Crystal

jhass:
See commits.


jhass:
Mh, should this actually go against one of the `round-` branches? I wasn't sure.

kbrock:
This is nice.

1. Are there still benchmarks for redis?
2. Is there a reason we are escaping parenthesis?

jhass:
> Are there still benchmarks for redis?

Is that a question for me? If so I'm not sure I follow. If you're asking whether I checked whether those still pass before enabling them, yes. You can also see that the CI build passed for them.

> Is there a reason we are escaping parenthesis?
Crystal's `HTML.escape`, which is used by the HTML builder in the Kemal examples, escapes parenthesis. This was introduced by https://github.com/crystal-lang/crystal/pull/2175 which based it on http://wonko.com/post/html-escaping

knewmanTE:
@jhass could you switch this PR over to round-14? Because of the long delay in getting round 13 out, we're doing some switching around of how our branches are organized. You can see a detailed explanation of what we're doing at the bottom of [this thread](https://groups.google.com/forum/#!topic/framework-benchmarks/AGHUyMnR_0o), but the end-goal is that we'd love to have working Crystal tests listed on our next results page.

PR #2175, opened 2016-07-30
Add missing php tests

knewmanTE:
<b>NOTE:</b> I'm not entirely sure why there are some non-PHP changes in this PR. So far as I can tell, the rbx and vertx changes should already be in round-14. I'd advise against merging this in until we get to the bottom of it.

<b>UPDATE:</b> I've looked through all of the non-PHP changes and they are all on round-14, so I'm chalking it up to being some sort of Github bug.

<b>UPDATE 2:</b> Non-PHP files no longer showing up in the diff!

Adds the following missing PHP tests:

- cygnite
  - fortunes
- lithium
  - plaintext
  - updates
  - fortunes
- slim
  - plaintext
  - updates
  - fortunes
- symfony2
  - plaintext
  - updates
- symfony2-stripped
  - plaintext
  - update
- yaf
  - plaintext
  - update
- yii2
  - plaintext
- zend
  - plaintext
  - updates
- zend1
  - plaintext
  - updates
  - fortunes

The following PHP tests are still missing:
- micromvc
  - see #2174 
- slim (hhvm)
  - fortunes (having trouble getting this to render the Japanese characters)
- zend
  - fortunes (can't for the life of me get the templating to work here)

PR #2190, opened 2016-08-10
Update crystal

jhass:
See #2189


toolset/benchmark/fortune_html_parser.py
```diff
@@ -74,20 +74,30 @@ def handle_charref(self, name):
     # Not sure why some are escaping '/'
     if val == "47" or val == "047" or val == "x2f":
       self.body.append("/")
+    # "&#40;" is a valid escaping of "(", but
+    # it is not required, so we need to normalize for out
+    # final parse and equality check.
+    if val == "40" or val == "040" or val == "x28":
```
knewmanTE: I'm seeing some Fortune fails on some tests for Fortune 11 (the Javascript alert that should be escaped). Could you try locally removing the additions to this file and running a test to see if these lines are what is affecting frameworks like [C++/ffead-cpp](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/151374139#L2674) or [C++/treefrog-mongodb](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/151374134#L3081)? It's possible that those frameworks are the problem and need to implement more rigorous sanitation on the DB values, but I'd like to figure out if the two things are somehow connected before merging this in.

jhass: Note that there's an identical failure in the build for #2175 

This: https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/151374139#L2664-L2680
#2175: https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/151376090#L5173-L5189

Or:

This: https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/151374134#L3068-L3087
#2175: https://s3.amazonaws.com/archive.travis-ci.org/jobs/151376071/log.txt (Travis log view is corrupted here for me)

knewmanTE: Hmm. Very strange. I will do some investigation as to why this might be happening. But otherwise, your pull request looks good!

knewmanTE: @jhass I just locally tested ffead-cpp and treefrog on the following branches:

- round-14
- add-missing-php-tests
- jhass/update_crystal

and am seeing Fortune pass there, so it's possible there's just something funky going on with Travis right now. I've also been seeing some weird MongoDB fails the past day or two. I'm still not sure what the issue is, but I'm good on merging your branch in now.

Issue #2191, opened 2016-08-11
MongoDB failing to start on Travis


knewmanTE:
If you look at recent Travis runs, we're seeing a bunch of fails on frameworks that were previously passing because MongoDB isn't starting up.

Some examples:
- [Go/go-std](https://s3.amazonaws.com/archive.travis-ci.org/jobs/151376128/log.txt) (jump to the bottom)
- [Haskel/yesod](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/151376141#L2540)
- [Java/Dropwizard](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/151376151#L4229)
- [PHP/phalcon](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/151376224#L2328)

I'm only seeing this on Travis runs from the past couple of days, so it's possible that it is something that will end up working itself out.

knewmanTE:
Resolved with #2238

PR #2199, opened 2016-08-17
Use Go 1.7

methane:


methane:
Tast fails are caused by MongoDB setup.

knewmanTE:
@methane yeah, I made an issue #2191 about MongoDB failing to start up. I'm still not totally sure if this is something that we did or if it's a bug on the Travis side of things.

Anyway, thanks for updating Go!

PR #2205, opened 2016-08-19
Fix issues with servant.

jkarni:
DB host was not being used.

### DIFF
--- a/frameworks/Haskell/servant/driver/Main.hs
+++ b/frameworks/Haskell/servant/driver/Main.hs
@@ -1,13 +1,16 @@
 {-# LANGUAGE OverloadedStrings #-}
 module Main (main) where
 
-import Data.ByteString
+import Data.ByteString.Char8
 import ServantBench
 import Hasql.Connection (settings)
+import System.Environment (getArgs)
 
 main :: IO ()
-main = run 7041 dbSettings
+main = do
+  [host] <- getArgs
+  run 7041 $ dbSettings (pack host)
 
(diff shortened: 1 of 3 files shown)


jkarni:
Note that these arose from someone pointing out [failures in the preliminary round 13 results](https://www.techempower.com/benchmarks/previews/round13/#section=data-r13&hw=azu&test=db&l=8vn073). The previous PR was again round-14; I'm not sure what this should be against.

knewmanTE:
Sorry for the confusion! Brian made a mention of our change in [this Google Group thread](https://groups.google.com/forum/#!topic/framework-benchmarks/AGHUyMnR_0o), saying "we have decided to base Round 13 on the current 'Round 14' branch.  This is much the same as Jonathan Dunlap has suggested above, but we won't skip the name 'Round 13.'  Instead, given the delays, we'll just move the "bar" so to speak to the current state of the art.  The net effect is the same, but we'll retain the name 'Round 13.'  So don't be surprised if you see movements in that direction at Github."

If you want to close this merge request and open it up against round-13, that is the branch we're running our preview tests on, but I will also be pulling round-14 into round-13 before our next run as well, to accommodate some of this confusion.

<b>Long term plan:</b> we will be folding round-13 into master, round-15 into round-14, and then use round-14 for the actual next official test release.

jkarni:
@knewmanTE Thanks for the prompt reply! I opened a new PR against round-13 in #2207.

PR #2227, opened 2016-08-23
Decrease backlog to 128

### LINKED ISSUE #2213: net.core.somaxconn is 128
http://tfb-logs.techempower.com/round-13/preview-1/latest/logs/bottle-nginx-uwsgi/out.txt.gz
bottle-nginx-uwsgi failed to startup.

The log said:
```
Server bottle-nginx-uwsgi: Listen queue size is greater than the system max net.core.somaxconn (128).
```

methane:
uwsgi tests were failed again and again because of this issue.
All other frameworks run with 128 backlog.  So just decrease it. 
fixes #2213

### DIFF
--- a/frameworks/Python/bottle/uwsgi.ini
+++ b/frameworks/Python/bottle/uwsgi.ini
@@ -2,7 +2,7 @@
 master
 ; Increase listen queue used for nginx connecting to uWSGI. This matches
 ; net.ipv4.tcp_max_syn_backlog and net.core.somaxconn.
-listen = 65535
+listen = 128
 ; for performance
 disable-logging
 ; use UNIX sockets instead of TCP loopback for performance
--- a/frameworks/Python/flask/uwsgi.ini
+++ b/frameworks/Python/flask/uwsgi.ini
@@ -2,7 +2,7 @@
 master
 ; Increase listen queue used for nginx connecting to uWSGI. This matches
 ; net.ipv4.tcp_max_syn_backlog and net.core.somaxconn.
-listen = 65535
+listen = 128
 ; for performance
 disable-logging
 ; use UNIX sockets instead of TCP loopback for performance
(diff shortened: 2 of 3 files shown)


frameworks/Python/bottle/uwsgi.ini
```diff
@@ -2,7 +2,7 @@
 master
 ; Increase listen queue used for nginx connecting to uWSGI. This matches
 ; net.ipv4.tcp_max_syn_backlog and net.core.somaxconn.
-listen = 65535
+listen = 128
```
Dith3r: this should be adjusted to variable from sysctl net.core.somaxconn and should be much higher than 128.

knewmanTE: ping @methane

methane: 128 is enough for concurrency=128, off course.
And it may be not enough for concurrency=256, but I don't know until measure.

Because it uses unix domain socket, backlog shortage cause "502 Bad Gateway" soon.
I'll think about it after seeing next result.

@knewmanTE Could you ensure `net.core.somaxconn` is more than 1024 for all setup?

knewmanTE: @methane it is 65535 for all setups. It is possible that it was set to 128 on one of the 3 machines during our preview round, but we will make sure it's correctly set for our next run of the suite.

methane: @knewmanTE OK, thanks. Let's increase it.

methane: @knewmanTE All nginx-uwsgi tests are still failing in Round14 preview.
Is this already fixed?

knewmanTE: From what I can tell, [it appears to be higher on master](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Python/bottle/uwsgi.ini#L5) , which, as far as I know, is what we're using for our Round 14 preview rounds.

Ping @nbrady-techempower @msmith-techempower

nbrady-techempower: uwsgi should be fixed for the Round 14 2nd preview

Issue #2221, opened 2016-08-22
Some of the Scala libraries don't play fair


vkostyukov:
I was quite surprised to see that some of the Scala libraries we test here don't respect one of the major rules: a response payload should be created/allocated on each request.

Examples: 
 - [colosus](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Scala/colossus/src/main/scala/example/Main.scala#L20) serves predefined payloads for both text and JSON (the JSON object is preallocated)
 - [s-server](https://github.com/TechEmpower/FrameworkBenchmarks/blob/round-13/frameworks/Scala/s-server/src/main/scala/woshilaiceshide/sserver/tfb/Server.scala#L33) serves preallocated plain text responses

I think it's fair to get those fixed before publishing Round 13 results. I'm happy to open a PR with fixes. Let me know if you have any questions and sorry for pointing fingers. I understand we can game some of the components (and we all do), but I wish we'd all respect at least basic rules.

bhauer:
@vkostyukov Thank you for bringing these to our attention!  If you find others, please feel free to document them as well.

Of these, I believe the Colosus implementation of JSON should be resolved.  Implementations of the JSON test type should incur the cost of allocating memory/instantiating objects.  Please submit a PR to fix this if you have the time.  Thank you!

However, a pre-allocated buffer of "Hello, World" in plaintext is acceptable for the Plaintext test.  Because that test was designed to exercise the utmost high-water mark of HTTP request handling, in a previous round we clarified the requirements of that test to be clear that it was not necessary to allocate new memory for the response payload of each request.  See requirement 4 from Test Type 6 below:

https://www.techempower.com/benchmarks/#section=code

> This test is not intended to exercise the allocation of memory or instantiation of objects. Therefore it is acceptable but not required to re-use a single buffer for the response text (Hello, World). However, the response must be fully composed from the response text and response headers within the scope of each request and it is not acceptable to store the entire payload of the response, or an unnaturally large subset of the response, headers inclusive, as a pre-rendered buffer. "Buffer" here refers to a byte array, byte buffer, character array, character buffer, string, or string-like data structure. The spirit of the test is to require the construction of the HTTP response as is typically done by a framework or platform via concatenation of strings or similar. For example, pre-rendering a buffer with `HTTP/1.1 200 OK<cr>Content-length: 15<cr>Server: Example<cr>` would not be acceptable.

vkostyukov:
[...]

PR #2231, opened 2016-08-23
Make Colossus building a response object on each request

vkostyukov:
This addresses #2221 and changes the way Colossus responds a JSON payload. The response object is now allocated on each request as per benchmark rules.


nbrady-techempower:
Looks good, passes relevant test on travis. Nice catch! Merging

PR #2233, opened 2016-08-23
Update Finagle to 6.36

vkostyukov:
This PR updates Finagle to 6.36 and also makes some misc changes:

1. Make `DateTimeFormat` a thread-local since it's not thread-safe and it's not fair to share it
2. Use one-jar plugin to run the program instead of reusing the sbt's JVM instance

Please let me know if there anything I can do/change to make this PR make it to the Round 13.

### DIFF
--- a/frameworks/Scala/finagle/src/main/scala/Main.scala
+++ b/frameworks/Scala/finagle/src/main/scala/Main.scala
@@ -1,4 +1,3 @@
-import java.text.{DateFormat, SimpleDateFormat}
 import java.util.Date
 
 import com.fasterxml.jackson.databind.ObjectMapper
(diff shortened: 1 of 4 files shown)


frameworks/Scala/finagle/src/main/scala/Main.scala
```diff
@@ -13,14 +13,17 @@ import com.twitter.io.Buf
 object Main extends App {
 
   val mapper: ObjectMapper = new ObjectMapper().registerModule(DefaultScalaModule)
-  val dateFormat: DateFormat = new SimpleDateFormat("E, dd MMM yyyy HH:mm:ss z")
+
+  val dateFormat: ThreadLocal[DateFormat] = new ThreadLocal[DateFormat] {
+    override def initialValue: DateFormat = new SimpleDateFormat("E, dd MMM yyyy HH:mm:ss z")
```
cacoco: The format here should be "E, dd MMM yyyy HH:mm:ss 'GMT'". See the issue we hit and the pull-request here: https://github.com/twitter/finatra/pull/312

vkostyukov: Thanks @cacoco! I updated the PR to use request API `req.date = ???` instead that uses the right date format.

knewmanTE:
@vkostyukov looks good! we'll do our best to get this into round 13!

vkostyukov:
Thank you, @knewmanTE!

PR #2234, opened 2016-08-23
Update Finch to 0.11-M2

vkostyukov:
This RB updates Finch to 0.11-M2 as well as changes the implementation to reflect its best practices.


frameworks/Scala/finch/src/main/scala/Main.scala
```diff
@@ -0,0 +1,47 @@
...
+import com.twitter.util.Await
+
+import io.circe.Json
+import io.finch._
+import io.finch.circe._
+
+object Main extends App {
+
+  val dateFormat: ThreadLocal[DateFormat] = new ThreadLocal[DateFormat] {
+    override def initialValue: DateFormat = new SimpleDateFormat("E, dd MMM yyyy HH:mm:ss z")
```
martin-g: Why don't you use Java 8 DateTimeFormatter instead ?
It is thread-safe, so you'll need just one instance.

vkostyukov: Good point! Thanks!

Let me update the PR.

knewmanTE:
For some reason Travis fired off 2 runs for your pull request, so I cancelled [the one that seemed to be a bug](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/154543626). The other Travis run is still scheduled to run and can be followed [here](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/154580050) (or by clicking the yellow circle next to your commit above).

vkostyukov:
Somehow this failed on the Erlang test. @knewmanTE do you want me to restart the build?

knewmanTE:
@vkostyukov don't worry about the Erlang test. Misultin has been removed from the benchmark tests, but your Travis run thinks it still exists, so it's trying to check the directory but not finding anything there. Won't be an issue on master.

Issue #1622, opened 2015-05-25
Confusion about rules for database update test


denkab:
http://frameworkbenchmarks.readthedocs.org/en/latest/Project-Information/Framework-Tests/#database-updates

In a way, rules 9 and 10 seem to be in conflict. It is very typical that individual updates in a batch for traditional RDBMS (and NoSQL is not different if similar optimization concepts are followed) can not be separated into individual transactions.

Possibly, a clarification and revision of existing code is needed to bring all implementations into compliance.

zapov:
It's also very typical to do selects in batches, at least on DB which support it, such as MS Sql Server. And is unfair to disallow patterns and frameworks which have workarounds on other DBs.
I find it highly unlikley that the rules will be fixed, but at the very minimum tests should be renamed to multiple roundtrips instead of multiple queries. That should remove any confusion

msmith-techempower:
The rules in question have been updated to better represent their meaning.

PR #2237, opened 2016-08-24
Postgres/MySQL: Restrict database permissions to just what is required

k-r-g:
Some frameworks were misbehaving and changing the schema.

Also remove upper-case verions of MySQL tables as those are not needed.


config/create.sql
```diff
@@ -3,66 +3,15 @@ DROP DATABASE IF EXISTS hello_world;
...
-INSERT INTO Fortune (message) VALUES ('A bad random number generator: 1, 1, 1, 1, 1, 4.33e+67, 1, 1, 1');
-INSERT INTO Fortune (message) VALUES ('A computer program does what you tell it to do, not what you want it to do.');
-INSERT INTO Fortune (message) VALUES ('Emacs is a nice operating system, but I prefer UNIX. — Tom Christaensen');
-INSERT INTO Fortune (message) VALUES ('Any program that runs right is obsolete.');
-INSERT INTO Fortune (message) VALUES ('A list is only as strong as its weakest link. — Donald Knuth');
-INSERT INTO Fortune (message) VALUES ('Feature: A bug with seniority.');
-INSERT INTO Fortune (message) VALUES ('Computers make very fast, very accurate mistakes.');
-INSERT INTO Fortune (message) VALUES ('<script>alert("This should not be displayed in a browser alert box.");</script>');
-INSERT INTO Fortune (message) VALUES ('フレームワークのベンチマーク');
-
```
msmith-techempower: I think we need the fortunes table.

LadyMozzarella:
@nkasvosve (BeyondJ), @jberger (Mojolicious), @zapov (RevenJ) - Heads up! This pull request changes the permissions for the database and, based on Travis-CI results, it appears that frameworks you have worked on are impacted by it. Could you look over your framework implementations and see what is happening? Thanks.

zapov:
Revenj will read database metadata at startup (pg_class, pg_namespace, pg_attribute, pg_type and pg_description). Without seeing the actual error I would assume you disallowed reading from those system tables. So grant select onto those tables and it should be back to running again.

zapov:
Ah.. I see. It fails to do an update due to missing insert/delete permissions.
Revenj calls generic save method which creates SQL for doing insert/update/delete in a single roundtrip, which will fail even if there is no insert/delete data since you have removed permission for doing those stuff.

It will keep failing until you give back those permissions (or run it through different user)

jberger:
I'll take a look, should be today sometime. I needed to do a dependency-pin bump anyway so this is  a good motivator.

jberger:
Interestingly mojolicious doesn't seem to care about this change. What IS breaking things is that the server deployment script cannot create files or folders in the mojolicious/ project directory. Since we use carton to deploy dependencies this is breaking dependency installation. 

@LadyMozzarella do you have any suggestions about why this is happening? Clearly it didn't used to happen. It seems that the default directory permissions are 0775. Is the user that runs the server setup not the 'vagrant' user (and also not root)?

jberger:
So I'm guessing that the user `testrunner` is the one that sets up and starts the servers though the directory is owned by `vagrant` and the perms are 0775. What can be done to address this? I'm also on IRC if this should be discussed a little more synchronously.

LadyMozzarella:
@jberger I'm not sure what the issue was, but I think this Travis run may have been a fluke. Merging this PR into master kicked off a new Travis run based off of master. [Mojolicious passed there](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/155458061). I set up a new Vagrant environment based on current master and ran Mojolicious and it also passed there. I can't seem to replicate the issue locally.

@zapov Oh I see, we will have to look into/consider that particular situation specifically.

jberger:
That is interesting too, but if I can't run it in vagrant then something is still not kosher.

jberger:
BTW I think I have a patch that I think we be acceptable. Working now on seeing if I can make Perl versions be specified by frameworks. I'd like to get a more modern perl than the one we configured back several years ago.

zapov:
@LadyMozzarella dont try to make special rules for Revenj. I have to change the submission anyway since you've changed the rules to disallow current Revenj implementation. Hopefully before the next deadline

jberger:
@zapov I read your email comments and I have to say, the persecution complex is unbecoming. My Mojolicious app could open 100 connections and parallel query it and yet I understood from the rules that that was against the spirit; so I make individual queries in a for loop. If parallel (or bundled) queries were allowed mine would be faster too. If it means this much to you, propose a new test; I'll second it.

zapov:
@jberger ;( There was no need to hijack this thread for that discussion.
First of all, calling people names is immature and will lead nowhere.
Second of all, there are multiple implementations which do a parallel query inside the benchmark.
Third, I did complain about this more than a year ago (suggested to change the name of test/rules of the test): https://github.com/TechEmpower/FrameworkBenchmarks/issues/1622
And lastly, this benchmarks facilitates a myth that if you want to squeeze some more performance out your DB interaction you should start several parallel DB queries. By doing a Revenj submission I showed just how wrong that myth is. But there is no much point about discussing that. Especially not in this thread.

nbrady-techempower:
@zapov Are you submitting a PR for a batched query test?

zapov:
yeah, I will try to find some time and add a version which does a loop (and fix this permission issue in the process)

nbrady-techempower:
@zapov thanks :)

PR #2207, opened 2016-08-19
Fix issues with servant.

jkarni:
DB host was not being used. Max threads were being ignored.

PR #2240, opened 2016-08-25
Fix issues with servant.

jkarni:
DB host was not being used. Max threads were being ignored.


jkarni:
Originally #2207

jkarni:
Again, failures have nothing to do with servant or these changes.

PR #2241, opened 2016-08-25
Upgrading to Curacao 4.2.0

markkolich:
Upgrade benchmarks to Curacao 4.2.0


markkolich:
@knewmanTE would you mind squeezing this into round-13?

PR #2242, opened 2016-08-25
Fix broken Round 13 PHP tests (clancats, lumen, yii2, yii2-hhvm) 

knewmanTE:
4 PHP tests (clancats, lumen, yii2, yii2-hhvm) were failing on the preview run. This is because we weren't properly setting the IP address for the database. Because local and Travis tests are only on 1 machine, this error won't show up and the test will pass fine, but as soon as we moved to one of our production setups where the database and application server are on separate machines, the tests failed.

Anyway, this fixes that.


msmith-techempower:
Looks good to me; LGTM!

PR #2245, opened 2016-08-26
Fix setup_server function

knewmanTE:
I'm not sure why this is registering so many line changes. this is a 1 line fix, but Github is registering a bunch of missing/additional spaces.

Anyway, the __setup_server() function in benchmarker.py wasn't running its subprocess calls because the first call was always failing. This will fix that and alloow us to properly set the max number of connections to 65535.


LadyMozzarella:
Looks like the line changes were just trailing spaces that were automatically removed by your text editor.

knewmanTE:
@LadyMozzarella whomp :(

PR #2249, opened 2016-08-29
Disabled bulk reading for Revenj.

zapov:
Use simple loop instead of bulk reading for loading data N times from database.
Update dsl-compiler to latest version (to fix update permission).
Update Revenj.Java to 0.9.9 due to API change and sync with latest compiler.


knewmanTE:
Looks good to me! We've already started our 2nd preview round, but we'll get this in before round 13 final!

nbrady-techempower:
Passing local and travis, merging. Thanks for updating @zapov! Lmk if you get the chance to look at batch query tests. Would love to start opening the dialogue for a range of new tests!

PR #2254, opened 2016-08-31
Merging master fixes into round-14

nbrady-techempower:
Merging master fixes into round-14


knewmanTE:
My guess is this Travis run isn't going to tell us much, since it's basically just running the latest from master, but I think this should still be merged in because 

1) round-14 is a dev branch that needs to stay up to date 

2) we're reasonably happy with the stability of master at the moment.

PR #2256, opened 2016-09-02
Added benchmarks for Rapidoid using the high-level API.

nmihajlovski:


nbrady-techempower:
@nmihajlovski I appreciate your efforts here, especially with transparency. Moving forward, we would love to be able to show people snippets of what realistic implementations of tests would look like. 

Is there any way we can combine this within the same rapidoid folder? We don't want to have separate framework directories for "low" and "high-level" implementations, but if you could keep it in rapidoid and then name the tests accordingly that would work great.

Would not be opposed to putting some snippets/examples in the README as well.

nmihajlovski:
Thanks for your quick reply.
I combined the low-level and high-level implementations, and added some basic documentation.

nbrady-techempower:
Great stuff @nmihajlovski! I will test this out locally by end of day tomorrow and get it in to round-14. This is definitely the gold standard of adding framework tests! Thanks again!

nmihajlovski:
Thanks, @nbrady-techempower, that sounds great! I am looking forward to seeing and comparing the high-level / low-level results.

nmihajlovski:
@nbrady-techempower Is there any reason why this PR hasn't been marked with any label (e.g. Round-14)?

nbrady-techempower:
Looks like that's all set @nmihajlovski Lmk if you need anything else. I've had my head in some travis and toolset work.

nmihajlovski:
Thanks! :)

PR #2257, opened 2016-09-02
Realistic low-level approach with "rapidoid-http-fast"

nmihajlovski:
Some users complained that the "rapidoid" tests are implemented using low-level techniques, despite the expected one-liners presented in the docs.

I renamed the apparently misleading "rapidoid" tests to "rapidoid-http-fast", to avoid the confusion and set the right expectations.

rapidoid-http-fast is the low-level HTTP and generic network protocol framework:
http://www.rapidoid.org/http-fast.html

I really believe that now the "rapidoid-http-fast" benchmark should be marked as realistic, because:
- rapidoid-http-fast is used using low-level bytes manipulation, exactly as officially advertised on the web site,
- it has production quality (rounds 11 and 12 demonstrated its high performance and 0 errors),
- low-level doesn't imply "striped" (e.g. Netty).

But, to be really fair, I already opened a PR using the high-level Rapidoid API:
https://github.com/TechEmpower/FrameworkBenchmarks/pull/2256

I hope we will be able to see both the high-level and low-level approaches and compare the results very soon...


knewmanTE:
@nmihajlovski, given that this merge request doesn't address the concerns that were initially raised about Rapidoid's implementation, I don't think just changing the name of the test is enough to merit reclassify it from Stripped to Realistic, as it is still manually constructing the byte arrays for all of the response headers.

nmihajlovski:
I took care of that. The old code is replaced by a new and simple code.
(Basically a made copy of the rapidoid-http-fast implementation of PR #2256.

nmihajlovski:
@knewmanTE Is it too late to include this in the upcoming "preview 2" round?

knewmanTE:
@nmihajlovski we ran preview 2 over the weekend, so this pull request was not in the code base for our suite run.

nmihajlovski:
@knewmanTE Is there any reason why this PR hasn't been marked with a "Round-13" label?

knewmanTE:
Hey @nmihajlovski, sorry for the delay. We've been busy with some internal tasks preparing for preview 2 and haven't had the time to do a full pass-through of the additional changes you've made. We'll be taking a look today though!

frameworks/Java/rapidoid/benchmark_config.json
```diff
@@ -1,24 +1,26 @@
...
-      "platform": "Rapidoid",
-      "webserver": "None",
-      "os": "Linux",
-      "database_os": "Linux",
-      "display_name": "rapidoid",
-      "notes": "",
-      "versus": "rapidoid"
+  "tests": [
+    {
+      "http-fast": {
```
knewmanTE: could you change this back to 'default'? The way the suite is built,  each `benchmark_config.json` must contain at least a 'default' test.

nmihajlovski: Thanks @knewmanTE for the quick reply. I changed the test name to 'default'.

nmihajlovski:
Thanks!

PR #2260, opened 2016-09-05
H2O: Optimize the implementation

volyrique:
Some optimizations based on the round 13 preview data.


volyrique:
All test failures occur in unrelated code.

PR #2263, opened 2016-09-08
Update web2py tests to production quality

abastardi:
The current web2py tests are based on a typical *development* setup and are therefore a very poor representation of the framework's performance. The current code uses the built-in development server rather than a production server (most other Python frameworks are using gunicorn+meinheld), unnecessarily loads a .ini file *on every request*, fails to use database connection pooling, leaves database migrations turned on (which adds overhead to every request), and fails to compile the application (typically done in production to improve performance).

This pull request corrects all of these issues with a setup and code typical of production deployment. Running the tests via the Vagrant VirtualBox VM, the new code leads to a 3-5x improvement in the database tests, a 6x improvement in the JSON test, and a 300x improvement in the plain text test.

It would be great if these changes could be incorporated in time for the Round 13 final release.


abastardi:
Also, note that with the current setup, the plain text test is completely broken, completing only 2 requests in the Round 13 preview data. This pull request fixes that, yielding almost 7000 completed requests when testing in the Vagrant VM.

frameworks/Python/web2py/app/standard/modules/database.py
```diff
@@ -0,0 +1,56 @@
...
+class Dal(object):
+    def __init__(self, table=None, pool_size=8):
+        self.db = DAL(DATABASE_URI, migrate_enabled=False, pool_size=pool_size)
+        if table == 'World':
+            self.db.define_table('World', Field('randomNumber', 'integer'))
+        elif table == 'Fortune':
+            self.db.define_table('Fortune', Field('message'))
+
+    def get_world(self, wid):
+        return self.db(self.db.World.id == wid).select(cacheable=True)[0].as_dict()
```
msmith-techempower: Does `cacheable=True` here imply that it's okay for the database to cache the result, or the ORM?

frameworks/Python/web2py/setup.sh
```diff
@@ -2,13 +2,18 @@
...
+git clone --recursive --branch master https://github.com/web2py/web2py.git
+cd web2py
+git checkout 326335c3cd027a97b9c2b83d880b2b1f8f62321d
+cd ..
+cp -r app/standard/ web2py/applications/
+cp -r app/optimized/ web2py/applications/
+cp app/wsgi.py web2py/
 cp app/routes.py web2py/
+touch web2py/__init__.py
+python compile_apps.py
```
msmith-techempower: This sort of stuff should be done inside of the `IROOT` directory if possible.

If that is NOT possible, then at least include a `.gitignore` in `frameworks/Python/web2py` that ignores everything gotten/created.

frameworks/Python/web2py/app/standard/modules/database.py
abastardi: No, `cacheable=True` simply tells the DAL to return a `Rows` object that *could be* cached. The only reason for setting that to `True` here is that in addition to allowing the results to be cached (with this code is *not* doing), it also speeds things up a bit because it foregoes creation of some additional attributes.

frameworks/Python/web2py/setup.sh
abastardi: This approach is unchanged from the previous version, and the frameworks/Python/web2py/.gitignore file already ignores the created /web2py subdirectory.

Note, web2py is a little different in that it is not pip-installable, and its application folders go inside of the framework parent folder.

frameworks/Python/web2py/app/standard/modules/database.py
knewmanTE: Yeah, seems like you need to actually set the "cache" attribute in order to have the results cached.

"The select method also takes a cache argument, which defaults to None. For caching purposes, it should be set to a tuple where the first element is the cache model (cache.ram, cache.disk, etc.), and the second element is the expiration time in seconds."

http://web2py.com/books/default/chapter/29/06/the-database-abstraction-layer#Caching-selects

abastardi:
Thanks for the quick review and getting this into Round 13.

PR #2264, opened 2016-09-09
Fintrospect: Upgrade to lib version, and tweak JVM parameters

daviddenton:


daviddenton:
This is to be added to Round 13 preview please

frameworks/Scala/fintrospect/src/main/scala/Fortunes.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  private val dbClient = Mysql.client
+    .withCredentials("benchmarkdbuser", "benchmarkdbpass")
+    .withDatabase("hello_world")
+    .configured(Param(low = 0, high = 10, idleTime = fromSeconds(5 * 60), bufferSize = 0, maxWaiters = Int.MaxValue))
+    .withStatsReceiver(NullStatsReceiver)
+    .withMonitor(NullMonitor)
+    .withTracer(NullTracer)
+    .withMaxConcurrentPrepareStatements(256)
+    .newRichClient("localhost:3306")
```
knewmanTE: Could you add a line to your setup.sh file that replaces `localhost` here with the system variable `$DBHOST`? Here's [an example] (https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Scala/akka-http/setup.sh#L5) of another framework doing something similar. We ask that you do this because localhost will work for Travis and for local testing, but on our production runs, we have different IP addresses for the App, Database, and Client servers.

frameworks/Scala/fintrospect/benchmark_config.json
```diff
@@ -1,23 +1,28 @@
...
+        "database_os": "Linux",
+        "setup_file": "setup",
+        "json_url": "/json",
+        "fortune_url": "/fortunes",
+        "plaintext_url": "/plaintext",
+        "port": 9000,
+        "approach": "Realistic",
+        "classification": "Micro",
+        "database": "MySQL",
+        "db_url": "/db",
```
knewmanTE: Should this be here? It doesn't look like you have a DB test implemented. If that's the case, I'd take this line out.

frameworks/Scala/fintrospect/src/main/scala/Fortunes.scala
knewmanTE: Also, your /fortunes test appears to be failing even on Travis and on a single local machine, so perhaps there's something that is misconfigured here?

daviddenton:
@knewmanTE Sorry - my bad. The DB tests aren't ready to be plugged in yet (meant for R14), so I've removed them from the config. If you could merge the rest then that'd be awesome.

:)

PR #2265, opened 2016-09-12
set the DB host for nodejs-postgres

knewmanTE:
nodejs-postgres tests were failing because we weren't changing the database IP address from 127.0.0.1 to ${DBHOST}.


knewmanTE:
closing in favor of a single bulk pull request fixing this type of error

PR #2266, opened 2016-09-12
set the DB host for jooby

knewmanTE:
jooby tests were failing because we weren't changing the database IP address from localhost to ${DBHOST}.


knewmanTE:
closing in favor of a single bulk pull request fixing this type of error

PR #2267, opened 2016-09-13
set DBHOST properly to fix broken database tests

knewmanTE:
I noticed that a few frameworks were assuming that database tests were running against localhost or 127.0.0.1, when in actuality the production environment has a separate machine hosting all of the databases. This fixes the broken database tests in `Java/jooby`, `Java/wicket`, and `Javascript/nodejs` that were a result of this problem.


martin-g:
I have the feeling that we will have official results for Wicket again! :fireworks: 
Thanks @knewmanTE !

knewmanTE:
@martin-g, always happy to be able to get more frameworks passing!

Also Travis fails appear to just be the usual bunch of unrelated tests.

Issue #2235, opened 2016-08-24
Some frameworks are overwriting the mongoDB world table: ffead-cpp, kelp, sabina


knewmanTE:
Before running certain tests test, I am seeing results like this in my local mongoDB database (freshly dropped and rebuilt):

```
{ "_id" : 1, "id" : 1, "randomNumber" : 7229 }
{ "_id" : 2, "id" : 2, "randomNumber" : 5321 }
{ "_id" : 3, "id" : 3, "randomNumber" : 9144 }
{ "_id" : 4, "id" : 4, "randomNumber" : 9358 }
{ "_id" : 5, "id" : 5, "randomNumber" : 4414 }
{ "_id" : 6, "id" : 6, "randomNumber" : 5760 }
{ "_id" : 7, "id" : 7, "randomNumber" : 6196 }
{ "_id" : 8, "id" : 8, "randomNumber" : 5389 }
{ "_id" : 9, "id" : 9, "randomNumber" : 5945 }
{ "_id" : 10, "id" : 10, "randomNumber" : 6208 }
{ "_id" : 11, "id" : 11, "randomNumber" : 7997 }
{ "_id" : 12, "id" : 12, "randomNumber" : 6508 }
{ "_id" : 13, "id" : 13, "randomNumber" : 7238 }
{ "_id" : 14, "id" : 14, "randomNumber" : 834 }
{ "_id" : 15, "id" : 15, "randomNumber" : 153 }
{ "_id" : 16, "id" : 16, "randomNumber" : 9355 }
{ "_id" : 17, "id" : 17, "randomNumber" : 6444 }
{ "_id" : 18, "id" : 18, "randomNumber" : 5860 }
{ "_id" : 19, "id" : 19, "randomNumber" : 974 }
{ "_id" : 20, "id" : 20, "randomNumber" : 596 }
```

After running the tests, I am seeing results like this:
 ```
{ "_id" : 1, "id" : 1, "randomNumber" : 7776 }
{ "_id" : 2, "id" : 2, "randomNumber" : 7637 }
{ "_id" : 3, "randomNumber" : 3636 }
{ "_id" : 4, "randomNumber" : 4423 }
{ "_id" : 5, "id" : 5, "randomNumber" : 9595 }
{ "_id" : 6, "randomNumber" : 5993 }
{ "_id" : 7, "id" : 7, "randomNumber" : 3051 }
{ "_id" : 8, "id" : 8, "randomNumber" : 5884 }
[...]

PR #2042, opened 2016-05-11
Remove sabina

jaguililla:
Delete Sabina benchmark as the project is unmaintained (I'm the framework's developer).

PR #2268, opened 2016-09-13
remove sabina

knewmanTE:
In a recent discussion, the maintainer of sabina requested that we remove the framework (see [here](https://github.com/TechEmpower/FrameworkBenchmarks/issues/2235#issuecomment-244538379))

There was also an issue opened up about this a few months ago (#2042) that didn't end up getting merged in (I think because of confusion about which branch it should target?). If there was a specific reason it wasn't merged in, please let me know!


msmith-techempower:
Do it

PR #2270, opened 2016-09-14
Fix falcore benchmark

### LINKED ISSUE #2269: Go/falcore breaking on concurrency tests
If you start up a Go/falcore server and hit its endpoints, you get the correct results back, so the framework is passing all of the Verify checks. But if you start up a server and run a wrk simulation like this:
```
wrk -H 'Host: localhost' -H 'application/json,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 32 --timeout 8 -t 8 http://localhost:8080/db
```

You get this error: `fatal error: concurrent map read and map write`

The result is that falcore gives results for the first couple levels of concurrency on its first test (JSON at -c 8 and -c 16) but then it shuts down with the concurrency error and all other tests just return `unable to connect to 127.0.0.1:8080 Connection refused`.

If you don't want to manually run wrk, you can also see these results by just doing a full (non-verify) run of falcore, which automatically runs wrk at various levels of concurrency against all of the test endpoints:

```
toolset/run-tests.py --test falcore
```

methane:
fixes #2269

### DIFF
--- a/frameworks/Go/falcore/src/framework_benchmarks/falcore.go
+++ b/frameworks/Go/falcore/src/framework_benchmarks/falcore.go
@@ -148,14 +148,10 @@ var requiredHeaders = falcore.NewResponseFilter(func(req *falcore.Request, res *
 	res.Header.Set("Date", time.Now().Format(time.RFC1123))
 })
 
-var applicationJson = http.Header{"Content-Type": []string{"application/json"}}
-var textPlain = http.Header{"Content-Type": []string{"text/plain"}}
-var textHtml = http.Header{"Content-Type": []string{"text/html"}}
-
 // Test 1: JSON serialization
 var jsonFilter = falcore.NewRequestFilter(func(req *falcore.Request) *http.Response {
 	if req.HttpRequest.URL.Path == "/json" {
-		resp, _ := responder.JSONResponse(req.HttpRequest, 200, applicationJson, &Message{helloWorldString})
+		resp, _ := responder.JSONResponse(req.HttpRequest, 200, nil, &Message{helloWorldString})
 		return resp
 	}
 	return nil
@@ -170,7 +166,7 @@ var dbFilter = falcore.NewRequestFilter(func(req *falcore.Request) *http.Respons
 			log.Fatalf("Error scanning world row: %s", err.Error())
 		}
 
-		resp, _ := responder.JSONResponse(req.HttpRequest, 200, applicationJson, &world)
+		resp, _ := responder.JSONResponse(req.HttpRequest, 200, nil, &world)
 		return resp
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


knewmanTE:
@methane thanks for the quick turnaround on the fix! Out of curiosity, how did removing the hardcoded content types fix the issue?

methane:
http.Header is typedef of map.  Some handlers returns same map instance.

Then, following filter adds headers to the http.Header which http handlers returns.
```
var requiredHeaders = falcore.NewResponseFilter(func(req *falcore.Request, res *http.Response) {
 	res.Header.Set("Server", "falcore")
  	res.Header.Set("Date", time.Now().Format(time.RFC1123))
  })
```

knewmanTE:
Ahh, makes sense. Yeah, I figured there was something that was returning the same instance across threads.

Did some test runs locally and it seems to be running okay for me! the only thing I noticed was that, when I run the plaintext test with the pipeline.lua script like it is in the suite:

```
 wrk -H 'Host: localhost' -H 'text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 256 --timeout 8 -t 4 http://10.0.0.1:8080/plaintext -s ~/pipeline.lua -- 16
```

I get repeated output that looks like this:
```
2016/09/14 11:17:57 [EROR] 46156 ERROR writing response: <*net.OpError write tcp 10.0.0.1:8080->10.0.0.3:48382: write: broken pipe>
```

Plaintext is still returning with responses, but perhaps its performance is less than it should be. In cases where I get that error, wrk is showing very high latency. I think this might be a result of the pipeline.lua script, which, as I understand it, generates ~16x more requests. Generally, max latency increases the more connections that are added, so the pipeline.lua script may just be exacerbating the problem. For comparison, these are my latency results with and without the lua script:

<b>without pipeline.lua</b>
<table>
<tr>
<th>concurrency</th>
<th>max latency</th>
<tr>
<td>8</td>
<td>29.89ms</td>
</tr>
<tr>
<td>256</td>
<td>80.40ms</td>
</tr>
<tr>
<td>1024</td>
<td>272.20ms</td>
</tr>
<tr>
<td>4096</td>
<td>885.84ms</td>
</tr>
</table>
<b>with pipeline.lua</b>
<table>
<tr>
<th>concurrency</th>
<th>max latency</th>
<tr>
<td>8</td>
<td>28.21ms</td>
</tr>
<tr>
<td>256</td>
<td>437.88ms</td>
</tr>
<tr>
<td>1024</td>
<td>1.44s</td>
</tr>
<tr>
<td>4096</td>
<td>6.49s</td>
</tr>
</table>

Anyway, those are just my observations in my local testing. If you're satisfied with the falcore results, I'm happy to merge this in!

methane:
The error may happens while wrk is stopping. (wrk may not wait all response of pipelined request.)
And I think the latency is because of overload.

If falcore result is very worse than other Go frameworks, I'll investigate.  Go ahead.

PR #2271, opened 2016-09-14
fasthttp: fixed setup scripts

valyala:
Hot fix for fasthttp setup scripts.
Fasthttp setup in round 13 is broken without this fix.


knewmanTE:
Looks good to me!

PR #2272, opened 2016-09-14
Fix web2py compile and race condition bugs

abastardi:
Ensure web2py applications are compiled with the same version of Python used to execute the tests. Also, update web2py to a more recent commit from "master" in order to take advantage of a fix for a possible framework-level race condition (which occurred once during a test run of the benchmarks).

Also includes some clarifying comments in the code.

### DIFF
--- a/frameworks/Python/web2py/setup.sh
+++ b/frameworks/Python/web2py/setup.sh
@@ -7,6 +7,9 @@ pip install --install-option="--prefix=${PY2_ROOT}" -r $TROOT/requirements.txt
 rm -fr web2py
 git clone --recursive --branch master https://github.com/web2py/web2py.git
 cd web2py
+# The following commit is checked out in order to take advantage of a particular
+# bug fix that was made after the latest stable release. This can be changed
+# in future rounds once a new stable release has been made.
 git checkout 326335c3cd027a97b9c2b83d880b2b1f8f62321d
 cd ..
 cp -r app/standard/ web2py/applications/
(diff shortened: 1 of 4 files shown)


frameworks/Python/web2py/setup.sh
```diff
@@ -7,13 +7,16 @@ pip install --install-option="--prefix=${PY2_ROOT}" -r $TROOT/requirements.txt
...
+# in future rounds once a new stable release has been made.
+git checkout 44362aa42948dca08ddf1b2f86ac03c69b0d48e4
 cd ..
 cp -r app/standard/ web2py/applications/
 cp -r app/optimized/ web2py/applications/
 cp app/wsgi.py web2py/
 cp app/routes.py web2py/
 touch web2py/__init__.py
-python compile_apps.py
+$PY2_ROOT/bin/python compile_apps.py
```
msmith-techempower: This line, while correct, should be unnecessary. The `fw_depends python2` above executes [this file](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/languages/python2.sh) which ensures that `$PY2_ROOT/bin` is the last entry on the `PATH`, which **should** mean that `python` is functionally equivalent to `$PY2_ROOT/bin/python` under these conditions.

If this is not the case, then we have a bigger bug at play that needs to be addressed.

Specifically, I expect this change to have no effect.

abastardi: Good point. I suppose that line isn't really necessary. I guess I was just thinking better safe (and explicit) than sorry. Do you want to revert that change or just leave it as is?

msmith-techempower: I think that it would be better to revert.

My only reason for this is to maintain the same simple code across the python codebases for when contributors pick one at random to reference.

abastardi: OK, done. Thanks.

PR #2274, opened 2016-09-15
45216 continuousbenchmarking 20160609 asb 2

ashawnbandy-te-tfb:
The work in this PR introduces a basic facility for continuous benchmarking.  run-continuously is essentially a while-loop that makes 'life-cycle' calls in addition to initiating a benchmarking run.  Additionally, an example of an upstart process configuration is included.

Below are some relevant notes from a recent commit:

    toolset/run-continuously.sh is simply a while loop that
    removes and rebuilds the framework benchmark suite and
    then runs the benchmarks.
   
    There are five environment variables to be set:
   
    * TFB_REPOPARENT      the absolute path to the folder containing
      the repository
    * TFB_REPONAME          the name of the repository folder
    * TFB_REPOURI              the URI of the repository
    * TFB_REPOBRANCH    the branch of the repository to use
    * TFB_LOGSFOLDER    the folder to archive log files
   
    There are four life cycle stages in run-continuously:
   
    * tear-down-environment.sh  called upon to remove the existing
      environment
    * rebuild-environment.sh    called upon to rebuild the environment
    * pre-run-tests                      scripts in this folder are called
                                                    before the benchmarks are run
    * post-run-tests                    scripts in this folder are called
                                                    after the benchmarks are run
   
    - Notes -
   
    run-continuously.sh generally assumes that a clone of the
    appropriate repo and branch exist and have the list
    scripts available and an appropriate copy of benchmark.cfg
    in place.  Some effort is made to support starting states
    that differ but those are not intended to be the general
    case for use.

At the end of each run (post-run-tests) there are two python scripts.  One zips the results.json file and sends it to a specified email address.  The other makes copies of the logs for each frameworks (each are independently zipped while being copied).

TODO:

    Tagging a commit for milestones.
    Testing - particularly on the environment this will eventually run on.
    Generating test metadata and sending it out.  (I need some sense of what is appropriate here.)


config/upstart.example/tfb.conf
```diff
@@ -0,0 +1,18 @@
+# /etc/init/tfb.conf
+env TFB_REPOPARENT="/private"
+env TFB_REPONAME="FrameworkBenchmarks"
+env TFB_REPOURI="git@github.com:ashawnbandy-te-tfb/FrameworkBenchmarks.git"
```
knewmanTE: Should this be the master repo URI? Or would you rather restrict it to your repo until we have all of the kinks ironed out?

toolset/lifecycle/post-run-tests/email-results.py
```diff
@@ -0,0 +1,70 @@
...
+zipFile.close()
+temporaryFile.seek(0)
+#
+#Set up the email to be sent to the mailing list
+#
+message = MIMEMultipart()
+message['From'] = os.environ['TFB_MAILINGLIST']
+message['To'] = os.environ['TFB_MAILINGLIST']
+message['Date'] = formatdate(localtime=True)
+message['Subject'] = subprocess.check_output(["git", "describe", "--always"])
```
knewmanTE: Is there a reason we seem to have this whole block twice? (once here and once up on line 22?

config/upstart.example/tfb.conf
knewmanTE: I guess you can ignore this comment. After actually setting up continuous benchmarking on Server Central I realized that we need to be using your repo until this is merged into master on the base repo.

knewmanTE: HOWEVER, my actually valuable comment on this chain is that TFB_REPOURI should probably be the https URL rather than the ssh URL. In my local testing, I was getting public key permission errors trying to clone down from your fork.

knewmanTE:
FYI, I have a branch [here](https://github.com/knewmanTE/FrameworkBenchmarks/tree/continuous-benchmarking-add-commit-tag) that creates a file `.commit`, which contains the most recent commit of the checked out branch, adds it to the results directory, and attaches it to the email. I'm unable to open up a pull request against your fork, perhaps because you haven't been established as a FrameworkBenchmarks contributor yet?

I've done some testing on ServerCentral and was able to get continuous benchmarking working pretty easily, so that's cool. Good work, Shawn!

As for the test metadata, I have a larger pull request open at #2283 which automatically generates the test metadata in `results/[timestamp]/test_metadata.json`. If we want to get continuous benchmarking running before we merge in my pull request, then we might want a workaround. Another major change my pull request makes is that it removes the `latest` directory entirely, so it's probably worth thinking of a way to access the `results/[timestamp]/` directory instead of `results/latest/` when fetching the `results.json` file.

config/upstart.example/tfb.conf
```diff
@@ -0,0 +1,18 @@
...
+env TFB_MAILINGLIST="sbandy@techempower.com"
```
knewmanTE: This needs to be split into 2 variables, one for the "To" field and one for the "From" field. Otherwise, if you try to send the results to multiple people, you'll end up with multiple email addresses in the "From" field. I don't know how email servers generally handle this, but it seems to me that, for the sake of email conventions, we should only be sending the email from a single address.

ashawnbandy-te-tfb:
Closing.  Will re-open under another pull request.

PR #2275, opened 2016-09-15
remove Scala/plain framework

knewmanTE:
Scala/plain consistently fails on all environments (local, Travis, and production) and appears to be unmaintained (I can't even find any documentation for it online), so I don't think it should still be in the repo.


nbrady-techempower:
Any searches for it bring our repo up before anything else remotely relevant. :+1:

PR #2277, opened 2016-09-16
Jruby removal

nbrady-techempower:
In preparation for some changes I have coming up, and due to jRuby having major issues prior to switching from Java 7 to Java 8 (Now it's completely borked) this PR removes jRuby. It's been out of date and broken for some time. I will probably revisit it in round-14 with updated versions.

Also, Ruby with travis will be fixed for round-14.


msmith-techempower:
:+1:

nbrady-techempower:
I'm going to merge this. The current implementation of ruby for travis doesn't work because of an issue with writing over the previous installation of the rvm. I've solved that in my new setup.

PR #2278, opened 2016-09-17
round-13 local postgresql fix

nbrady-techempower:
This PR fixes a problem where the postgresql service was being killed during the database installation step before the "Fortunes" table was fully populated. It also removes precise support and other fixes that now make it travis compatible for the upcoming round-14 travis update.

### DIFF
--- a/toolset/setup/linux/database.sh
+++ b/toolset/setup/linux/database.sh
@@ -95,37 +95,36 @@ rm create.sql
 
 ##############################
 # Postgres
+# Version: 9.*
 ##############################
-echo "Setting up Postgres database"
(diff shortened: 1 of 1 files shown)


nbrady-techempower:
@knewmanTE The reason it was working on one machine and not the other was just variability in how fast the table populated on that machine. When reproducing the problem the "Fortunes" table would have between 0 and 4 rows on my vm.

nbrady-techempower:
fyi this is a travis run of nodejs from my work. This includes the postgres changes made here as well as travis installing client and database via the scripts we use. It's still a work in progress for a few things but postgresql is fixed: https://travis-ci.org/nbrady-techempower/FrameworkBenchmarks/jobs/160604384

PR #2279, opened 2016-09-19
Rails fixes

nbrady-techempower:
This PR fixes a problem with rails after the postgres permission changes. The benchmarkdbuser does not have access to do inserts, while activerecord's batch updating uses sql insert statements behind the scenes.

Also, rbx 2.4 was not working properly on both local and my travis instance. Switching to the 2.5 rbx setup that was already included in the project fixes this issue.


nbrady-techempower:
Applying work in progress to this atm so I can apply changes to the others

nbrady-techempower:
More updates including adding the Gemfile.lock files to the repo once the framework was in a running state. A few frameworks required updating to ruby 2.2.2 because of activesupport 5.0 which may not have been in the framework originally, but since there was no gemfile snapshot, it's hard to tell.

nbrady-techempower:
@knewmanTE This is ready to go. Passes locally, passes on my local travis runs but it won't pass here. I got all rbx working except for padrino, so I removed that one as well.

knewmanTE:
Wooooooo looks good to me!

PR #2283, opened 2016-09-20
Reorganize Results Directories

knewmanTE:
<b>Note: my text editor trims the whitespace at the ends of some lines, which makes it seem like there are more changed lines than there are. Sorry about that!</b>

---

Our current results directory structure looks like this:

```
results/
--latest/
----results.json
----logs/
------nodejs/
--------out.txt
--------verification/
----------json.txt
----------plaintext.txt
----------etc.
--------benchmark/
----------json.txt
----------plaintext.txt
----------etc.
--20160920161349/
----results.json
------json/
--------nodejs/
----------raw
----------stats
----------stats.json
------plaintext/
--------nodejs/
----------raw
----------stats
----------stats.json
--20160920175232/
----test_metadata.json
```

There are a number of things that I don't like about this:

1) The `latest` directory is confusing
2) It's redundant to have `results.json` twice
3) The organization of [test-type]/[test-name] (e.g. json/nodejs, json/go, db/undertow) seems a bit counter-intuitive, since more often than not you're looking for results by framework, not by test type (e.g. nodejs/json, undertow/db)
4) Having certain information split up between the `latest` and the `20160920161349` directory seems a bit pointless
5) We currently have the toolset built such that, if you try to run a test but still have old results in your `latest` directory, the suite will output those results rather than doing a fresh run. While I understand the original intent of that, I think it's an unneeded tool and is more often frustrating than it is helpful.
6) The `test_metadata.json` file has to be generated manually and therefore ends up in a separate timestamp directory from the actual test results

So! With this pull request, I am proposing the following new directory structure:

```
results
--20160920161349/
----results.json
----test_metadata.json
----logs/
------nodejs/
--------out.txt
--------json/
----------benchmark.txt
----------raw
----------stats
----------stats.json
----------verification.txt
--------plaintext/
----------benchmark.txt
----------raw
----------stats
----------stats.json
----------verification.txt
------undertow/
--------etc.
```

With this change, you have a central hub for each test's results located at:
`results/20160920161349/logs/[test-name]/`
(e.g. `results/20160920161349/logs/nodejs/`)
And all of the relevant information for each individual test can be found at:
`results/20160920161349/logs/[test-name]/[test-type]/`
(e.g. `results/20160920161349/logs/nodejs/json/`)

This PR also automatically generates the `test_metadata.json` file with reach run, so everything ends up neatly packaged in the same directory.


msmith-techempower:
I don't have a problem with the restructure as long as it doesn't break anything.

Also, your pull request exemplifies why auto-formatters are bad.

knewmanTE:
@msmith-techempower yeah, I've been using Atom recently instead of Sublime. It has some nice features, but stuff like this and a few other weird quirks it has really get on my nerves.

msmith-techempower:
Merged into master

knewmanTE:
@msmith-techempower @ashawnbandy-te-tfb just double checking, but I don't think this actually ended up in master.

Issue #2284, opened 2016-09-21
Error with aspnetcore service index


knewmanTE:
The aspnetcore test is currently failing with the following error:
```
Setup aspnetcore: Welcome to .NET Core!
Setup aspnetcore: ---------------------
Setup aspnetcore: Learn more about .NET Core @ https://aka.ms/dotnet-docs. Use dotnet --help to 
see available commands or go to https://aka.ms/dotnet-cli-docs.
Setup aspnetcore: Telemetry
Setup aspnetcore: --------------
Setup aspnetcore: The .NET Core tools collect usage data in order to improve your experience. The 
data is anonymous and does not include commandline arguments. The data is collected by Microsoft 
and shared with the community.
Setup aspnetcore: You can opt out of telemetry by setting a DOTNET_CLI_TELEMETRY_OPTOUT 
environment variable to 1 using your favorite shell.
Setup aspnetcore: You can read more about .NET Core tools telemetry @ https://aka.ms/dotnet-
cli-telemetry.
Setup aspnetcore: Configuring...
Setup aspnetcore: -------------------
Setup aspnetcore: A command is running to initially populate your local package cache, to improve 
restore speed and enable offline access. This command will take up to a minute to complete and will 
only happen once.
Setup aspnetcore: Decompressing 100% 4074 ms
Setup aspnetcore: Expanding 100% 29116 ms
Setup aspnetcore: log  : Restoring packages for /home/techempower/FrameworkBenchmarks
/frameworks/CSharp/aspnetcore/Benchmarks/project.json...
Setup aspnetcore: error: Unable to load the service index for source https://dotnet.myget.org/F/aspnet1
/api/v3/index.json.
Setup aspnetcore: error:   Response status code does not indicate success: 404 (Feed does not exist.).
aspnetcore: setup.sh and framework processes have terminated
Setup aspnetcore: Setup has terminated
Setup aspnetcore: Status: Poll: 1, Port 8080 bound: False, Time Left: 1:37:31.604706
Setup aspnetcore: setup.sh process exited naturally with 1
INFO:root:Executed setup.sh, returning 1
verify: 100% |########################################################################| Rough Time: 00:07:33
==================================================================
  Verification Summary

| Test: aspnetcore
|      NO RESULTS (Did framework launch?)
==================================================================
```

Of note appear to be these lines:
```
Setup aspnetcore: error: Unable to load the service index for source https://dotnet.myget.org/F/aspnet1
[...]

PR #2285, opened 2016-09-21
update nuget feeds

nathana1:
Resolves https://github.com/TechEmpower/FrameworkBenchmarks/issues/2284.


knewmanTE:
Awesome, thanks for the quick fix @nathana1!

PR #2288, opened 2016-09-25
ffead-cpp - Added support for mysql and postgresql along-with nginx webserver module

sumeetchhetri:
Added support for nginx webserver
Added mysql tests
Added postgresql tests
Fixed issue in the benchmarker.py file to stop apache webserver (as it runs with root privileges and cannot be stopped with the killall -s 9 -u testrunner command)


nbrady-techempower:
great catch @sumeetchhetri! I'm wondering if there's a better way we can go about this, making sure that all services that should be stopped are stopped. Maybe one location for something like that. I'm working on cleaning up the toolset right now, so I'm going to merge this in and think about that. Thanks for all your work!

PR #2293, opened 2016-09-26
Flavors to none

ashawnbandy-te-tfb:
In cases where there's no flavor distinction to be made within a langauge, change the label to 'None'.


knewmanTE:
ping @msmith-techempower

PR #2298, opened 2016-10-06
45216 continuousbenchmarking nuketmp 20161006 5

ashawnbandy-te-tfb:
As it stands currently, many frameworks are writing to /tmp. So many, in fact, that in our testing environment the pool of inodes are exhausted which causes tests needing to write anything to fail. This commit introduces work to remove files that are generated by a test after that test is done. Note: there is no attempt to remove all files from /tmp (i.e. sudo is not used, nor is the search recursive).

### DIFF
--- a/toolset/benchmark/benchmarker.py
+++ b/toolset/benchmark/benchmarker.py
@@ -650,6 +650,19 @@ class Benchmarker:
         time.sleep(5)
 
         ##########################################################
+        # Remove contents of  /tmp folder
+        ##########################################################
+        try:
+          filelist = [ f for f in os.listdir("/tmp") ]
+	  for f in filelist:
+            try:
+	      os.remove("/tmp/" + f)
+            except OSError as err:
+              print "Failed to remove " + str(f) + " from /tmp directory: " + str(err)
+        except OSError:
+          print "Failed to remove contents of /tmp directory."
+
+        ##########################################################
         # Save results thus far into the latest results directory
         ##########################################################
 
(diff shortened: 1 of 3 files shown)


knewmanTE:
Could this cause any problems for users who aren't running the testing suite in an isolated VM?

knewmanTE:
@ashawnbandy-te-tfb document telling users that they should be running in an isolated environment, or telling users that it will only be deleting tmp files owned by testrunner?

ashawnbandy-te-tfb:
I would say probably a caution that one should not use one's regular user to run tests in an environment where erasing files out of /tmp would be a problem.  The worst case should not be all that bad: /tmp is meant to be temporary in nature.

toolset/benchmark/benchmarker.py
```diff
@@ -650,6 +650,19 @@ def exit_with_code(code):
         time.sleep(5)
 
         ##########################################################
+        # Remove contents of  /tmp folder
+        ##########################################################
+        try:
+          filelist = [ f for f in os.listdir("/tmp") ]
+	  for f in filelist:
```
LadyMozzarella: I just wanted to let you know that there are a couple of tab characters mixed in here. There's one at the beginning of the line here and two lines down (`os.remove("/tmp/ + f")`).

ashawnbandy-te-tfb: Good catch!  Will correct.

ashawnbandy-te-tfb:
@knewmanTE suggested adding a command line switch to turn on this feature (i.e. by default no files are removed).  I will add that and re-open this PR when it is complete.

knewmanTE:
@ashawnbandy-te-tfb you can also just add "WIP" to the title of the task to indicate that it is a work in progress!

PR #2304, opened 2016-10-15
add s-server v1.1 to Round 15

woshilaiceshide:
s-server is an nio socket server / http server /websocket server written in scala.

add it to tfb


knewmanTE:
Thanks for the pull request @woshilaiceshide! I have a couple of quick notes: 
1) it looks like we already have an s-server v2 test in our suite. Is your intention to have both v1 and v2 tests in the suite, or is it sufficient to have the most recent version being tested?
2) Any substantial changes/new frameworks can actually be opened up against round-14. Our round-15 branch is actually quite out of date and not representative of the current state of the benchmark suite.

knewmanTE:
ping @woshilaiceshide

nbrady-techempower:
@woshilaiceshide Along with @knewmanTE request, this needs to be rebased. 
You will also need to add `mysql` as a `fw_depends` dependency in the setup file.
Please open a new pull request with those changes if you want to be included in the next round! Thanks for the contributions :)

PR #2311, opened 2016-10-21
PERF: django/hello/views.py: use generators

westurner:
- With generators, there's no need to build large intermediate lists.
- Note: I haven't tested these two changes. The performance impact may be negligible.

### DIFF
--- a/frameworks/Python/django/hello/world/views.py
+++ b/frameworks/Python/django/hello/world/views.py
@@ -57,7 +57,7 @@ def dbs(request):
   # by creating dicts, we don't need to user the model serializer, which is probably slow and only appropriate
   # for complicated serializations of joins and crazy query sets etc
   # test xrange vs range if the query number is gigantic
-  worlds = uj_dumps([{'id' : r, 'randomNumber' : g(id=r).randomnumber} for r in [rp() for q in xrange(queries)]])
+  worlds = uj_dumps([{'id' : r, 'randomNumber' : g(id=r).randomnumber} for r in (rp() for q in xrange(queries))])
   return HttpResponse(worlds, content_type="application/json")
 
 def fortunes(request):
@@ -75,7 +75,7 @@ def update(request):
   rp = partial(random.randint, 1, 10000)
 
   worlds = []
-  for r in [rp() for q in xrange(queries)]:
+  for r in (rp() for q in xrange(queries)):
     w = g(id=r)
     w.randomnumber=rp()
     w.save()


methane:
generator comprehension is slower than list comprehension in common case.
In case of the list is large, generator comprehension wins.

In this case, using generator comprehension doesn't make sense, because queries
must not so large.

```
$ python3.5 -m timeit 'sum([i for i in range(100)])'
100000 loops, best of 3: 5.53 usec per loop

$ python3.5 -m timeit 'sum((i for i in range(100)))'
100000 loops, best of 3: 7.52 usec per loop
```

westurner:
Good point.

RAM utilization might be a useful benchmark to add.

methane:
This code generates list only up to 20.
Generator can't reduce RAM usage significantly when size =20.

ashawnbandy-te-tfb:
Hi @westurner - can we close this PR and re-open against the round 14 branch?  We're very near the end of the round 13 cycle.

westurner:
Sure

knewmanTE:
@westurner and @methane, I did some local testing and was seeing roughly equivalent results with and without this change. Looking at your discussion, do you think it's worth trying to merge this in, or should we just leave Django as-is?

westurner:
- PyPy may make a significant difference in req/s with Django and with
other frameworks
- These benchmarks do not account for RAM utilization (GC is at a
disadvantage here; assuming no programmer error in non-GC languages)
- As example code, demonstrating use of generators may be worth clicking
merge (even if there is not a significant performance difference)

A few Python performance resources:

- http://pandas.pydata.org/pandas-docs/stable/enhancingperf.html
- http://scikit-learn.org/stable/developers/performance.html
- https://wiki.python.org/moin/PythonSpeed/PerformanceTips
- https://github.com/python/performance

methane:
> - PyPy may make a significant difference in req/s with Django and with
> other frameworks

I use PyPy for some frameworks.  If you like to use for Django, please make pull request.

> - These benchmarks do not account for RAM utilization (GC is at a
> disadvantage here; assuming no programmer error in non-GC languages)

This pull request doesn't make significant RAM usage change, because "queries"
is small enough.

> - As example code, demonstrating use of generators may be worth clicking
> merge (even if there is not a significant performance difference)

I don't care. (-0)

frameworks/Python/django/hello/world/views.py
```diff
@@ -75,7 +75,7 @@ def update(request):
   rp = partial(random.randint, 1, 10000)
 
   worlds = []
-  for r in [rp() for q in xrange(queries)]:
+  for r in (rp() for q in xrange(queries)):
     w = g(id=r)
```
methane:
```
for _ in xrange(queries):
   r = rp()
  w = g(id=r)
...
```

msmith-techempower:
Is this PR ready to be merged?

westurner:
Probably ready to close because it's not advantageous for req/s performance.

knewmanTE:
Closing this PR for now. If you decide you'd like it re-opened and merged, just let us know. 

@westurner @methane

PR #2312, opened 2016-10-21
Remove claim vagrant from apt-get is too old

raboof:
On recent versions of Ubuntu (Xenial, Yakkety) and Debian (Testing, Unstable) vagrant 1.8.x is available which AFAICT should be fine

### DIFF
--- a/deployment/vagrant-aws/README.md
+++ b/deployment/vagrant-aws/README.md
@@ -17,8 +17,7 @@ production setup costs about TODO.
 
 ## Prerequisites
 
-* **A recent version of Vagrant**, like 1.6.3 (NOTE: `apt-get` is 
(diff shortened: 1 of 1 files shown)


ashawnbandy-te-tfb:
@raboof Do you know which version of Ubuntu has a version of vagrant that is too old to be useable?  It would be to provide that information, but I agree, if later versions of ubuntu are fine, we should probably drop the blanket warning.

raboof:
The message suggests 1.6.3 is recent enough, http://packages.ubuntu.com/search?keywords=vagrant&searchon=names&suite=all&section=all suggest 'precise' and 'trusty' still have pre-1.6.3 versions, but xenial (16.04LTS) is OK.

PR #2319, opened 2016-10-27
akka-http use hikaricp connection pool

fedgehog:


knewmanTE:
LGTM! Did some local testing and was seeing results that were as good or better on this branch than on master.

PR #2320, opened 2016-10-27
Cleanup metadata

msmith-techempower:
METADATA IS LIFE; METADATA IS LOVE!


ashawnbandy-te-tfb:
METADATA GO!

martin-g:
This PR also removed some non-config files for undertow-edge.
Was this intentional?

msmith-techempower:
Yes, it was our understanding that `undertow-edge` was an old undertow build that could be removed.

zhong-j-yu:
@msmith-techempower - in commit [Cleaned up a lot of stuff](https://github.com/TechEmpower/FrameworkBenchmarks/pull/2320/commits/ef8ba5d2742a2ad223efb329e878bc4ed1f7836d), you changed `bayou` framework from "Realistic" to "Stripped". Can you give a source of the reason for the change? Thanks.

msmith-techempower:
I believe `bayou` was changed to stripped because the code does not appear to exemplify a realistic approach to writing a webapp. Specifically, there does not appear to be routing outside of a naked switch statement and all of the business logic for the test implementations are done inline.

I realize that this is a hazy explanation of the `realistic` vs. `stripped` classification, but it's basically an eye-ball for our contributors and it appeared to be stripped in this case.

zhong-j-yu:
Thank you for the reply. I'm quite surprised by the reasoning, because several frameworks that I'm familiar of are doing the same thing and they are not categorized as 'stripped'

https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/netty/src/main/java/hello/HelloServerHandler.java

https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/vertx/src/main/java/vertx/WebServer.java

https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/JavaScript/nodejs/routing.js

(note that in the case of `nodejs`, it's a hashmap created in app code, which is no different from `switch`)

I'm sure there are more frameworks doing the similar thing. My understanding is that this is a tests of HTTP frameworks, i.e. frameworks providing abstractions over HTTP semantics. Some frameworks provide just that, while leaving higher functions, like request routing, view rendering, to others. And some applications need exactly a bare-minimum HTTP engine and prefer to handle the rest by themselves.

In any case, if the categorization of 'stripped' is done by arbitrary contributors, without giving a notification to the framework, let alone a simple explanation, it is likely to be abused. I wonder why my framework is singled out and not many other similar ones.

PR #2328, opened 2016-11-02
add octopus lua benchmark

cyberz-eu:

### DIFF
--- a/frameworks/Lua/octopus/config.lua
+++ b/frameworks/Lua/octopus/config.lua
@@ -34,5 +34,6 @@ return {
 		sourceCtxPath = "",
 		requireSecurity = false,
 		sessionTimeout = 3600,
+		usePreparedStatement = false,
 	},
 }
--- a/frameworks/Lua/octopus/config.lua
+++ b/frameworks/Lua/octopus/config.lua
@@ -21,7 +21,7 @@ return {
 	
 	databaseConnection = {
 		rdbms       =   "mysql",
-		host        =   "127.0.0.1",
+		host        =   "DBHOSTNAME",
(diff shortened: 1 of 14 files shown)


cyberz-eu:
I faced problems connecting to postgres. It looked like there was service at 127.0.0.1:5432 but it did not respond to my authentication correctly, so I switched to mysql and everything went well.

frameworks/Lua/octopus/config.lua
```diff
@@ -0,0 +1,39 @@
...
+	errorLog = "error_log logs/error.log;",
+	accessLog = "access_log logs/access.log;",
+	includeDrop = [[#include drop.conf;]],
+	maxBodySize = "50k",
+	minifyJavaScript = false,
+	minifyCommand = [[java -jar ../yuicompressor-2.4.8.jar %s -o %s]],
+	
+	databaseConnection = {
+		rdbms       =   "mysql",
+		host        =   "127.0.0.1",
```
knewmanTE: Could you add a line to your setup.sh file that replaces this with the value in `$DBHOST`? On our actual test systems, we have 3 separate machines for the app server, the database, and the client load-generator, so we can't count on them all being `127.0.0.1`.

Here's an [example](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Lua/lapis/setup.sh) of what I mean. You can probably just use that sed code and just drop in the appropriate file (config.lua).

knewmanTE:
Thanks for the PR @cyberz-eu! I left you a couple quick notes for things to fix up. Once those are addressed, this LGTM!

PR #2332, opened 2016-11-04
Fix Nginx as the front end tests by putting uwsgi socket in /var/tmp …

dantti:
Most (maybe all) frameworks using Nginx and uwsgi are failing due /tmp being privately visible to uwsgi, the "proper" fix would be to put it in /run but it isn't writable by testuser so I'm putting it in /vat/tmp
bottle and flask are two I saw that need this kind of fix

### DIFF
--- a/frameworks/C++/cutelyst/config/config_socket.ini
+++ b/frameworks/C++/cutelyst/config/config_socket.ini
@@ -3,9 +3,10 @@ master
 ; Increase listen queue used for nginx connecting to uWSGI. This matches
 ; net.ipv4.tcp_max_syn_backlog and net.core.somaxconn.
 ; for performance
+listen = 65535
 disable-logging
 ; use UNIX sockets instead of TCP loopback for performance
-socket = /tmp/uwsgi.sock
+socket = /var/tmp/uwsgi.sock
 ; allow nginx to access the UNIX socket
 chmod-socket = 666
 ; Avoid thundering herd problem http://uwsgi-docs.readthedocs.org/en/latest/articles/SerializingAccept.html .
--- a/frameworks/C++/cutelyst/nginx.conf
+++ b/frameworks/C++/cutelyst/nginx.conf
@@ -41,7 +41,7 @@ http {
         server_name  localhost;
 
         location / {
-            uwsgi_pass unix:/tmp/uwsgi.sock;
+            uwsgi_pass unix:/var/tmp/uwsgi.sock;
             include /usr/local/nginx/conf/uwsgi_params;
         }
     }
--- a/frameworks/C++/cutelyst/setup.sh
+++ b/frameworks/C++/cutelyst/setup.sh
@@ -8,14 +8,12 @@ sed -i 's|SendDate=.*|SendDate=false|g' config/config.ini
 cd $IROOT
 mkdir cutelyst-benchmarks || true
 cd cutelyst-benchmarks
-rm -rf *
 
 QT_VERSION_MM=56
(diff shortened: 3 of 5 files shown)


methane:
Python's uwsgi benches are suffered same issue.

I didn't understand what cause this issue.
Systemd has PrivateTmp feature. But we start nginx without systemd, am I wrong?

If PrivateTmp cause this, `/var/tmp` has same issue.
How about making `${FWROOT}/run` directory?

dantti:
@methane I really dunno how exactly PrivateTmp works, but what I could see is that this is the exactly behavior happening in these runs, once I changed the directory nginx worked fine, I also needed to increase uwsgi listen backlog othewise there are some of resource unavailable

msmith-techempower:
> failing due /tmp being privately visible

Wat. Are you saying that the permissions on `/tmp` need to be relaxed so the suite can write to it?

ping @ashawnbandy-te-tfb

ashawnbandy-te-tfb:
@methane @dantti I'm not sure that I am following.  Will opening up the permissions on /tmp to make it writable by all negate the need for a workaround?  If, so I just made the change in our testing environment that should be included in the next preview run which is coming up shortly.

dantti:
@aschneider-techempower the nginx log lists that it coult not connect to the unix socket, giving no such file or directory, so in fact I'm not sure if it's just a permissions issue. It looks like there is some different feature in use here, that gets better described by "temporary namespaced directory" which is a feature distros like Fedora has by default so that one application doesn't see other app's /tmp. This improves security as lots of CVE are due insecure /tmp writes...

msmith-techempower:
This FEELS like a permissions issue on the `/tmp` dir and nothing more.

dantti:
For the record I wrote this workaround for Cutelyst due:
http://stackoverflow.com/questions/32974204/got-no-such-file-or-directory-error-while-configuring-nginx-and-uwsgi
Which has a link to another SO page containing this other link:
http://fedoraproject.org/wiki/Features/ServicesPrivateTmp
Which feels like this issue, because if it was a permission issue the error would be permission denied:

> 2014/02/17 03:42:20 [crit] 5455#0: *1 connect() to unix:/tmp/uwsgi.sock failed (13: Permission denied) while connecting to upstream, client: xx.xx.x.xxx, server: localhost, request: "GET /users HTTP/1.1", upstream: "uwsgi://unix:/tmp/uwsgi.sock:", host: "EC2.amazonaws.com"

msmith-techempower:
OOOOOOOOOOOOOOOOOOOOOOOOOOKAY

I get it now.

@ashawnbandy-te-tfb We MAY need to look deeper into this... may be the cause of some of our handle-problems.

methane:
@ashawnbandy-te-tfb @msmith-techempower Should I update Python frameworks too?

msmith-techempower:
@methane I think we need to do an audit across **any** framework that uses nginx... scary.

ashawnbandy-te-tfb:
@methane @msmith-techempower I agree.  At a minumum, we should start with uwsgi and nginx.

dantti:
@aschneider-techempower @msmith-techempower I was checking the new preview run to see how Cutelyst threaded was performing now that when it segfaults (as master if freeze for the fix) the --master workaround will restart it and found the following line:
  Error: Could not empty /tmp
Isn't your scripts cleaning /tmp at the wrong time so when nginx runs the socket isn't there?
Because when --mode test runs it works...

msmith-techempower:
That is a known issue, but should not impact your run.

PR #2336, opened 2016-11-06
Upgrade CPython + PyPy2 and add PyPy3

GMLudo:
Hi,

This PR upgrades CPython + PyPy2 and add PyPy3, because PyPy3 becomes useable to launch benchmarks.

Have a nice day.

### DIFF
--- a/toolset/setup/linux/languages/pypy.sh
+++ b/toolset/setup/linux/languages/pypy.sh
@@ -6,11 +6,11 @@ RETCODE=$(fw_exists ${IROOT}/pypy.installed)
   return 0; }
   
 PYPY_ROOT=$IROOT/pypy
-PYPY_VERSION=5.0.1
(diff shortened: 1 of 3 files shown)


toolset/setup/linux/languages/pypy.sh
```diff
@@ -6,11 +6,11 @@ RETCODE=$(fw_exists ${IROOT}/pypy.installed)
   return 0; }
```
msmith-techempower: This is a pedantic thing, but could we change this file (and every instance of `fw_depends` that points to `pypy` to be `pypy2` for clarity?

nbrady-techempower:
Hi @GMLudo . I'm going to close this due to inactivity. We're going to start previews for the next round after the new year. If you want this to be included, please address the issues above in a new pull request targeting master. Thanks for the contribution!

PR #2341, opened 2016-11-08
Travis work (WIP) (RND 14)

nbrady-techempower:
**Disclaimer: There's a lot going on here in the diff because it hasn't been rebased yet. As per some internal discussion, just showing some work with travis.**

The purpose of the travis changes, are to get travis to work the same way you would bootstrap the suite locally. Travis provides a lot of additional features and pre-installed services. Because of this, we can often encounter errors on travis that don't happen in local test environments and vice-versa. This strips travis down to some bare minimums, and then installs the suite the way you would here: http://frameworkbenchmarks.readthedocs.io/en/latest/Development/Installation-Guide/ and uses the same install scripts we would use locally instead of separate scripts for travis. This also means we will be able to test changes to those install scripts with travis.

The core changes can be found in the `travis.yml` file and the `toolset/travis` dir

A few todos left:
- [ ] Figure out a cleaner way to test only the frameworks who's files have changed.
- [ ] Rebase and catch-up with any toolset changes


knewmanTE:
Exciting stuff!!!

nbrady-techempower:
Closing this for now for rebasing / cleanup

PR #2342, opened 2016-11-08
Origin/with permeagility update to 0.8.0 framework version

glennirwin:


knewmanTE:
Hey @glennirwin, we're unfortunately no longer taking Round 13 pull requests as we prepare for our final release. Would you mind re-opening this request against round-14? Now that we've solidified our testing environments and are moving towards continuous benchmarking, we're shooting to have a much shorter wait before Round 14 than there was between rounds 12 and 13.

PR #2344, opened 2016-11-10
Round14 update

glennirwin:
Update for PermeAgility version 0.8.0


knewmanTE:
@glennirwin I'm fine with the changes you've made in this PR, but going forward Permeagility will need to either access OrientDB on the database machine or disable its DB tests, as [we discussed](https://github.com/TechEmpower/FrameworkBenchmarks/issues/2346) previously.

I believe we're planning on rolling out a new (and hopefully simpler) way of handling database setup soon, so I can ping you once that is implemented if you'd like.

knewmanTE:
Permeagility is also failing the first time I run it and succeeding on subsequent runs, which I think has to do with it not instantiating the OrientDB database in time.

glennirwin:
ok, thanks.  Yes, please ping me once the db changes are implemented.

Thanks for the heads up on the failure, is there any way I can increase the
startup timeout?

Thanks,
Glenn.

knewmanTE:
@glennirwin, currently, the timeout is set globally for all frameworks in the `benchmark.cfg` file (see the [sleep](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/benchmark.cfg.example#L26) property). But, in the long run, the OrientDB setup will take place in the `fw_depends` section of the setup file, which will wait until the whole process finishes before starting up the framework and running the test.

nbrady-techempower:
Continued work on this will have to be rebased

nbrady-techempower:
Hi @glennirwin ! I'm going to close this PR. If you'd like to address the issues above, our new setup for databases works a little differently. If you want to make an orientdb setup, you can look how it's being done in `toolset/setup/linux/database` and make an orientdb.sh that can then be added to the framework's setup file with `fw_depends`. If you want this to be included in the next round, which we will start preview runs for right after the new year, please open a new pull request with the necessary changes targeting the master branch. Hope to see it in there and thanks for your contributions! :)

PR #2345, opened 2016-11-10
Fix cutelyst threaded tests

dantti:
The preview runs raised a segfault when running
cutelyst in threaded mode, an unsafe use of atribution
of a static object, the fix got included in the 1.0.0 release

### DIFF
--- a/frameworks/C++/cutelyst/setup_thread.sh
+++ b/frameworks/C++/cutelyst/setup_thread.sh
@@ -17,6 +17,5 @@ cmake $TROOT -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$IROOT
 make -j $MAX_THREADS
 
 export LD_LIBRARY_PATH=/opt/qt${QT_VERSION_MM}/lib:${IROOT}/lib/x86_64-linux-gnu/
-export CUTELYST_CONFIG=${TROOT}/config/config.ini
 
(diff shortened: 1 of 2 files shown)


knewmanTE:
Hey @dantti, we're unfortunately no longer taking Round 13 pull requests as we prepare for our final release. Would you mind re-opening this request against round-14? Now that we've solidified our testing environments and are moving towards continuous benchmarking, we're shooting to have a much shorter wait before Round 14 than there was between rounds 12 and 13.

dantti:
@knewmanTE I thought you were still accepting bug fixes (as it's not a new tests), as I got a fix for nginx merged last days.
Cutelyst 0.13.3 to 1.0.0 has only a bug fixes, without this cutelyst-thread benchmaks will always crash

knewmanTE:
@dantti Brian Hauer made a [post in the TFB Google Group](https://groups.google.com/forum/#!topic/framework-benchmarks/hsIThZR8Lo4) last week announcing that November 8th would be the last day we are accepting pull requests for Round 13. Sorry for any confusion! It's a bit harder to post announcements and keep people up-to-date on GitHub than it is on the Google Group.

dantti:
I saw that mail, but thought it was still ahead :( with too few preview runs and actually even different environment between runs (nginx was fine on 1st) IMO a 3rd preview would be best... anyway will reopen against the other one

knewmanTE:
@dantti as I mentioned, we've got some really cool continuous benchmarking tools in the works that should smooth out the process of posting official rounds, so your fix will still show up on the sooner side of things!

PR #2347, opened 2016-11-11
Fix cutelyst threaded tests

dantti:
The preview runs raised a segfault when running
cutelyst in threaded mode, an unsafe use of atribution
of a static object, the fix got included in the 1.0.0 release


knewmanTE:
LGTM!

PR #2351, opened 2016-11-12
Update round-14 with latest from master

knewmanTE:
There have been a number of bug fixes and framework enhancements on master that should be on round-14.


ashawnbandy-te-tfb:
This seems reasonable and desirable to me.

PR #2352, opened 2016-11-12
[Round-14] Upgrade openresty to latest version(1.11.2.1) 

halfcrazy:
Using latest Openresty and compile with -O3 parameter.


knewmanTE:
Thanks for being so accommodating @halfcrazy! This LGTM!

PR #2353, opened 2016-11-12
python: Use /var/tmp instead of /tmp for unix socket.

methane:
ref #2332


msmith-techempower:
Do we need to check that `/var/tmp` exists prior to a run (I'm not sure if this exists on our ubuntu environments, basically)?

methane:
`/var/tmp` is standard directory like `/var/run` or `/tmp`.
So normal Linux system should have it.

msmith-techempower:
Looks good to me; LGTM!

PR #2327, opened 2016-11-01
[Java/Servlet] web.xml deployment descriptor changes

zloster:
Modify `servlet-mapping` tags in `web.xml` to use `url-pattern` not something proprietary as `url-regexp`.
Modify `web.xml` to conform to WebApp 2.4 XML schema.

PR #2354, opened 2016-11-12
[Java/Servlet] Modify servlet-mapping tags in web.xml to use url-pattern not somethi…

zloster:
…ng propriatory as url-regexp.

Modify web.xml to comform to WebApp 2.4 XML schema

This pull request is replacing #2327

### DIFF
--- a/frameworks/Java/servlet/benchmark_config.json
+++ b/frameworks/Java/servlet/benchmark_config.json
@@ -45,7 +45,7 @@
       "setup_file": "setup",
       "db_url": "/servlet/postgres",
       "query_url": "/servlet/postgres?queries=",
-      "fortune_url": "/servlet/postgre-fortunes",
+      "fortune_url": "/servlet/postgres-fortunes",
       "update_url": "/servlet/postgres-update?queries=",
       "port": 8080,
       "approach": "Realistic",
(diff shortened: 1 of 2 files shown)


zloster:
There is a problem with Resin and the deployment descriptor. I'm trying to understand what's going on.

zloster:
Now everything works as expected on a local test run.

knewmanTE:
@zloster, the Postgres Fortunes test is failing. It looks like there's a typo in the [benchmark_config.json](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/servlet/benchmark_config.json#L50) that was also in the old [web.xml](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/servlet/src/main/webapp/WEB-INF/web.xml#L39), but now that you've updated web.xml, the test is failing. If you fix the typo in `benchmark_config.json`, that should solve the problem.

zloster:
@knewmanTE Thank you for your help.

Updated and this time I've run the correct test entry locally (`dropwizard-postgres`). Embarrassingly previously I've run only the `dropwizard` test which only calls the `MySQL` endpoints.

PR #2355, opened 2016-11-12
Round 14: Tweaking of Fintrospect implementation and JVM params

daviddenton:


knewmanTE:
This looks good to me! Sorry for the inconvenience, and thanks for being so understanding about the branch changes!

daviddenton:
No problem. Keep up the great work :)

martin-g:
It seems this PR removed the impl for Fortunes test.
Was this intentional? Neither the title nor the description mention this.

daviddenton:
It was intentional. The fortunes test didn't run anyway, so I thought I'd tidy it up and take another run at it later. :)

martin-g:
Why Scala devs prefer to remove the implementation instead of fixing/improving it ?
Same with Play and Akka-HTTP/Spray (https://twitter.com/rolandkuhn/status/704620217104736256).
This doesn't leave good impression!

daviddenton:
I can't comment on other cases, but i think it's better to not have broken code sitting around that might be mistaken for working code, especially when said code is meant to be an example for other devs to follow.

I'll look at reimplementing it in due course.

PR #2343, opened 2016-11-09
H2O: Optimize the implementation further

volyrique:
Here's another set of optimizations based on the round 13 preview data. I hope I have made it before the deadline.

PR #2356, opened 2016-11-12
H2O: Optimize the implementation further

volyrique:
Reopen #2343 against round-14, and make some additions.

### DIFF
--- a/frameworks/C/h2o/src/main.c
+++ b/frameworks/C/h2o/src/main.c
@@ -24,6 +24,7 @@
 #include <netdb.h>
 #include <signal.h>
 #include <stdarg.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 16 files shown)


frameworks/C/h2o/src/main.c
```diff
@@ -177,9 +183,7 @@ static int initialize_global_data(const config_t *config, global_data_t *global_
 	sigset_t signals;
 
 	memset(global_data, 0, sizeof(*global_data));
-	global_data->config = config;
 	global_data->memory_alignment = get_maximum_cache_line_size();
```
knewmanTE: Just to confirm, are you actually caching data, or is your use of "cache line size" an aspect of memory management within h2o?

volyrique: Cache line size relates to CPU caches, not HTTP request, database or any other higher-level caching mechanism, and yes, it is used as a memory management parameter.

PR #2357, opened 2016-11-13
Update dropwizard to 1.0.3 and Java 8 compilation

zloster:


jplock:
I wonder if the performance would improve on the postgres-related tests if we tried using https://github.com/impossibl/pgjdbc-ng and https://github.com/brettwooldridge/HikariCP

zloster:
@jplock I've played today a little with https://github.com/impossibl/pgjdbc-ng. Here are [some results](https://github.com/zloster/logs/tree/master/FrameworkBenchmarks/dropwizard-postgresql-driver) with one test run. Basically it is beaten quite bad from https://github.com/pgjdbc/pgjdbc. The test is on the standard development Vagrant box from the FrameworkBenchmark running on a VirtualBox VM. I was following closely the development of postgre-jdbc the last 18 months and there are massive improvements in performance in comparison of say 9.4-1204. FrameworkBenchmark code is using postgres jdbc 9.4-1208 in the moment.
 
About HikariCP: Is there a `dropwizard module` with `HikariCP` instead of `Tomcat jdbc pool`? I'm not sure if there will be some measurable difference when tested on a local VM - I have to adjust the concurrency quite low not to overload the CPU for the VM. The database pool is tuned for a real DB server with a lot of CPU cores. Max connections are set to 256.

jplock:
@zloster ok that's cool.  I personally haven't tried either but I know they seem new-ish.

knewmanTE:
This looks to me if you're happy merging it in, @zloster.

zloster:
@knewmanTE It's a good framework. Actually I enjoy reading the source code - I've learned a lot from it. Also it is one of the frameworks that needed some love.

PR #2362, opened 2016-11-18
Disable kernel polling

josevalim:
Because the pollset is single core, on benchmarks that are mostly
waiting on the pollset (almost no CPU work), it may reduce throughput.

Therefore we are disabling the pollset and see how it will affect
future results.

### DIFF
--- a/frameworks/Elixir/phoenix/setup.sh
+++ b/frameworks/Elixir/phoenix/setup.sh
@@ -12,4 +12,4 @@ mix local.rebar --force
 mix deps.get --force --only prod
 mix compile --force
 
-elixir --erl "+K true" --detached -S mix phoenix.server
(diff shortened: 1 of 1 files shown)


knewmanTE:
Thanks for the PR @josevalim! I did some local testing of Phoenix on master and Phoenix on your patch-3 branch and wasn't seeing a huge difference in results, but this change doesn't seem to make things worse, and I'm using pretty low-power VMs, so it's possible that your changes will have a larger impact on our actual production machines.

josevalim:
Supposedly the change only matters if you are having a lot of IO contention
so it should only matter for larger machines indeed. It should not affect
benchmarks that are CPU bound as well.

knewmanTE:
Cool, I look forward to seeing how it performs on future production runs then!

PR #2369, opened 2016-11-21
Cutelyst: Split setup files for each DB

dantti:
Split setup files for each DB and avoid connecting to a db that will not be used
this saves resources and also makes it easy to test without any DB

### DIFF
--- a/frameworks/C++/cutelyst/benchmark_config.json
+++ b/frameworks/C++/cutelyst/benchmark_config.json
@@ -2,7 +2,7 @@
     "framework": "cutelyst",
     "tests": [{
             "default": {
-                "setup_file": "setup",
+                "setup_file": "setup_pf",
                 "json_url": "/json",
                 "plaintext_url": "/plaintext",
                 "port": 8080,
@@ -21,7 +21,7 @@
                 "versus": ""
             },
             "postgres-raw": {
-                "setup_file": "setup",
+                "setup_file": "setup_pf_pg",
                 "db_url": "/db_postgres",
                 "query_url": "/query_postgres?queries=",
                 "update_url": "/updates_postgres?queries=",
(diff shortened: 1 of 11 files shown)


knewmanTE:
@dantti, it looks like a lot of the scripts are pretty similar. From what I can tell, other than the sed script differences, they all pretty much contain the following:
```
cd $IROOT
mkdir cutelyst-benchmarks || true
cd cutelyst-benchmarks

QT_VERSION_MM=56
export CMAKE_PREFIX_PATH=/opt/qt${QT_VERSION_MM}:${IROOT}

cmake $TROOT -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$IROOT

make -j $MAX_THREADS

nginx -c $TROOT/nginx.conf

export LD_LIBRARY_PATH=/opt/qt${QT_VERSION_MM}/lib:${IROOT}/lib/x86_64-linux-gnu/
uwsgi --ini ${TROOT}/config/config_socket.ini --cutelyst-app ${IROOT}/cutelyst-benchmarks/src/libcutelyst_benchmarks.so -p $MAX_THREADS &
```

Is it possible to refactor some of these lines into a single file and then call `source` on that file? That way, if you or someone else decides to make a change to nginx or another aspect of the framework's startup, they will only have to change it in one place.

dantti:
@knewmanTE sure, I was just afraid of source foo would work or would I need source $IROOT/foo. 

in fact I'd preffer to have a singe .sh and have the json file to support arguments.

knewmanTE:
@dantti hmm, you'll probably want to source $TROOT/foo, as $TROOT gets set to the root directory for the current framework.I don't believe our benchmark config JSON file currently supports that type of configuring (its mostly used by the python scripts that run the suite), but you could accomplish something similar by creating separate bash files that each set the arguments/environment variables that you need for a given test and then use a single master bash file that contains all of the business logic and plugs in the appropriate variables when necessary.

nbrady-techempower:
hi @dantti ! I'm going to close this due to inactivity and because we've made some changes to the setup files. Now you will need to add the db as an `fw_depends` dependency. So either: `mysql postgresql mongodb`. If you'd like to get this included in the next round, which we should be starting previews for just after the new year, please include the necessary changes in a new pull request targeting master. As always, thanks for your contributions! :)

dantti:
@nbrady-techempower where the fw_depends should be? in the setup script?

PR #2371, opened 2016-11-22
Revenj.NET simplified. DSL compiler upgraded. Latest Revenj.Java

zapov:
Switch to latest Revenj.NET (1.4.1)

Use simpler response types (instead of Stream) in .NET
Move use of Random to ThreadLocal


knewmanTE:
This looks good to me!

PR #2378, opened 2016-11-25
weppy updates

gi0baro:
This includes updates for python weppy framework:
- align uwsgi configuration to the other ones
- updated weppy to 0.8.4


knewmanTE:
Looks good to me!

PR #2379, opened 2016-11-28
Increase wrk threads count from 8 to 16.

valyala:
This improves /plaintext and /json benchmark results
for top performers on a test machine with 32 CPU cores:

- Plaintext raised from 6.7M to 8.8M requests per second.
- Json raised from 880K to 980K requests per second.


knewmanTE:
@valyala, the values in the benchmark.cfg file are there for you to test with depending on the specs of your development environment. On our production machines, we have tweaked the file to be optimized for each environment, so changing the example values won't affect the performance for the results site at all.

valyala:
@knewmanTE , it would be great to see benchmark.cfg files for each production environment in the repository.
Could you try increased `wrk` threads count (`-t` argument) on a test machine with 40 CPU cores in the next round?

msmith-techempower:
We run `wrk` with `-t` equal to the number of CPU cores available on the machine in our benchmarks already.

PR #2382, opened 2016-11-29
Round 14: Fintrospect updates including re-adding missing benchmark tests

daviddenton:
... plus a performance tweak for populating the Date header.


knewmanTE:
@daviddenton thanks for the PR! I think there might be a bit of confusion about how the database tests are run. From what I can tell, you appear to be starting up a docker image with a copy of the MySQL database running in it. All of the tests should be able to run standalone with the existing instance of MySQL that is created during the suite setup process (e.g. the "Setting up the database server" script found [here](http://frameworkbenchmarks.readthedocs.io/en/latest/Development/Installation-Guide/)).

Although both Travis tests and local tests run everything on a single machine, our production environments actually run on 3 machines: 1 for running the framework server, 1 for hosting the databases, and 1 for generating the client requests with wrk. 

There is also a $DBHOST system variable that gets set based on your benchmark_config file. Please add a line to your setup.sh file that updates [your localhost](https://github.com/TechEmpower/FrameworkBenchmarks/pull/2382/files#diff-efe37ddde55ec469cc5a1e2c81144733R18) to the $DBHOST value. You can see an example of this [here](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Scala/akka-http/setup.sh#L5).

Thanks again for the PR, and I'm looking forward to getting this merged in once you make the fixes to the database!

daviddenton:
yep - sorry. that was my bad. I was using a Docker container to test the functionality. 
Thanks for the catch! :)

PR #2361, opened 2016-11-17
Attempting to fix Java installation in Travis-CI

msmith-techempower:
In Travis-CI, the Java installation was hanging on waiting for user input; this **should** avoid that problem.

PR #2384, opened 2016-11-30
Clean up apt get installs in prerequisites.sh

knewmanTE:
This PR is in response to an idea that @msmith-techempower proposed [here](https://github.com/TechEmpower/FrameworkBenchmarks/pull/2361#pullrequestreview-9496462). With this change, if there is a dependency that is specified with `fw_depends XXX` but does not exist in the `toolset/setup/linux/` directory, the suite will attempt to install it with `apt-get`.

### DIFF
--- a/toolset/setup/linux/bash_functions.sh
+++ b/toolset/setup/linux/bash_functions.sh
@@ -131,8 +131,8 @@ fw_depends() {
       set -x
       . $FWROOT/toolset/setup/linux/frameworks/${depend}.sh
     else
-      echo WARN: No installer found for $depend, installing with 'apt-get'
-      sudo apt-get install -o Dpkg::Options::="--force-confold" --force-yes -s ${depend}
+      echo WARN: No installer found for $depend, attempting to install with 'apt-get'...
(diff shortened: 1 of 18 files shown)


toolset/setup/linux/bash_functions.sh
```diff
@@ -132,7 +131,8 @@ fw_depends() {
       set -x
       . $FWROOT/toolset/setup/linux/frameworks/${depend}.sh
     else
-      echo WARN: No installer found for $depend
+      echo WARN: No installer found for $depend, attempting to install with 'apt-get'...
+      sudo apt-get install -o Dpkg::Options::="--force-confold" --force-yes ${depend}
```
msmith-techempower: There's another `Dpkg::Options` flag we should use in addition to the one here: `Dpkg::Options::="--force-confdef"`

msmith-techempower:
I love everything about this; ideally, `prerequisites.sh` would only be those dependencies on which the suite relies and nothing else.

toolset/setup/linux/bash_functions.sh
knewmanTE: When I tested with a fake package (e.g. `fw_depends abcde`), it would fail as expected.

PR #2385, opened 2016-11-30
upgrade akka http stable version

sxend:
The latest and stable version of akka-http has been released so we upgrade it

http://akka.io/news/2016/11/22/akka-http-10.0.0-released.html


knewmanTE:
Looks good to me!

PR #2386, opened 2016-11-30
Round 14: Fintrospect updates including re-adding missing benchmark tests 

daviddenton:
Including fixes for using DBHOST instead of hardcoding to localhost


knewmanTE:
Hmmm, it still appears to be failing on the database tests when I run the fintrospect locally. Plaintext and JSON are passing, so the problem must be with how it's connecting to the database.

daviddenton:
Darn it! I'll take another look. Guess I'll be spinning up a vagrant in my near future..

Thanks @knewmanTE for your efforts :)

daviddenton:
Weird... when I run it locally from a clean install, it completes ok (although there are warnings because fintrospect includes the charset in the content type responses). @knewmanTE I've added a clean step to the setup file just in case - could you please run it again and see if it works now?

```verify: 100% |#############################################################################################| Rough Time: 00:02:13
  Verification Summary

| Test: fintrospect
|       fortune     : PASS
|       plaintext   : WARN
|       db          : WARN
|       update      : WARN
|       json        : WARN
|       query       : WARN
================================================================================```

knewmanTE:
Alright, so what I'm seeing now is that the DB tests pass when I use a local environment that is a single machine (i.e. the app server, db server, and client-load generation are all on a single VM), but when the 3 servers are on separate machines with their own IP addresses, it isn't able to reach the DB.

Often this is an issue with the DBHOST, but it appears to be loading the value properly. I'll try poking around a bit more to see if I can find anything.

knewmanTE:
@daviddenton I'm not very familiar with Scala or Fintrospect. Is there some simple debugging code that I could throw in that might help us figure out where the DB connection is getting messed up? Something in `Database.scala` or `DatabaseRoutes.scala` maybe?

martin-g:
`println("=============DBHOST: " + theNameOfTheVariable)`

daviddenton:
Sure. If you add:

`println(dbHost)`

... after line 28 of FintrospectBenchmarkServer.scala, you'll see the value it's using for the DBHOST (which is loaded from the environment). Since the code defaults to using localhost when it can't find the env variable, I assume that this is the problem, and you'll see "localhost" or equivalent in the output.

Could it be there is a problem with the DBHOST env variable not being available to the JVM? We can try exporting it to another variable and loading that env variable instead on line 28 of FintrospectBenchmarkServer.

Thanks again for your help in getting this sorted :)

knewmanTE:
@daviddenton happy to help! There's a fun kind of mystery to trying to debug a framework that I'm unfamiliar with.

Also, as soon as I threw in a debug line, it started working on my testing machines. Perhaps it wasn't properly updating my `target/` directory, so it was running against an older build of the framework.

daviddenton:
I had that earlier, which was why I put the `sbt clean` line into the script, which should clean the target dir. You can tell that it's booting an old version because it sysouts  `Finagle v6.37.0` instead of the newer `Finagle v6.40.0` when you run the server.

knewmanTE:
@daviddenton Travis is currently broken, but I am satisfied having tested this locally.

PR #2388, opened 2016-12-01
Rapidoid: fixed the pipeline limits, added Postgres Fortunes test.

nmihajlovski:
This PR fixes the HTTP pipelining limits that caused the terrible performance of Rapidoid in Round 13.
Also, a Fortunes test is added for Postgres, and Rapidoid's latest release is used.


knewmanTE:
Travis is currently experiencing some issues that are causing all of our tests to fail, but I ran this locally and it looks good to me!

nmihajlovski:
Thanks for the quick response!

PR #2390, opened 2016-12-02
python: Update meinheld

methane:
This update contains important performance improvements
when using with gunicorn.


knewmanTE:
Our Travis is currently broken, but I tested locally and this looks good to me!

PR #2392, opened 2016-12-02
Round-14 (Fintrospect): Fix interpretation of empty "queries" parameter in db tests

daviddenton:
An empty "queries" param now defaults to a value of 1 instead of returning a 400 Bad Request.

…nstead of returning a 400 (Bad request)


knewmanTE:
LGTM. Travis is still a bit broken, but I tested locally and this seems fine to me.

PR #2396, opened 2016-12-03
Phoenix is a Fullstack framework - and its overhead should be compare…

orestis:
…d to Cowboy.

I believe that Phoenix should not be classified as Micro, but should be instead classified as Fullstack, since:

> Fullstack, meaning a framework that provides wide feature coverage including server-side templates, database connectivity, form processing, and so on.

* server-side templates: check, using EEx
* database-connectivity: check, using Ecto & Phoenix-Ecto
* form-processing: check

It also provides session tracking, CRSF protection, WebSocket support etc etc so by no means it should be consider a "Micro" framework.


knewmanTE:
Seems reasonable to me!

PR #2397, opened 2016-12-04
Update Sails app  (this is currently a WIP)

mikermcneil:
### Update

After further study, it looks like there are a few other things we need to update in the Sails benchmark (esp. setting up clustering so that the metrics are relevant vs. other frameworks in the list).   So rather than peppering y'all with individual pull requests, @irlnathan, @sgress454 and I will just put everything into this one PR so you can review it all at once.  I'll post a comment and edit the title of this PR when it's ready.  Thanks again!

>### Original post (about NODE_ENV)
>
>Hey guys, apologies if this is taken care of elsewhere, but I noticed it wasn't in here so wanted to raise the question.  For Sails.js to perform as expected in production, it's important to set the NODE_ENV environment variable accordingly.  See http://sailsjs.com/documentation/concepts/deployment#?set-the-nodeenv-environment-variable-to-production for more information.
>
>Thanks!
>
>P.S. I opened a [separate PR](https://github.com/TechEmpower/FrameworkBenchmarks/pull/2398) for making the same change in the Express benchmark, where it is [also expected in production](https://expressjs.com/en/advanced/best-practice-performance.html#set-nodeenv-to-production).

PR #2398, opened 2016-12-04
Start express in production mode

mikermcneil:
This is a follow-on to [my original PR where I set up the same thing for the Sails.js benchmark](https://github.com/TechEmpower/FrameworkBenchmarks/pull/2397)-- I just happened to notice the same thing was true for Express, so I wanted to make sure and track this somewhere.

For Express to perform as expected in production, it's important to set the NODE_ENV environment variable.  See https://expressjs.com/en/advanced/best-practice-performance.html#set-nodeenv-to-production for more information.

Thanks!

### DIFF
--- a/frameworks/JavaScript/express/setup.sh
+++ b/frameworks/JavaScript/express/setup.sh
@@ -8,4 +8,4 @@ sed -i 's|localhost|'"${DBHOST}"'|g' app.js
 # install dependencies
 npm install
 # run app
-node app &
+NODE_ENV=production node app &


knewmanTE:
@mikermcneil could you update your repo with the latest from master? We recently merged in a fix that should get Travis working again so we can verify your test.

mikermcneil:
@knewmanTE thanks!  And sorry for the slow reply.  We're just now getting time to come back to this, so @sgress454 will follow up with another PR with a few other tweaks some time in the next couple of weeks.

knewmanTE:
@mikermcneil no worries! Thanks again for your contributions to TFB!

PR #2399, opened 2016-12-06
Use Revenj routes instead of .NET ones

zapov:
Revenj route can specify synchronous response.

### DIFF
--- a/frameworks/CSharp/revenj/Revenj.Bench/RestService.cs
+++ b/frameworks/CSharp/revenj/Revenj.Bench/RestService.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
-using System.ServiceModel.Web;
 using System.Text;
 using System.Threading;
 using FrameworkBench;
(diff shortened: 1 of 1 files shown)


knewmanTE:
@zapov could you update your repo with the latest from master? We recently merged in a fix that should get Travis working again so we can verify your test.

zapov:
I did.
btw. you've changed metadata so revenj.net is now called revenj, while revenj.jvm is still called revenj.jvm
Also, I think Mono missing from Platform is confusing people thinking that it's probably running on .NET core

knewmanTE:
@zapov thanks! I'm not too familiar with the Platform/Flavor distinction, but perhaps @msmith-techempower can shed some light on it.

Sorry about the test names. Would you mind opening a PR to change it back? Also, could you update both of the test names in that PR to be lowercase so they better match the style of the other tests?

zapov:
Well, I'm reluctant to change names at this point since this would mean results would drop from previous round.

knewmanTE:
Hmm... @msmith-techempower @bhauer thoughts?

PR #2389, opened 2016-12-01
temporary fix for travis breaking on MySQL 

knewmanTE:
A temporary workaround to get Travis working again. Likely related to [this](https://github.com/travis-ci/travis-ci/issues/6961)

PR #2402, opened 2016-12-07
Fix #2389 apt error for MySQL in Travis

ashawley:
According to the Travis documentation the way to depend on mysql is

    services:
      - mysql

https://docs.travis-ci.com/user/database-setup/#MySQL

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -210,12 +210,12 @@ services:
   - postgresql
   - redis-server
   - mongodb
+  - mysql
 
 addons:
   postgresql: "9.3"
   apt:
     packages:
-      - mysql-server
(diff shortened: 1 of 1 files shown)


msmith-techempower:
If Travis-CI gets working because of this, LGTM.

Note: we're going to shake up the Travis-CI setup **a lot** in a bit.

ashawley:
The apt command succeeded in the few builds that have already been kicked off in the matrix so far.

As #2389 by @knewmanTE showed, there's no problem with running the tests against MySQL 5.6, so this should get Travis working again.

knewmanTE:
This LGTM.

PR #2405, opened 2016-12-11
Rust Iron updates

scottmmjackson:
Wrote Rust Iron benchmarks for cases 2, 3, 4, and 5 utilizing raw Postgres as ORM and rust-mustache as templating engine.

Switched from rustc-serialize to serde which is approximately 3x faster at serializing JSON.

Also bumped rust version to 1.13.0, as the previous version didn't allow postgres to compile against it.

Rust is still a really young language, but I'm quite impressed at how it's come along so far.


scottmmjackson:
Hmm, Travis seems to be failing randomly. It passed the Rust/iron job though.

knewmanTE:
@scottmmjackson yeah, when you make a change to anything in `toolset/` it forces all tests to run on Travis, and there are a handful of tests that are currently broken. This looks good to me, save for one small change that I will mark in the code.

frameworks/Rust/iron/src/main.rs
```diff
@@ -1,36 +1,226 @@
...
 
 fn main() {
-    let mut router = Router::new();
-    router.get("/json", json_handler);
-    router.get("/plaintext", plaintext_handler);
-
-    Iron::new(router).http("0.0.0.0:8080").unwrap();
+    let r2d2_config = r2d2::Config::default();
+    let pg_conn_manager = PostgresConnectionManager::new(
+        "postgres://benchmarkdbuser:benchmarkdbpass@localhost/hello_world",
```
knewmanTE: Could you add a line in your `setup.sh` file that replaces "localhost" here with the current value of the environment variable $DBHOST? Rust/Iron will pass on Travis and on your local instance because they are running on a single machine, but our production setups use separate machines for the app server and the database server, so this connection will fail in that case. Here is an [example](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Elixir/phoenix/setup.sh#L5) of what your `setup.sh` line should look like.

knewmanTE:
Actually, one more thing. Make sure to add the endpoints for your additional tests to your `benchmark.config` file. Currently, even though you have the additional endpoints implemented, it is only running the plaintext and json tests because those are the only ones listed in `benchmark.config`.

knewmanTE:
Looks like all the tests are passing on Travis, so once you make the `$DBHOST` change, you should be good to go!

scottmmjackson:
@knewmanTE sounds good. I'm also going to fix a latent deadlock issue that cropped up when testing it myself the other night.

scottmmjackson:
```
let dbhost = match option_env!("DBHOST") {
        Some(it) => it,
        _ => "localhost"
    };
```
This was added as part of the service initialization portion. Basically, it gets DBHOST if it exists, and if it doesn't, it sets the value to localhost. Then it sticks it into the database connection string at the hostname portion- so if it's ultimately some "host:port" or "ip;port" it should work fine.

knewmanTE:
@scottmmjackson great! Yeah, that works just fine. Did a quick test on a 3-machine setup and everything is passing. This LGTM if Travis agrees.

Issue #2204, opened 2016-08-19
RevenJ making bulk requests for Queries and Update tests


knewmanTE:
RevenJ is getting very high results for Queries and Update because it is making bulk requests, rather than making each request individually, as the TFB docs specify ([see number 6](http://frameworkbenchmarks.readthedocs.io/en/latest/Project-Information/Framework-Tests/#multiple-database-queries)). We're currently hiding the RevenJ results from the official listings until this is resolved.

```
public World[] loadWorlds(final int count) throws IOException {
		bulkReader.reset();
		for (int i = 0; i < count; i++) {
			callables[i] = bulkReader.find(World.class, Integer.toString(getRandom10k()));
		}
		bulkReader.execute();
		try {
			for (int i = 0; i < count; i++) {
				buffer[i] = ((Optional<World>) callables[i].call()).get();
			}
		} catch (Exception e) {
			throw new IOException(e);
		}
		return buffer;
	}
```

([Source](https://github.com/TechEmpower/FrameworkBenchmarks/blob/round-14/frameworks/Java/revenj/src/main/java/hello/Context.java#L66))

zapov:
Rule 6 states: "Since this test is designed to exercise multiple queries, each row must be selected individually by a query. It is not acceptable to retrieve all required rows using a SELECT ... WHERE id IN (...) clause."

Let me explain why Revenj is not violating this nor in the spirit, nor in the rules as stated above:
1) Revenj doesn't translate that query into id IN clause, but into multiple subqueries where each subquery is a separate "column"
2) If purpose of that test is to simulate loading various data from various different sources, bulk API used by Revenj will work without any modification; therefore Revenj follows the spirit of the test in doing more DB operations (while knowing how to do it at once)
3) Revenj is doing something no other framework knows how to do. This alone is not good enough reason to dismiss it's behavior. If anything others should learn from it, instead of accusations of breaking the rules
4) Revenj will translate that query into something along the lines of
SELECT ROW( (subquery1), (subquery2),...)
where specific subquery can be arbitrary complex (singe row/multiple rows)...
The resulting response from Postgres is not something you can easily parse, so trying to do a manual implementation of that is futile

[...]

PR #2406, opened 2016-12-12
H2O: Optimize the multiple queries and database updates tests

volyrique:
These changes are mostly about query parallelization. There are more details in the commit descriptions.

### DIFF
--- a/frameworks/C/h2o/src/database.c
+++ b/frameworks/C/h2o/src/database.c
@@ -66,8 +66,10 @@ static const struct {
 } prepared_statement[] = {
 	{FORTUNE_TABLE_NAME, "SELECT * FROM " FORTUNE_TABLE_NAME ";"},
 	{WORLD_TABLE_NAME,
-	 "SELECT * FROM " WORLD_TABLE_NAME " "
-	 "WHERE " WORLD_TABLE_NAME "." ID_FIELD_NAME " = $1::integer;"},
+	 "SELECT * FROM " WORLD_TABLE_NAME " WHERE " ID_FIELD_NAME " = $1::integer;"},
+	{UPDATE_QUERY_NAME,
+	 "UPDATE " WORLD_TABLE_NAME " SET randomNumber = $2::integer "
+	 "WHERE " ID_FIELD_NAME " = $1::integer;"},
 };
 
 static void do_execute_query(db_conn_t *db_conn)
--- a/frameworks/C/h2o/src/database.h
+++ b/frameworks/C/h2o/src/database.h
@@ -22,10 +22,12 @@
 #define DATABASE_H_
 
 #include <h2o.h>
+#include <inttypes.h>
 #include <stdint.h>
 #include <postgresql/libpq-fe.h>
 
 #include "list.h"
+#include "utility.h"
 
 #define DB_REQ_ERROR "too many concurrent database requests\n"
 #define DB_TIMEOUT_ERROR "database timeout\n"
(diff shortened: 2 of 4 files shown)


knewmanTE:
This LGTM. Are you ready to merge it in?

volyrique:
Yes, I am.

knewmanTE:
Excellent! Merging it in then!

volyrique:
Thanks! Are there any plans to do a preview round soon?

methane:
Is parallel query execution allowed?
ref https://github.com/TechEmpower/FrameworkBenchmarks/issues/2204

knewmanTE:
Hmm, that's actually a good question. I may have misinterpreted what exactly h2o was doing in this change. I know that we don't allow batch requests (i.e. each DB request is expected to make a full round-trip to the database server). I'm not sure how that applies to parallel execution. I apologize if I merged this prematurely. 

ping @msmith-techempower @bhauer

volyrique:
I don't see how #2204 applies to this case: The implementation is not making bulk requests of any kind - every row is selected by exactly one query, and there are exactly N roundtrips to the database.

methane:
https://groups.google.com/forum/#!msg/framework-benchmarks/nePDNY9jp-4/EtG6SswmFQAJ

> The use-case this is approximating is an application behavior where you need to read item A and then based on A's value and other logic, you need to read item B and then based on B's value and other logic, you now need to read item C, and so on.  Imagine branching code.

volyrique:
The same message says as well:
> The N iterations are intended to be the equivalent of doing the database work of the Single-query test in its entirety N times, but without the overhead of an additional the HTTP request. [...] The intent has always been to do N round-trips to the database server to fully exercise the connection pool, database driver, ORM (where applicable), and other elements of database connectivity.

This is exactly what the implementation does; in fact, what prompted the change was the fact that the database connection pool was not exercised properly.

methane:
I don't understand full chnage of this pull request.
Couldn't you exercise connection pool properly without parallel query execution?

volyrique:
No, as explained in the first commit description (the current physical hardware environment has 40 cores).

methane:
Uh, I though "exercise" means "simulate real usage", not "saturate the pool".

As I mentioned above, this test simulates sequential N query roundtrip
(one query is depends on result of previous query).

No problem about can't saturate all connections in pool.
If you don't like free conns in pool, you can reduce conns/thread.
All frameworks must follows same rule.

bhauer:
For the multiple-query test, I feel it is acceptable to run the N queries in parallel *as long as* each query is executed with a complete round-trip communication with the database server.  What we want to avoid is tactical optimization for this benchmark that uses either a batch or a single statement to run all N queries, and in so doing, avoids the overhead of N round-trips to the database server.

The test is using a very simple query as a least-complex proxy for more complex queries that have nothing to do with one another and would be run separately in a real-world application—that is, not batched.  So I feel it is acceptable to run them in parallel as long as they are not sharing statements and avoiding the round-trip communication costs.  (Executing them in parallel would, in most cases, actually ensure that they were running on separate connections, so I think it would be exceptional for any sharing to be happening in a parallel implementation approach.)

As for the Updates test type, I would say parallelism is okay here as well, except that each parallel task should run a read then a write.  In other words, the read-then-write work load should be considered an atomic sequence.  It is not acceptable to parallelize N reads and N writes as a unified set of tasks numbering 2*N.

Also, the intent of this test is still to exercise separate statements for each read and write.  In other words, we want 2*N round-trips to the database server in total: N reads and N writes.

I hope that helps clarify, but please let me know if anything is unclear or inconsistent.  Thanks!

volyrique:
Thanks for the explanation! I just have one comment for the updates test:
> In other words, we want 2*N round-trips to the database server in total: N reads and N writes.

Doesn't that contradict the current requirements page, and in particular the statement that batch updates are acceptable? Such approach would imply less than 2*N round-trips.

bhauer:
Ah yes.  Sorry, that was a recent change that I made to the requirements after a conversation with someone else about the same matter.  The *original* intent was 2*N round-trips, but you are correct, the requirements as they are set now allow for batches on the updates (writes) side of the Updates test.

The [requirements](https://www.techempower.com/benchmarks/#section=code) should always trump anything that I say here or on the mailing list.

PR #2407, opened 2016-12-14
Remove testrunner

msmith-techempower:
As the name suggests - this pull request aims at removing `testrunner` as a requirement for the suite.


msmith-techempower:
This task will lead the way toward the database setup improvements.

PR #2414, opened 2016-12-17
Update Hexagon Benchmark

jaguililla:
* Upgrade Hexagon framework version
* Add test over Resin
* Refactor DB access to support MySQL in the near future
* Change start up script memory (probably the cause of the poor results last round :( )


jaguililla:
The only failed job is this: `Java/servlet3-cass` (in both builds). It seems reasonable to think that this has nothing to be with `Hexagon`.

One thing I noticed is that all tests pass in 1 - 2 min (closer to 1), but mine ran for 7' 48" Is this because of my configuration?

Thank you very much for your fabulous benchmark!

knewmanTE:
@jaguililla the time difference you're seeing is because of a script we wrote that checks the commits of a pull request and only executes a Travis run on frameworks that have been modified in those commits. Since your PR only changed Hexagon, that was the only framework that ran a full test.

jaguililla:
Thanks for the explanation! if you need anything from me, just let me know.

knewmanTE:
I'm running into this error when I try running the Hexagon test locally:

```
Setup hexagon: FAILURE: Build failed with an exception.
Setup hexagon: 
Setup hexagon: * What went wrong:
Setup hexagon: Could not create service of type CachingFileSnapshotter using TaskExecutionServices.createFileSnapshotter().
Setup hexagon: > java.io.FileNotFoundException: /home/techempower/FrameworkBenchmarks/frameworks/Kotlin/hexagon/.gradle/3.2.1/taskArtifacts/taskArtifacts.lock (Permission denied)
Setup hexagon: 
Setup hexagon: * Try:
Setup hexagon: Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.
Setup hexagon: 
Setup hexagon: BUILD FAILED
```

I believe the error might be related to the changes that were made to the gradle script, as I'm seeing this error pop up when I manually go through the `setup.sh` file and try  to run `gradle/wrapper`. When I run it with sudo, it starts executing some tests and then I see:
```
85 tests completed, 1 failed, 84 skipped
:test FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':test'.
> There were failing tests. See the report at: file:///home/techempower/FrameworkBenchmarks/frameworks/Kotlin/hexagon/build/site/testng/hexagon/index.html
```

but it isn't clear which test is failing. Also, for what it's worth, we are already executing a few warmup runs of wrk before each test, so I'm not sure how much it will actually help to run your own warmups.

jaguililla:
Done

frameworks/Kotlin/hexagon/src/main/kotlin/co/there4/hexagon/Benchmark.kt
```diff
@@ -1,65 +1,55 @@
...
 
-private val DB_HOST = getenv("DBHOST") ?: "localhost"
-private val DB = setting<String>("database") ?: "hello_world"
-private val WORLD: String = setting<String>("worldCollection") ?: "world"
-private val FORTUNE: String = setting<String>("fortuneCollection") ?: "fortune"
-
-private val database = mongoDatabase("mongodb://$DB_HOST/$DB")
-private val worldRepository = repository(WORLD, World::_id)
-private val fortuneRepository = repository(FORTUNE, Fortune::_id)
+private val fortune = Fortune(0, "Additional fortune added at request time.")
```
jaguililla: Done

jaguililla:
The problem you mention in the last comment is fixed adding write permission to the framework directory (`hexagon`). It seems the user that runs the tests can not write in that directory and that is required by Gradle.

Is it failing for other Gradle modules?

knewmanTE:
@jaguililla ahh, thanks for the info, that helps a lot. We recently made some changes that remove the need for a secondary `testrunner` user, which should alleviate the issue. Could you try rebasing your branch with the latest from master? 

There's also one small change you'll have to make as well. We no longer have a single script that installs every database for every test. Going forward, you will just add the database(s) you are using to the `fw_depends` call at the top of your `setup.sh` file. So, for example, Hexagon would have something like `fw_depends mongodb java`.

jaguililla:
I just completed the changes. Let me know if there is something else. Thanks!

knewmanTE:
@jaguililla that did the trick! It looks like my issues were coming from the `build` and `.gradle` directories still belonging to the testrunner user, when the new setup tools run with the default user. When I deleted those and let the gradle/wrapper create a fresh build, I got PASS on everything.

jaguililla:
Thanks for your support and feedback! and good luck for the round 14

PR #2419, opened 2016-12-22
Remove /config directory and more toolset cleanup

nbrady-techempower:
"A place for everything, and everything in its place."

This PR removes the config directory and moves the files into the proper folders. 
I've updated `fw_depends` to search through `toolset/setup/linux` at any depth for `.sh` files.

What this allows for now is to be able to group files that belong together in subdirectories.

Examples:
```
toolset/
  setup/
    linux/
      databases/
        mongodb/
          -- mongodb.sh (the installation file)
          -- mongodb.conf
          -- create.js

       languages/
         ruby/
           -- ruby-2.0.sh
           -- ruby-2.1.sh
           -- ruby-2.2.sh
```

Not only does this make it easier to find what you're looking for, but the old config directory was also mixing in config files for separate os installations. For the `eventual` addition of full windows support, the structure of `toolset/setup/windows` would look exactly like `toolset/setup/linux`.


nbrady-techempower:
Going to merge this in as it also includes a fix to the apt-get installation in fw_depends that are causing only a few frameworks to fail.

PR #2421, opened 2016-12-23
Hexagon

nbrady-techempower:
Fixed the malformed benchmark_config file so that hexagon isn't being reported as not having a default test.

Tied down `gradle.properties` to a specific commit. Recent pushes to master are breaking tests.

@jaguililla The resin tests are no longer working. Would you be able to explore?

### DIFF
--- a/frameworks/Kotlin/hexagon/benchmark_config.json
+++ b/frameworks/Kotlin/hexagon/benchmark_config.json
@@ -1,60 +1,51 @@
 {
-    "framework" : "hexagon",
-    "tests" : [
-        {
-            "default" : {
-                "json_url" : "/json",
-                "db_url" : "/db",
-                "query_url" : "/query?queries=",
-                "fortune_url" : "/fortunes",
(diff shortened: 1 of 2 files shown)


knewmanTE:
@nbrady-techempower are any Resin tests working since the database setup changes?

nbrady-techempower:
@knewmanTE Yeah, tapestry and wicket are both using resin and they're passing.

knewmanTE:
While you're updating the Hexagon metadata, could you change the webserver on the Resin test from "None" to " Resin" ?

jaguililla:
I didn't see the config error, neither the problem with the Gradle scripts. Thanks for the update!

I'll check the error with Resin. It is working on my machine, so I guess there should be something with my `setup_resin.sh` script.

Anyway, that would be next week. I wish you a merry christmas, enjoy it with your family and friends :)

nbrady-techempower:
Closing this in favor of the framework owner's pr

PR #2422, opened 2016-12-27
Apply fixes for Resing test as stated in #2421

jaguililla:


nbrady-techempower:
Hi @jaguililla! I closed my PR in favor of yours here since it does the same thing, however this only fixes the default test. The resin test is still failing.

LadyMozzarella:
Merging this to fix the default test. Note that the resin implementation is still failing and will need some attention.

jaguililla:
I made everything just as Tapestry (only one test, add write permission to project files, deploy the war with 'hexagon.war' instead 'ROOT.war') and it did not work.

The Resin logs looks good also.

I also downloaded the WAR from the Vagrant development machine and run it in a local Resin and it worked.

Could it be because it is in the `Kotlin` folder instead the `Java` folder?

I need some help here, I'm not able to see the problem.

nbrady-techempower:
@jaguililla Could you open an issue for the problems you're having? For starters, are you trying to hit port 9090? The benchmark_config for the resin test is 8080. If you don't come up with anything, we can carry on in an issue.

PR #2424, opened 2016-12-27
Hapi fix

nbrady-techempower:
The default test for hapi was also trying to spin up redis. We currently do not have a working redis setup.

### DIFF
--- a/frameworks/JavaScript/hapi/benchmark_config.json
+++ b/frameworks/JavaScript/hapi/benchmark_config.json
@@ -2,27 +2,9 @@
   "framework": "hapi",
   "tests": [{
     "default": {
-      "setup_file": "setup",
+      "setup_file": "setup-mongodb",
       "json_url": "/json",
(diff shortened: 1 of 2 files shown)


LadyMozzarella:
Could we just delete the implementation instead of commenting it out?

nbrady-techempower:
@LadyMozzarella We may be adding redis back in, so for tests that are failing because redis was removed I was just planning on keeping the author's original implementation in code. It may be helpful for people who are looking at the current implementation, wondering what the intent was. Since I don't know if they would have done things differently had they known they couldn't use redis, this might be the best way for now.

nbrady-techempower:
Nonetheless, this one passed locally but isn't passing travis so I will have to revisit in a moment.

LadyMozzarella:
In that case, can you add a comment with that information while you're in there?

nbrady-techempower:
Will do.

nbrady-techempower:
All green! merging

PR #2430, opened 2016-12-27
Cakephp fix

nbrady-techempower:
Cakephp was using redis (which is no longer implemented) but was also using it as an in memory cache for some of the database endpoints which is not permitted.


LadyMozzarella:
Nice comment. Thanks!

nbrady-techempower:
Travis agrees! The failure is the pimf Travis mistake from earlier.

PR #2431, opened 2016-12-27
PHP Laravel framework fix

nbrady-techempower:
Was missing the updated location for the fpm config


nbrady-techempower:
Laravel passing!

PR #2432, opened 2016-12-27
pimf framework removal

nbrady-techempower:
pimf hasn't been working for a few rounds and the current implementation is far outdated versus their recent releases. Also most of the framework source was included.

ping @gjerokrsteski - since you were the original contributor of the test, wanted to let you know in case you wanted to create an updated version.


.travis.yml
```diff
@@ -129,7 +129,6 @@ env:
     - "TESTDIR=PHP/fat-free"
     - "TESTDIR=PHP/fuel"
     - "TESTDIR=PHP/kohana"
-    - "TESTDIR=PHP/laravel"
```
LadyMozzarella: Was this supposed to be `PHP/pimf` instead of laravel?

nbrady-techempower:
OH NO! haha yes, nice catch!

PR #2435, opened 2016-12-28
Update clang compilers

nbrady-techempower:
Adds a missing tool in clang-3.8 and adds clang-3.9 to the toolset.

The latter will be used in an upcoming C++/silicon fix.


nbrady-techempower:
This is interesting, and again makes me wonder how travis actually works. This test started running before the crystal/kemal benchmark_config error was merged in to master. Then once it was, all of the tests starting failing even though this was in the middle of the tests and hadn't been rebased. So strange.

nbrady-techempower:
All relevant tests passed travis!

PR #2436, opened 2016-12-28
Elli and Mochiweb framework fixes

nbrady-techempower:
Both of these erlang frameworks were missing the mysql dependency


nbrady-techempower:
All relevant tests passed travis

PR #2441, opened 2016-12-28
Java/restexpress framework fix

nbrady-techempower:
restexpress default test was testing a mongo connection. Combined the default test to include the mongo db tests since all the other meta data is the same.


nbrady-techempower:
travis agrees!

PR #2442, opened 2016-12-28
RevenJ framework testing fixes

nbrady-techempower:
In my quest to figure out why travis was reporting failures for CSharp/revenj, I realized that it was running both revenj frameworks - CSharp/revenj and Java/revenj - because they have the same name. Since the display name in Java/revenj is RevenJ-JVM I renamed that test to Java/revenj-jvm. In the process of testing, I had moved the default test for the CSharp version to the top. I'll leave that as is, since it's more uniform with the others.


LadyMozzarella:
This PR LGTM. However, I'm curious for the future if it is possible (or if we would want) to support frameworks with the same name in different languages. Should we create an issue for this?

nbrady-techempower:
I think, for many reasons, that would be hard. A lot of the tools, and including the results structure, rely on unique framework names without caring about the language folder.

LadyMozzarella:
Interesting, thanks. Good to know!

knewmanTE:
I don't know all of the ins and outs of how the metadata works on the results site, but I was talking with the Revenj maintainer recently and his concern with renaming his tests is that the metadata change would prevent them from displaying properly in past rounds, since the `results.json` file only lists frameworks by their name. I still am in favor of getting all of the frameworks to conform to a lowercase-hyphen naming convention, but it is worth considering as a potential consequence of updating the name of a framework. Perhaps it's something we'll want to consider when we try to get a continuous benchmarking results site working, or perhaps there's a solution that involves tying a specific metadata file to a specific round of results (do we already do this? I feel like we don't, but am hoping that I am wrong).

nbrady-techempower:
@knewmanTE sadly we don't do that right now but that's a good idea as part of reformatting the results data. Each framework could have a unique id. Or even something like formatting them all to have a name that looks like <language>-<framework> all lowercase. As far as the revenj rename goes, if it's that big of a concern (I don't know how many rounds he's in right now) it shouldn't be too hard to go in and change the meta data for previous rounds. Going forward, unless and until we make some big changes with the toolset, we'll have to make sure framework names are unique.

With travis fixed and soon only passing tests in master, we can catch things like this in the future. That's my first concern, to fix these breaking tests.

nbrady-techempower:
So, for now, this simple change fixes both revenj tests on travis and locally. We can probably look at doing a `sed` type replace in the meta-data for the java one. Merging

PR #2448, opened 2016-12-29
Go/echo framework removal

nbrady-techempower:
Go/echo is pulling in go packages from a number of different repos without being tied down to any specific versions. I've tried a couple of combinations of versions without any luck.

@methane Just going to ping you here in case you wanted to take a look at bringing this back in, since I believe you were the one that updated it to v2.


methane:
Please wait.  I'm checking what happens in echo.
At least, it drops fasthttp support and APIs for selecting http server are removed.
That's why build fails.

nbrady-techempower:
I can hold off on this but the fix would have to include specific versions of the packages being used. Pulling packages direct from github repos at master causes occasional failures and doesn't keep a clean history of performance improvements.

methane:
> Pulling packages direct from github repos at master causes occasional failures and doesn't keep a clean history of performance improvements.

Sadly speaking, all Go frameworks doesn't pin the version.
There are no defacto standard tool for pinning yet.

methane:
https://github.com/methane/FrameworkBenchmarks/pull/8

nbrady-techempower:
Can you 'go get' from local git repos? In that case we could pull in the source from a specific commit at runtime.

methane:
I'm trying glide to lock dependency versions.

nbrady-techempower:
Excellent! Hopefully that works out. If not I can explore more tomorrow. Thanks for all your work @methane!

nbrady-techempower:
Closing in favor of #2450

PR #2449, opened 2016-12-29
Java/play2 framework fix

nbrady-techempower:
Fixes a typo in the `play2-ebean-hikaricp` test that was causing failure.


nbrady-techempower:
travis says let's go!

PR #2451, opened 2016-12-30
Octane: Adding /json benchmark and cleaning up some build issues.

simongui:
This PR adds the `/json` benchmark implementation for octane and cleans up some build issues.

Do we really need to implement all the code inside the Techempower repository? I'm spending many hours trying to sync this implementation with the one in my repository. I don't see a lot of entries doing this.

I'm actually confused at the use of `toolset/setup/linux/frameworks/octane.sh` here because it's actually compiling the octane repo and the source code in `frameworks/C/octane` aren't being used.


nbrady-techempower:
@simongui Source code for the framework shouldn't be in the `frameworks/C/octane` directory. I may have originally thought it was part of the top layer that would be specific to this implementation, but it's not. It should be pulling in a specific version (in this case from the github repo) that anyone else would use to build on top of.

simongui:
Octane is a bit raw right now so the application and protocol code ends up being in the same place and I am just copying source/build between the Octane repo + the FrameworkBenchmarks repo. Could I just make it git clone octane and use the built in application that implements the FrameworkBenchmarks and as the octane API evolves I'll add a simpler implementation into this repo at a later date? Would that be alright temporarily?

nbrady-techempower:
Maybe I'm a bit confused. Isn't that what the `octane.sh` file is doing?

simongui:
Yeah pretty much, but by your request I added the implementation code in the frameworks directory. Should I just clean that out?

Also another question, this implementation is raw in that it only implements the bare minimum requirements defined. There's going to be a fuller featured implementation of the same benchmarks.

Should I have the following or some similar naming convention for both entries?
`frameworks/C/octane-stream`
`frameworks/C/octane-web`

nbrady-techempower:
The src that would implement the routes/handlers specifically for testing has to live within the `frameworks/C/Octane` directory. You can do any cloning from the `toolset octane.sh` file and move it into the `frameworks/C/Octane` directory and combine it with the src you have there if you need to, but we would need to see the implementation specific to the FrameworkBenchmarks there.

nbrady-techempower:
For `stream` and `web` you can create subdirectories within `C/octane` and have a single benchmark config that points to those.

simongui:
So it'll run `plaintext` and `json` twice, once for each? How do you define it to run a different executable for different passes?

nbrady-techempower:
Just point to a different `setup.sh` file within the subdirectories. So in your `benchmark_config` you would have something like:

```
{
  "framework": "octane",
  "tests": [{
    "default": {
      "setup_file": "web-setup",
      "plaintext_url": "/",
      "port": 8000,
      "approach": "Realistic",
      "classification": "Platform",
      "database": "None",
      "framework": "None",
      "language": "C",
      "flavor": "None",
      "orm": "Raw",
      "platform": "None",
      "webserver": "None",
      "os": "Linux",
      "database_os": "Linux",
      "notes": "",
      "versus": "octane"
    },
    "stream": {
      "setup_file": "stream-setup",
      "plaintext_url": "/",
      "port": 8000,
      "approach": "Realistic",
      "classification": "Platform",
      "database": "None",
      "framework": "None",
      "language": "C",
      "flavor": "None",
      "orm": "Raw",
      "platform": "None",
      "webserver": "None",
      "os": "Linux",
      "database_os": "Linux",
      "notes": "",
      "versus": "octane"
    }
  }]
}

```

nbrady-techempower:
That's just a quick example changing the setup files, not anything else. If you'd like quicker chat, I'm available in the #techempower-fwbm channel on irc. freenode server.

PR #2452, opened 2016-12-30
Scala/unfiltered removal

nbrady-techempower:
Scala/unfiltered hasn't produced results in the last 3 rounds. Several packages have moved/apis need to be updated, etc.

*note* if anyone does revert this, there was a typo in the `setup_unfiltered.sh`. should be `fw_depends mysql java`

nonetheless, framework still hasn't worked in some time.


nbrady-techempower:
This is gtg!

PR #2453, opened 2016-12-30
Scala framework removals

nbrady-techempower:
`lift-stateless` and `scalatra` are 3 year old implementations that also have not worked for some time. Scala is still well represented with 11 other frameworks.


nbrady-techempower:
Travis did so agree!

PR #2455, opened 2016-12-30
Ur/urweb framework fix

nbrady-techempower:
urweb's default test relied on testing a postgresql connection, so I've combined the default and postgresql tests in to 1. The mysql test would only work after the default test was run, and would not work on it's own. It was relying on urweb already being installed. I've moved the installation of the framework to `toolset/setup/linux/frameworks/urweb.sh` so it can be `fw_depends` by both setup files that need it. All tests pass now.


nbrady-techempower:
urweb back to passing!

PR #2458, opened 2016-12-30
ffead-cpp framework fix

nbrady-techempower:
Attempting to fix some of the failing tests. The nginx tests will probably still fail.

@sumeetchhetri pinging you here in case you wanted to take a look at it before I remove the nginx tests. This could be a broader problem with the nginx setup after the toolset change, though some framework implementations that use nginx are still fine.

### DIFF
--- a/frameworks/C++/ffead-cpp/setup.sh
+++ b/frameworks/C++/ffead-cpp/setup.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-fw_depends ffead-cpp
+fw_depends mongo ffead-cpp
 
 export FFEAD_CPP_PATH=$TROOT/ffead-cpp-2.0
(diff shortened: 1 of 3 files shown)


frameworks/C++/ffead-cpp/setup.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-fw_depends ffead-cpp
+fw_depends mongo ffead-cpp
```
LadyMozzarella: Why are the others `mongodb` and this is just `mongo`?

nbrady-techempower: because jk! thanks ;)

nbrady-techempower:
Closing in favor of #2460

PR #2459, opened 2016-12-31
Update Finatra framework version to release 2.7.0

cacoco:
Problem

We'd like to update the Finatra framework version and
also bring the /plaintext test inline with the Finagle
version for consistency between the frameworks.

Solution

Update the Finatra dependency and the test Controller.


nbrady-techempower:
Tests pass. This will be in the next round. Thanks @cacoco !

PR #2460, opened 2016-12-31
Fixed failing ffead-cpp framework nginx tests Attempt 1

sumeetchhetri:
Copied changes from https://github.com/TechEmpower/FrameworkBenchmarks/pull/2458
nginx was getting built before mongo c driver was installed and hence the failing tests, reordered the sequence of installation of mongo c driver and nginx

### DIFF
--- a/frameworks/C++/ffead-cpp/setup-apache2.sh
+++ b/frameworks/C++/ffead-cpp/setup-apache2.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-fw_depends apache ffead-cpp-apache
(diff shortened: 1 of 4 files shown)


nbrady-techempower:
Oh  interesting! Hopefully this does the trick.

nbrady-techempower:
@sumeetchhetri this fixes the first test but the nginx tests are still failing and now I know why. This is across the board. Adding the runner user (just the regular user) to the root group solves this problem and fixes a few other tests. This is something we overlooked when removing the testrunner user.

Gonna merge this and make a PR for that.

PR #2462, opened 2017-01-02
C++/cutelyst framework removal

nbrady-techempower:
@dantti Wanted to ping you in this since you've done the most work. I'm removing all frameworks/tests that aren't currently passing travis and local testing. It looks like this is trying to install packages that no longer exits - QT56? Am going to remove for now but feel free to fix if you have some time. We'll probably be starting preview runs for round-14 soon.


dantti:
@nbrady-techempower sorry but this is really non-sense to me.
If the framework doesn´t pass tests create a way to disable it like adding a DISABLED file or changing the json config, not removing all sources and making git history a mess.
The fix is trivial, its because I deppend on a PPA to get a decent new Qt version even the newest Ubuntu LTS includes a very old release, I require 5.6 and the PPA now has only 5.7 so I just need to update a single line.

nbrady-techempower:
@dantti I'm sorry that is non-sense to you. If that's the fix, that's great. Works great. Moving forward,something like DISABLED would also be great. But things have been broken for a long time and travis is now inline with our testing environments.

nbrady-techempower:
And just an fyi, though disabling the fw will probably be ok for a round, we will not be keeping frameworks that don't pass tests in the repo longer than that. We're at the point now with travis that if it doesn't pass travis it won't give us any results when we run the entire suite, and it will hold up a run for several hours. This initial clean up is just to make sure that travis gives us green check marks across the board (which is something we've never been able to accomplish) So though it seems unreasonable to completely remove a framework, this won't be happening in the future. Once we get to all passing, we'll be able to identify things that break right away and address it ourselves or with the contributor.

PR #2465, opened 2017-01-02
Cutelyst framework update

nbrady-techempower:
@dantti Looking at your repo, seems this is the correct ppa. Fixed at least the default test for me locally.


nbrady-techempower:
Passes all tests

dantti:
Thanks, the problem is that the PPA author completely removes and renames his PPA when a new Qt release is out, sadly I don't know a better (more stable) PPA. Tomorow I'll do a new Cutelyst release and would update this to use his 5.7. Sometimes my Travis tests also fails for Cutelyst due some maintanance on PPA infrastructure...

PR #2468, opened 2017-01-03
remove redis from databases

nbrady-techempower:


nbrady-techempower:
lwan and mojolicious queries = 501 both broke. Should be unrelated but restarting those jobs.

nbrady-techempower:
And passing

PR #2469, opened 2017-01-03
Update Cutelyst to 1.2.0

dantti:


dantti:
One of the tests is failing since the new Cutelyst version now reads uwsgi config files, this is fixed in #2474

PR #2471, opened 2017-01-03
fw_installed bash function 

nbrady-techempower:
In an effort to make the setup scripts a little cleaner, I've created an `fw_installed` function that will source the install file if it exists. Provide a working example with the `octane.sh` setup.


msmith-techempower:
Loving it

nbrady-techempower:
`aspnetcore` timing out is unrelated and appearing in all travis runs. will look into as a separate issue.

PR #2475, opened 2017-01-03
Use /etc/hosts aliases instead of $DBHOST and other

nbrady-techempower:
I noticed that `TFB-client` `TFB-database` and `TFB-server` were already being set up in the `/etc/hosts` file during vagrant bootstrapping. I've updated the travis setup to do the same. Before something like this gets merged in, a note in the [Installation Guide](http://frameworkbenchmarks.readthedocs.io/en/latest/Development/Installation-Guide/) for those not using vagrant is needed.

Ideally, we would no longer need to use $DBHOST and other environment variables for IP addresses. We would also no longer need to use `sed` in setup files to replace IPs. I've edited a few frameworks here as a proof of concept, but also, the suite uses `benchmark.cfg.example` if a `benchmark.cfg` doesn't exist. So for the sake of this travis run, `$DBHOST=TFB-database`.

### DIFF
--- a/benchmark.cfg.example
+++ b/benchmark.cfg.example
@@ -1,9 +1,9 @@
 [Defaults]
 # Available Keys: 
-client_host=127.0.0.1
+client_host=TFB-client
 client_identity_file=None
 client_user=techempower
-database_host=127.0.0.1
+database_host=TFB-database
 database_identity_file=None
 database_os=linux
 database_user=techempower
@@ -20,7 +20,7 @@ query_levels=[1, 5,10,15,20]
 threads=8
 mode=benchmark
 os=linux
-server_host=127.0.0.1
+server_host=TFB-server
 sleep=60
 test=None
 type=all
--- a/frameworks/C/h2o/setup.sh
+++ b/frameworks/C/h2o/setup.sh
@@ -5,7 +5,7 @@ fw_depends postgresql h2o mustache-c yajl
 H2O_APP_HOME="${IROOT}/h2o_app"
 BUILD_DIR="${H2O_APP_HOME}_build"
 H2O_APP_PROFILE_PORT="54321"
-H2O_APP_PROFILE_URL="http://TFB-database:$H2O_APP_PROFILE_PORT"
+H2O_APP_PROFILE_URL="http://127.0.0.1:$H2O_APP_PROFILE_PORT"
 
 # A hacky way to detect whether we are running in the physical hardware or the cloud environment.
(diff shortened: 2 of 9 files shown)


benchmark.cfg.example
```diff
@@ -20,7 +20,7 @@ query_levels=[1, 5,10,15,20]
 threads=8
 mode=benchmark
 os=linux
-server_host=127.0.0.1
+server_host=TFB-server
```
msmith-techempower: There was once, long ago, an implementation similar to this; we removed it because some frameworks could not resolve names for reasons we could not explain. Hopefully, this is no longer the case....

nbrady-techempower: If it is the case, those frameworks can still do the `sed` stuff they're doing now and pull in the ip from the `/etc/hosts` file.

msmith-techempower: I see this being done in travis and vagrant, but I don't see this setup happening in the general case (like our production code): you should plan on adding that in.

nbrady-techempower: This would be part of the setup. So it would be something done manually (see the linked installation guide) or did you have something else in mind?

nbrady-techempower:
@msmith-techempower Here is the corresponding documentation: https://github.com/TechEmpower/TFB-Documentation/pull/58/files

All of the modified tests pass. We can leave `$DBHOST` and other environment variables available for now but push future tests and updates toward `TFB-database`, `TFB-client`, `TFB-server`.

If this is good for you, we can merge this in and I can update a lot of the others.

frameworks/C/h2o/setup.sh
```diff
@@ -5,7 +5,7 @@ fw_depends postgresql h2o mustache-c yajl
 H2O_APP_HOME="${IROOT}/h2o_app"
 BUILD_DIR="${H2O_APP_HOME}_build"
 H2O_APP_PROFILE_PORT="54321"
-H2O_APP_PROFILE_URL="http://127.0.0.1:$H2O_APP_PROFILE_PORT"
+H2O_APP_PROFILE_URL="http://TFB-database:$H2O_APP_PROFILE_PORT"
```
volyrique: This is wrong: The URL has nothing to do with the database (note that it is used as a parameter for curl); it is just the location at which the HTTP server is temporarily accessible while gathering the profiling data.

nbrady-techempower:
@volyrique Thanks. Going through some of the setups, to test this change. At a quick glance I misread the port as `5432` and associated it with `postgres`.

PR #2476, opened 2017-01-04
Simplify FWROOT and remove benchmark_profiler

nbrady-techempower:
Rebased and learned how to `git`.

Added distro information that was being done in the `benchmark_profiler` and causing some frameworks that use it, like `wt` to break.


toolset/run-tests.py
```diff
@@ -63,11 +64,12 @@ def main(argv=None):
...
     # Update environment for shell scripts
-    fwroot = setup_util.get_fwroot()
-    if not fwroot: 
-        fwroot = os.getcwd()
-    setup_util.replace_environ(config='toolset/setup/linux/config/benchmark_profile', root=fwroot)
-    print "FWROOT is %s"%setup_util.get_fwroot()
+    os.environ['FWROOT'] = setup_util.get_fwroot()
+    os.environ['IROOT'] = os.environ['FWROOT'] + '/installs'
+    # 'Ubuntu', '14.04', 'trusty' respectively
+    os.environ['TFB_DISTRIB_ID'], os.environ['TFB_DISTRIB_RELEASE'], os.environ['TFB_DISTRIB_CODENAME'] = platform.linux_distribution()
```
msmith-techempower: This line **should** only work if you issue `source /etc/lsb-release` first.

toolset/setup/linux/config/benchmark_profile
```diff
@@ -1,50 +0,0 @@
...
-  FWROOT=$(echo $FWROOT)
-fi
-IROOT=${FWROOT}/installs
-
-# Determine which OS we are using and set the appropriate variables
-source /etc/lsb-release
-export TFB_DISTRIB_ID=$DISTRIB_ID
-export TFB_DISTRIB_RELEASE=$DISTRIB_RELEASE
-export TFB_DISTRIB_CODENAME=$DISTRIB_CODENAME
-export TFB_DISTRIB_DESCRIPTION=$DISTRIB_DESCRIPTION
```
msmith-techempower: These 5 lines, for reference.

PR #2480, opened 2017-01-06
Remove redis from nodejs test

nbrady-techempower:
Removes the redis tests from nodejs


nbrady-techempower:
passing travis and no more redis issues

Issue #2478, opened 2017-01-04
ffead not installing correctly


msmith-techempower:
@sumeetchhetri We are seeing errors when trying to build `ffead-ccp-nginx`

Specifically, it seems there is a linking error when trying to issue `make install` for `nginx-1.11.3` when searching for `libmongoc-1.0.so.0` and `libbson-1.0.so.0`.

Additionally, you should not need to use `sudo` for the majority of the `make install` calls since they should all be rooted in `$IROOT`.

PR #2481, opened 2017-01-06
remove breaking tests from ffead-cpp

nbrady-techempower:
Removes the failing nginx tests and the apache2-mongo test from ffead's `benchmark_config` until resolved. See #2478


nbrady-techempower:
Closing in favor of ongoing work in #2487

PR #2483, opened 2017-01-07
CSharp/aspnetcore update

mikeharder:
- Update framework and packages to 1.1
- Remove dependency on net451 and mono
- Update PostgreSQL connection string for latest npgsql
  - Prepared statements are now persisted by default, so the PersistPrepared option has been removed.
  - npgsql/npgsql#1237
- Make shell scripts executable
- Change scripts to use `dotnet Benchmarks.dll`
  - Scripts currently use `dotnet run` which is intended for iterative development rather than production.
  - Should not cause a measurable change in performance.
- Add DB scenarios to README.md


nbrady-techempower:
@mikeharder thanks! Was going to ping you guys yesterday because the dapper implementation is no longer working. It looks like this is also not passing.

frameworks/CSharp/aspnetcore/run-linux.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-fw_depends mono dotnetcore
+fw_depends dotnetcore
```
msmith-techempower: Wooo!

msmith-techempower:
The error is odd; when you run it manually in the foreground, you get the following:

```
dotnet bin/Release/netcoreapp1.1/Benchmarks.dll server.urls=http://*:8080 scenarios=$1 server=kestrel threadCount=$threadCount NonInteractive=true
The specified framework 'Microsoft.NETCore.App', version '1.1.0' was not found.
  - Check application dependencies and target a framework version installed at:
      /usr/share/dotnet/shared/Microsoft.NETCore.App
  - The following versions are installed:
      1.0.1
  - Alternatively, install the framework version '1.1.0'.
```
Which returns a 131 exit code.

Interestingly, when run in the background, it waits for user input before exiting and then printing the error message. This may be how bash handles error exits from background processes by default... I'm not sure.

mikeharder:
Thanks @nbrady-techempower and @msmith-techempower.

To fix the `Microsoft.NETCore.App` error, we will need to either:

* Update the `dotnetcore` install script to install 1.1 instead of 1.0
* Create a versioned script like `dotnetcore-1.1` to install 1.1, and leave `dotnetcore` to install 1.0

I will check if any other frameworks depend on dotnetcore.  If not, I will just update the existing install script, and if so, I will create a versioned install script.

I was not aware of the Dapper failures, I will investigate those as well.

msmith-techempower:
@mikeharder I believe yours is the only test that currently relies upon `dotnetcore`, so go ahead and update that to 1.1. If it breaks anything, Travis will let us know and we'll go from there.

toolset/setup/linux/languages/dotnetcore.sh
```diff
@@ -2,11 +2,11 @@
...
 fw_installed dotnetcore && return 0
 
-sudo sh -c 'echo "deb [arch=amd64] https://apt-mo.trafficmanager.net/repos/dotnet-release/ trusty main" > /etc/apt/sources.list.d/dotnetdev.list' 
-sudo apt-key adv --keyserver apt-mo.trafficmanager.net --recv-keys 417A0893
+sudo sh -c 'echo "deb [arch=amd64] https://apt-mo.trafficmanager.net/repos/dotnet-release/ trusty main" > /etc/apt/sources.list.d/dotnetdev.list'
+sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 417A0893
 sudo apt-get update
 
-sudo apt-get install dotnet-dev-1.0.0-preview2-003131 -y
+sudo apt-get install dotnet-dev-1.0.0-preview2.1-003177 -y
```
msmith-techempower: I thought this was going to be 1.1?

mikeharder: This does install the 1.1 runtime.  See https://github.com/dotnet/core/blob/master/release-notes/1.1/1.1.md.  The version number in the package name is the SDK version (1.0.0 Preview 2.1) rather than the runtime version (1.1).

msmith-techempower: Got it, thanks.

mikeharder:
@nbrady-techempower: You mentioned the dapper tests were no longer working.  Do you have an error message?  On my machines, the aspnetcore dapper tests are working fine both before and after this PR.

nbrady-techempower:
@mikeharder before this PR this is what the dapper tests looked like both on my local tests and on travis: https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/189603954

Everything looks good here though, merging! Thanks!

mikeharder:
@nbrady-techempower: Thanks for the error message.  This was caused by the pre-release version of `npgsql` removing support for the `PersistPrepared` connection string option, which is something I fixed in this PR.

Issue #2482, opened 2017-01-07
Cannot run nodejs-postgres without running nodejs


rkraman:
Downloaded the latest code today and created a vagrant development environment and ran the test for nodejs-postgres
**toolset/run-tests.py --test nodejs-postgres**
This gives an error for MongoDB and the test fails and stops.
The output log file can be found here:  http://sprunge.us/XUQC

Deleted the vagrant image, created a new one and this time first ran the default test
**toolset/run-tests.py --test nodejs**

That ran successfully.  Later I **ran toolset/run-tests.py --test nodejs-postgres**
and everything worked fine now.
Need to remove the dependency for MongoDB.
-rkraman

nbrady-techempower:
At the time this implementation was written, our test suite had all databases already loaded and accepting connections. We recently changed this so that frameworks would only load the databases they need and also make sure there was fresh starting data in the db. Because there's only 1 `package.json` and imports are happening for all dependencies on all tests, each test seems to be testing for other db connections even if not using it. Rather than rewrite the framework implementation, which we leave up to contributors/maintainers, loading all dbs for each test is a quick fix. See #2484

PR #2485, opened 2017-01-07
Big Travis run - Do Not Merge

nbrady-techempower:
**Problem** 
#2482 made me think of this. After doing the db split and other toolset changes in the past, I realize that some tests might not work correctly unless one of the previous tests in that framework already ran. This shouldn't be the case; each test should work on its own.

**Test for Problems**
I made a *temporary* change to travis to run every test individually, spinning up a new vm every time. This isn't meant to be merged, but just as a one time pass to see if there are any other tests that are failing if they're not run after the other tests in their framework.

So, let's see what travis do.

##### Some tests will break since I didn't take the time to test for os. Just listed all the tests with `--list-tests`, output to a file, quick `sed` to turn it into a travis matrix.


methane:
#2481 is "big travis run" already.
And #2486 will be too.

nbrady-techempower:
@methane no this  is different. Those are one vm per framework directory.  This is one vm per test.

methane:
Ah, I see.

msmith-techempower:
This would be a mean thing to do; going from 174 to 360 Travis-tests is just painful.

nbrady-techempower:
@msmith-techempower I noted it's not meant to ever be merged or changed. I wanted to do a one time pass to see what other tests like `nodejs` were failing. Found and fixed `hapi` from this.

msmith-techempower:
Ah, I think I see.

This is really interesting. Ideally, we WOULD be able to set up our Travis-CI this way, but it is just too punitive generally. Let's add this is a step to take every now and then (open a new branch with `travis.yml` built this way, throw it against the wall, and see what sticks).

nbrady-techempower:
@msmith-techempower: i can make a script and throw it in the `travis` directory that will generate a `.travis.yml` file like this based on the existing tests.

PR #2486, opened 2017-01-07
Travis Update & tfb command

nbrady-techempower:
Updates travis to source the `prerequisites.sh` file like a normal installation and adds a `tfb` command to `prerequisites.sh` such that you can do:

`tfb --mode verify --test gemini`

from ~~fwroot~~ anywhere.


toolset/setup/linux/prerequisites.sh
```diff
@@ -56,6 +56,11 @@ sudo sh -c "echo '*               -    nofile          65535' >> /etc/security/l
 sudo sh -c "echo '*            hard    rtprio             99' >> /etc/security/limits.conf"
 sudo sh -c "echo '*            soft    rtprio             99' >> /etc/security/limits.conf"
 
+# Create a tfb command alias for running the toolset
+# For now, this still ensures you have to be in the framework root to run it
+alias tfb="./toolset/run-tests.py"
+echo 'alias tfb="./toolset/run-tests.py"' >> ~/.bash_profile
```
msmith-techempower: If you delete the `installs` directory (which I do for testing fairly regularly), then this will get rerun and append several times.

Alternatively, I **believe** you can use `/etc/profile.d` to create a file that will be executed in new shells (would require sudo, but meh).

toolset/travis/travis_setup.sh
```diff
@@ -41,3 +29,4 @@ echo 127.0.0.1 TFB-database | sudo tee --append /etc/hosts
 echo 127.0.0.1 TFB-client   | sudo tee --append /etc/hosts
 echo 127.0.0.1 TFB-server   | sudo tee --append /etc/hosts
 
+source ./toolset/setup/linux/prerequisites.sh
```
msmith-techempower: 👍

toolset/setup/linux/prerequisites.sh
```diff
@@ -56,6 +56,11 @@ sudo sh -c "echo '*               -    nofile          65535' >> /etc/security/l
...
+
 # Sudo in case we don't have permissions on IROOT
 sudo touch fwbm_prereqs_installed
```
msmith-techempower: Also, update this line - I don't want it to be `sudo`

Also, ensure that this is in `$IROOT`

nbrady-techempower:
@msmith-techempower I have documentation ready here when you're good with this. [#60](https://github.com/TechEmpower/TFB-Documentation/pull/60)

.gitignore
```diff
@@ -42,7 +42,7 @@ php-kohana/application/logs/
 php-fuel/fuel/app/logs/
 results/
 benchmark.cfg
-current_benchmark.txt
+tmp/
```
msmith-techempower: This file doesn't need any changes, right?

toolset/setup/linux/prerequisites.sh
```diff
@@ -3,12 +3,7 @@
 set -x
 export DEBIAN_FRONTEND=noninteractive
 
-source $FWROOT/toolset/setup/linux/bash_functions.sh
-
-RETCODE=$(fw_exists fwbm_prereqs_installed)
-[ ! "$RETCODE" == 0 ] || { \
-  echo "Prerequisites installed!"; 
-  return 0; }
+source ./toolset/setup/linux/bash_functions.sh
```
msmith-techempower: Is this necessary? I don't think you're actually using any of the functions in here, are you?

toolset/travis/travis_setup.sh
```diff
@@ -2,18 +2,6 @@
...
-sudo apt-get install python-dev
-sudo pip install pyopenssl --upgrade
-sudo pip install ndg-httpsclient --upgrade
-sudo pip install pyasn1 --upgrade
-sudo pip install colorama
-# Version 2.3 has a nice Counter() and other features
-# but it requires —-allow-external and -—allow-unverified
-sudo pip install progressbar
-echo "127.0.0.1 " `hostname` | sudo tee /etc/hosts
-echo "127.0.0.1 localhost" | sudo tee /etc/hosts
```
msmith-techempower: These two hosts things fixed a few issues a while back; are you sure they can be removed now?

nbrady-techempower: This was in the travis setup and it doesn't seem to have caused any issues.

toolset/setup/linux/prerequisites.sh
nbrady-techempower: hm, right.

.gitignore
nbrady-techempower: right

toolset/travis/travis_setup.sh
msmith-techempower: Let's allow a full Travis-CI run and we'll make sure nothing obvious breaks from removing that from `/etc/hosts`

PR #2489, opened 2017-01-09
JavaScript/hapi fix tests & remove redis handlers

nbrady-techempower:
Same as issue #2482 and found with #2485 
Combine all db installs to one setup since some dependencies for the framework rely on db's that we might not be testing. Should be able to run any test now without running one prior. 
Also removed redis stuff.


nbrady-techempower:
passes le travis.

Issue #2494, opened 2017-01-12
Remove run-tests.py arguments


msmith-techempower:
1. ~~`run-tests.py` accepts a number of command line arguments which serve the same purpose as keys in `benchmark.cfg`; remove those arguments from `run-tests.py` and leave only arguments which augment the running of the application (clean, test, mode, etc).~~

2. ~~Add argument to run all tests in a framework directory (e.g. `tfb --framework gemini` will be functionally equivalent to `tfb --test gemini gemini-mysql gemini-postgresql` etc)~~

3. ~~Drop `--list-test-metadata` and instead just always output this file to the results timestamp dir for every run.~~

4. `--parse` is probably broken since we have done so many directory changes; rename this to `--merge` and give it the implementation and description that it will merge all results timestamp directories. This will involve the difficulty in merging `results.json` files and `metadata.json` files (as well as `dstat` and whatnot), so it will be difficult.

nbrady-techempower:
@msmith-techempower We can probably close this after you review and merge #2502 and maybe hold off on `--parse` until the `results.json` structure is done.

msmith-techempower:
Agreed

nbrady-techempower:
So these are all done. Would love to work on a new `results.json` structure soon!

PR #2495, opened 2017-01-12
Travis Smart Diffing & tfb --test-dir argument

### LINKED ISSUE #2491: Make Travis-CI script smarter about dependencies
Update our Travis-CI scripts to be more intelligent about situations in which dependencies are updated.

For example, if `toolset/setup/linux/languages/dotnetcore.sh` is updated, only collect tests which have `dotnetcore` in the `fw_depends` line.

nbrady-techempower:
Here we go.

#### Problems encountered

The travis build cycle creates some issues. Exiting 0 in the `before_script` lifecycle won't stop the `script` lifecycle from running. Exiting 0 in the `script` lifecycle will only stop the current script, and move to the next item if there is one so we couldn't move everything there.

You can't set environment variables in python for the shell that started the script, only the subshell running the python script. Solution? Just return something from the script and set the environment variable from the shell? Unfortunately, no. Any return/exit code other than 0 from a script being run by travis will immediately terminate the build. My solution: grep the output of the python script, searching for some success output, in this case `travis-diff-continue` and saving `'1'` to the `CONTINUE_TEST` env var. 

One of my goals here is to keep travis completely separate from toolset files. In other words, I didn't want any conditional flows in `run-tests.py` relying on any travis environment variables. If one day we unplug travis for something else, we won't have to scrub through `run-tests.py` to remove anything.

#### Now for some new goodies

If an `fw_depends` installation file has been modified, instead of all tests running, only the tests that rely on that file (or the files relied upon by that file), will run.

Travis provides `[ci skip]` which if found in the latest commit message will not trigger any builds from `.travis.yml`
Keeping with this theme I've added the following commit message hooks:

`[ci run-all]`: Will force every test in the matrix to run, despite files changed.

`[ci fw <testdirs>]`: Will force a particular test to run, even if its files haven't been changed.
  examples: `[ci fw Java/gemini]` or multiple tests `[ci fw Java/gemini CSharp/aspnetcore]`

`[ci fw-only <testdirs>]`: No matter what has changed, please only run these tests. This is useful for quickly testing one test while making larger toolset changes without having to modify the `.travis.yml` file. Obviously, this won't be accepted for the final run of a pull request but can be used while testing on your own branch.

#### --test-dir argument

I added this here so that I could use it within `.travis.yml`. This works the same way as `--test` as it can take any number of arguments and it will run all of those test directories. If one directory doesn't exist, it will error right away and let you know which was the problem.

Example:
`tfb --mode verify --test-dir Java/gemini CSharp/aspnetcore`

resolves #2491 
and begins work toward #2494


msmith-techempower:
You might be a wizard.

nbrady-techempower:
#wizardmerge

PR #2501, opened 2017-01-13
Fix Rails 4.1.4 failures in Travis

mwpastore:
Rails sees a test environment where none exists, and fails the sanity
check of said environment. Puma, Unicorn, and Thin all set RACK_ENV to
production, so it should be safe to create a dummy test environment to
placate the sanity check. We won't use it.


nbrady-techempower:
Awesome! Thank you!

mwpastore:
I left in the magic Travis tag. I think it's safe to merge because the changes are local to the Ruby/rails directory? Let me know if I need to take it out.

nbrady-techempower:
@mwpastore perfect! Thanks again!

PR #2502, opened 2017-01-13
Use benchmark.cfg for defaults and remove args from run-test.py

nbrady-techempower:
This PR removes a lot of the args that should be provided by the `benchmark.cfg` file as mentioned in #2494.

Also removes `--list-test-metadata` which now runs all the time by default.

Updated to work with travis and vagrant as well.

(Editor picked up some hard tabs and replaced with spaces it looks like as well)


msmith-techempower:
If there isn't, there needs to be a "view diff ignoring whitespace"

nbrady-techempower:
There's probably a way with the git cli. Most of the work here was in run-tests

nbrady-techempower:
Note I tested locally but had to edit my benchmark.cfg file. Tested on a fresh vagrant install which auto updated the cfg file for vagrant and works in Travis.

msmith-techempower:
Odd, the `lua` database tests started failing on this build... probably not related, but odd that it happened only to them.

knewmanTE:
What was the reasoning behind switching `benchmark.cfg.example` to `benchmark.cfg`? It seems that now it will show up as a changed file on Git if a user has any changes from the default (such as running as a user other than techempower).

nbrady-techempower:
@knewmanTE there were a couple reasons, you'll have to have @msmith-techempower remind me but it's in the .gitignore right? Basically the toolset was always falling back on the example which isn't good practice either.

knewmanTE:
@nbrady-techempower interesting. I see it in my .gitignore, but it's still popping up when I check git status.

msmith-techempower:
I am, apparently, wrong about how `.gitignore` works. We will have to add back the example file and have users create their own to ignore in git.

PR #2505, opened 2017-01-16
Update Readme

nbrady-techempower:
Updated readme to include `tfb` command and removed `Issues Stats` badges, which are either really slow to load and old, or broken.


cjnething:
I just went to the Issue Stats page and it seems to be loading fine. Perhaps it was just down when you went to check it?

nbrady-techempower:
@cjnething You can see the image loading in our README on the front page?

cjnething:
Yes, I can

nbrady-techempower:
Hmm, weird. I always get this:

PR #2506, opened 2017-01-18
Expand tokio-minihttp with /db and /queries

alexcrichton:
This commit expands the Rust framework tokio-minihttp with routes for both `/db`
and `/queries`, implementing those two benchmarks. The Tokio framework in Rust
is asynchronous but there aren't many asynchronous database drivers just yet, so
these routes are implemented with blocking I/O on an external thread pool.

I also refactored the existing code a bit to be more amenable to the addition of
the new routes, and I also added a `Cargo.lock` file to ensure reproducible
builds in development and on the release servers. A `Cargo.lock` is similar to a
`Gemfile.lock` in Ruby which I've seen checked in for other frameworks in this
repo, so I figured it'd be a good idea to check it in here as well.


frameworks/Rust/tokio-minihttp/Cargo.toml
```diff
@@ -11,3 +11,13 @@ num_cpus = "1.0"
...
+
+postgres = "0.13"
+r2d2 = "0.7"
+futures-cpupool = "0.1"
+r2d2_postgres = "0.11"
+rand = "0.3"
+url = "1.0"
+
+[replace]
+'futures:0.1.8' = { path = "/home/alex/code/futures" }
```
nbrady-techempower: Where should this be pointing to?

alexcrichton: Oops sorry about that! I meant to remove that before pushing, I've now removed.

PR #2507, opened 2017-01-18
Fixes the hostname issues with Lua

msmith-techempower:
Octopus is still broken for other reasons


msmith-techempower:
@cyberz-eu Any chance you could give me a hand and figure out why octopus is suddenly breaking?

msmith-techempower:
Alright, well this has sat a while - merging in; someone fix Octopus.

PR #2515, opened 2017-01-24
Add Rack+Sequel benchmarks

mwpastore:
This "hard fork" of sinatra-sequel swaps out Sinatra and Slim for plain ol' Rack and an inline string template. Otherwise it is almost exactly the same!

[ci fw-only Ruby/rack-sequel]


knewmanTE:
This looks good to me. Thanks for the contribution!

Issue #2074, opened 2016-06-16
Broken Frameworks


nbrady-techempower:
The following frameworks are not passing Travis in one or both of Round 13 (Master) and our Round 14 target branch. There could be a number of reasons. The goal is to get all frameworks included in the current benchmarks to pass both Travis and local tests. We are considering removing frameworks that are not passing and not being maintained. This would speed up Travis runs and give more accurate snapshots for pull requests.

:heavy_multiplication_x: means the Framework is not passing Travis for that round. :heavy_check_mark:, it has. Local refers to pulling and testing the Framework in its most recent contribution in a vagrant environment. A lot of these issues stem from dependency problems, while others are strictly Travis problems. Some of the Frameworks pass some of the tests. (A lot of the Ruby frameworks are failing on jruby tests). So like I said, there are many different reasons a Framework might appear on this list.

If we can get these all to :heavy_check_mark:, we will be able to better identify when a Framework breaks and if a toolset change may have caused it.

Please feel free to add anything here and if you have a fix for any of these, certainly submit a pull request targeting the next round.

| Framework | Rnd 13 | Rnd 14 | Local | Notes |
| :-: | :-: | :-: | :-: | --- |
| C/lwan | :heavy_multiplication_x: |  | :heavy_check_mark: |  |
| C/haywire | :heavy_multiplication_x: |  | :heavy_check_mark: | Submitted #2099 with a fix for round-14 |
| CSharp/servicestack | :heavy_multiplication_x: |  | :heavy_multiplication_x: |  |
| C++/cpoll_cppsp | :heavy_check_mark: |  | :heavy_check_mark: |  |
| C++/treefrog | :heavy_multiplication_x: | :heavy_check_mark: | :heavy_check_mark: |  |
| C++/silicon | :heavy_check_mark: |  | :heavy_check_mark: | Submitted #2082  with a fix for round-14 |
| Crystal/crystal-raw | :heavy_multiplication_x: | :heavy_check_mark: | :heavy_check_mark: |  |
| Crystal/kemal | :heavy_multiplication_x: |  | :heavy_check_mark: |  |
| Haskell/spock | :heavy_multiplication_x: | :heavy_check_mark: | :heavy_check_mark: |  |
| Haskell/yesod | :heavy_check_mark: |  | :heavy_check_mark: |  |
| Java/activeweb | :heavy_multiplication_x: |  | :heavy_check_mark: | Submitted #2118 with a fix for round-14 |
[...]

PR #2516, opened 2017-01-24
Add Roda+Sequel benchmarks

mwpastore:
This "hard fork" of sinatra-sequel swaps out Sinatra and Slim for Roda and Erubi. Otherwise it is almost exactly the same!

[ci fw-only Ruby/roda-sequel]


knewmanTE:
This looks good to me! 

@mwpastore, out of curiosity, do the different platforms and servers make a noticeable difference in benchmark results? I ask just because Ruby tests have historically had a large number of tests per framework and given that each additional test implementation adds to the total time it takes to run the suite, it would be great if we could find a way to trim down on the total number of Ruby tests. From a quick look at our Round 13 results, it looks like Thin consistently performs the worst across all Ruby tests, Unicorn performs the best, and Puma is a little lower than Unicorn.

mwpastore:
That's fair, and the number of Ruby frameworks is ever-increasing (sorry), so we have some quadratic growth here. Puma, Unicorn, and Passenger are the three most popular Ruby app servers, so I would definitely like to keep those, and JRuby+TorqueBox seems to be the best massively-multithreaded option today (given the somewhat sorry states of Rubinius and JRuby web servers such as Trinidad, Jubilee, Fishwife). I would be comfortable removing JRuby+Puma, would that be helpful?

knewmanTE:
Yeah, that would probably be helpful! I'll make an Issue where maybe we can solicit some feedback from any other Ruby developers who might have strong opinions one way or another.

knewmanTE:
@mwpastore actually, looking at some of our other Ruby tests, it looks like we no longer have any JRuby or Puma tests anyway, so maybe just open up a new PR that removes those tests from your two recent framework additions?

mwpastore:
@knewmanTE @nbrady-techempower removed JRuby in #2277 with the intention of adding it back in for Round 14. I [offered](https://github.com/TechEmpower/FrameworkBenchmarks/issues/2074#issuecomment-263682641) to add it back in for sinatra-sequel and carried the change through to rack-sequel and roda-sequel.

knewmanTE:
Ahh, right you are! I'm fine keeping it as-is for now then. (sorry for the confusion, I'm still catching myself up on all of the setup and Travis changes that were implemented over the past couple months).

mwpastore:
No worries! It's been quite a whirlwind of activity. The new setup and Travis configuration is pretty fantastic.

knewmanTE:
Yeah, we're pretty stoked about it! Having an all-green Travis is a dream come true!

PR #2517, opened 2017-01-25
Add a simple blaze sample

bryce-anderson:
Add an sample employing the http4s blaze native server backend.


knewmanTE:
Looks good to me!

PR #2518, opened 2017-01-25
Update Finagle and Finch benchmarks

vkostyukov:
This PR updates Finagle and Finch benchmarks:

 - Finch is bumped to 0.12
 - Circe is bumped to 0.7
 - Finagle is bumped to 6.41

These are still using Scala 2.11 by a very simple reason - we haven't tested the effect of switching to 2.12 internally.

In addition to dependencies update, this extracts the date-time machinery into a standalone object `currentTime` that also maintains the previous value and only recalculates it when needed (ideally only once a second). I learned this trick from the [Netty benchmark](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/netty/src/main/java/hello/HelloServerHandler.java#L74-L81) where similar (but not the same) approach is taken to minimize the number of times the date object is allocated and formatted.

### DIFF
--- a/frameworks/Scala/finagle/build.sbt
+++ b/frameworks/Scala/finagle/build.sbt
@@ -2,11 +2,11 @@ name := "finagle"
 
 scalaVersion := "2.11.8"
 
-version := "1.0.2"
+version := "1.0.3"
 
 com.github.retronym.SbtOneJar.oneJarSettings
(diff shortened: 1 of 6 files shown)


knewmanTE:
Looks good to me!

PR #2519, opened 2017-01-25
Update Bundler version in Ruby languages

mwpastore:
It looks like 1.14 fixes a bunch of things without any multithreading
regressions!

[ci fw-only Ruby/rack Ruby/rack-sequel Ruby/rails]


nbrady-techempower:
@mwpastore Could relaxing the version cause future problems if not kept up with? In other words, if bundler makes some breaking change, this benchmark test would break, right?

msmith-techempower:
I'm fine with using 1.14 for bundler, but please lock it down with `-v` so that our install files have a built in history for our runs.

mwpastore:
Done!

nbrady-techempower:
This passes with bundler tied to v 1.14. Merging

PR #2520, opened 2017-01-25
Extract common logic for Ruby Sequel benchmarks

mwpastore:
This should make it easier to keep all the benchmarks in the suite
up-to-date and in-sync! And easier to compare and contrast the
frameworks.

(This literally just takes the stuff that's common to all three and moves it into a boot.rb file in the same directory. No dragons here.)

[ci fw-only Ruby/sinatra-sequel Ruby/rack-sequel Ruby/roda-sequel]

### DIFF
--- a/frameworks/Ruby/rack-sequel/boot.rb
+++ b/frameworks/Ruby/rack-sequel/boot.rb
@@ -1,0 +1,68 @@
+# frozen_string_literal: true
+require 'bundler'
+require 'time'
+
+MAX_PK = 10_000
+QUERIES_MIN = 1
+QUERIES_MAX = 500
+SEQUEL_NO_ASSOCIATIONS = true
+
+Bundler.require(:default) # Load core modules
+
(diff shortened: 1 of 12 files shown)


msmith-techempower:
Would it be appropriate to update `grape`, `padrino`, `rails`, and/or `rails-stripped` with this same idea?

mwpastore:
I haven't delved into those codebases (yet) but I'm assuming so. `rails` is probably already modularized out so maybe not that one. If you fine folks are comfortable with it, I'm willing to keep chipping away at them to bring them into something resembling alignment. An ActiveRecord-based `sinatra` rewrite is already in progress.

nbrady-techempower:
@mwpastore I think we are collectively more than comfortable with that.

nbrady-techempower:
I'll merge this one in. Thanks again @mwpastore We're all excited we have a ruby enthusiast fixing up these benchmarks!

PR #2522, opened 2017-01-26
=akka-http remove unnecessary copying of entity

ktoso:
Just a small improvement which I noticed while skimming the code.

The HttpEntity call made on this line already makes the content type be `text/plain` (UTF-8), the additional call made here created an entire copy of the entity object for each request and did not really have to do that since it is the right content type already. See here: https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala#L253


knewmanTE:
Good catch! This looks good to me if Travis agrees!

ktoso:
Thanks for merging, I'll poke around the other ones to see if there's more small things like that somewhere.

PR #2523, opened 2017-01-28
Bring sinatra into alignment with other Ruby frameworks

mwpastore:
[ci fw-only Ruby/sinatra]


nbrady-techempower:
@mwpastore You're a hero!

mwpastore:
Hardly!

PR #2527, opened 2017-01-29
rename asyncio > api_hour

samuelcolvin:
Renaming `python/asyncio` to `python/api_hour`.

Asyncio is part of the python standard library, not a web framework. This benchmark is based on API hour which is a framework build on top of aiohttp which is in term just one of the web frameworks which utilize asyncio.

Copying in @GMLudo who I think built API Hour and submitted it's benchmarks.

Let me know if there's any other changes required.


samuelcolvin:
humm, this seems to be failing on `pip install` is it possible the results are normally cached so the failure would happen with an uncached build on master?

Lots of the packages here are really out of date.

nbrady-techempower:
Looks good!

GMLudo:
@samuelcolvin : api-hour isn't a Web framework, but a gunicorn worker for any AsyncIO Server, not only HTTP.
It's why I called that asyncio to test any asyncio server and not api-hour itself.
If asyncio had a builtin worker, I've used that, but it isn't the case.

methane:
> If asyncio had a builtin worker, I've used that, but it isn't the case.

Gunicorn has aiohttp worker for now.
http://docs.gunicorn.org/en/stable/design.html#asyncio-workers

PR #2528, opened 2017-01-29
upgrade python3 to 3.6

samuelcolvin:
Python 3.6 was released in December, it has numerous new features and is supposed to improve asyncio performance.


samuelcolvin:
I think `Python/asyncio` will fail but that's a different issue I've hopefully solved on #2527.

nbrady-techempower:
Great! And the asyncio issue was addressed in your other PR. Merging

PR #2529, opened 2017-01-29
Python aiohttp

samuelcolvin:
Adding the aiohttp framework with two database access methods:
* [aiopg+sqlalchemy+psycopg2](http://aiopg.readthedocs.io/en/stable/sa.html)
* [asyncpg](https://github.com/MagicStack/asyncpg)

From my testing locally the "raw" variant should be faster than other python frameworks.

Would be good to get some feedback from @asvetlov who is the main developer of aiohttp.


nbrady-techempower:
Looks good! Thanks for all the contributions @samuelcolvin !

samuelcolvin:
No problem, it felt like a productive way to spend a hungover Sunday.

When will the next round be published?

nbrady-techempower:
@samuelcolvin We are working on that now. Been busy for us here.

PR #2532, opened 2017-01-30
Ruby Framework Reduction Act of 2017

mwpastore:
Disable Puma+JRuby variants of ruby-sequel benchmarks (in favor of
TorqueBox+JRuby) in an attempt to curb the somewhat out-of-control
proliferation of Ruby framework benchmarks.

[ci fw-only Ruby/roda-sequel Ruby/rack-sequel Ruby/sinatra-sequel]


nbrady-techempower:
This will probably get bi-partisan support.

PR #2533, opened 2017-01-30
Round 14 JVM tuning for ruby-sequel/sinatra benchmarks

mwpastore:
Switch the garbage collector and simplify MemAvailable calculations. No
more tweaking, I promise!

[ci fw-only Ruby/sinatra Ruby/sinatra-sequel Ruby/rack-sequel Ruby/roda-sequel]


nbrady-techempower:
Tweak as much as you think is necessary! Appreciate your work on this and hope the results payoff!

mwpastore:
I know, but I can only worry the same four lines of code so many times. At some point you have to say, "Good enough!"

PR #2535, opened 2017-01-30
Use jemalloc for Ruby 2.2 and 2.4

mwpastore:
[ci fw-only Ruby/sinatra Ruby/sinatra-sequel Ruby/rails Ruby/grape]


nbrady-techempower:
@mwpastore just a note, that if you are changing one of the toolset/setup files that may affect multiple frameworks, let the whole thing run. Just so we can verify it doesn't break anything else.

PR #2538, opened 2017-02-01
Added framework: Rust/rouille

gkbrk:


nbrady-techempower:
@gkbrk Thanks again for your work! I'm happy to add this in. From here on, it would be great to see more db tests for the rust frameworks if you feel inclined!

PR #2540, opened 2017-02-01
Vagrant benchmark.cfg

nbrady-techempower:
I think the decision was to add back the benchmark.cfg.example.

In the meantime this ensures that the bencmark.cfg won't be tracked for vagrant users.


nbrady-techempower:
Forget this, will just put the example back now and copy to the benchmark.cfg.

PR #2541, opened 2017-02-01
Python/api-hour fix

nbrady-techempower:
For two of the updates tests, api_hour is returning the original object. Per [Update Rule #12](http://frameworkbenchmarks.readthedocs.io/en/latest/Project-Information/Framework-Tests/#database-updates) the object needs to be updated, then sent back.


GMLudo:
It looks good to me.

PR #2542, opened 2017-02-01
add benchmark.cfg.example back

nbrady-techempower:
This adds the `benchmark.cfg.example` file back and removes the `benchmark.cfg` file from the repo.

Also updates travis and vagrant to copy from the example file and then make changes instead of having the toolset fallback on the example file.


nbrady-techempower:
Passes travis and a vagrant up, merging

PR #2544, opened 2017-02-01
JavaScript/express fix

nbrady-techempower:
Because isNaN null, is not is a !!NaN.

Anyway, when queries was null in the request, it was being set to NaN.

### DIFF
--- a/frameworks/JavaScript/express/app.js
+++ b/frameworks/JavaScript/express/app.js
@@ -224,7 +224,6 @@ if (cluster.isMaster) {
     var selectFunctions = [],
         queries = Math.min(parseInt(req.query.queries) || 1, 500);
 
-    console.log('QUERIES IS: ' + queries);
     for (var i = 1; i <= queries; i++ ) {
       selectFunctions.push(function(callback) {
         World.findOne({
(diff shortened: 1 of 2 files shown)


frameworks/JavaScript/express/app.js
```diff
@@ -8,7 +8,7 @@ var cluster = require('cluster')
   , express = require('express')
   , Sequelize = require('sequelize')
   , mongoose = require('mongoose')
-  , conn = mongoose.connect('mongodb://localhost/hello_world')
+  , conn = mongoose.connect('mongodb://TFB-database/hello_world')
```
knewmanTE: Is this an intentional change, or should we revert it back to localhost?

frameworks/JavaScript/express/app.js
```diff
@@ -36,7 +36,7 @@ var FortuneSchema = new mongoose.Schema({
   MFortune = conn.model('Fortune', FortuneSchema);
 
 var sequelize = new Sequelize('hello_world', 'benchmarkdbuser', 'benchmarkdbpass', {
-  host: 'localhost',
+  host: 'TFB-database',
```
knewmanTE: Is this an intentional change, or should we revert it back to localhost?

nbrady-techempower: It was unintentional but ok nonetheless. We can actually begin changing all of the statements like this to TFB-database and get rid of the messy sed statements in the setup files.

PR #2545, opened 2017-02-02
Remove capitalized mongo tables

knewmanTE:
For a long time now, we've had 4 MongoDB collections: world, *World*, *fortune*, and *Fortune*. This is causing me a lot of headache on #2536. However, the benefit of this is that #2536 made it very easy to determine which frameworks were using the uppercase *World* and *Fortune* collections because they would fail on their /update tests.

This PR removes the *World* and *Fortune* collections, leaving us with just *world* and *fortune*. It also updates what is hopefully all of the offending frameworks to use the lowercase version.

Issues/weirdness that I'm currently aware of:

- **C++/ffead-cpp**: we are unable to change this at all because the implementation code is currently [hosted on the contributor's repo](https://github.com/sumeetchhetri/ffead-cpp/tree/master/web/te-benchmark). I'm honestly in favor of just removing the framework entirely and pinging the creator asking them to re-submit the framework with a local implementation so we can stay true to the open source nature of our project and prevent any changes that aren't approved by Pull Request.
- **C++/ulib-mongodb**: on my personal Travis run, I was seeing the /update test PASS, but all other tests FAIL. But when I run ulib-mongodb on my local VM, they all PASS. Not really sure what the issue might be there.


knewmanTE:
My guess is that the ulib issue stems from [these travis-specific lines](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/C%2B%2B/ulib/setup_mongodb.sh#L6), though it isn't clear why these are suddenly being problematic when they worked fine before and aren't throwing any errors for the other ulib tests.

knewmanTE:
So I did a bunch of test builds, and what I'm seeing is that ulib is only passing on Travis with the lowercase table names when I also install the uppercase tables, which leads me to believe that it is not respecting the capitalization of the table passed into it. However, this still doesn't explain why /update still works. 😓

knewmanTE:
@stefanocasazza do you have any idea why ulib's UMongoDBClient might be behaving this way when I try to it to use the lowercase `world` and `fortune` collections in favor of the `World` and `Fortune` collections?

greenlaw110:
@knewmanTE Java/act also need to do the same thing. Before I commit the change, should I also change the table name of mysql and postgresql cases?

knewmanTE:
@greenlaw110 as far as I know, if your MySQL and Postgres implementations are passing at the moment, then you shouldn't have to worry about changing those.

@stefanocasazza it looks like we're getting [a segfault](https://travis-ci.org/knewmanTE/FrameworkBenchmarks/jobs/197794759#L4284), which leads me to believe there might be something a little funky with the driver?

Also, I've figured out why /update passes when the others fail. See #2546 for more info.

greenlaw110:
@knewmanTE https://github.com/TechEmpower/FrameworkBenchmarks/pull/2547 created for updating Java/act mongodb collection names

knewmanTE:
C++/ulib-mongodb passing on my local Travis!!!! Hoping this should be merge-able once the Travis run finishes.

knewmanTE:
Update on the currently broken tests:

- **C++/ffead-cpp**: we can't change the source code, so I will just disable the Fortune tests until they are fixed.
- **Haskell/yesod**: will hopefully be fixed in my most recent push.
- **Java/act**: addressed in #2553
- **Lua/octopus**: not sure what the deal is. Tests are not using MongoDB and it's failing on master as well.
- **Perl/kelp**: fixed MongoDB test in my recent push. Not sure why the base MySQL test is failing on Travis, as it's passing locally for me.
- **Python/tornado**: fixed in my recent push.

frameworks/Python/tornado/server_py3.py
```diff
@@ -22,7 +22,7 @@
 class SingleQueryHandler(JsonHandler):
 
     async def get(self):
-        world = await db.World.find_one(randint(1, 10000))
+        world = await db.dorld.find_one(randint(1, 10000))
```
cjnething: I don't think you meant to change it to "dorld" :)

knewmanTE:
Great! With @nbrady-techempower's dorld fix, I think the only remaining issue will be Lua/octopus.

knewmanTE:
@nbrady-techempower the only failing test in the most recent run is Lua/octopus. Are you cool merging this in with the caveat that #2557 is open?

nbrady-techempower:
Definitely. Octopus was failing before too. Awesome work!

Issue #2546, opened 2017-02-02
C++/ulib-mongodb only making a single trip with its Update handler


knewmanTE:
I'm not particularly familiar with C++, but from what I can tell reading the ulib-mongodb Update handler, it looks like it's doing the following:

1. [Creating a Vector to hold up to 500 World objects](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/C%2B%2B/ulib/src/mupdate.usp#L37)
2. [Instantiating id and rnum variables](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/C%2B%2B/ulib/src/mupdate.usp#L62)
3. [Making a single call to both set id and rnum and update the database](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/C%2B%2B/ulib/src/mupdate.usp#L72)
4. [Creating a new world object from id and rnum](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/C%2B%2B/ulib/src/mupdate.usp#L79)
5. [Pushing the World object to the Vector](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/C%2B%2B/ulib/src/mupdate.usp#L81)
6. [Returning the Vector](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/C%2B%2B/ulib/src/mupdate.usp#L96)

There are a couple issues with this implementation:

- The Update test specifies that you need to make 2 trips: one to retrieve an object from a random ID and one to update it in the database. So far as I can tell, this test is doing all of that in a single request.
- It doesn't account for cases where the database request fails. I came to this conclusion in my testing in #2545. Because we are setting *id* and *rnum* within the updateOneBulk function call, even if the call fails, the variables are still updated, so the World object is instantiated without a hitch. This is definitely problematic, as it means that Travis could receive 500 {id, randomNumber} pairs back and emit a PASS when no items have actually been retrieved or updated in the database.

PR #2548, opened 2017-02-03
Ulib: code update + address issue #2546

stefanocasazza:


knewmanTE:
Awesome. Thanks so much for the quick turnaround on the fixes, @stefanocasazza! We really appreciate your help!

PR #2549, opened 2017-02-03
JavaScript/nodejs fix

nbrady-techempower:
The node mongo-raw handler was doing a `findAndModify` which is against our update test [rules](http://frameworkbenchmarks.readthedocs.io/en/latest/Project-Information/Framework-Tests/#database-updates). Resolved the problem from #2536.

I also es6-yfied the handlers and removed the `sed` statements while I was in there.


knewmanTE:
That good good ES6 :squirrel:

knewmanTE:
Hmmm. Weird fails from other frameworks. 😕

nbrady-techempower:
After restarting the job, looks like some keyserver issues for a few of the java frameworks.. I'd like to get to the bottom of that today, and then the ringo-convenient one is failing because it's no longer in master.

PR #2550, opened 2017-02-03
JavaScript/ringojs combine with ringojs-convenient

nbrady-techempower:
This combines the ringojs and ringojs-convenient tests into one test directory.


knewmanTE:
lgtm

PR #2551, opened 2017-02-03
correcting returned value in flask-mysql-raw - updates

samuelcolvin:
Previously the newly generated random number was returned, not the value returned by the select query.

By the way (and I'm well aware it's extremely difficult to change now) in these tests it's impossible for you to validate that the server has returned (or updated) the correct value. If there was a pattern to the expected values or those being set the `validate` check the server is acting correctly.

Currently, you have to take it on trust or read the code very carefully.

### DIFF
--- a/frameworks/Python/flask/app.py
+++ b/frameworks/Python/flask/app.py
@@ -173,10 +173,11 @@ def raw_updates():
         worlds = []
         rp = partial(randint, 1, 10000)
         for i in xrange(num_queries):
-            world = connection.execute("SELECT * FROM World WHERE id=%s", (rp(),)).fetchone()
+            world_id = rp()
+            world = connection.execute("SELECT randomnumber FROM World WHERE id=%s", (world_id,)).fetchone()
+            worlds.append({'id': world_id, 'randomNumber': world['randomnumber']})
             randomNumber = rp()
-            worlds.append({'id': world['id'], 'randomNumber': randomNumber})
-            connection.execute("UPDATE World SET randomNumber=%s WHERE id=%s", (randomNumber, world['id']))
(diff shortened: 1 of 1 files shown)


knewmanTE:
Thanks for the PR @samuelcolvin! I'm actually in the midst of adding in some better validation for the Update test that will verify that values in the database have actually been changed. You can see #2536 for more info.

samuelcolvin:
Makes sense, but I guess it's still impossible to know that the values returned were actually taken (correctly) from the database?

Without reading the code you can't know that the returned values weren't randomly generated without ever touching the db. Or am I missing something?

knewmanTE:
We circumvent this by ignoring the framework's response entirely (at least for this subset of the verification test). Instead, we use the framework's benchmark_config file to determine which database it is using, grab all 10,000 World items before the /update URLs are hit, and then grab the 10,000 World items again afterwards, comparing the two sets to ensure that we have roughly as many changes as we're expecting (with a small margin of error allowing for poor pseudo-random number generators or situations where the same ID happens to be updated multiple times in one request).

knewmanTE:
@samuelcolvin looking through this, your /update response should contain the updated randomNumber value, not the previously held value. So the process should be:

1. Get a random World from the database
2. Update that World's randomNumber
3. Persist that change to the database
4. Return a JSON list of all of the updated World objects that you created in step 2

samuelcolvin:
Makes sense, my mistake. I'll update aiohttp.

knewmanTE:
No worries! We try our best to keep our specifications as clear as possible, but I admit that some nuanced aspects (like what World to return in the update request) can be a bit confusing.

PR #2552, opened 2017-02-03
PHP fixes

nbrady-techempower:
This addresses issues found while working on the update verification. Also removes `sed` statements related to `DBHOST`


nbrady-techempower:
All pass except octopus

PR #2559, opened 2017-02-07
Lua/octopus fix

nbrady-techempower:
This is a bit of a hacky solution to get octopus working again. The zlib change needs to be made with the framework owner's repo, as it's currently broken.

This will only work until zlib updates again, so if we see Octopus breaking in the future we should consider removing this benchmark until it's resolved by the owner.

### DIFF
--- a/frameworks/Lua/octopus/config.lua
+++ b/frameworks/Lua/octopus/config.lua
@@ -21,7 +21,7 @@ return {
 	
 	databaseConnection = {
 		rdbms       =   "mysql",
-		host        =   os.getenv("DBHOST"),
+		host        =   "TFB-database",
 		port        =   3306, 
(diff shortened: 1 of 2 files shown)


knewmanTE:
While we're at it, why not change this line of *server.sh*:

```
zlib_url=http://zlib.net/zlib-$zlib_version.tar.gz
```

to this:
```
zlib_url=https://github.com/madler/zlib/archive/v$zlib_version.tar.gz
```
so that even if the zlib version is upgraded again, it won't break.

nbrady-techempower:
Why not? I'll tell you why not!

I have no reason why not. In fact, I'll replace the other sed line with that such that it's still using the version the framework author originally intended.

knewmanTE:
Yay hacks!

PR #2564, opened 2017-02-12
Upgrade Finagle and Finch benchmarks

vkostyukov:
As of Finch 0.13, there is no need to set `Date` header manually so the `currentTime` thing is removed. This PR also bumps Finagle's dependency to 6.42 (which was released a week ago).


nbrady-techempower:
lgtm!

PR #2566, opened 2017-02-13
Vagrant Overhaul

### LINKED ISSUE #2562: Clean up deployment directory
I haven't used any of the deployment scripts in a while, but given that we've had some substantial toolset changes recently and there are some subdirectories of the [deployment directory](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/deployment) that haven't been updated in 2-3 years, I think we can pretty safely remove some of them?

@nbrady-techempower, I know you've worked recently on updating the vagrant-common build. Is there any utility in having the other 4 vagrant builds? 

We have an Azure build there, but it's 3 years old and for Round 13 we had a good amount of success using the standard manual build process to set up our Azure environment.

Also ping @msmith-techempower, who may have some thoughts on the matter.

nbrady-techempower:
This is currently a work in progress.

Most of the implementations we had for deployments are broken. The goal of this PR is to simplify the deployment process for vagrant. I've removed a lot of the cruft, and am loading vagrant with the same setup we've been doing for travis and what we ask others to do in the installation guide.

This is pretty close. One issue that I'm having is getting syncing working for Windows. There have been some updates to vagrant that I'm looking in to, however the `smb` option doesn't seem like a good approach from the cli. 

Using `vm.synced_folder` with the default options seems to work fine with unix but when it syncs with windows it changes all the files to executable on the vm. This may not matter in the long run if the host vm is the one making the changes and pushing to a git repo.

Doing more testing and will update accordingly. Any comments/suggestions welcome!

Resolves #2562


nbrady-techempower:
Getting some issues with running tests while synced to a windows host: http://sprunge.us/VbMg

nbrady-techempower:
Ok here's where some of the problems lied and the solutions:

### Shared folders for everyone

One of the things we've been trying to tackle for some time was allowing windows and *nix user the ability to edit files on their hosts machines (or the vagrant vm) and have those changes reflected in both environments. We ran into several problems with this. Frameworks like node try to do symlinking (npm is trying to do it in this case) with shared directories, which would just fail. Also removing the installs/results directory was a hassle if it was part of the shared folders on windows.

Here's the solution: When vagrant spins up a new vm we are **only** sharing the `frameworks/` and `toolset/` directory. The `benchmark.cfg` is created on the fly specifically for the vm so the original on the host machine is not touched as well. None of the other files are needed on the vm.. `.git/`, `.travis.yml`, etc. 

With this setup, the vm still has everything it needs to run the suite, the `results/` and `installs/` directories will only exist in the vm for the tests run on the vm, and if the host is also an ubuntu box, you can run the suite there without conflicting results.

The only change here for windows users is that you need to run `vagrant up` from a shell with administrator privileges. If you don't, everything will look like it set up fine, but tests, like nodejs, will fail.

This implementation is also faster. There's no pulling from git. The repo/branch that you do the `vagrant up` in will be the one that exists in the vm. However, because they're shared folders, if you switch branches on the host, it will be reflected remotely.

nbrady-techempower:
Merging. See https://github.com/TechEmpower/TFB-Documentation/pull/64 for corresponding docs

PR #2569, opened 2017-02-14
Act optimize update test

greenlaw110:


nbrady-techempower:
Nice work @greenlaw110 ! This will fix act for #2536

PR #2575, opened 2017-02-21
Update all the JavaScripts

nbrady-techempower:
Node Server
- Upgrade from v6 to v7.5

HapiJS
- Updated framework to newest major version
- Updated code to available es2015 for the v8 engine
- Updated all package dependencies

Express
- Updated code to available es2015 for the v8 engine
- Updated all package dependencies
- const Promise = require('bluebird');

NodeJS
- Updated code to available es2015 for the v8 engine
- Fixed WARN on plaintext test
- Use native Promise instead of bluebird
- Updated all package dependencies

Koa
- Updated code to available es2015 for the v8 engine (sadness that fat arrow generators are not available yet)
- Updated all package dependencies
- Added proper source_code file
- Added README

RingoJS
- Update framework to first major release
- Update code to available es2015 (no arrow functions available yet)
- Remove deprecated mustache package (fortune for non-convenvient test will have to be re-implemented)


frameworks/JavaScript/ringojs/app/models.js
```diff
@@ -1,26 +1,18 @@
...
         "password": "benchmarkdbpass",
         "minimumIdle": 10,
         "maximumPoolSize": 30
     });
 });
-var store = exports.store = new Store(connectionPool);
-var queryCache = module.singleton("queryCache", function() {
-    return new Cache(10000);
-});
-store.setQueryCache(queryCache);
```
grob: Please don't remove the query cache. It only caches parsed queries, not database results (see https://github.com/grob/ringo-sqlstore/wiki/Querycache for an explanation).

PR #2580, opened 2017-02-28
Go/fasthttp: fix update test violations

### LINKED ISSUE #2579: [Go/fasthttp] Benchmark requirement violation(s)
The `updateHandler()` method in `server-postgresql.go` and `server-mysql.go` uses a single transaction to perform multiple writes. Per the benchmark requirements, "a transaction should only encapsulate a single iteration, composed of a single read and single write."

nbrady-techempower:
As pointed out by @mwpastore here, fasthttp was not making multiple transactions as dictated by the update rules.

resolves #2579

Will leave open in case a `Go/fasthttp` contrib wants to do this a different way, but we will need to get in before round 14.

### DIFF
--- a/frameworks/Go/fasthttp/src/server-mysql/server.go
+++ b/frameworks/Go/fasthttp/src/server-mysql/server.go
@@ -117,26 +117,19 @@ func updateHandler(ctx *fasthttp.RequestCtx) {
 
 	worlds := make([]common.World, n)
 	for i := 0; i < n; i++ {
+		txn, err := db.Begin()
 		w := &worlds[i]
+		if err != nil {
+			log.Fatalf("Error starting transaction: %s", err)
+		}
 		fetchRandomWorld(w)
 		w.RandomNumber = int32(common.RandomWorldNum())
-	}
-
-	// sorting is required for insert deadlock prevention.
-	sort.Sort(common.WorldsByID(worlds))
-	txn, err := db.Begin()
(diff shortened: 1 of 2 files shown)


frameworks/Go/fasthttp/src/server-mysql/server.go
```diff
@@ -117,26 +117,19 @@ func updateHandler(ctx *fasthttp.RequestCtx) {
 
 	worlds := make([]common.World, n)
 	for i := 0; i < n; i++ {
+		txn, err := db.Begin()
 		w := &worlds[i]
+		if err != nil {
+			log.Fatalf("Error starting transaction: %s", err)
+		}
```
methane: Maybe, `db.Exec()` instead of explicit transaction is preferred.

valyala:
The original code in `fasthttp-postgresql` shouldn't violate any rules. The transaction is used here only for updates' batching - this is well-known and the most easiest technique for performing batch updates in `postgresql`. The transaction doesn't cover rows' fetching - each row is read by a distinct query at [this line](https://github.com/TechEmpower/FrameworkBenchmarks/blob/a2e309f4ba72e1230dac0b3c79adb0588a5be51a/frameworks/Go/fasthttp/src/server-postgresql/server.go#L109) outside the transaction.

nbrady-techempower:
thanks @valyala for clarification

PR #2582, opened 2017-02-28
[Ruby] Align to clarified benchmark requirements

mwpastore:
[ci fw-only Ruby/rack-sequel Ruby/roda-sequel Ruby/sinatra-sequel Ruby/sinatra]


mwpastore:
Thanks @nbrady-techempower!

PR #2584, opened 2017-03-07
[Ruby] Misc updates to Sinatra/Sequel benchmarks

mwpastore:


nbrady-techempower:
thanks @mwpastore ! Looks like we'll have preview data out today/tomorrow

mwpastore:
Thank you! That's exciting!

PR #2586, opened 2017-03-08
Add CPU_COUNT env variable / get MAX_THREADS for load_gen

nbrady-techempower:
Removes the command line options for threads, concurrency_levels, and query_levels. The latter two remain in the `benchmark.cfg` file. 

### Note
WIP: Moving away from making a `max_threads` env variable available and instead will make a `cpu_count` variable available to frameworks. Will continue to use cpu_count*4 for the `-t` option of the `wrk` commands.


toolset/run-tests.py
```diff
@@ -115,9 +115,9 @@ def main(argv=None):
     if defaults['server_host'] is None:
         defaults['server_host'] = defaults['client_host']
 
-    maxThreads = 8
+    defaults['threads'] = 8
     try:
-        maxThreads = multiprocessing.cpu_count()
+        defaults['threads'] = multiprocessing.cpu_count() * 2
```
msmith-techempower: Doesn't this count the number of cpus on the `application` server and not the `load` server (where it is used)?

nbrady-techempower:
@msmith-techempower please review changes. works for me on vagrant with 2 virtual cores:

```
 Concurrency: 256 for gemini
 wrk -H 'Host: localhost' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 60 -c 256 --timeout 8 -t 8 http://TFB-server:8080/plaintext -s ~/pipeline.lua -- 16
```

nbrady-techempower:
This may need more work, since extra frameworks fail on travis with this.

frameworks/C++/cutelyst/config.sh
```diff
@@ -47,9 +48,9 @@ sed -i "s|SendDate=.*|SendDate=${SEND_DATE}|g" ${CROOT}/config.ini
 export LD_LIBRARY_PATH=/opt/qt${QT_VERSION_MM}/lib:${CROOT}/lib/x86_64-linux-gnu/
 
 if [ -n "${UWSGI}" ]; then
-  uwsgi --ini ${CROOT}/config.ini --plugin ${CROOT}/lib/uwsgi/plugins/cutelyst_plugin.so --cutelyst-app ${CROOT}/benchmarks/src/libcutelyst_benchmarks.so ${PROCESS_OR_THREAD} $MAX_THREADS &
+  uwsgi --ini ${IROOT}/config.ini --cutelyst-app ${IROOT}/cutelyst-benchmarks/src/libcutelyst_benchmarks.so ${PROCESS_OR_THREAD} $CPU_COUNT &
 else
-  ${CROOT}/bin/cutelyst-wsgi --ini ${CROOT}/config.ini -a ${CROOT}/benchmarks/src/libcutelyst_benchmarks.so ${PROCESS_OR_THREAD} $MAX_THREADS --socket-timeout 0 ${BALANCER} &
+  ${IROOT}/bin/cutelyst-wsgi --ini ${IROOT}/config.ini -a ${IROOT}/cutelyst-benchmarks/src/libcutelyst_benchmarks.so ${PROCESS_OR_THREAD} $CPU_COUNT --socket-timeout 0 &
```
michaelhixson: This undid some other recent changes in the cutelyst config.  I suggest starting over with a copy of this file from master, then doing a search and replace for MAX_THREADS with CPU_COUNT.

frameworks/C++/wt/setup.sh
```diff
@@ -6,4 +6,4 @@ sed -i 's|INSERT_DB_HOST_HERE|'"${DBHOST}"'|g' benchmark.cpp
 
 g++-4.8 -O3 -DNDEBUG -std=c++0x -L${BOOST_LIB} -I${BOOST_INC} -L${WT_LIB} -I${WT_INC} -o benchmark.wt benchmark.cpp -lwt -lwthttp -lwtdbo -lwtdbomysql -lboost_thread -lboost_system
 
-./benchmark.wt -c wt_config.xml -t ${MAX_THREADS} --docroot . --http-address 0.0.0.0 --http-port 8080 --accesslog=- --no-compression
+./benchmark.wt -c wt_config.xml -t ${CPU_COUNT} --docroot . --http-address 0.0.0.0 --http-port 8080 --accesslog=- --no-compression &
```
michaelhixson: Could you remove the trailing `&` again?  I removed it intentionally here https://github.com/nbrady-techempower/FrameworkBenchmarks/commit/84d97ea28a8b11d8da9932adf75a1b4cf220c4c7

frameworks/C++/wt/setup_postgres.sh
```diff
@@ -6,4 +6,4 @@ sed -i 's|INSERT_DB_HOST_HERE|'"${DBHOST}"'|g' benchmark.cpp
 
 g++-4.8 -O3 -DNDEBUG -DBENCHMARK_USE_POSTGRES -std=c++0x -L${BOOST_LIB} -I${BOOST_INC} -L${WT_LIB} -I${WT_INC} -o benchmark_postgres.wt benchmark.cpp -lwt -lwthttp -lwtdbo -lwtdbopostgres -lboost_thread -lboost_system
 
-./benchmark_postgres.wt -c wt_config.xml -t ${MAX_THREADS} --docroot . --http-address 0.0.0.0 --http-port 8080 --accesslog=- --no-compression
+./benchmark_postgres.wt -c wt_config.xml -t ${CPU_COUNT} --docroot . --http-address 0.0.0.0 --http-port 8080 --accesslog=- --no-compression &
```
michaelhixson: ditto for this `&`

toolset/run-tests.py
```diff
@@ -149,9 +145,6 @@ def main(argv=None):
     parser.add_argument('--list-tests', action='store_true', default=False, help='lists all the known tests that can run')
 
     # Benchmark options
-    parser.add_argument('--concurrency-levels', default=[8, 16, 32, 64, 128, 256], help='Runs wrk benchmarker with different concurrency value (type int-sequence)', action=StoreSeqAction)
-    parser.add_argument('--query-levels', default=[1, 5,10,15,20], help='Database queries requested per HTTP connection, used during query test (type int-sequence)', action=StoreSeqAction)
```
michaelhixson: With this, `concurrency_levels` and `query_levels` will no longer be configurable, right?  If so, how about removing those lines from `benchmark.cfg.example`?

michaelhixson:
`frameworks/Dart/dart-raw/setup.sh` uses `MAX_THREADS` and I don't see it here.  Is it possible others were missed?

frameworks/C++/wt/setup.sh
nbrady-techempower: Ah, sorry about that. When I was fixing conflicts I missed that in the diff.

nbrady-techempower:
@michaelhixson Ok, I believe I've addressed everything. There might be another framework out there but I think I got them all. Some are defining there own MAX_THREADS variable so i couldn't just replace all. Didn't want those to lose their meaning.

michaelhixson:
LGTM then

nbrady-techempower:
@michaelhixson fyi, the concurrency and query levels need to stay in the `cfg`. I had forgotten that @msmith-techempower had wanted to just remove some extra noise from the cli args.

PR #2591, opened 2017-03-13
cutelyst: Update to 1.5.0

dantti:

### DIFF
--- a/toolset/setup/linux/frameworks/cutelyst.sh
+++ b/toolset/setup/linux/frameworks/cutelyst.sh
@@ -2,7 +2,7 @@
 
 fw_installed cutelyst && return 0
 
-CUTELYST_VER=r1.4.0
+CUTELYST_VER=v1.5.0
 QT_VERSION_MM=56
 QT_VERSION_FULL=562-trusty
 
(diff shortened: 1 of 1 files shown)


nbrady-techempower:
hey @dantti !

So this failed here, and it also failed on the preview run we just did (as did a few other C/C++ fw's)

dantti:
Did the previous version ran fine on the preview or you tested only this one?

nbrady-techempower:
It was the previous version that failed. It's also failed on a full travis run recently here: https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/209848247

dantti:
@nbrady-techempower Fixed the issue, and the link you pointed out failed due out of memory to build cutelyst, since the build script uses the number of cores to build maybe it would be a good idea to fix CPU cores to 1 when running on Travis.

nbrady-techempower:
Hmm, been trying my best to not have any Travis specific code polluting the toolset. Will see if there's another way to resolve. Thanks for all your support!

dantti:
@nbrady-techempower I can also change the make line to not use all cores if this keeps happening, the ideal solution would be to have a PPA with cutelyst so it would only build the test app, but current Ubuntu LTS doesn't even has the minimum Qt required and the toolset is using even older LTS...

PR #2594, opened 2017-03-16
Add flag to suppress most stuff written to stdout and stderr

michaelhixson:
This is designed for use with continuous benchmarking.

Currently, we run continuous benchmarking as an upstart script.  By
default it puts everything sent to stdout into a file like
/var/log/upstart/tfb.log.  This log file is absolutely enormous -- it
contains everything printed by every framework.  It's so big that it
fills up the disk of our test environments before a single iteration of
the full test suite completes.

We work around this by disabling the upstart log completely (declaring
"console none" in its config), but that's not ideal because then we have
no logs at all.

With this change, with quiet mode enabled, we print much, much less to
stdout.  There should only be a few lines printed per framework.  This
will let us enable the upstart logs again, which should be useful when
trying to answer questions like, "Which framework are we testing right
now?"

### DIFF
--- a/toolset/benchmark/benchmarker.py
+++ b/toolset/benchmark/benchmarker.py
@@ -24,6 +24,8 @@ import threading
 import textwrap
 from pprint import pprint
 
+from contextlib import contextmanager
+
 from multiprocessing import Process
 
 from datetime import datetime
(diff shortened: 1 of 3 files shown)


nbrady-techempower:
Oh, awesome!

PR #2596, opened 2017-03-20
[Ruby] Bump Roda and add :optimize_model_load to Sequel benchmarks

mwpastore:
[ci fw-only Ruby/rack-sequel Ruby/roda-sequel Ruby/sinatra-sequel]


mwpastore:
Thank you!

PR #2597, opened 2017-03-21
Implement prepared statement updating in PO (plain old) Servlets

zloster:
- Implement `SQL` prepared statement updates 
- Update `Jackson` library to latest (2.7.8) - on my local DEV virtual machine gives around 15% improvement in throughput
- Update `Apache Commons Lang` to latest (3.5)
- Switch to `Java8` byte-code compilation
- Disable debug info in compiled byte-code *
- Enable optimized compilation *
- Change test metadata to allow showing of information in `Framework overhead` tab for tests with database operations

_*_ These two give some small but measurable improvement in throughput


zloster:
That was fast :+1:

PR #2603, opened 2017-03-22
Fix broken wicket db tests by downgrading the mysql connector

michaelhixson:
Version 6.x of the connector is apparently more opinionated about time zones.  All of the db tests were failing with exceptions like this:

```
java.sql.SQLException: The server time zone value 'CDT' is
unrecognized or represents more than one time zone. You must
configure either the server or JDBC driver (via the serverTimezone
configuration property) to use a more specifc time zone value if you
want to utilize time zone support.
```

Maybe it's making a valid point that we should address, maybe not.  But right now mysql-connector-java 6.x is a "development release".  Let's use the stable release (5.x) instead, which is what every other framework that uses Java and mysql is doing already.

### DIFF
--- a/frameworks/Java/wicket/pom.xml
+++ b/frameworks/Java/wicket/pom.xml
@@ -74,7 +74,7 @@
 		<dependency>
 			<groupId>mysql</groupId>
 			<artifactId>mysql-connector-java</artifactId>
-			<version>6.0.5</version>
+			<version>5.1.41</version>
 		</dependency>
 	</dependencies>
 	<build>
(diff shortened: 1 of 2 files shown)


martin-g:
The tests pass here and AFAIK they pass at Travis too.
I guess your local timezone breaks it somehow.
But I don't mind downgrading it to 5.x!

martin-g:
It seems the DB tests fail for Round 14 Preview 1: https://www.techempower.com/benchmarks/previews/round14/r13-vs-r14p1.html
So definitely please merge this PR!
Thank you, @michaelhixson !

PR #2607, opened 2017-03-22
H2O: Improve load balancing and other enhancements

volyrique:
The main change is to improve the way incoming connections are distributed to the worker threads started by the framework using the SO_REUSEPORT socket option, but there are also some error logging enhancements and various other minor fixes.


volyrique:
Only the cutelyst tests fail, but they are not related to my changes at all.

PR #2609, opened 2017-03-23
upgrade aiohttp to major new version 2

samuelcolvin:
This should give a significant improvement in performance.


nbrady-techempower:
r14 preview 2 was just posted.. normally i'd want to avoid a major version change between a preview round and final, but it looks like it took a bit of a hit in performance so we'll get this in and see if it corrects anything.

samuelcolvin:
Thanks, this should make a fair bit of difference.

PR #2610, opened 2017-03-23
Isolate cutelyst install to installs/cutelyst, let it use sudo again

michaelhixson:
The use of sudo was breaking other frameworks because it created
root-owned directories in installs that the other frameworks could not
use.  However, removing sudo broke the cutelyst install.

This change lets cutelyst use sudo again but puts all of the installed
files in installs/cutelyst, meaning that it should no longer break other
frameworks (tested by running "toolset/run-tests.py --mode verify --test
cutelyst ffead-cpp").

### DIFF
--- a/frameworks/C++/cutelyst/config.sh
+++ b/frameworks/C++/cutelyst/config.sh
@@ -21,40 +21,41 @@ elif [ "${DRIVER}" == "QPSQL" ]; then
   fw_depends postgresql
 fi
 
+CROOT=${IROOT}/cutelyst
 
-mkdir ${IROOT}/cutelyst-benchmarks || true
-cd ${IROOT}/cutelyst-benchmarks
+mkdir -p ${CROOT}/benchmarks || true
+cd ${CROOT}/benchmarks
 
 # build
-export CMAKE_PREFIX_PATH=/opt/qt${QT_VERSION_MM}:${IROOT}
-cmake $TROOT -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$IROOT
+export CMAKE_PREFIX_PATH=/opt/qt${QT_VERSION_MM}:${CROOT}
+cmake $TROOT -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$CROOT
 make -j $MAX_THREADS
 
 if [ -n "${UWSGI}" ]; then
-  cp -v ${TROOT}/config/config_socket.ini ${IROOT}/config.ini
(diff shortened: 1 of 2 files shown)


dantti:
Thanks @michaelhixson for finding the issue but you can fix this without using `sudo`:
Just set the two lines like CMAKE_INSTALL_PREFIX=$IROOT/cutelyst
and 
export LD_LIBRARY_PATH=/opt/qt${QT_VERSION_MM}/lib:${IROOT}/cutelyst/lib/x86_64-linux-gnu/

dantti:
hmmm thinking again having a CROOT will make things looks cleaner, does the sudo is needed in this case too? I'm asking because it *should* work without root rights so maybe there's some fix I need to do to the build system or to the tests...

michaelhixson:
@dantti Without sudo, it looks like it fails trying to add a file to the uwsgi plugins directory (which is in `/usr/lib/uwsgi` and is owned by root).  Here's a snippet of the output:

```
Setup cutelyst: -- Installing: /usr/lib/uwsgi/plugins/cutelyst_plugin.so
Setup cutelyst: CMake Error at uwsgiEngine/cmake_install.cmake:50 (file):
Setup cutelyst:   file INSTALL cannot copy file
Setup cutelyst:   "/home/techempower/FrameworkBenchmarks/installs/cutelyst/cutelyst-1.5.0/build/uwsgiEngine/cutelyst_plugin.so"
Setup cutelyst:   to "/usr/lib/uwsgi/plugins/cutelyst_plugin.so".
Setup cutelyst: Call Stack (most recent call first):
Setup cutelyst:   cmake_install.cmake:90 (include)
Setup cutelyst:
Setup cutelyst:
Setup cutelyst: make: *** [install] Error 1
```

msmith-techempower:
Why is `uwsgi` in `/usr/lib`?

It should **also** be in `$IROOT`

dantti:
Hmm ok, well that's where `uwsgi` is going to look for it, I can fix it in CMakeLists.txt, but then uwsgi needs the full cutelyst_plugin.so path to load it. Is a point Cutelyst release acceptable at this time? Or should I care only for r15?

michaelhixson:
We would like to fix the sudo issue eventually, but for the sake of getting Round 14 out the door, I'm going to merge this PR as-is.

dantti:
Oh, I just remembered that there is actually a flag for that plugin be installed elsewhere:
UWSGI_PLUGINS_DIR
I'll create a PR that reverts this and set the plugin full path so that uwsgi finds it.

PR #2611, opened 2017-03-24
Fix hyper

polachok:
Upgrades hyper to latest master

### DIFF
--- a/frameworks/Rust/hyper/src/main.rs
+++ b/frameworks/Rust/hyper/src/main.rs
@@ -5,15 +5,14 @@ extern crate pretty_env_logger;
 extern crate log;
 extern crate net2;
 extern crate tokio_core;
+extern crate tokio_proto;
 extern crate num_cpus;
 
 use hyper::{Get, StatusCode};
-use hyper::header::{ContentLength, ContentType};
-use hyper::server::{Server, Service, Request, Response};
+use hyper::header::{self, ContentLength, ContentType};
+use hyper::server::{Http, Service, Request, Response};
 
-use net2::TcpBuilder;
-use net2::unix::UnixTcpBuilderExt;
-use tokio_core::net::TcpListener;
+use tokio_proto::TcpServer;
 
 static HELLOWORLD: &'static [u8] = b"Hello, world!";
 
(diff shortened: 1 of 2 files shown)


frameworks/Rust/hyper/src/main.rs
```diff
@@ -33,8 +32,9 @@ impl Service for TechEmpower {
                 Response::new()
                     .with_header(ContentLength(HELLOWORLD.len() as u64))
                     .with_header(ContentType(Mime(TopLevel::Text, SubLevel::Plain, vec![(Attr::Charset, Value::Utf8)])))
+                    .with_header(header::Server("hyper/async".to_owned()))
```
steveklabnik: standard rust style is `to_string` not `to_owned` (but they do the exact same thing)

alexcrichton:
Oh one strategy this may wish to take as well is to check in `Cargo.lock` and avoid the need for `rev = "..."` in the dependency specification, but that's completely optional!

steveklabnik:
@alexcrichton I don't remember why I didn't do it that way, but I think there was some kind of issue. Can't remember; that was like two years ago or something. I could have messed it up too.

alexcrichton:
Ah yeah [tokio-minihttp](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Rust/tokio-minihttp/Cargo.lock) nowdays has `Cargo.lock`, so I figured it'd be ok to throw in here as well (but again, just an optional extension)

frameworks/Rust/hyper/src/main.rs
polachok: fixed

arthurprs:
It's a bit naive to allocate a string on every reques just to set the server header. I'm sure it doesn't matter in practice but it may show in a hello-world benchmark

polachok:
@arthurprs well, that's how Server is defined https://hyper.rs/hyper/0.8.0/hyper/header/struct.Server.html

arthurprs:
I see, `Server` is also required by the test rules.

nbrady-techempower:
Looks like there's agreement to close this in favor of #2619 

Closing this and merging that. Thanks guys!

PR #2612, opened 2017-03-24
Cutelyst no sudo

dantti:
This should allow for Cutelyst make install to work without `sudo`

### DIFF
--- a/frameworks/C++/cutelyst/config.sh
+++ b/frameworks/C++/cutelyst/config.sh
@@ -46,7 +46,7 @@ sed -i "s|SendDate=.*|SendDate=${SEND_DATE}|g" ${CROOT}/config.ini
 export LD_LIBRARY_PATH=/opt/qt${QT_VERSION_MM}/lib:${CROOT}/lib/x86_64-linux-gnu/
 
 if [ -n "${UWSGI}" ]; then
-  uwsgi --ini ${CROOT}/config.ini --cutelyst-app ${CROOT}/benchmarks/src/libcutelyst_benchmarks.so ${PROCESS_OR_THREAD} $MAX_THREADS &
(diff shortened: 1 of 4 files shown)


michaelhixson:
Thanks for looking into this.

The `cutelyst-nginx-mysql-raw` and `cutelyst-nginx-postgres-raw` tests appear to fail with this change.  I see errors like this when the tests are running:

```
Server cutelyst-nginx-mysql-raw: default[warning] QSqlQuery::prepare: database not open
Server cutelyst-nginx-mysql-raw: cutelyst.utils.sql[critical] Failed to prepare query: "SELECT id, message FROM fortune" "Driver not loaded"
```

```
Server cutelyst-nginx-postgres-raw: default[warning] QSqlQuery::prepare: database not open
Server cutelyst-nginx-postgres-raw: cutelyst.utils.sql[critical] Failed to prepare query: "SELECT id, message FROM fortune" "Driver not loaded"
```

The `cutelyst-thread-mysql-raw` and `cutelyst-thread-postgres-raw` tests still pass though.

Also, we would still like to keep cutelyst's installed files as isolated as possible.  Putting the files in IROOT/cutelyst was a desirable change by itself, separate from the sudo issue.  We don't want frameworks to create files or directories like IROOT/bin, IROOT/lib, IROOT/include, IROOT/config.ini, IROOT/nginx.conf, for example.  (It's not important that the directory is named exactly IROOT/cutelyst or that there's a variable in the code named CROOT.)

dantti:
Ok, I'll remove the revert then and investigade why did it failed..

dantti:
I could not reproduce the issue on vagrant, let's see if it works now that I didn't reverted your changes on Travis

dantti:
great it passed... :stuck_out_tongue:

michaelhixson:
It passes on our in-house TFB machines now too.  Thanks!

PR #2614, opened 2017-03-24
More idiomatic symbol-to-proc usage in Rails controller

nateberkopec:
Also use destructive variant of sort_by, grab presorted fortunes.

### DIFF
--- a/frameworks/Ruby/rails/app/controllers/hello_world_controller.rb
+++ b/frameworks/Ruby/rails/app/controllers/hello_world_controller.rb
@@ -24,9 +24,9 @@ class HelloWorldController < ApplicationController
   end
   
   def fortune
-    @fortunes = Fortune.all
+    @fortunes = Fortune.order(:message)
     @fortunes << Fortune.new(:id => 0, :message => "Additional fortune added at request time.")
-    @fortunes = @fortunes.sort_by { |x| x.message }
+    @fortunes.sort_by!(&:message)
   end
 
   def update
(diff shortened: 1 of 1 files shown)


frameworks/Ruby/rails/app/controllers/hello_world_controller.rb
```diff
@@ -24,9 +24,9 @@ def query
   end
   
   def fortune
-    @fortunes = Fortune.all
+    @fortunes = Fortune.order(:message)
     @fortunes << Fortune.new(:id => 0, :message => "Additional fortune added at request time.")
-    @fortunes = @fortunes.sort_by { |x| x.message }
+    @fortunes.sort_by!(&:message)
```
rjnienaber: I might be reading the code incorrectly but isn't this sorting twice now? Once in the database and then again in the Ruby?. I also have a feeling that I benchmarked this and it was faster to sort in code, rather then to do it in the database.

nateberkopec: My hypothesis was that it would make the Ruby-sorting faster, but it looks like using ORDER BY is actually prohibited by the rules anyway, so I'll remove it.

nbrady-techempower:
@nateberkopec Thanks for the contribution. Looks like you're going to take care of the order by?

nbrady-techempower:
@nateberkopec I've removed order by for you. If there's anything else, let me know. Otherwise will merge if it passes travis

nbrady-techempower:
All of the fortune tests still failing here anyway, so going to close this since it's only a one line change that looks like it needs to be reworked.

PR #2616, opened 2017-03-24
Rails: skip validation on updates

nateberkopec:
Also, only select the fields required.

I wish there was a way to generate X select queries and then just a single update, but I don't think that's possible.


rjnienaber:
> I wish there was a way to generate X select queries and then just a single update, but I don't think that's possible.

When I updated the test I used [activerecord-import](https://github.com/zdennis/activerecord-import) to do that but it looks like it got removed on [the version update](https://github.com/TechEmpower/FrameworkBenchmarks/commit/1bb9fbc3c8946462b7ed5a64a0ac37e15b8a513c#diff-ef6d4483fde2c49f4ca8e7e5cee7059a).

nbrady-techempower:
@rjnienaber @nateberkopec I remember there being several issues and we were just trying to get these tests back to a working state. I'll merge this in but feel free to re-implement activerecord-import if you'd like.

PR #2617, opened 2017-03-24
Capitalize "Rust" for consistency with other language IDs

bstrie:
I've noticed that on the preliminary results page for round 14 that frameworks written in Rust are abbreviated as "rus" under the Language column and "Rus" under the Platform column, which mildly annoys me. :) I didn't find the string "rus" anywhere (or other abbreviations like "ngx") so I'm assuming that you have an algorithm somewhere that generates abbreviations based on benchmark_configs, where the string "rust" was present as the "Language" field for every benchmark but hyper.


scottmmjackson:
LGTM, thanks for contributing! :+1:

PR #2619, opened 2017-03-25
Rust: hyper on tokio-proto/service

nayato:
Refresh for hyper benchmark based on tokio-proto and tokio-service.
serde-derive is used for json serialization as that's what people
will do in real app.

### DIFF
--- a/frameworks/Rust/hyper/Cargo.toml
+++ b/frameworks/Rust/hyper/Cargo.toml
@@ -1,16 +1,18 @@
 [package]
 name = "hello"
-version = "0.2.0"
+version = "0.3.0"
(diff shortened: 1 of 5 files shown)


nayato:
@seanmonstar @steveklabnik @alexcrichton @polachok PTAL

nayato:
@polachok just noticed a couple hours back that there's #2611. I aligned my changes to your PR now and will rebase once yours is in.

alexcrichton:
Ah yeah I was otherwise just going to recommend https://github.com/TechEmpower/FrameworkBenchmarks/pull/2611, but other than that looks great!

PR #2620, opened 2017-03-25
[ci fw-only Java/act] update act to 1.0.6; set connection pool number

greenlaw110:


nbrady-techempower:
thanks @greenlaw110 this should be reflected in r14 final

PR #2622, opened 2017-03-25
Ruby: Address the perf issues exposed by R14P1.1

mwpastore:


mwpastore:
Uhh, hold off on merging this for a bit. Thanks.

nbrady-techempower:
@mwpastore you're a beast! getting this in now. should be in r14 final but just missed preview 2. thanks as always!

mwpastore:
@nbrady-techempower Thank you sir, I feel like I just keep yak shaving the same three lines of code over and over again, but hopefully it is nearing some approximation of finished!

PR #2623, opened 2017-03-25
Java JSON fix. C# connection fix

zapov:
Separate DB and JSON context.
Update Revenj Java to latest version.

Release connection when thread dies.


nbrady-techempower:
lgtm; @zapov this should be reflected in r14 final, not r14 preview 2. thanks as always!

PR #2624, opened 2017-03-25
Jooby 1.1.0 + Rocker + Requery

jknack:


maxdemarzi:
LGTM let's get this in before Round 14 Preview 2

nbrady-techempower:
lgtm; we actually just posted r14p2 but this should make it for r14 final

PR #2450, opened 2016-12-30
fix Go/echo

methane:
This pull request uses [glide](https://github.com/Masterminds/glide) to version pinning.
I hope all other Go frameworks use same tool.

PR #2625, opened 2017-03-27
[ci fw-only Go/revel] Repair revel tests, make future failures less punishing

michaelhixson:
A couple of function/property names have changed in the revel framework, causing our test code to not compile.  I updated our code to use the new names.

Also, the fact that `setup.sh` was doing compilation in the background meant the failures went unnoticed by our toolset (specifically by `framework_test.py`), causing a long period of inactivity in the test suite.  Removing the `&` fixes that problem, which will be useful in case the revel framework ever changes in a way that breaks our test code again.

There are more details in the commit messages.

### DIFF
--- a/frameworks/Go/revel/setup.sh
+++ b/frameworks/Go/revel/setup.sh
@@ -8,4 +8,4 @@ go get -u github.com/revel/cmd/revel
 go get -u github.com/coocood/qbs
 go get -u github.com/eaigner/jet
 
-bin/revel run benchmark prod &
+bin/revel run benchmark prod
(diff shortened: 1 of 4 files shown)


methane:
@michaelhixson Could you add `[ci fw-only Go/revel]` prefix in pull request title?
And could you use glide to pinning dependency versions? (see #2450).

michaelhixson:
@methane Thanks for looking at this. 

> could you use glide to pinning dependency versions?

That seems like a good idea, but I'm going to leave that for someone else to implement.

PR #2629, opened 2017-03-29
Newer compiler

fredrikwidlund:


nbrady-techempower:
@fredrikwidlund sorry preview 2 was already up and running, but i'll get this in now for r14 final. thanks for your work

PR #2633, opened 2017-03-30
Added API Star to Python frameworks.

tomchristie:
Hi folks,

This PR adds [API Star](https://github.com/tomchristie/apistar) to the Python frameworks.
So far I've only add the JSON and Plaintext tests.

Feel free to let me know if there's anything else I need to address here before this could be considered.

Thanks! ✨🚀✨🌟


nbrady-techempower:
Hi @tomchristie -

Thanks for contributing a framework. This is missing a couple of key things. First, it needs a `benchmark_config.json`. Also, it needs a line added to the `travis.yml` file so travis can pick it up, and a `README.md` similar to [this one](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Perl/mojolicious/README.md) so we can get some more information about the framework.

tomchristie:
Thanks @nbrady-techempower appreciate your guidance there. Have made some updates in line with that now.

nbrady-techempower:
Hey @tomchristie no problem. So:
`Exception: benchmark_config.json for test default is invalid, please amend by adding the following required keys: [orm, database_os, database]
`
Throw those keys in the `benchmark_config.json` with "None" if they don't apply

tomchristie:
Thanks, have added those keys now. I'll be updating this benchmark as things progress, and including the additional benchmark types as soon as possible.

What's the best way to stay fully informed about progress towards the next benchmark? I'd like to make sure I get as many test types as possible included in time for Round 14.

nbrady-techempower:
@tomchristie if you're still doing work, it's not likely this will make it in time for Round 14 final, as we've already done two previews. [Here is a good place to stay informed](https://groups.google.com/forum/#!forum/framework-benchmarks)

tomchristie:
Okay, thanks @nbrady-techempower.

It shouldn't take long for me to add the "queries" test types, so perhaps there's still an outside chance of having those added too, but let's go one step at a time...

I think I've got everything addressed now, just waiting to see if the tests run green.

Will there be any remaining blockers to at least getting these JSON and Plaintext results for API Star into the Round 14 publication, once the tests here pass?

methane:
test failed.
```
Exception: Invalid `database_os` value specified for test "default" in framework "API Star"; suggestions:
  `linux` -- Our best-supported host OS, it is recommended that you build your tests for Linux hosts
  `windows` -- TFB is not fully-compatible on windows, contribute towards our work on compatibility: https://github.com/TechEmpower/FrameworkBenchmarks/milestones/Windows%%20Compatibility
```

frameworks/Python/apistar/app.py
```diff
@@ -0,0 +1,25 @@
+from apistar import App, Route, wsgi
+import ujson as json
+
+
+def json_view() -> wsgi.WSGIResponse:
+    content = json.dumps({'message': 'Welcome to API Star!'}).encode('utf-8')
```
methane: wrong message.

tomchristie:
Thanks @methane! Pushed up changes to address both of those two issues. Crossing fingers that we're there this time.

tomchristie:
Test failure due to Gunicorn missing from requirements.
Added to requirements, and pinned requirements to fixed versions.

nbrady-techempower:
Ah right, good catch @methane 
@tomchristie looks like you're almost there. `plaintext` passing but `json` is not

tomchristie:
Phew, all the way there now. 👍  😎 

Do let me know if there's any last blockers getting this in, in time for round 14.

I might attempt to address the database test types too, but will treat that as a separate PR if I do so, as that shouldn't block us getting the basics in, and making a showing.

nbrady-techempower:
Thanks guys. Since we're doing continuous benchmarking we probably won't be doing hard cutoff dates for fixes/improvements like in the past. I'm going to merge this in, and if it doesn't make Round 14 final, Round 15 will have a much faster turnaround time.

tomchristie:
Wonderful, thanks for helping get this through, folks. ✨ 

Do feel free to get in touch if there's any Q's the team might have about API Star that would affect if it makes in into 14 or not.

Cheers.

pavelholvi:
Hello @nbrady-techempower I was about to work on tests for queries, namely:
Single query, Multiple queries, Fortunes and Data updates.

However, since currently APIStar supports SQLAlchemy and Django ORM backends I wonder if it's something we should test:
https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Python

Here it says: 

- If your framework uses SQLAlchemy, Flask may be enough to know performance of SQLAlchemy.
- If you want to know about performance difference between MySQL and PostgreSQL, there are no need to all frameworks implement both of them.
- Additionally, Python is not a good language to measure performance of DBs because web application written in Python is slower than DBs.

Big thanks,
Pavel

methane:
If it's asyncio based, it may show unique, interesting result.
If it's WSGI based, synchronous,, SQLAlchemy overhead is very same to other frameworks.
It just increase number of result without any interesting information.

methane:
Main concern of the recommendation is (1) time to run all benchmarks, and (2) maintenance cost.

For time to run benchmarks, I think TFB highly automate it already.  So it's not big problem now.

For maintenance cost, I dislike many people:

1. Create own toy framework
2. Register it to TFB benchmark to promote it
3. Lose interest about it
4. Then, dozen garbage frameworks no one using are remained in TFB

If you believe apistar is different from them and you will maintain it in TFB,
I don't against adding it.

pavelholvi:
Thanks a lot @methane I'll come up with something.

PR #2635, opened 2017-03-30
[ci fw-only Erlang/chicagoboss] Accept non-localhost connections in chicagoboss

michaelhixson:
Without this, the client machine running wrk cannot connect to the server and so chicagoboss gets zero RPS.

It still seems to have issues with the plaintext test, but this should at least fix the json test.


nbrady-techempower:
Going to merge these in. Travis won't really give us what we're looking for here since it's doing a local connection.

PR #2636, opened 2017-03-30
Use mysql connector 5.x instead of 6.x in light-java-mysql

michaelhixson:
This repairs the light-java-mysql test.

Our ServerCentral dedicated hardware environment is set up in such a way that breaks mysql connector 6.x with exceptions like this:

```
java.sql.SQLException: The server time zone value 'PDT' is
unrecognized or represents more than one time zone. You must
configure either the server or JDBC driver (via the serverTimezone
configuration property) to use a more specifc time zone value if you
want to utilize time zone support.
```

As far as I know, 6.x is presently an unstable development release whereas 5.x is the stable, production-ready release.  Other Java frameworks in the benchmarks use 5.x.  It doesn't seem worthwhile to try to fix this "problem" with our environment right now...  Maybe later when 6.x is stable or we aren't trying to get a new round of benchmarks out the door.

### DIFF
--- a/frameworks/Java/light-java/pom.xml
+++ b/frameworks/Java/light-java/pom.xml
@@ -37,7 +37,7 @@
         <version.httpasyncclient>4.1.2</version.httpasyncclient>
         <version.swagger>1.5.10</version.swagger>
         <version.hikaricp>2.5.1</version.hikaricp>
-        <version.mysql>6.0.4</version.mysql>
+        <version.mysql>5.1.41</version.mysql>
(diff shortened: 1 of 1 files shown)


stevehu:
@michaelhixson Thanks for your help. It looks good to me.

greenlaw110:
I've encountered the same issue while updating ActFramework benchmark project to 6.x. And I got it worked around by adding `&serverTimezone=UTC` into the jdbc URL connection string.

PR #2641, opened 2017-03-31
Remove transactions as they aren't required for these tests

codylerum:


codylerum:
Not possible with EE. Transaction will be required for updates but not selects.

PR #2642, opened 2017-03-31
Fix db host for mochiweb, enable db tests

michaelhixson:
Having the wrong db host was causing the server to crash, which made the json and plaintext tests fail.

I enabled the four db-related tests since they seem to be implemented. They all pass except fortunes.  Unfortunately that implementation appears to live in another github repo (utkarshkukreti/erl_bench) so we can't fix it.

I removed the sed line from setup.sh since the file it was targeting (mochiweb_bench_sup.erl) didn't have any of the strings it was trying to replace.


nbrady-techempower:
#2656

PR #2648, opened 2017-04-01
Update of dlang and vibe-d framework

tchaloupka:
Use of the new vibe-core package.
Also removed the dub tool dependency as it is now included with dmd compiler itself and it is not needed to be installed separately.

Modified the binding settings to bind to all ipv4 and ipv6 addresses (default) instead of just all ipv4 as I don't get where the current problem in preview1 and 2 is, as it compiles fine but test can't connect to the server so maybe ipv6 was used there?

Anyway, it worked for me with previous version and this one too in Vagrant environment.


s-ludwig:
Looks good. Will be interesting to see how performance will change. I haven't done performance tuning on the new code yet (except for avoiding some previous design-based bottle necks), so I don't expect a big improvement yet.

PR #2649, opened 2017-04-02
Ruby: Add invokedynamic to JRuby tuning script

mwpastore:
[ci fw-only Ruby/sinatra-sequel Ruby/roda-sequel Ruby/rack-sequel]


mwpastore:
Thank you!

PR #2652, opened 2017-04-02
Adjusting JVM parameters for Dropwizard, Jetty, Resin, Undertow

zloster:
Since these projects don't have explicit settings for the JVM they are picking the defaults for the current version of the JVM and the OS. The `-server` mode will be enabled on 64-bit OS with more than 2GB RAM and `UseParallelGC` is the default garbage collector. Significant change is to enable NUMA architecture support which helps the GC threads to be local on the CPU for which RAM area they are responsible cleaning. `-XX:+UseNUMA` support works only with `+UseParallelGC`. Why this is needed? Both the physical server at ServerCentral and the Azure D3v2 machines are using NUMA architecture. If the machine is not using NUMA there shouldn't be any penalties when this option is provided, it just will be ignored. `-Xms1G -Xmx1G` restricts the JVM to 1GB of heap memory. `-Xss320k` will set the thread's stack size to 320 KB, the default is 2 MB for 64-bit OSes, 320 KB for 32-bit OSes. `XX:+AggressiveOpts` enables additional bytecode optimizations which are experimental and not enabled by default for the current JVM release version. The current version of the JVM (java version "1.8.0_121") changes only two options:

> intx AutoBoxCacheMax                         = 20000
> intx BiasedLockingStartupDelay                 = 500

How to verify what is enabled/disabled see [this StackOverflow answer](http://stackoverflow.com/questions/2959878/what-flags-are-enabled-by-xxaggressiveopts-on-sun-jre-1-6u20#21361242).

Additional notes:

- Credit for the decreasing thread stack sizes goes to [amichair](https://github.com/zloster/FrameworkBenchmarks/commit/d78dfdb6041c472cbcbbb2d6a4798661b0317207#diff-1d283148ad090e8126999400a7b3a81eR11). I suspect that by decreasing the thread stack the CPU caches are utilized slightly better and you get a small increase in performance.
- The decrease of the heap memory could affect the results if a default heap is relatively big and the framework is not producing enough garbage to trigger the major GC events. I can't find the memory size of the ServerCentral machines to calculate the default heap-stack size. All of the frameworks hit 100K+ rps in the JSON and plaintext test (except Dropwizard which is around 90K+ rps) so I don't expect to hit such a problem.
- [The resin configuration reference](http://www.caucho.com/resin-4.0/admin/config-resin-properties.xtp) doesn't mention anything about ` -jvm_mode` property key. That's why I've deleted it and moved the configuration to the `-jvm_args` key.

### DIFF
--- a/frameworks/Java/README.md
+++ b/frameworks/Java/README.md
@@ -19,11 +19,11 @@ should have an `install.sh` that contains at least
 
     fw_depends java
 
-This installs the Oracle Java 8 JDK. It also provides a shell variable `JAVA_OPTS` with tweaked default parameters for the JVM. Sample use:
+This installs the Oracle Java 8 JDK. It also provides a shell variable `JAVA_OPTS_TFB` with tweaked default parameters for the JVM. Sample use:
 
     #!/bin/bash
 
-    java -server $JAVA_OPTS
+    java -server $JAVA_OPTS_TFB
 
 ## Get Help
 
--- a/frameworks/Java/README.md
+++ b/frameworks/Java/README.md
@@ -19,7 +19,11 @@ should have an `install.sh` that contains at least
 
     fw_depends java
 
-This installs the Oracle Java 8 JDK.
+This installs the Oracle Java 8 JDK. It also provides a shell variable `JAVA_OPTS` with tweaked default parameters for the JVM. Sample use:
+
+    #!/bin/bash
+
+    java -server $JAVA_OPTS
 
 ## Get Help
 
--- a/frameworks/Java/dropwizard/setup_mongo.sh
+++ b/frameworks/Java/dropwizard/setup_mongo.sh
@@ -6,4 +6,4 @@ sed -i 's|host: 127.0.0.1|host: '"${DBHOST}"'|g' hello-world-mongo.yml
 
 mvn -P mongo clean package
 
-java -jar target/hello-world-0.0.1-SNAPSHOT.jar server hello-world-mongo.yml &
+java -server $JAVA_OPTS -jar target/hello-world-0.0.1-SNAPSHOT.jar server hello-world-mongo.yml &
--- a/frameworks/Java/dropwizard/setup_mongo.sh
+++ b/frameworks/Java/dropwizard/setup_mongo.sh
@@ -6,4 +6,4 @@ sed -i 's|host: 127.0.0.1|host: '"${DBHOST}"'|g' hello-world-mongo.yml
 
 mvn -P mongo clean package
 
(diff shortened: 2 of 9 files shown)


msmith-techempower:
I don't have any qualms with the suggested args, but I would like them to be introduced more cleanly. Everything that uses `java` should do so via `fw_depends java` which should take care of setting environment variables as needed.

Specifically, [this file](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/languages/java.sh#L14) should be updated such that the `$IROOT/java.installed` file has an additional line for the jvm options. This way, any time `fw_depends java` is called, that variable will be exported.

tl;dr - drop all the `JAVA_OPTS="-Xms1G -Xmx1G -Xss320k -XX:+UseNUMA -XX:+UseParallelGC -XX:+AggressiveOpts"` and instead add `echo "export JAVA_OPTS="-Xms1G -Xmx1G -Xss320k -XX:+UseNUMA -XX:+UseParallelGC -XX:+AggressiveOpts"" >> $IROOT/java.installed` to `java.sh`

zloster:
Thank you for the suggestion. Indeed it is a lot better place.

Note: I've squashed and force-pushed the commits - that's why the comment looks like not in place.

zloster:
Clojure/pedestal is consistently failing with the following: [`Setup pedestal: Conflicting collector combinations in option list; please refer to the release notes for the combinations allowed`](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/218180208#L1970-L1973) Which is probably related to the `JAVA_OPTS`. I'll check.

zloster:
It seems that leiningen is [doing](https://github.com/technomancy/leiningen/issues/856) [`magic`](https://github.com/technomancy/leiningen/issues/326) with the `JAVA_OPTS` environment variable. Add to this that pedestal project is [defining JVM options](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Clojure/pedestal/project.clj#L25-L52). I don't understand Closure but it seems they are **appending** and not replacing the JVM arguments - see `^:replace` after the `jvm-opts`. The [intro docs](https://github.com/technomancy/leiningen/blob/master/doc/TUTORIAL.md#setting-jvm-options) suggest this because they don't mention the `^:replace` part.

So the variable is renamed to `JAVA_OPTS_TFB` so it will not clash with tools that rely on `JAVA_OPTS`.

nbrady-techempower:
@zloster nice work, good idea on the rename. Ping @msmith-techempower for review

msmith-techempower:
Yeah, I think that as long as that is getting the Java options through (and I have no idea if that is true) then this is good to merge.

PR #2654, opened 2017-04-03
optimize plaintext and router

stevehu:
Someone pointed me the preliminary round 14 result and I realized that my plaintext and json endpoints are not optimized. If possible, could you please merge the changes and rerun the test? This is my very first time to work on small responses and there are more room to optimize even further if there is more time. Also the multiple queries and updates can be rewritten to have better performance. Do you know when the window for pull requests will be closed? Thanks.


nbrady-techempower:
Hi @stevehu thanks I can get this in now. It may make it for R14 final

stevehu:
@nbrady-techempower Thanks a lot for your help. We have very good performance on single query and fortunes but fall behind in multiple queries and updates. I will rewrite it for the next round.

PR #2660, opened 2017-04-04
Do miscellaneous cleanup of the grizzly-jersey code

michaelhixson:
* Implement MustacheViewProcessor ourselves, allowing us to correctly use UTF-8 to render fortunes.  The third-party library we used for this previously relied on the default charset.
* Upgrade dependencies that are trivial to upgrade.  I did not attempt larger upgrades like Jersey 1->2 or Hibernate 4->5.
* Use Collections.singletonMap instead of HashMap for the JSON endpoint, which should be more efficient.
* Use an array instead of a ConcurrentHashMap for the db/queries endpoint, which should be more efficient.
* Other minor changes.


michaelhixson:
> About the encoding problem

I want to correct these encoding issues in the source because IMO they are errors in the source code.  Relying on `Charset.defaultCharset()` is almost always a bad idea (that a good static analysis tool will warn you about), but it's even worse in this case because using UTF-8 is a functional requirement.

We do also have an environment-level "fix"...  We started noticing these errors when we started running the benchmarks from an upstart script.  Although we configure the script to run as our normal user, it doesn't automatically inherit environment variables such as `LANG`, so that leads to a different default charset in Java.  We've changed this script to set `LANG` to `en_US.UTF-8`.  That change should kick in the next time we restart the script.

> Will the instability be cured?

No, it still prints the broken pipe errors during the plaintext test.  I would like to fix that eventually, but I don't know how yet.

zloster:
> I want to correct these encoding issues in the source because IMO they are errors in the source code. Relying on Charset.defaultCharset() is almost always a bad idea (that a good static analysis tool will warn you about), but it's even worse in this case because using UTF-8 is a functional requirement.

I don't know how many resources your team have. My main idea was if there are other more pressing matters you could use the forced encoding for the JVM and open issues for fixing the frameworks. I agree about the Charset.defultCharset(). Maybe adding a reminder in the documentation for using explicitly UTF-8 when reading resources from the file system will be a good idea. It's very easy to forget.

> No, it still prints the broken pipe errors during the plaintext test. I would like to fix that eventually, but I don't know how yet.

:+1:

zloster:
As a fun fact about the pitfalls of the default encoding: http://mail.openjdk.java.net/pipermail/jmh-dev/2017-March/002527.html It could happen to anyone, even if you know the [JDK inside out](https://shipilev.net/)

PR #2661, opened 2017-04-04
Remove verbose GC logging to file for Java/wildfly-ee7

zloster:
This logging doesn't follow the project recommendation about the logs and IMO is affecting the benchmark results of Wildfly. But this gave me an idea about a new test type: which framework can fill a SSD disk faster. Or which framework logs are the biggest for given test run :)

### DIFF
--- a/frameworks/Java/wildfly-ee7/setup.sh
+++ b/frameworks/Java/wildfly-ee7/setup.sh
@@ -2,7 +2,7 @@
 
 fw_depends mysql java maven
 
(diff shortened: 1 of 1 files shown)


codylerum:
👍

zloster:
No better results: https://www.techempower.com/benchmarks/previews/round14/r14p3-vs-r14p4.html#wildfly-ee7.json Just the normal variation for the test results.
So the detailed garbage collection logging is basically free for this setup and conditions.

codylerum:
@zloster I'm guessing their are gains to be made by tweaking the thread settings. I believe the environment is 40 cores now and I don't know how many wildfly is configured to use by default.

nbrady-techempower:
@codylerum the server environment is 80 cores, and you can access that via the `CPU_COUNT` env variable if you need to.

PR #2664, opened 2017-04-05
Make wicket encode its fortunes response in UTF-8

michaelhixson:
(This replaces a previous PR https://github.com/TechEmpower/FrameworkBenchmarks/pull/2662, incorporating the feedback.)

This should repair the wicket fortune test on ServerCentral, which is currently failing.

It appeared to be using the system default charset instead, which was not necessarily UTF-8.

I removed the `initialCapacity` argument for the `ArrayList` because there's no good reason to set it to 10,000.  Since the test requirements state that the list shouldn't be sized using foreknowledge of the row count, we might as well trust `ArrayList`'s defaults.


martin-g:
Thank you, @michaelhixson !

PR #2671, opened 2017-04-11
New Framework added to the list of tests

Helidium:
I have added a new NodeJS test, using a custom server addon. All the tests have passes and travis build is a go.

Please add this test for the Round #14 if possible. 

Thank you,
Marko Biškup


nbrady-techempower:
@Helidium Thanks for the submission. Can you point out some of the differences between this and the main nodejs tests? This seems like a very similar reproduction. These implementations may be better off within the nodejs directory with tests named `nodejs-mitol-mysql` etc, since mitol itself is not a framework.

Helidium:
@nbrady-techempower First off I would like to thank you for considering to accept this tests. The differences of the tests are the server used. This tests use the Mitol server as a C++ addon. I wasn't planning of creating new tests, as some code changes are required, and I didn't want to change the original NodeJS tests. Is this still acceptable?

nbrady-techempower:
Thanks @Helidium I will take a closer look at this shortly. I think since it's "nodejs" framework tests but with a different server addon it will end up in the nodejs directory with a few different implementations. I will keep you posted.

Helidium:
Thank you @nbrady-techempower, if there is anything else I have to do, please let me know.

Helidium:
Hi @nbrady-techempower, I am wondering what is the state of this PR. Should I prepare additional tests instead of new framework?

nbrady-techempower:
@Helidium That would be great. If you could prepare additional tests in the nodejs directory with a separate setup file, like `setup_midol.sh`. I'll close this in favor of doing that. Been working on some high priority fixes. Thanks!

PR #2678, opened 2017-04-13
jooby: get back netty

jknack:

### DIFF
--- a/frameworks/Java/jooby/benchmark_config.json
+++ b/frameworks/Java/jooby/benchmark_config.json
@@ -15,13 +15,13 @@
       "language": "Java",
       "flavor": "None",
       "orm": "Raw",
-      "platform": "Undertow",
+      "platform": "Netty",
       "webserver": "None",
       "os": "Linux",
       "database_os": "Linux",
       "display_name": "Jooby",
       "notes": "",
-      "versus": "undertow"
+      "versus": "netty"
     }
   }]
 }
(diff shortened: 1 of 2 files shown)


maxdemarzi:
Why pick, why not have both?

jknack:
Too late for adding both, will be a great addition for R15

nbrady-techempower:
Yeah, seems like we're going back and forth with this. You could duplicate the tests for undertow as well. No problem there.

jknack:
Will do in R15

zloster:
@jknack @maxdemarzi It is quite easy to support both especially when your code doesn't require changes in it. Make additional `setup-undertow.sh`. Add two maven profiles in the `pom.xml` (or default with `netty` as dependency and another with `undertow` as dependency). In the first script you call maven with the profile that is packaging `netty`, the other script will call maven with the profile which is packaging `undertow`. Add the new setup script as configuration in `benchmark_config.json`. Update the metadata. All done. Take a look in the [Dropwizard test](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Java/dropwizard) for details.

jknack:
@zloster Thank you. I didn't know this was possible.

nbrady-techempower:
@jknack Do you want this in for Round 14 and then work on putting multiple implementations together or leave as is until that's complete?

jknack:
@nbrady-techempower In for R14 (if possible).

Issue #2554, opened 2017-02-04
Test implementations should live within the FrameworkBenchmarks repo


knewmanTE:
I've noticed recently that there are a handful of frameworks that only have a few files living on this repo and then maintain their actual test implementations on a separate repo (often as a subdirectory of the framework itself).

I'd really like to avoid this as much as possible for the following reasons:

1. It makes it difficult/impossible for anyone other than the framework's maintainers to update their implementation, regardless of how small a change needs to be made. 
2. In some cases, no version or tag is being specified, allowing the maintainer to make changes that don't go through the Pull Request process.
3. It feels a bit contradictory to the spirit of the Framework Benchmarks project, where I would love for everyone to have full access to inspect and improve the code wherever they see a place that could be enhanced.

Frameworks I've noticed so far:

- [C++/ffead-cpp](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/C%2B%2B/ffead-cpp) (ping @sumeetchhetri)
- [C/haywire](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/C/haywire) (ping @kellabyte)
- [C/libreactor](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/C/libreactor) (ping @fredrikwidlund)
- [C/lwan](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/C/lwan) (ping @lpereira)

If you're one of those maintainers, or if you implemented a different framework using this methodology, please try to shift your code over so it's living on our repo!

Thanks!

lpereira:
Hi, @knewmanTE! I'm working on Lwan's first release, and that will enable me to move the benchmark code to the FB repository. I know this has been promised for a long time, but life got in the way.

knewmanTE:
@lpereira no worries! Thanks for the update, and best of luck as you move towards your release!

nbrady-techempower:
@lpereira Great news!

fredrikwidlund:
Hi, @knewmanTE. Duly noted! I'm fairly new here so I simply followed some existing patterns. I'll move the project as soon as possible, but it might take a few weeks before I find some time to work on this. Is this something that will disqualify the framework for the next round?

Kind regards,
Fredrik

knewmanTE:
[...]

PR #2683, opened 2017-04-13
Revert back to old compiler since gcc-6 for some very strange reason destroyed performance

fredrikwidlund:
Revert back to old compiler since gcc-6 for some very strange reason destroyed performance.


nbrady-techempower:
@fredrikwidlund Are you sure this is the reason? If so, you might want to throw a comment above that line saying so, that way no one comes along and tries to upgrade it.

fredrikwidlund:
I am not at all sure, but it is the only relevant change and the result
varied greatly. What can I do but to guess. Ulib increased what... 50% in
the plaintext test, without any changes since the last preview?

F

nbrady-techempower:
@fredrikwidlund I hadn't realized until now but libreactor was one of the frameworks that does not have the test implementations living within this repo. see #2554 

So we can maybe dig in a little deeper when that happens, but also the 4 frameworks that do not do this will probably be removed after this round. I'm hesitant to make any changes here until we can get that done.

fredrikwidlund:
As is discussed in the issue it will indeed be moved since this is
requested. How is this related to reverting back to gcc 4.9?

nbrady-techempower:
@fredrikwidlund It's related because we're not sure it's a gcc issue at all and because we don't have the test implementations. For instance, were you using the MAX_THREADS env variable that we provided? When we make a fix to the toolset that requires the TE staff grepping through all the test implementations to make some tweaks so fw's aren't penalized, nothing for libreactor will show up. In that sense, we've made several improvements and changes to the toolset since round 13.

Also you're cloning from the master branch where your test implementations live, so any changes that you make will skip travis and wouldn't be able to be reviewed by the team since we won't even know they happened.

I'll take a look through that repo and see if I can identify anything, and just check over the implementation since it looks like you've made some recent changes.

I'll get this in tonight if you can add a line in `libreactor.sh` that checks out a specific commit and then would you mind making the changes that were requested in February for the next round?

Not trying to single you out; there are 3 others that are doing the same. Just happened to remember the fw name and see you doing some work with it. Thought it might be a good time to make that change.

Cheers! 🏃

fredrikwidlund:
I do understand the reasoning behind this, and did reply to this in the
issue mentioned. As a parenthesis this is also true for a number of other
candidates in the benchmark, not just the ones you mention, with references
to tags and so forth. The reason I haven't changed this since the
discussion in February is because I specifically asked if this would be an
issue for this round, was given a clear answer that it would not be, and
therefore decided to wait until after this round. Again, I will change this
after this round.

nbrady-techempower:
@fredrikwidlund We've been addressing the tags/releases and believe we have most of them. That's an immediate fix that needs to be made when found.

fredrikwidlund:
Are we ok then to go ahead and downgrade gcc version? To be quite honest I regardless find it hard to believe that this is the reason behind the degraded performance. Also since the json/plaintext benchmarks are so broken I'm not sure it really matters anyhow.

PR #2684, opened 2017-04-13
Turn off batch updates for servlet-postgres-raw

zloster:
The batch updates are leading to transaction deadlocks for higher concurrences in Java/Servlet-PostgreSQL - see the results for Round 14 preview 2 and 3: `0 rps`. This effect is not visible when using Vagrant/VirtualBox.

### DIFF
--- a/frameworks/Java/servlet/src/main/java/hello/PostgresUpdateServlet.java
+++ b/frameworks/Java/servlet/src/main/java/hello/PostgresUpdateServlet.java
@@ -61,8 +61,7 @@ public class PostgresUpdateServlet extends HttpServlet {
 		try (Connection conn = source.getConnection()) {
 			try (PreparedStatement statement = conn.prepareStatement(DB_QUERY,
 					ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY);
-					PreparedStatement statement2 = conn.prepareStatement(UPDATE_QUERY,
-							ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY)) {
+					PreparedStatement statement2 = conn.prepareStatement(UPDATE_QUERY)) {
 				// Run the query the number of times requested.
 				for (int i = 0; i < count; i++) {
 					final int id = random.nextInt(DB_ROWS) + 1;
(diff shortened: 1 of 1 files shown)


volyrique:
Some other frameworks, e.g. fasthttp, Revenj.JVM, and now h2o, are sorting the results by ID before executing the batch updates. Have you considered that too?

zloster:
@volyrique Thank you for the information.
Several months ago I've played with Dropwizard/JDBI implementation (it is still in branch and needs changes) and noticed the sorting thing in the Revenj.JVM implementation. But I couldn't make it work adhering to the test rules without transaction deadlocks. If I remember correctly the ordering is giving some performance gain even without batch updates. This is puzzling me.
When I find some time I'll get back to play with this (see below for the `autoCommit`).

Here are some of my observations about Revenj.JVM (the approach to the implementation deserves a nice review writing) and servlet implementations:
- it is skipping the Resin DB connection pool, the SQL connections are [directly obtained](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/revenj-jvm/src/main/java/hello/Context.java#L41-L42) from the driver and attached in [ThreadLocal variable](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/revenj-jvm/src/main/java/hello/Utils.java#L9) when a [request is processed](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/revenj-jvm/src/main/java/hello/UpdatesServlet.java#L19). [Resin.properties](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/webservers/resin/resin.properties#L89-L92) seems to limit the maximum number of threads to 256. 
- Notable setting here is the setting of "autoCommit" which I haven't noticed before.
- The update test numbers from Round14 preview 3: 1 query - 1,772; 5 queries - 809; 10 queries - 5,496; 15 queries - 4,750; 20 queries - 3,709 - Something is not right currently. However the numbers in Round 13 are OK.

In contrast the Servlet tests [are using the Resin DB connection pool](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/servlet/src/main/webapp/WEB-INF/resin-web.xml), which is using quite high [default setting for the maximum connections](http://caucho.com/resin-4.0/admin/database.xtp#database): 1024. But given the max thread limitation Servlet and Revenj.JVM basically use max 256 connections to the DB.

vlsi:
> If I remember correctly the ordering is giving some performance gain even without batch updates. This is puzzling me

Well, batching enables to shave network roundtrips.

@zloster , could you please clarify if you did try sort+batch update?

zloster:
@vlsi I have played with `sort+batch` update in Servlet+PostgreJDBC only on my limited (in terms of computing resources) environment. This environment is skewing the results (because it is a VM on an underpowered hardware) and small improvements (let's say +5% or +10% percent) are translated in bigger wins on a real hardware. `sort+batch` checks were done when I was implementing the PR with `batch updates` here some time ago. Unfortunately I don't have my results stored somewhere for reference.

PR #2691, opened 2017-04-17
Fix serious socket leakage introduced by optimization

fredrikwidlund:
Fix serious socket leakage introduced by optimization


fredrikwidlund:
@nbrady-techempower, found reason why libreactor breaks when scaling from 256 to 512 in concurrency. This needs to be merged for the results regarding libreactor to make any sense.

PR #2698, opened 2017-04-20
[ci fw-only Java/act-beetlsql] Add actframework with beetlsql as database ORM tool

greenlaw110:


nbrady-techempower:
Hey @greenlaw110 ! Looks like this is just adding support for a new ORM for act, is that correct? If so we're going to want to keep this in the `Java/act` directory and add the test to the benchmark_config there. You can create separate setup files and even a subdirectory if it doesn't reuse a lot of code.

greenlaw110:
@nbrady-techempower There is no problem for me to merge it into the same project. However the concern is it will need to configure a different data source and it might used up available data connections. Because this ORM will connect to postgresql, which is the same database as the one used by original ebean based solution. Note when I start Act project it really activate all data connections for all tests in one process, in other words you don't need to restart the project in order to test on another database connection. And it is the way Act benchmark project is coded

nbrady-techempower:
@greenlaw110 I see. Though you may have coded it that way, before each test the databases required by that test are restarted, so it shouldn't be a problem:

https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/benchmark/benchmarker.py#L546-L560

greenlaw110:
@nbrady-techempower I've removed `act-beetlsql` project and merged it into `act`. Can you merge this PR now?

PR #2701, opened 2017-04-20
increase gunicorn worker count for aiohttp to 3

samuelcolvin:
As per conversation with on of the aiohttp authors @fafhrd91 this PR increases the gunicorn worker count to 3 * cpu count to match other frameworks.

### DIFF
--- a/frameworks/Python/aiohttp/app/main.py
+++ b/frameworks/Python/aiohttp/app/main.py
@@ -46,8 +46,9 @@ async def startup(app: web.Application):
     dsn = pg_dsn()
     # number of gunicorn workers = multiprocessing.cpu_count() as per gunicorn_conf.py
     # max_connections = 2000 as per toolset/setup/linux/databases/postgresql/postgresql.conf:64
-    # give 10% leeway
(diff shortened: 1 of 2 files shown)


methane:
For sync workers, I used `cpu_count * 3` because each worker can't do anything while waiting reply from DB.  It means there are at most `cpu_count * 3` DB connections.

In case of asyncio + asyncpg, worker can work on another request while waiting response from DB.
So `cpu_count` workers, 3 DB connection in each workers will be good at start line.

I think current configuration and this pull request seems using too much connections.

samuelcolvin:
I've just tried this locally and it gives a 5-15% performance improvement with the `/json` and `/plaintext` tests.

But gives a ~20% reduction in performance with `/queries/10`.

This is what I had feared, and aiohttp is always going to perform best in tests requiring database connections - that's where it's at it's best. I'm closing this PR.

@fafhrd91, feel free to see if you can find a way which doesn't damage db dependent tests. Personally, I think 1 gunicorn worker per CPU is a good compromise.

PR #2702, opened 2017-04-20
correct naming of api-hour

samuelcolvin:
In #2527 I forgot to change the platform, framework and display name for API hour. Meaning the 14 preview results are still saying "asyncio" which doesn't really make sense.

By the way, for me api hour doesn't meet [FrameworkBenchmarks test requirements](http://frameworkbenchmarks.readthedocs.io/en/latest/Project-Information/Framework-Tests/):

> All test implementations **should be production-grade**

It uses `aiohttp` `0.16.3` when the framework is now at `2.0.7` and customised old version of `aiohttp_jinja2` and `api_hour`. Sorry to be blunt, but I sure wouldn't use this setup in production.

Either way, this just corrects the names.


nbrady-techempower:
@GMLudo I would like to get your input here on `production-grade`. @samuelcolvin thanks for all your efforts. Certainly if there's a discussion to be had about whether these results should be included in Round 14 and representative of the framework we should have it now while there's still a few days left before the final run. Thanks guys!

nbrady-techempower:
Ok I'm gonna go ahead and merge this in.

PR #2704, opened 2017-04-20
[ci fw-only Java/act] make beetlsql dbservice share the datasource with pgsql dbservice

greenlaw110:


greenlaw110:
@nbrady-techempower this PR is important as I've mentioned previously whenever the app start it will agreesively initialize all configured db service and in turn initialize all attached datasource with configured DB connections. The other more critical issue fixed in this PR is previous version doesn't allow beetlsql test working with remote db server.

PR #2712, opened 2017-04-22
Share the results folder

joanhey:
in vagrant, for easier view and manage.
Still the contents in the folder are ignored by Git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,7 +40,8 @@ bin/
 azure.err
 php-kohana/application/logs/
 php-fuel/fuel/app/logs/
-results/
+results/*
+!results/.gitkeep
 benchmark.cfg
 
 # Mac filesystem dust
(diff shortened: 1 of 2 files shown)


nbrady-techempower:
@joanhey I'd have to play with this on monday. My concern is what happens on windows when you run the `tfb --clean-all` command or otherwise try and remove the results directory. I don't think this would work.

joanhey:
It is really useful, perhaps `tfb --clean-all`, will delete the contents in the folder and not the folder.
Because remove the directory, will give problems, the same that remove the toolset directory or any other.

nbrady-techempower:
@joanhey Those are fair points. I believe there were some reasons for not doing this before, some other intricacies of the toolset that may be resolved now. I'll do some testing on this and make appropriate changes. I agree with you.

nbrady-techempower:
Hey @joanhey closing for #2732 - Thanks for pointing this out. I did it in a way that doesn't require the results directory being in the repo as well.

PR #2715, opened 2017-04-23
benchmark for the Macchiato ClojureScript framework

yogthos:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ env:
     - "TESTDIR=Clojure/compojure"
     - "TESTDIR=Clojure/http-kit"
     - "TESTDIR=Clojure/luminus"
-      "TESTDIR=Clojure/macchiato"
+    - "TESTDIR=Clojure/macchiato"
     - "TESTDIR=Clojure/pedestal"
     - "TESTDIR=Clojure/aleph"
     - "TESTDIR=Crystal/crystal"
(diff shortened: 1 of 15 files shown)


nbrady-techempower:
Hey @yogthos! This is looking good.. it just needs a line in the `.travis.yml` file

yogthos:
ah right missed that part, just added it in

.travis.yml
```diff
@@ -36,6 +36,7 @@ env:
     - "TESTDIR=Clojure/compojure"
     - "TESTDIR=Clojure/http-kit"
     - "TESTDIR=Clojure/luminus"
+      "TESTDIR=Clojure/macchiato"
```
nbrady-techempower: You need the `-` to precede it. See the lines above and below

yogthos:
oops fixed

frameworks/Clojure/macchiato/setup.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+fw_depends postgresql nodejs
```
nbrady-techempower: You'll need to add leiningen here to use `lein package`

yogthos:
Right makes sense, just updated.

yogthos:
@nbrady-techempower looks like everything's good to go :)

nbrady-techempower:
@yogthos Yes, I've added the proper label to it. Round 14 final is currently running, so this won't make it in for that. Am going to merge right after that run is done. Thanks!

yogthos:
Thanks, sounds great. Since it won't make it in for the current round, I might play around with it to see if I can improve performance and update the PR if I get any improvements.

PR #2722, opened 2017-04-24
Php update to v7.1.4 stable

joanhey:


joanhey:
setup-mysql.sh and framework processes have terminated
ERROR 1298 (HY000) at line 3: Unknown or incorrect time zone: '00:00'
Setup has terminated

All tests using mysql fail for the commit #2720

nbrady-techempower:
Should be good now @joanhey thanks!

joanhey:
The update is ready.
Now I will try to update de php5 to 5.6.30. The v5.5 is not supported anymore.

Issue #2646, opened 2017-04-01
Frameworks using gzip encoding which is prohibited by the rules


zloster:
**Update 1**: pull request for `Dropwizard` submitted by me
**Update 2**: pull request for `hapi` submitted by nbrady-techempower
**Update 3**: pull request for most of the `PHP` frameworks submitted by joanhey
**Update 4**: pull request for `Tapestry` submitted by me

**Update**: I've a quick look at the PR by joanhey. I think only `yii2-hhvm` is not covered by it and it will be the last one left. The other `php` frameworks seems to be covered.

It will be good someone else to check this information also.
It is possible I've missed something and/or to have false positives but here are the steps how I've checked.

I've downloaded the R14 preview 2 logs:

> wget \
>      --recursive \
>      --no-clobber \
>      --page-requisites \
>      --html-extension \
>      --convert-links \
>      --restrict-file-names=windows \
>      --domains tfb-logs.techempower.com \
>      --no-parent \
>          http://tfb-logs.techempower.com/round-14/preview-2/

Using grep: `~/tmp/r14p2$ grep -rl --include="verification.txt" "'content-encoding': 'gzip'" | sort`.
Results are:

> zloster@box:~ /tmp/r14p2$ grep -rl --include="verification.txt" "'content-encoding': 'gzip'" | sort
> tfb-logs.techempower.com/round-14/preview-2/codeigniter-raw_hhvm/fortune/verification.txt
> tfb-logs.techempower.com/round-14/preview-2/codeigniter-raw_hhvm/query/verification.txt
> ~~tfb-logs.techempower.com/round-14/preview-2/dropwizard/fortune/verification.txt
> tfb-logs.techempower.com/round-14/preview-2/dropwizard-mongodb/fortune/verification.txt
> tfb-logs.techempower.com/round-14/preview-2/dropwizard-mongodb/query/verification.txt
> tfb-logs.techempower.com/round-14/preview-2/dropwizard-mongodb/update/verification.txt
> tfb-logs.techempower.com/round-14/preview-2/dropwizard-postgres/fortune/verification.txt
> tfb-logs.techempower.com/round-14/preview-2/dropwizard-postgres/query/verification.txt
> tfb-logs.techempower.com/round-14/preview-2/dropwizard-postgres/update/verification.txt
> tfb-logs.techempower.com/round-14/preview-2/dropwizard/query/verification.txt
[...]

PR #2739, opened 2017-04-27
HHVM disabled gzip encoding

joanhey:
and enabled Keep Alive 
.hdf is deprecated
Partiallly solves for php (hhvm) #2646


nbrady-techempower:
Excellent! Thank you!

PR #2745, opened 2017-04-28
Upgrade Tapestry to 5.4.3, use JDK 8

jochenberger:
Upgrade to the latest release


zloster:
@jochenberger You could also update the compile settings in the `pom.xml` to Java 8. `JDK 8` is the default for quite a while in this project.

jochenberger:
Thanks for the hint

PR #2750, opened 2017-05-02
Update elixir version to 1.4.1

Jesterovskiy:


frameworks/Elixir/phoenix/web/controllers/page_controller.ex
```diff
@@ -5,17 +5,20 @@ defmodule Hello.PageController do
 
   def index(conn, _params) do
     conn
+    |> put_resp_content_type("application/json", nil)
```
josevalim: This should not be necessary. If it is, it is a Phoenix bug.

frameworks/Elixir/phoenix/web/controllers/page_controller.ex
```diff
@@ -69,6 +74,7 @@ defmodule Hello.PageController do
 
   def plaintext(conn, _params) do
     conn
+    |> put_resp_content_type("text/plain", nil)
```
josevalim: Same here.

frameworks/Elixir/phoenix/web/controllers/page_controller.ex
Jesterovskiy: @josevalim `utf-8` added by default to all `content-type`. And I see warning from FB. Maybe we can ignore it. What do you think?

josevalim: Weird, why isn't FB requiring a charset?

Jesterovskiy:
```
Content encoding found in "text/plain; charset=utf-8" where "text/plain" is acceptable.
```
this warnings I see in Ruby frameworks too

josevalim: If we are going to explicitly set the content type, then I would use send_resp and give up on the `json` and `text` functions altogether. What do you think?

Jesterovskiy: Agree, good idea

frameworks/Elixir/phoenix/web/controllers/page_controller.ex
```diff
@@ -31,7 +34,8 @@ defmodule Hello.PageController do
     end
 
     conn
-    |> json(Enum.map(1..q, fn _ -> Repo.get(World, :rand.uniform(10000)) end))
+    |> put_resp_content_type("application/json", nil)
+    |> send_resp(200, Poison.encode!(Enum.map(1..q, fn _ -> Repo.get(World, :rand.uniform(10000)) end)))
```
josevalim: Are we required per the spec to perform multiple queries? Or can we perform a single one?

Jesterovskiy: @josevalim  http://frameworkbenchmarks.readthedocs.io/en/latest/Project-Information/Framework-Tests/#multiple-database-queries

nbrady-techempower:
@Jesterovskiy Looks like Elixir/Cowboy had issue with this

frameworks/Elixir/phoenix/web/controllers/page_controller.ex
josevalim: Thanks!

josevalim:
@Jesterovskiy looks good to me, thank you!

Jesterovskiy:
@nbrady-techempower https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/230251855 I think it fixed now. Check plz

nbrady-techempower:
@Jesterovskiy looks good now, thank you!

PR #2755, opened 2017-05-03
Update Ruby to 2.4.1 in Rails, Rack and Grape

Jesterovskiy:


frameworks/Ruby/grape/Gemfile
```diff
@@ -1,20 +1,21 @@
 source 'http://rubygems.org'
 
 platforms :jruby do
-  gem "activerecord-jdbcmysql-adapter", "~> 1.3.9", :require => false
-  gem 'torqbox', '0.1.7'
-  gem "trinidad", "1.4.6"
+  gem 'activerecord-jdbcmysql-adapter', '~> 1.3.9', :require => false
+  gem 'torquebox', '3.2.0'
```
rjnienaber: Does this work? I'm not sure `torquebox` is a like-for-like replacement for `torqbox`.

frameworks/Ruby/grape/config.ru
```diff
@@ -3,25 +3,26 @@ require 'active_record'
...
   class HelloWorld < Grape::API
     get '/json' do
-      {message:"Hello, World!"}
+      {message:'Hello, World!'}
     end
   end
 
   class PlainText < Grape::API
-    content_type :plain, "text/plain; charset=utf-8"
+    content_type :plain, 'text/plain'
```
rjnienaber: There seems to be a bit of ambiguity here. The requirements say the following:

```
The response content type must be set to text/plain.
```
by the example response is this:

```
HTTP/1.1 200 OK
Content-Length: 15
Content-Type: text/plain; charset=UTF-8
Server: Example
Date: Wed, 17 Apr 2013 12:00:00 GMT

Hello, World!
```

I guess it depends if the smoke tests pass.

Jesterovskiy: @rjnienaber when you run `tfb`:
```
Content encoding found in "text/plain; charset=utf-8" where "text/plain" is acceptable.
```

frameworks/Ruby/grape/Gemfile
Jesterovskiy: @rjnienaber we don't have `run_jruby_*` files for now, but in Sinatra I see https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Ruby/sinatra/Gemfile#L9
this is will be next step, add JRuby tests for all Ruby frameworks

mwpastore: @rjnienaber It is!

PR #2758, opened 2017-05-03
Implement parallelism for Crystal

RX14:
This PR updates the Crystal version to 0.22.0, and implements parallelism by way of `SO_REUSEPORT`. We ensure to set `SO_REUSEPORT`, and spawn a process on each core. Incoming connections are then load balanced across all processes.

What might be a problem is that I see your environment has 40 cores, but wrk makes far fewer connections than that. Preferably 2 or 3 connections would be made per core, is there a way to make this change (only for crystal)?

### DIFF
--- a/frameworks/Crystal/kemal/shard.yml
+++ b/frameworks/Crystal/kemal/shard.yml
@@ -4,15 +4,12 @@ version: "0.0.1"
 dependencies:
   pg:
     github: "will/crystal-pg"
-    version: "0.9.0"
-  pool:
-    github: "ysbaddaden/pool"
-    version: "0.2.3"
+    version: "0.13.3"
   kemal:
     github: "sdogruyol/kemal"
-    version: "0.16.0"
+    version: "0.18.3"
   redis:
     github: "stefanwille/crystal-redis"
-    version: "~> 1.6.6"
+    version: "1.8.0"
 
 license: "MIT"
--- a/frameworks/Crystal/kemal/shard.yml
+++ b/frameworks/Crystal/kemal/shard.yml
@@ -7,7 +7,7 @@ dependencies:
     version: "0.13.3"
   kemal:
     github: "sdogruyol/kemal"
-    branch: master
+    version: "0.19.0"
   redis:
     github: "stefanwille/crystal-redis"
     version: "1.8.0"
--- a/frameworks/Crystal/kemal/shard.yml
+++ b/frameworks/Crystal/kemal/shard.yml
@@ -7,7 +7,7 @@ dependencies:
     version: "0.13.3"
   kemal:
     github: "sdogruyol/kemal"
-    version: "0.18.3"
+    branch: master
   redis:
     github: "stefanwille/crystal-redis"
     version: "1.8.0"
--- a/frameworks/Crystal/crystal/server.cr
+++ b/frameworks/Crystal/crystal/server.cr
@@ -9,7 +9,7 @@ server = HTTP::Server.new("0.0.0.0", 8080) do |context|
   when "/json"
     response.status_code = 200
     response.headers["Content-Type"] = "application/json"
-    response.print({message: "Hello, World!"}.to_json)
+    {message: "Hello, World!"}.to_json(response)
   when "/plaintext"
     response.status_code = 200
     response.headers["Content-Type"] = "text/plain"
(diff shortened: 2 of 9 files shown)


RX14:
Actually I think i'm confused about the `wrk` concurrency. What is the max concurrency used on the actual rounds?

RX14:
Sorry, I only just found https://frameworkbenchmarks.readthedocs.io/ :(

My question has been answered.

RX14:
Please don't merge this, disabling logging is broken on kemal due to https://github.com/kemalcr/kemal/issues/362

nbrady-techempower:
@RX14 Adding the WIP tag. Let me know when you think this is ready.

RX14:
I think this is all fixed now. The log output [here](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/228544893#L1629-L1633) is gone on the latest build [here](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/228819464#L1589). Tell me when you want this PR is reviewed and ready to be squashed.

As an additional question, will there be another preview for round 14?

nbrady-techempower:
@RX14 I should be able to take a look at this later today and merge if all is good. Unfortunately we've finalized results for Round 14 and they'll be out soon. However, we are committed to faster round turnaround now that continuous benchmarking is running in our environments. Round 15 preview data will be right around the corner. :)

Thanks for your contribution!

RX14:
@nbrady-techempower Although it's unfortunate that this PR won't make it into round 14 I'll be waiting eagerly for the continuous benchmarking! This benchmarking suite has been an itch I've wanted to scratch for a long while.

frameworks/Crystal/kemal/shard.yml
```diff
@@ -7,7 +7,7 @@ dependencies:
     version: "0.13.3"
   kemal:
     github: "sdogruyol/kemal"
-    version: "0.18.3"
+    branch: master
```
Sija: This should point to next release containing `logging` bugfix, @sdogruyol would u be so kind? :)

RX14: I think it's fine to leave it, it is pinned to a specific git sha after all.

frameworks/Crystal/kemal/shard.yml
```diff
@@ -4,15 +4,12 @@ version: "0.0.1"
...
     github: "will/crystal-pg"
-    version: "0.9.0"
-  pool:
-    github: "ysbaddaden/pool"
-    version: "0.2.3"
+    version: "0.13.3"
   kemal:
     github: "sdogruyol/kemal"
-    version: "0.16.0"
+    branch: master
```
nbrady-techempower: Everything here looks good, but we need kemal tied to a specific commit or release.

sdogruyol:
I just release Kemal `0.19.0` @RX14 @nbrady-techempower

frameworks/Crystal/kemal/shard.yml
kbrock: Locking down the version number keeps things stable and documents to users of future benchmarks the differences.

So much is changing in Crystal that it would be nice to know this code will work.
If possible, please lock down to a version and/or

kbrock:
On a side note:

Reading through this code can give developers some tricks to speeding up their own app.
Very easy to read.

Thanks so much for putting this together.

frameworks/Crystal/kemal/shard.yml
RX14: The code would work forever as-is, because the version is locked down to a git sha in `shards.lock`. However as `0.19.0` has been released, we might as well use that version instead.

RX14:
Pushed a fix

Sija:
@sdogruyol _kemal_ version hasn't been bumped along with the release, so it still equals to `0.18.3`, just FYI :)

faustinoaq:
@RX14 Hi!

Your current bash script implementation just do one iteration.

See [stackoverflow](http://stackoverflow.com/questions/6191146/variables-in-bash-seq-replacement-1-10)

```sh
for i in {1..$(nproc --all)}; do
  echo "server"
done
# server
```

The above `for` syntax works well in `zsh` not `bash`.

To archive more than one iteration in bash, need to use `seq`.  See #2766

```sh
for i in $(seq 1 $(nproc --all)); do
  echo "server"
done
# server
# server
# server
# server
# server
# ...
```

RX14:
Haha, too used to zsh syntax, thanks for testing/finding/fixing this!

kbrock:
thanks again - very excited to see how this stacks up against other frameworks.

PR #2765, opened 2017-05-09
Please add http4k framework to benchmarks.

daviddenton:
If we could get this into Round 14 that would be awesome. Plaintext and Json benchmarks are implemented with more to come! 

:)


nbrady-techempower:
@daviddenton Round 14 has been finalized. We're preparing the site for displaying the data.

daviddenton:
No problem. Out of interest, is there a timeline on the first previews of R15? 

Thanks :)

nbrady-techempower:
No timeline just yet, but I predict they'll be faster and more often than previous rounds now that continuous benchmarking is up and running in house

daviddenton:
That's brilliant news. Keep up the awesome work! :)

frameworks/Kotlin/http4k/source_code
```diff
@@ -0,0 +1,7 @@
+fintrospect/src/main/scala/
```
martin-g: fintrospect ?
Scala ?
:-)

daviddenton: Whoops! I also look after the fintrospect benchmark and copied the general for structure from there. I'll sort that out asap.


Thanks ! :)

nbrady-techempower:
@daviddenton can we pull `gradle-wrapper.jar` in at test time. I am going through other tests to remove large files. We don't want to store them in the repo.

daviddenton:
No problem. How do you want to proceed - aka what changes should I make changes to the implementation to make this work? It would be good to be able to run the suite locally without any extra setup for vagrant.

nbrady-techempower:
@daviddenton I'm not familiar with gradle wrapper but it looks like you should be able to pull it in in your setup files. https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:wrapper_generation

daviddenton:
@nbrady-techempower I've just checked and the wrapper is only 50k in size - it's only the bootstrapping bit of gradle and not all of gradle.

Given this, do you still want me to come up with a way to remove it? :)

nbrady-techempower:
No worries @daviddenton I'll get this in for now and we can work on something down the road. The project, as a whole, is rather large because of some jar files and other things, that should eventually be removed.

daviddenton:
Thanks :)

PR #2767, opened 2017-05-11
MySQL flush-hosts fix

nbrady-techempower:
Fixes a problem with host-caching between framework runs


msmith-techempower:
LGTM!

nbrady-techempower:
ok confirmed this passes mysql tests on travis. mergin

PR #2775, opened 2017-05-12
Akka http version upgrade

sxend:
akka-http and dependent libraries upgrade.


ktoso:
Thanks for the upgrade @sxend. The new round with the old version is somewhat disappointing - hoped it to include the latest; it's misleading to our communities otherwise.

So thanks for the update!

PR #2776, opened 2017-05-12
Improve akka-http code

jrudolph:
* update dependencies to the latest version
 * enable pipelining
 * better code structure
 * a few thread pool settings optimized for performance
 * automatic code formatting
 * idiomatic usages for marshalling


nbrady-techempower:
@jrudolph Looks like this failed travis. I also merged in an upgrade earlier on #2775 so you will need to rebase to continue.

jrudolph:
I see, will try to fix on the weekend. Seems I broke the mysql part of things with the refactoring.

jrudolph:
I fixed the problem with the DB tests. I made the DB access more flexible, though, that will conflict with https://github.com/TechEmpower/FrameworkBenchmarks/pull/2773.

nbrady-techempower:
@jrudolph Not a problem. All I care about is that the `sed` statements are gone.

jrudolph:
Cool, thanks for merging!

frameworks/Scala/akka-http/src/main/scala/com/typesafe/akka/http/benchmark/handlers/JsonHandler.scala
```diff
@@ -1,36 +1,24 @@
...
+}
 
-class JsonHandler(components: {
-
-}) {
+trait JsonHandler {
+  import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 
-  import JsonHandler.Protocols._
+  lazy val jsonResponse = JsonResponse("Hello, World!") // domain object
```
sxend: https://www.techempower.com/benchmarks/#section=code
```
Requirements

    For each request, an object mapping the key message to Hello, World! must be instantiated.
```
follow this rule?

PR #2778, opened 2017-05-12
Add benchmarks for the VSGI platform

arteymix:


nbrady-techempower:
Hi @arteymix!

So if I understand this correctly, VSGI belongs to valum, and is more of a stripped implementation? Since it belongs to valum, it would be best to add it to the `vala/valum` folder and add a second entry to the `benchmark_config.json`

arteymix:
VSGI is closer to WSGI and Node.js in the sense that it provides a platform for writting frameworks and simple Web applications using Vala, C and other languages targeted by GObject Introspection (Python, JavaScript, ...). It also supports various server backends to ease deployment.

It's just that so far, there's not so many Web frameworks in Vala, so I do not want to maintain two distinct projects.

nbrady-techempower:
Great, thanks for the explanation. 

Also, looks like valum failed on this run

arteymix:
I have to investigate this because it occasionally happen on the VM environment. Likely due to an old libsoup-2.4 version.

nbrady-techempower:
hey @arteymix 

We are releasing preview 1 data within the next day or two. Was wondering if there's anything we can do to get this in for the next preview?

arteymix:
@nbrady-techempower I think the best would be to include this as it works and possibly postpone Valum until I get some time to investigate the issue.

Issue #2783, opened 2017-05-14
Octane/Round 14 should be considered stripped


fredrikwidlund:
Please note that "Octane" (currently #1 in plaintext) does not even attempt to implement the HTTP protocol at all. It works as a TCP daemon that count the X occurrences of "\n\r" (supposedly unique as part of the double newline \r\n\r\n) to assume a request is finished, assumes a specific request without parsing the input, and dumps X times a raw prebuilt "reply"

methane:
How about just remove it?

I feel "stripped" approach means disable some option or middleware in the framework
which usually enabled in production.

nbrady-techempower:
@simongui Can you speak to this please?

stevehu:
There are several frameworks only implement plaintext so they are not using any routing rules to find the handler. It is unfair to other frameworks which implement all tests as routing takes a lot of CPU time. I think we should remove them or add an indicator for partial implementation. What do you think?

methane:
I dislike personal projects which has no (or only one) production usage are listed and maintained.
[Octane](https://github.com/simongui/octane/commits/master) seems personal, experimental,
one-month project.

stevehu:
@methane I totally agree with you. It is not a framework as there are only two files in the project.

fredrikwidlund:
Stevehu, libreactor implements json and plaintext, and has a request routing layer.

fredrikwidlund:
The main issue with Octane is that it does not parse HTTP.

stevehu:
@fredrikwidlund Thanks for the clarification. Most framework will have a routing layer to handler HTTP request. The most simple one will be hard-coded if/else but a real router should support pattern matching in order to parse uri parameters. In later case, the sequence of the routes and the number of the routes will impact performance especially plaintext as the handler is too small. That is the reason I recommend to have flag to indicate partial or full implementation.

fredrikwidlund:
@stevehu I understand the reasoning, agree that request routing should be included. However I do not understand why you include libreactor (my framework) in this context?

stevehu:
@fredrikwidlund I didn't read the code but just looked at the test result to check if the framework is partial implementation or not. My main point is about partial implementation not about if the framework has a routing layer or not. libreactor only implements json, plaintext so it is included in my list.

fredrikwidlund:
[...]

PR #2789, opened 2017-05-15
C/octane removal

nbrady-techempower:
Per discussion in #2783


nbrady-techempower:
Since @simongui is around, let's continue this discussion in the issue.

PR #2795, opened 2017-05-18
Several Toolset Verification Updates

### LINKED ISSUE #2634: "run-tests.py --mode verify" should confirm the client machine can reach the server
Over the past two days, I've noticed three frameworks that passed the verification step but got zero RPS in actual test runs.  It turns out they all were accepting connections from localhost but not from the client machine that runs wrk.

The verification step should include a test for connectivity from the client machine.  A single successful request to any one of the server's endpoints should be sufficient for passing.

Other notes:
* Accepting connections from localhost only is the default setting for many frameworks / HTTP servers, so it's easy to run into this issue by accident.
* When this issue occurs, there usually aren't any useful clues in the server logs.  The most useful clue appears in the raw wrk output, where there should be some "connection refused" messages.

nbrady-techempower:
resolves #2634 

Now, during the verification step, there will be a check to see if the server is reachable from the client machine. The first step remains, which does header verification using python from the app server, and will throw errors as normal with useful information. If this step doesn't fail, then the client check will be made. Seeing that check fail will usually mean the server was set up correctly but only accepting connections from localhost.

### Queries

This PR revealed that only some frameworks are not sanitizing the query param in the queries and updates test. Since the [Test Requirements](http://frameworkbenchmarks.readthedocs.io/en/latest/Project-Information/Framework-Tests/#multiple-database-queries) make it clear that missing or non-integer parameters should be interpreted as `1`, no response in these cases will now be marked as failing.

_Those are now fixed_

### JSON

JSON now warns if response bytes are greater than 27 -> {"message":"hello, world!"}

### DIFF
--- a/frameworks/Erlang/cowboy/src/query_handler.erl
+++ b/frameworks/Erlang/cowboy/src/query_handler.erl
@@ -12,18 +12,19 @@ init(_Transport, Req, []) ->
 
 handle(Req, State) ->
         random:seed(erlang:now()),
-        {JSON, Req2} = case cowboy_req:qs_val(<<"queries">>, Req) of
-    {undefined, Req1} ->
-      {result_packet, _, _, [[ID, Rand]], _} = emysql:execute(test_pool, db_stmt, [random:uniform(10000)]),
-      {[{[{<<"id">>, ID}, {<<"randomNumber">>, Rand}]}], Req1};
-    {N, Req1} ->
-      I = list_to_integer(binary_to_list(N)),
-      Res = [ {[{<<"id">>, ID}, {<<"randomNumber">>, Rand}]} || 
+    {N, Req1} = cowboy_req:qs_val(<<"queries">>, Req, <<"1">>),
+
+      I = try binary_to_integer(N) of
+            X when X > 500 -> 500;
+            X when X < 1 -> 1;
+            X -> X
+          catch error:badarg -> 1 end,
+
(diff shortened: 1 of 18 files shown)


nbrady-techempower:
@daviddenton If you have a chance, would you mind updating fintrospect  to account for missing queries paramater? (Fine to do in a separate PR if you have time.)

daviddenton:
Sure I'll take a look :)

nbrady-techempower:
@daviddenton You rock! thanks!

daviddenton:
@nbrady-techempower that's strange. From the code it looks like this case should be handled in fintrospect.
Can you please just confirm the request URL that the framework isn't responding correctly to and I'll take a closer look.

Thanks :)

nbrady-techempower:
@daviddenton I tried for a while to get that test working. It appears that when there is no parameter given, the `numberOfQueries` and `multipleRoute` vars are never evaluated. I assumed it had to do with `Query.optional` and tried `Query.required` but couldn't quite figure out the api for it.

daviddenton:
@nbrady-techempower I've worked out that it's the empty query that is causing the problem - it behaves ok when there is an invalid or missing query - which confused me for a while!

I'm not at a machine at the moment, but if you change line 40 of DatabaseRoutes to:

```    
val numberOfQueries = Query.optional(ParameterSpec.string(StringValidations.EmptyIsValid).map {
```

... it should start to work ok. If it can wait then I'll PR tomorrow.

Thanks!

nbrady-techempower:
@daviddenton that did it! thanks! :)

PR #2801, opened 2017-05-19
Only Prepare DbCommand once

markrendle:
Moved the `cmd.Prepare()` call from `ReadSingleRow` to `CreateReadCommand` so it's only executed once instead of repeatedly.

### DIFF
--- a/frameworks/CSharp/aspnetcore/Benchmarks/Data/RawDb.cs
+++ b/frameworks/CSharp/aspnetcore/Benchmarks/Data/RawDb.cs
@@ -39,9 +39,6 @@ namespace Benchmarks.Data
         
         async Task<World> ReadSingleRow(DbConnection connection, DbCommand cmd)
         {
-            // Prepared statements improve PostgreSQL performance by 10-15%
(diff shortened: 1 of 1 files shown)


markrendle:
OK, that's weird, that ran OK on my PC.

nbrady-techempower:
Thanks! I'm going to wait for @nathana1 to review and approve this before merging since the Microsoft guys have been working on these tests.

nathana1:
Mike, can you take a look at this?

@mikeharder

mikeharder:
I recommend ignoring whitespace in the diff: https://github.com/TechEmpower/FrameworkBenchmarks/pull/2801/files?w=1

This seems reasonable.  @anpete: What do you think?

CC: @dougbu

anpete:
@mikeharder Looks good. A further improvement here could be to re-use the DbCommands themselves - via pooling or similar.

mikeharder:
@nbrady-techempower: We approve this change.

nbrady-techempower:
Thanks guys!

PR #2803, opened 2017-05-19
ULib: code and setup update

stefanocasazza:


nbrady-techempower:
Thanks @stefanocasazza ! I will review on Monday! :)

PR #2805, opened 2017-05-19
Add a version of the vertx-web postgres test that uses the com.github.susom:database library and standard JDBC driver

garricko:


pmlopes:
This database module is not official part of *Eclipse Vert.x* so the metadata should be properly annotated to reflect that.

PR #2821, opened 2017-05-24
Update Continuous Benchmarking

nbrady-techempower:
This PR is meant to do a few things:

* Adds a `README.md` and `tfb.conf` to the repo to show how to set up continuous benchmarking should our current environment be lost.

* Separates `run-continuously.sh` from the repo when the `tfb service` is run, making tearing down and rebuilding (deleting and re-cloning) easier to follow.

* Separating the task list (`run-tasks.sh`) from the `run-continuously.sh` allows us to modify tasks such that changes to the continuous benchmarking toolset will be reflected in the next run. Now, only changes to `run-continuously.sh` will require manually restarting the service. That file should not require changes, as it simply rebuilds the environment and starts the task list.


toolset/continuous/run-continuously.sh
```diff
@@ -0,0 +1,41 @@
...
+  # so that it may tear down the environment and run without interruption.
+  # If this file is changed, the tfb service will need to be restarted.
+
+  # Now if tasks are updated in the continuous toolset, they will execute
+  # properly on the next full run.
+
+  # Start by tearing down the environment.
+  # Hold on to the benchmark.cfg it it exists
+  echo Tearing down and rebuilding the environment
+  cp $TFB_REPOPARENT/$TFB_REPONAME/benchmark.cfg $TFB_REPOPARENT/bencmark.cfg 2>/dev/null
```
RX14: Typo: `bencmark.cfg`

nbrady-techempower: Nice catch!

PR #2822, opened 2017-05-25
Fix http4k setup files, and add undertow server backend, database and fortunes tests

daviddenton:
If we could get this into the first R15 Preview round that's be awesome. Probably best to squash and merge this PR as there were several attempts at getting the config right :)


nbrady-techempower:
Shouldn't be a problem @daviddenton 
I'll merge when it passes travis! Thanks!

daviddenton:
@nbrady-techempower It looks like there are several builds stacked up and that there are spurious failures on other frameworks (at least on the last build). It might be worth at least cancelling all but the last of the build requests for this PR.

Is there a policy regarding merging PRs where other tests have failed?

nbrady-techempower:
@daviddenton I try to stay on top of the built up runs and just cancel everything but the most recent.

We were at a point recently where every test was passing on travis (which should be the case) however, some have started failing recently due to dependencies not being tied down correctly, etc. I know of the few that are failing on their own and then I just verify that nothing your PR has done could be causing other failures and merge when ready.

daviddenton:
Cool. Have TE looked into making the travis job a bit more intelligent, so it only runs tests for the frameworks that have changed? It seems pretty wasteful to run 150ish builds for every change made to a single framework, especially when there is some fragility there. :)

msmith-techempower:
@daviddenton The Travis-CI script checks and only runs the tests that have had changes.

nbrady-techempower:
^ We *try* to make sure this is the case, including if you're changing a setup script that might affect multiple frameworks, only those frameworks will run. Sometimes it misfires, but it generally works.

PR #2824, opened 2017-05-25
CachedQueryTestType Scaffolding

nbrady-techempower:
This provides all the scaffolding for the Cached Query test without any test implementations.


nbrady-techempower:
Calling on some volunteers to be the first to add cached query tests. See work in progress details here: http://frameworkbenchmarks.readthedocs.io/en/latest/Project-Information/Framework-Tests/#caching

The work would have to be done from forking this PR branch. And you'll need to add this line to your benchmark_config.json:

```
      "cached_query_url": "/cached_worlds?queries=",
```

Pinging recent contributors: @zloster @stefanocasazza @daviddenton @simongui @benmccann @volyrique @dantti

volyrique:
While the caching implementations will differ between frameworks, one parameter that is most probably going to be universal is the cache capacity. Do you intend to specify it? If not, is it sensible to set it to 10000 entries (which implies that the whole CachedWorld table might be cached)?

frameworks/JavaScript/nodejs/handlers/mysql-raw.js
```diff
@@ -21,6 +26,16 @@ const queries = {
   }
 };
 
+const populateCache = (callback) => {
+  if (cachePopulated) return callback();
+  connection.query(queries.ALL_WORLDS, (err, rows) => {
```
volyrique: Will the CachedWorld table be used (as the specification says right now), or the regular World table?

nbrady-techempower: I think we settled on just the regular world table. Will update the specifications soon. Still a work in progress.

nbrady-techempower: Also yes i think it's sensible to suggest the entire table might be cached

volyrique:
All right, I have done an initial implementation for h2o [here](https://github.com/volyrique/FrameworkBenchmarks/tree/cached-query), but I haven't tested it properly yet. I will open a pull request after this one is merged and the specification is finalized.

nbrady-techempower:
@volyrique Ok thanks! I'm going to merge this is now so it's easier to open pull requests for testing. Because there will probably be few test implementations by the time round 15 is released, it's likely that we'll provide preview feedback data but won't publish caching results. We'll discuss with the community when the time comes.

PR #2826, opened 2017-05-26
Changed column to retrieve documents.

whitingjr:
This PR solves why results for vertx-web and MongoDB are as they are. The vertx-web application has been changed to use the proper column for MongoDB queries.


pmlopes:
This looks good for me and the CI failures are not related or from the vertx-web code at all.

PR #2827, opened 2017-05-27
H2O: Implement the cached queries test type

volyrique:
The implementation uses the [in-memory caching mechanism](https://github.com/h2o/h2o/blob/v2.2.2/include/h2o/cache.h) provided by libh2o. Now, that mechanism is pretty simple, and I would like to know if it is acceptable as a _bona-fide cache library_, so let me mention what distinguishes it from a plain key-value map: First of all, it supports multi-threaded operation, i.e. several threads can use the same cache simultaneously. Also, it has a mechanism to limit its capacity; if the limit is exceeded, then the cache removes the least recently used element. In addition to that, it implements time-based expiration of cache entries (the maximum duration is specified at cache creation). Please, let me know if this implementation choice is unacceptable. FYI, H2O itself uses the same mechanism to implement its SSL session cache.

### DIFF
--- a/frameworks/C/h2o/benchmark_config.json
+++ b/frameworks/C/h2o/benchmark_config.json
@@ -9,6 +9,7 @@
       "fortune_url": "/fortunes",
       "update_url": "/updates?queries=",
       "plaintext_url": "/plaintext",
+      "cached_query_url": "/cached-worlds?queries=",
       "port": 8080,
       "approach": "Realistic",
       "classification": "Platform",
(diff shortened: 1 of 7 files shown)


bhauer:
@volyrique Thanks very much for contributing a test implementation for a new test type!

Also thank you for taking the time to explain the caching implementation in some detail.  I suspect there will be some iteration of the requirements, and with that some more clarity on what qualifies as a bona-fide cache library, as we see the variety of implementations and have additional people taking a look at each other's code and choices.  That said, the implementation that you've described sounds like a good match for what we had in mind.

PR #2829, opened 2017-05-30
Fix ruby hanami

Jesterovskiy:


nbrady-techempower:
Thanks for taking care of this!

PR #2834, opened 2017-05-31
Ktor: update benchmark

cy6erGn0m:
- sampling call logging
- run database tasks on a separate thread pool

### DIFF
--- a/frameworks/Kotlin/ktor/src/main/kotlin/org/jetbrains/ktor/benchmarks/Hello.kt
+++ b/frameworks/Kotlin/ktor/src/main/kotlin/org/jetbrains/ktor/benchmarks/Hello.kt
@@ -3,17 +3,19 @@ package org.jetbrains.ktor.benchmarks
 import com.google.gson.*
 import com.mysql.jdbc.*
 import com.zaxxer.hikari.*
+import kotlinx.coroutines.experimental.*
 import org.jetbrains.ktor.application.*
 import org.jetbrains.ktor.content.*
 import org.jetbrains.ktor.features.*
 import org.jetbrains.ktor.host.*
 import org.jetbrains.ktor.http.*
-import org.jetbrains.ktor.logging.*
 import org.jetbrains.ktor.netty.*
 import org.jetbrains.ktor.routing.*
 import org.jetbrains.ktor.util.*
-import java.sql.ResultSet.*
+import java.sql.ResultSet.CONCUR_READ_ONLY
+import java.sql.ResultSet.TYPE_FORWARD_ONLY
 import java.util.concurrent.*
+import java.util.concurrent.atomic.*
 import javax.sql.*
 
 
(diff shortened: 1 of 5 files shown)


frameworks/Kotlin/ktor/src/main/kotlin/org/jetbrains/ktor/benchmarks/Hello.kt
```diff
@@ -37,64 +50,72 @@ fun main(args: Array<String>) {
...
-                    val result = mutableListOf<World>()
-
-                    connection.prepareStatement("SELECT * FROM World WHERE id = ?", TYPE_FORWARD_ONLY, CONCUR_READ_ONLY).use { statement ->
-                        for (i in 1..(queries ?: 1)) {
-                            statement.setInt(1, random.nextInt(DbRows) + 1)
-
-                            statement.executeQuery().use { rs ->
-                                while (rs.next()) {
-                                    result += World(rs.getInt("id"), rs.getInt("randomNumber"))
+                val response = run(databaseDispatcher) {
```
martin-g: I am not sure this is allowed.
Recently @zloster reported a similar issue with another test where the DB update has been executed asynchronously.
@nbrady-techempower Am I correct ?

cy6erGn0m: I see reason to fail here as transaction is performed in one thread (no thread switch in the middle of transaction)

nbrady-techempower: Looking at the updates section, this _looks_ ok. Just to confirm @cy6erGn0m, `get("/updates")` is doing all sql executions synchronously, and `executeUpdate()` resolves before the result for that call is returned?

zloster: Warning: I'm not a Kotlin expert so feel free to correct me.
@martin-g Looking at [the docs](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.experimental/run.html) for `run(...)` it seems OK, because later there is the [following line](https://github.com/TechEmpower/FrameworkBenchmarks/pull/2834/files/b545caa5fbe918d0527048fde7c824b613b0d2ec#diff-4bb4adade5ee157dcce0e7cdc46088a2R78) The problem with the other Kotlin implementation was that the result of the updates was not served to the client AFTER the execution of the queries. IMO here we have the necessary wait to see if execution (of the SQL queries) will complete before serving the response to the client. Also this [seems to be the documentation](https://kotlinlang.org/docs/reference/coroutines.html) about the `coroutines`.

cy6erGn0m: @zloster @nbrady-techempower `run()` is waiting for the sub-coroutine completion so we asynchronously waiting for completion and after that do `call.respond` that actually does HTTP response processing, inside of `run` block we only prepare response object `TextContent` with the result of execution.

cy6erGn0m: the significant difference is that we run execution on a separate pool so requests execution pool is released and we do not block other http requests (for example plaintext or even non-existing urls that will procude 404)

PR #2839, opened 2017-06-02
Update symfony setup to use php7

raziel057:


nbrady-techempower:
@raziel057 The version of symfony for these wasn't supporting php7. We had tried all the php frameworks when we added php7 support. Please feel free to reopen your Pull requests if you manage to get them working. Also, symfony is one of the tests i'd like to have the `stripped` version moved under the same framework directory.

Thanks for the effort!

raziel057:
@nbrady-techempower I didn't realized it was not the last release of Symfony. I will try to upgrade to Symfony 3.3 this week end. I think there is no need to have a ``normal`` version that contains dependencies to unused bundles. For me only the ``stripped`` version make sense.

Best regards,

nbrady-techempower:
@raziel057 Awesome! Thanks!

PR #2841, opened 2017-06-02
Upgrade NodeJS to v8.0.0

nbrady-techempower:
This should see a small bump in performance for all the JavaScript frameworks. There was a move to get most of the code over to es2015, which did result in some memory leak issues found in node v7.x.

### DIFF
--- a/toolset/setup/linux/webservers/nodejs.sh
+++ b/toolset/setup/linux/webservers/nodejs.sh
@@ -2,7 +2,7 @@
 
 fw_installed node && return 0
 
-VERSION="7.10.0"
+VERSION="8.0.0"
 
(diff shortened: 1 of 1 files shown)


nbrady-techempower:
@Helidium Can you check out the mitol failures for version 8.0?

Helidium:
@nbrady-techempower I have updated the Makefile and added the build target.
Nothing is needed from your part.

Btw. is it possible to look at some draft preview results? When do you expect to release preview 15?

nbrady-techempower:
Thanks! Will have some preview results at the beginning of this week

Helidium:
Great! Thanks.

PR #2848, opened 2017-06-06
Uvicorn

tomchristie:
Hi folks, adding a new "platform" type framework here, for Python 3.

Some more details here: https://github.com/tomchristie/uvicorn


frameworks/Python/uvicorn/uvicorn_conf.py
```diff
@@ -0,0 +1,13 @@
...
+_is_travis = os.environ.get('TRAVIS') == 'true'
+
+workers = multiprocessing.cpu_count()
+if _is_travis:
+    workers = 2
+
+bind = "0.0.0.0:8080"
+keepalive = 120
+errorlog = '-'
+pidfile = 'uvicorn.pid'
```
nbrady-techempower: hey @tomchristie Thanks for the submission. One of the issues we're having with a few frameworks is where they are writing certain files. Where is `unicorn.pid` being stored? If it's within the framework directory, can we point it to `/tmp` or `$IROOT`?

tomchristie: Sure thing. Few questions off that:

* We could also alternatively avoid writing a pid file completely, if you prefer?
* Right now I'm just following the convention of other Python frameworks, which [write an equivalent `gunicorn.pid`](https://github.com/TechEmpower/FrameworkBenchmarks/search?utf8=%E2%9C%93&q=gunicorn.pid). I'd be happy to pull together a pull request changing those too, if that's helpful?
* Is either of `/tmp` or `$IROOT` preferable to the other?

nbrady-techempower: @tomchristie That would be fantastic if you could do that for all the frameworks! `/tmp` is preferable since it gets cleaned at the end of each framework test

tomchristie: @nbrady-techempower Updated this PR for now. I'll aim to address the other Python framework pid files in due course.

tomchristie:
Okay, I'm happy with how this is looking now, from my side anyways! 😄 

Planning to add adding extra test types in a later PR, and I'll treat the `/tmp` on the other python frameworks as a separate pull request.

nbrady-techempower:
Thanks @tomchristie for the help!

PR #2849, opened 2017-06-07
All Ruby frameworks use Puma 3.9.x

nateberkopec:
Hi, Puma maintainer here. Puma 3.9 contained some important changes which may affect application throughput. All Ruby frameworks should use Puma 3.9.


nbrady-techempower:
Hey @nateberkopec ! Thanks so much!

nateberkopec:
Hmm, some of the frameworks got hung up on bundler-related issues. The sequel frameworks don't use a Gemfile.lock, so I'll just remove them from all the frameworks and see what happens.

nateberkopec:
Hey @nbrady-techempower, tests pass - couple of questions:

1) If there are warnings in the test output in Travis, does that basically mean the framework won't be included in the final Round results? Trying to figure out why things like rails-stripped etc don't appear.
2) Is there a date for Round 15?

nateberkopec:
Also, is there any way to get some info on CPU and memory usage during a run on the actual benchmarking hardware?

nbrady-techempower:
@nateberkopec -
* Warnings only mean there are probably ways to improve the test for better results. For example, you have an extra byte in your response, etc.
* There is no hard date for Round 15. We're working on some MySQL updates and toolset issues and are getting a preview out ASAP.
* Some environment info: http://frameworkbenchmarks.readthedocs.io/en/latest/Project-Information/Environment/

nateberkopec:
Cool, thanks.

Re: CPU and memory usage, here's why I ask: MRI Ruby framework benchmarks will be deeply dependent on having the correct settings for the *total number of processes and threads* running on the machine. The only way to make sure these settings are actually correct (and not just a guess, based on total hyperthreads) is to look at CPU and memory usage and determine if the machine could support additional processes/threads or not.

nbrady-techempower:
@nateberkopec We try and supply some global ENV variables for you with some of that information, but if you have another way to get more accurate information, feel free to create other variables during the setup scripts we discussed earlier. And if you have any suggestions about other ENV variables we can set for you, certainly let me know. I'd like to make it as easy as possible for you to get the machine-specific information you need.

nateberkopec:
(this is ready to merge, btw)

What I'm concerned about isn't the characteristics of the machine, I definitely have enough information on that, what I want to know is the *actual resource consumption when this benchmark is run*.

As an example, this issue #2850 could have been avoided if we had some kind of CPU number that said "this framework only used 5% of the available CPU". That's a very valuable red flag for configuration issues that would be useful across all frameworks. Unless I'm wrong, *most* TE benchmarks, especially the non-database-related ones, will be CPU bound, no?

nbrady-techempower:
Yeah, that makes sense. We don't currently have a way to capture that but that certainly would be valuable information. I'll make a note of that in the roadmap and look for avenues to capture and record that without causing some negative observer effect.

nbrady-techempower:
@nateberkopec Now that I think about it, we do use dstat to capture a lot of extra statistics, including usage for each cpu. The stats file in the results folder at each test level

PR #2850, opened 2017-06-07
Rails: Auto-tune unicorn and puma

nateberkopec:
Copy-pasted the script used for Sequel.

Is there a place to keep code which can be used *across* frameworks? For Ruby, all frameworks can configure Puma, Unicorn, and Passenger in almost exactly the same way, but right now they all do it differently, which leads to more abandoned frameworks running at massively sub-optimal settings. It would be better if all the frameworks could use this auto-tune script without having to copy-paste it between the projects. Is there a good place to put code like that? I guess I could ship a Rubygem but it would be better if it just lived in this repository.

### DIFF
--- a/frameworks/Ruby/rails/config/auto_tune.rb
+++ b/frameworks/Ruby/rails/config/auto_tune.rb
@@ -1,0 +1,41 @@
+#!/usr/bin/env ruby
+# Instantiate about one process per X MiB of available memory, scaling up to as
+# close to MAX_THREADS as possible while observing an upper bound based on the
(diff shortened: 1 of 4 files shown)


nbrady-techempower:
Hey @nateberkopec 

There's `/toolset/setup/linux/` and then subdirectories `databases`, `languages`, `frameworks`, `systools`, etc. If you put something like `unicorn.sh` in one of those directories, you can call it in any setup scripts by doing `fw_depends unicorn`

nateberkopec:
Thanks @nbrady-techempower. I'll leave that for a separate PR.

Issue #2845, opened 2017-06-05
MySQL Performance Degredation


nbrady-techempower:
As we work on getting you guys some preview data for Round 15, we've noticed a significant performance decrease for MySQL. After monitoring the Database server, there have been no real changes between Round 14 and Round 15. We've made a few slight modifications to the conf file which you can see here: https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/databases/mysql/my.cnf

Another change was the currently supported version of MySQL Server in the `apt` repository is now `5.5.55` https://dev.mysql.com/doc/relnotes/mysql/5.5/en/news-5-5-55.html

Most frameworks that have MySQL tests are seeing a whopping 20% to 50% decrease in RPS from Round 14 final.

raziel057:
Do you tried to revert the modifications of config file to check if it change something?

Which Mysql version was installed for Round 14?

nbrady-techempower:
@raziel057 Thanks for checking in here. The config file has been the same for quite some time. The only difference was what mysql-server version was available in apt. It was only a patch version update from `5.5.54` to `5.5.55`. Trying to revert back was difficult since that version is no longer in the apt repository, but after some effort it produced the same results.

We've moved to `5.7` which helped a lot but still seeing performance decreases for mysql db and query tests across the board. After some help with `mysqltuner` and folks from the mysql community, this is the current conf on our database server: https://github.com/nbrady-techempower/FrameworkBenchmarks/blob/d42de20a4a2b3b632f02ce9b282d882a88364dab/toolset/setup/linux/databases/mysql/my.cnf

raziel057:
If you have the same result (degradation compared to Round 14) by reverting the Mysql version, we can estimate the problem doesn't comes from Mysql no? Maybe related to the server itself? It's a bit difficult to help...

nbrady-techempower:
@raziel057 It's a bit difficult to solve too. If it were related to the DB server itself you would likely see the degradation across the board. MongoDB, PostgreSQL are both fine. It's quite the mystery. Nonetheless, we're moving forward with MySQL 5.7 and will work on tweaking the conf. Some spot tests actually show an uptick in the Updates test. We'll be releasing numbers (assuming the current run finishes without issue) around Monday.

zloster:
@nbrady-techempower @raziel057 

Some frameworks with **impressive** boost in the `Updates` test, all are using MySQL:
```
[...]

PR #2851, opened 2017-06-07
MySQL Update to 5.7

nbrady-techempower:
Still experiencing performance issues pointed out here #2845, but this did turn it around some


nbrady-techempower:
Last commit gets this working on vagrant as well. Looks like some log locations and the socket file were writing to different locations. An update to `usr.sbin.mysqld` got this working there.

This is working now in our ServerCentral environment and vagrant. If it passes travis, will merge.

PR #2856, opened 2017-06-09
Upgrade HTTP4K version with associated API changes

daviddenton:


nbrady-techempower:
Looks good @daviddenton 

We've been addressing some issues on the DB server, but we now have a run going that should produce results. So this won't be included, but your other changes will be. This will be in preview 2. Thanks!

PR #2857, opened 2017-06-09
optimize flask

methane:
Disable SQLALCHEMY_TRACK_MODIFICATIONS option.  It increase 10%
performance.

Additionally, use lower case letter for table name, because `create.sql`
use it.


nbrady-techempower:
As always, thank you @methane for your contributions! :)

PR #2865, opened 2017-06-14
Rust/hyper: Update hyper to 0.11 and mime to 0.3

robinst:
(CC @seanmonstar)


nbrady-techempower:
@robinst This lgtm. Did you want to wait on review from sean?

PR #2868, opened 2017-06-15
Uvicorn remaining test types

tomchristie:
Adds single queries, multiple queries, and updates.

Thanks to @maximilianhurl 😎


nbrady-techempower:
Thanks for adding these test types! Would love to see more caching tests if you are able to implement that as well!

tomchristie:
Not sure what you mean wrt caching tests? Is that a different test type that we've missed, or something else?

tomchristie:
Ah gotcha I see that now. Yup I'll take a crack, tho will prob treat it as a seperate pull request. 👍

tomchristie:
@nbrady-techempower Having taken a look at the caching test I'm going to punt on that. We could use in-memory caching for the test, but it's not something that'd reflect production usage given that you'd typically want the ability to scale past a single process on a single instance, and at least two instances for redundancy. I'd be more than happy to implement a caching test type if distributed caching ends up getting added at a later point. Welcome to take any further discussion on this off of this thread and onto the discussion group if that's helpful! 😄

nbrady-techempower:
@tomchristie No worries, that all makes sense! Thanks again for the contributions! :)

PR #2870, opened 2017-06-16
Removed the two memory options for the java process.

whitingjr:
This commit removes the options for the process. Allowing the ergonomics system to make a selection suitable for the machine running the benchmark.
 This change includes changing vertx and the vertx-web platform. Making them similar to the Netty project jvm configuration.

### DIFF
--- a/frameworks/Java/vertx-web/setup.sh
+++ b/frameworks/Java/vertx-web/setup.sh
@@ -2,7 +2,7 @@
 
 sed -i 's|localhost|'"${DBHOST}"'|g' src/main/conf/config.json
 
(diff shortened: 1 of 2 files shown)


frameworks/Java/vertx-web/setup.sh
```diff
@@ -2,8 +2,8 @@
 
 sed -i 's|localhost|'"${DBHOST}"'|g' src/main/conf/config.json
 
-fw_depends java maven
+#fw_depends java maven
```
nbrady-techempower: You'll need to uncomment this

Issue #90, opened 2013-04-05
New Test: Warp for Haskell


ghost:
http://www.haskell.org/haskellwiki/Web/Frameworks

PR #2871, opened 2017-06-16
Vertx header optimizations

whitingjr:
The changes in this PR affect the PLAINTEXT and JSON operations.

 Static header field are prepared using more efficient types. Incurring fewer allocations and instances.

 The changes tested on a dedicated machines provide the following improvements.

Plaintext.
Concurrency, Before, After

```
8       594124	   616546
16      1133246	1149831
32      1786700	1841151
64      2341128	2519504
128     3408049	3501957
256     3531211	3641489
1024    3622122	3772576
4096    3510294	3718517
16384   3327707	3660664
```

Json

Concurrency, Before, After
```
8    605786    639702
256  3320293   3734758
```

### DIFF
--- a/frameworks/Java/vertx/src/main/java/vertx/WebServer.java
+++ b/frameworks/Java/vertx/src/main/java/vertx/WebServer.java
@@ -30,6 +30,11 @@ public class WebServer extends AbstractVerticle implements Handler<HttpServerReq
 
 	private static final CharSequence TEXT_MESSAGE = new AsciiString("message");
 	private static final byte[] HELLO_WORLD = "Hello, world!".getBytes();
+
+	private static final String JSON_HELLO_WORLD = Json.encode(Collections.singletonMap(TEXT_MESSAGE, "Hello, world!"));
+	private static final Buffer JSON_HELLO_WORLD_BUFFER = Buffer.buffer(JSON_HELLO_WORLD);
+	private static final CharSequence JSON_HELLO_WORLD_CONTENT_LENGTH = new AsciiString(String.valueOf(JSON_HELLO_WORLD.length()));
+	
 	private static final Buffer HELLO_WORLD_BUFFER = Buffer.buffer(HELLO_WORLD);
 	private static final CharSequence HELLO_WORLD_CONTENT_LENGTH = new AsciiString(String.valueOf(HELLO_WORLD.length));
 
@@ -78,13 +83,12 @@ public class WebServer extends AbstractVerticle implements Handler<HttpServerReq
 	private void handlePlainText(HttpServerRequest request) {
 		request.response()
 		.putHeader(CONTENT_TYPE , RESPONSE_TYPE_PLAIN).putHeader(SERVER, VERTX)
-		.putHeader(DATE, dateString).putHeader(CONTENT_TYPE, HELLO_WORLD_CONTENT_LENGTH).end(HELLO_WORLD_BUFFER);
+		.putHeader(DATE, dateString).putHeader(CONTENT_LENGTH, HELLO_WORLD_CONTENT_LENGTH).end(HELLO_WORLD_BUFFER);
 	}
 
 	private void handleJson(HttpServerRequest request) {
-		Buffer buff = Buffer.buffer(Json.encode(Collections.singletonMap(TEXT_MESSAGE, HELLO_WORLD)));
 		request.response().putHeader(CONTENT_TYPE, RESPONSE_TYPE_JSON).putHeader(SERVER,  VERTX)
-		.putHeader(DATE, dateString).end(buff);
+		.putHeader(DATE, dateString).putHeader(CONTENT_LENGTH, JSON_HELLO_WORLD_CONTENT_LENGTH).end(JSON_HELLO_WORLD_BUFFER);
 	}
 	
 	public static void main(String[] args) {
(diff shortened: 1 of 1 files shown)


frameworks/Java/vertx/src/main/java/vertx/WebServer.java
```diff
@@ -75,14 +82,13 @@ public void stop(){
...
+		.putHeader(CONTENT_TYPE , RESPONSE_TYPE_PLAIN).putHeader(SERVER, VERTX)
+		.putHeader(DATE, dateString).putHeader(CONTENT_LENGTH, HELLO_WORLD_CONTENT_LENGTH).end(HELLO_WORLD_BUFFER);
 	}
 
 	private void handleJson(HttpServerRequest request) {
-		Buffer buff = Buffer.buffer(Json.encode(Collections.singletonMap(TEXT_MESSAGE, HELLO_WORLD)));
-		request.response().putHeader(HEADER_CONTENT, RESPONSE_TYPE_JSON).putHeader(HEADER_SERVER,  SERVER)
-		.putHeader(HEADER_DATE, dateString).end(buff);
+		request.response().putHeader(CONTENT_TYPE, RESPONSE_TYPE_JSON).putHeader(SERVER,  VERTX)
+		.putHeader(DATE, dateString).putHeader(CONTENT_LENGTH, JSON_HELLO_WORLD_CONTENT_LENGTH).end(JSON_HELLO_WORLD_BUFFER);
```
zloster: Hi, you should check the test rules about JSON test: http://frameworkbenchmarks.readthedocs.io/en/latest/Project-Information/Framework-Tests/#json-serialization
IMO this is violation of rules 1 and 3 for the JSON test.

nbrady-techempower: @zloster is correct. The change in lines 90/91 can't be accepted.

zapov: There have been other implementations doing the same since the beginning: https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/netty/src/main/java/hello/HelloServerHandler.java#L129
I suggest not making fuss about it, mark it as stripped and stop hiding stripped implementation by default

RX14: @zapov while your example does use a fixed content length, it still serializes the json on every request. This pr changes it so that the json is serialized once and then the result is reused in each request. This means there is no difference between this test and the plaintext test, so doing that is disallowed.

zapov: Right. My bad.

whitingjr: @zapov
 > mark it as stripped and stop hiding stripped implementation
what do you mean by stripped ?

zapov: Well, it's a separate discussion. I think implementations which add timer to refresh date every second and put that explicitly in the headers, or which manually put in json length (previously cached) should be marked as stripped in techempower grid. But I don't think they should be hidden by default. This way stripped doesn't mean something bad, just that it's not really a convenient/high level way to do an implementation.

whitingjr: ok

whitingjr: @zapov @zloster @nbrady-techempower I am content if there is a level playing field for every project. Otherwise the benchmark results published by TechEmpower are less useful than they could be.
@zapov 

> should be marked as stripped in techempower grid

I agree.

whitingjr:
@zloster @zapov @nbrady-techempower the latest changes have backed out the JSON application caching the json response.

frameworks/Java/vertx/src/main/java/vertx/WebServer.java
nbrady-techempower: @whitingjr 
> Otherwise the benchmark results published by TechEmpower are less useful than they could be.
The reason why we write the rules is to do our best to create a level playing field for every project. Obviously, we can miss things. This is an open source project so we rely on the community to help vet the frameworks, exactly like has happened here. If you feel like any framework is in violation of any of the rules we have written, please feel free to open an issue, or a pull request to address it.

We don't have a favorite here. We want a level playing field. However, we also rely on community support. We're not going to make everyone happy. We understand that this project is wide-ranging in scope. We don't believe that there is a `winner` of the benchmarks. We want you to take from it what you can; hopefully a way to help you improve your framework and help others improve theirs.

whitingjr:
Thank you for accepting the PR.

frameworks/Java/vertx/src/main/java/vertx/WebServer.java
```diff
@@ -28,15 +26,11 @@
 	private static final CharSequence RESPONSE_TYPE_PLAIN = new AsciiString("text/plain");
 	private static final CharSequence RESPONSE_TYPE_JSON = new AsciiString("application/json");
 
-	private static final CharSequence TEXT_MESSAGE = new AsciiString("message");
-	private static final byte[] HELLO_WORLD = "Hello, world!".getBytes();
```
zloster: @whitingjr IMO This code is acceptable for the `plaintext` test - using a byte array with the values of the plaintext response and just returning it. It is used by other participants. The rules are **different** for it.
I thought that I was clear that only the JSON instantiation is the problem. Also @nbrady-techempower  commented the particular lines that are problematic (#90-91).

nbrady-techempower: ^ I would accept this. Did not realize you also removed that.

PR #2874, opened 2017-06-17
upgrade packages for aiohttp

samuelcolvin:
aiohttp tests all failed apparently due to an error [compiling Cython modules](http://tfb-logs.techempower.com/round-15/preview-1/aiohttp/out.txt).

It looks like wheezeweb-py3 [failed for the same reason](http://tfb-logs.techempower.com/round-15/preview-1/wheezyweb-py3/out.txt).

I suspect there's a problem with the setup generally but this can't hurt and could solve the problem.

### DIFF
--- a/frameworks/Python/aiohttp/requirements.txt
+++ b/frameworks/Python/aiohttp/requirements.txt
@@ -1,10 +1,10 @@
-aiohttp==2.0.2
+aiohttp==2.1.0
 aiohttp-jinja2==0.13.0
 aiopg==0.13.0
-asyncpg==0.9.0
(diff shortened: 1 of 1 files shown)


nbrady-techempower:
Excellent! I'll test this in our SC env on Monday to see if it takes care of the issue.

samuelcolvin:
@nbrady-techempower did you test this on the environment, was the problem sorted?

nbrady-techempower:
@samuelcolvin Yes! All tests passed and it will be included in preview 2 which should be out just after 4th of july.

PR #2880, opened 2017-06-19
cutelyst: Fix some benchmark_config.json classification: to "Fullstack"

dantti:
Hi, I user pointed me that some classification was as "Platform" and others as "Fullstack" by definition I think "Fullstack" is more correct and it was also a copy paste mistake


nbrady-techempower:
LGTM @dantti Don't need a travis run for this!

PR #2884, opened 2017-06-21
Upgrade jetty

zloster:
- Jetty updated from the ancient 9.2 and 9.3 RC versions to latest 9.4;
- Merged the two projects (`jetty` and `jetty-servlet`). Now only `jetty` is a project;
- Changed `Jetty handler` implementation to `Realistic`.


nbrady-techempower:
Oh excellent! Thanks for this @zloster ! We really appreciate all the work you're doing!!

PR #2885, opened 2017-06-21
update to latest kemalyst

drujensen:
This updates the Kemalyst Tests to address compilation errors.


nbrady-techempower:
Thanks @drujensen 
Would you mind also adding the `shards.lock` file here? We should make sure that all github dependencies are locked down to a specific version.

drujensen:
@nbrady-techempower aaah!  No wonder this keeps breaking.  done.

Issue #593, opened 2013-11-04
Grizzly: Review JSON test implementation; instantiate object and then serialize


bhauer:
As Daryl Robbins points out in the Google Group [1], the Grizzly JSON serialization test implementation should be audited and modified as needed to reflect the spirit of the test.  Most importantly, the test type asks that a object be instantiated and then serialized, but the Grizzly implementation uses the Jackson streaming API directly to formulate a JSON response without having actually instantiated a Java object.

The implementation should be modified to more closely match the implementation seen in the Servlet test.

[1] https://groups.google.com/d/msg/framework-benchmarks/s3IG2GXVUWQ/gyHYUHgSgh0J

zloster:
This issue should be closed. The current code seems to be fine: https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/grizzly-bm/src/main/java/org/glassfish/grizzly/bm/JsonHttpHandler.java#L25-L33

nbrady-techempower:
Looks like this has since been resolved

PR #2887, opened 2017-06-22
Update Grizzly dependencies and remove character encoding from response

zloster:
Another note: this old issue #593 should be closed - the test implementation seems OK now.


nbrady-techempower:
Thanks @zloster

PR #2891, opened 2017-06-23
kemal - Use env.response as IO to avoid intermediate string

sdogruyol:
This updates Kemal to use `env.response` as existing `IO` instead of creating intermediates for every `to_json` call.

### DIFF
--- a/frameworks/Crystal/kemal/server-postgres.cr
+++ b/frameworks/Crystal/kemal/server-postgres.cr
@@ -54,7 +54,7 @@ end
 # Test 1: JSON Serialization
 get "/json" do |env|
   env.response.content_type = CONTENT::JSON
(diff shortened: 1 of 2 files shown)


nbrady-techempower:
This is failing db & json tests https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/246103748

RX14:
@sdogruyol the return value of `to_json` is being returned and rendered. Kemal shouldn't write the return value of handlers to the response if the response has already been written to. You can check this by inserting a custom IO wrapper into `context.response.output`.

faustinoaq:
@sdogruyol is this PR still necessary?

sdogruyol:
This is not working properly anyway. Closing 👍

PR #2892, opened 2017-06-23
ktor: split to netty and jetty tests, configure mysql (disable ssl)

cy6erGn0m:
- run ktor tests with both jetty and netty backends
  * build separate jars
  * use ktor development launcher
- disable SSL
- upgrade kotlin (1.1.2-5) and ktor (0.3.3)


frameworks/Kotlin/ktor/benchmark_config.json
```diff
@@ -2,15 +2,15 @@
   "framework": "ktor",
   "tests": [
     {
-      "default": {
+      "netty": {
```
nbrady-techempower: We still need a test marked as `default`.

cy6erGn0m: so if I have one `"default"` and one `"jetty"` then what test will be started if I specify no test name? (like `tfb -m benchmark`) Both or only the default one?

nbrady-techempower: The `default` test is the one that gets run.:

`tfb --test ktor` - runs the default test
`tfb --test ktor-netty` - runs the netty test in this case
`tfb --test-dir Kotlin/ktor` - would run all the tests in your benchmark config

frameworks/Kotlin/ktor/benchmark_config.json
```diff
@@ -21,7 +21,30 @@
...
+        "classification": "Micro",
+        "database": "MySQL",
+        "framework": "ktor",
+        "language": "Kotlin",
+        "orm": "Raw",
+        "platform": "Jetty",
+        "webserver": "None",
+        "os": "Linux",
+        "database_os": "Linux",
+        "display_name": "ktor netty",
```
nbrady-techempower: You also probably wanted `"display_name": "ktor jetty",` here

cy6erGn0m:
@nbrady-techempower anything else to be done?

PR #2893, opened 2017-06-24
Adding tests for ffead-cpp (mongodb, mysql, postgresql) on

sumeetchhetri:
- Embedded server
- Apache
- Nginx
Resolves https://github.com/TechEmpower/FrameworkBenchmarks/issues/2554


nbrady-techempower:
@sumeetchhetri ffead was removed because the test implementations did not live in the repo. This does not do anything to address that. See issue #2554

sumeetchhetri:
Done as suggested, Please refer implementation files at https://github.com/sumeetchhetri/FrameworkBenchmarks/tree/master/frameworks/C%2B%2B/ffead-cpp/te-benchmark
Refer usage in the setup script at https://github.com/sumeetchhetri/FrameworkBenchmarks/blob/master/toolset/setup/linux/frameworks/ffead-cpp-framework.sh (line 12/13)

sumeetchhetri:
Finally all tests are successful

nbrady-techempower:
Thanks @sumeetchhetri ! I will get around to reviewing this today or tomorrow. If it's all good, it will make it for Round 15 Preview 2

toolset/setup/linux/frameworks/ffead-cpp-apache-mongo.sh
```diff
@@ -0,0 +1,28 @@
...
+rm -rf web/default web/oauthApp web/flexApp web/markers
+chmod 755 *.sh resources/*.sh rtdcf/autotools/*.sh
+./server.sh > ffead-cpp-apache-mongo.log 2>&1
+while ! echo exit | nc localhost 8080; do sleep 5; done
+rm -f serv.ctrl
+sleep 10
+
+cd ${IROOT}
+
+touch ${IROOT}/ffead-cpp-apache-mongo.installed
```
msmith-techempower: This line should be updated to set up the environment when used. For example, the line in the setup scripts:
`${IROOT}/httpd/bin/apachectl restart > ffead.log 2>&1`
should be
`apachectl restart > ffead.log 2>&1`

So, this line should be something like
`echo -e "export PATH=${IROOT}/httpd/bin:\$PATH" > $IROOT/ffead-cpp-apache-mongo.installed`

toolset/setup/linux/frameworks/ffead-cpp-apache-mongo.sh
```diff
@@ -0,0 +1,28 @@
...
+cp -f web/te-benchmark/config/sdormmongo.xml web/te-benchmark/config/sdorm.xml
+rm -rf ffead-cpp-2.0-bin
+make build-apps
+rm -rf ${IROOT}/ffead-cpp-2.0
+cp -rf ffead-cpp-2.0-bin ${IROOT}/ffead-cpp-2.0
+cd ${IROOT}/ffead-cpp-2.0
+rm -rf web/default web/oauthApp web/flexApp web/markers
+chmod 755 *.sh resources/*.sh rtdcf/autotools/*.sh
+./server.sh > ffead-cpp-apache-mongo.log 2>&1
+while ! echo exit | nc localhost 8080; do sleep 5; done
```
msmith-techempower: This loop is unnecessary; please remove.

sumeetchhetri: ffead-cpp generates a lot of dynamic code on initial startup, which is where it generates and compiles stuff on the fly and creates a dynamic library. This is a one time init step which is very much required.

PR #2894, opened 2017-06-24
Added DB endpoints to crystal-raw

foliot:
I adapted the DB endpoints from the Crystal/Kemal framework code.

### DIFF
--- a/frameworks/Crystal/crystal/server.cr
+++ b/frameworks/Crystal/crystal/server.cr
@@ -12,7 +12,7 @@ server = HTTP::Server.new("0.0.0.0", 8080) do |context|
 
   response.headers["Server"] = "Crystal"
   response.headers["Date"] = Time.utc_now.to_s
-
+  
   case request.path
   when "/json"
     response.status_code = 200
@@ -30,11 +30,17 @@ server = HTTP::Server.new("0.0.0.0", 8080) do |context|
     response.status_code = 200
     response.headers["Content-Type"] = "application/json"
 
-    results = (1..sanitized_query_count(request)).map do
-      random_world
+    JSON.build(response) do |json|
+      json.array do
+        (1..sanitized_query_count(request)).each do
+          world = random_world          
+          json.object do
+            json.field "id", world[:id]
+            json.field "randomNumber", world[:randomNumber]
+          end
+        end
+      end
     end
-
-    results.to_json(response)
   when "/fortunes"
     response.status_code = 200
     response.headers["Content-Type"] = "text/html; charset=UTF-8"
@@ -46,18 +52,24 @@ server = HTTP::Server.new("0.0.0.0", 8080) do |context|
     }
 
     data.push(additional_fortune)
-    data.sort_by! { |fortune| fortune[:message] }
+    data.sort! { |f1, f2| f1[:message] <=> f2[:message] }
 
     ECR.embed "views/fortunes.ecr", response
   when "/updates"
     response.status_code = 200
     response.headers["Content-Type"] = "application/json"
-    
-    updated = (1..sanitized_query_count(request)).map do
-      set_world({id: random_world[:id], randomNumber: rand(1..ID_MAXIMUM)})
-    end
 
-    updated.to_json(response)
+    JSON.build(response) do |json|
+      json.array do
+        (1..sanitized_query_count(request)).each do
+          world = set_world({id: random_world[:id], randomNumber: rand(1..ID_MAXIMUM)})          
+          json.object do
+            json.field "id", world[:id]
+            json.field "randomNumber", world[:randomNumber]
+          end
+        end
+      end
+    end
   else
     response.status_code = 404
   end
--- a/frameworks/Crystal/crystal/server.cr
+++ b/frameworks/Crystal/crystal/server.cr
@@ -32,12 +32,8 @@ server = HTTP::Server.new("0.0.0.0", 8080) do |context|
 
     JSON.build(response) do |json|
       json.array do
-        (1..sanitized_query_count(request)).each do
-          world = random_world          
-          json.object do
-            json.field "id", world[:id]
-            json.field "randomNumber", world[:randomNumber]
-          end
+        1.upto(sanitized_query_count(request)) do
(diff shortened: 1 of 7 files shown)


frameworks/Crystal/crystal/server.cr
```diff
@@ -14,9 +22,72 @@ server = HTTP::Server.new("0.0.0.0", 8080) do |context|
...
     response.print "Hello, World!"
+  when "/db"
+    response.status_code = 200
+    response.headers["Content-Type"] = "application/json"
+    random_world.to_json(response)
+  when "/queries"
+    response.status_code = 200
+    response.headers["Content-Type"] = "application/json"
+
+    results = (1..sanitized_query_count(request)).map do
```
RX14: You could use `JSON.build` to stream the array here without allocating, which would be faster.

frameworks/Crystal/crystal/server.cr
```diff
@@ -14,9 +22,72 @@ server = HTTP::Server.new("0.0.0.0", 8080) do |context|
...
+    response.headers["Content-Type"] = "text/html; charset=UTF-8"
+
+    data = fortunes
+    additional_fortune = {
+      id:      0,
+      message: "Additional fortune added at request time.",
+    }
+
+    data.push(additional_fortune)
+    data.sort_by! { |fortune| fortune[:message] }
```
RX14: `sort_by!` is slower than using `sort! { |f1, f2| f1[:message] <=> f2[:message] }` because `sort_by!` tries to call the block only once per item (it allocates an entirely new array!). Since `f1[:message]` essentially has zero cost it's faster to do it like this.

@asterite This seems like a wart in the stdlib, thoughts?

frameworks/Crystal/crystal/server.cr
```diff
@@ -14,9 +22,72 @@ server = HTTP::Server.new("0.0.0.0", 8080) do |context|
...
+
+    ECR.embed "views/fortunes.ecr", response
+  when "/updates"
+    response.status_code = 200
+    response.headers["Content-Type"] = "application/json"
+    
+    updated = (1..sanitized_query_count(request)).map do
+      set_world({id: random_world[:id], randomNumber: rand(1..ID_MAXIMUM)})
```
RX14: ditto on the JSON optimization

foliot:
@RX14 The changes oddly didn't seem to impact the results. I'm currently rerunning the original code to double-check, but here is the updated code along with the original and updated results: [https://gist.github.com/foliot/5dbae98909896e4f7c391ca24dcd4c59](https://gist.github.com/foliot/5dbae98909896e4f7c391ca24dcd4c59).

The arrays may be small enough that the amount of time to manipulate them is negligible compared to querying the DB and writing to IO. In that case, I think the original code is better, since it's less verbose.

foliot:
I ran the tests again and there was some improvement, but there seems to be some variation between tests. The code should be more efficient using your suggestions, so we might as well go with that.

frameworks/Crystal/crystal/server.cr
```diff
@@ -30,11 +30,17 @@ server = HTTP::Server.new("0.0.0.0", 8080) do |context|
     response.status_code = 200
     response.headers["Content-Type"] = "application/json"
 
-    results = (1..sanitized_query_count(request)).map do
-      random_world
+    JSON.build(response) do |json|
+      json.array do
+        (1..sanitized_query_count(request)).each do
+          world = random_world
```
RX14: Simply call `random_world.to_json(json)` and it'll work, and be *much* cleaner. See [here](https://github.com/crystal-lang/crystal/blob/c6e2077a934c84184844c53d10dee31847e1713b/src/json/to_json.cr#L8).

frameworks/Crystal/crystal/server.cr
```diff
@@ -30,11 +30,17 @@ server = HTTP::Server.new("0.0.0.0", 8080) do |context|
     response.status_code = 200
     response.headers["Content-Type"] = "application/json"
 
-    results = (1..sanitized_query_count(request)).map do
-      random_world
+    JSON.build(response) do |json|
+      json.array do
+        (1..sanitized_query_count(request)).each do
```
RX14: `1.upto(sanitized_query_count(request)) do`?

frameworks/Crystal/crystal/server.cr
```diff
@@ -46,18 +52,24 @@ server = HTTP::Server.new("0.0.0.0", 8080) do |context|
...
-    
-    updated = (1..sanitized_query_count(request)).map do
-      set_world({id: random_world[:id], randomNumber: rand(1..ID_MAXIMUM)})
-    end
 
-    updated.to_json(response)
+    JSON.build(response) do |json|
+      json.array do
+        (1..sanitized_query_count(request)).each do
+          world = set_world({id: random_world[:id], randomNumber: rand(1..ID_MAXIMUM)})
```
RX14: ditto with the `to_json` but it'll be cleaner to use
```cr
world = set_world({id: random_world[:id], randomNumber: rand(1..ID_MAXIMUM)})
world.to_json(json)
```
here.

nbrady-techempower:
@foliot @RX14 Let me know when you guys want to merge this in. Looks like you're doing iterative work on it, which is great. We won't be releasing a Preview 2 until just after 4th of July so you have some time.

Also, while you're in this space, if there is a canonical way to do caching in Crystal it would be great to see another caching test, as we only have a few implementations right now. [Test #7](http://frameworkbenchmarks.readthedocs.io/en/latest/Project-Information/Framework-Tests/#caching)

RX14:
@nbrady-techempower I saw you implemented that test but crystal doesn't have a "canonical" way to do it. Personally i'd use redis but that's really not a obvious choice like crystal-db is right now (crystal-db is written by the core team and all the ORMs use it). I wouldn't be comfortable calling it "crystal" without qualifying it such that other caching methods (in-memory) could be used too.

foliot:
@nbrady-techempower I'm ready to merge unless @RX14 has some more suggestions. I think we'll have to do some more thinking about how to implement the caching test, since there's not a canonical way of doing that right now.

PR #2899, opened 2017-06-26
Update Resin to 4.0.53

zloster:
This is the latest version accessible for download here: http://caucho.com/products/resin/download/gpl#download


zloster:
These are the failures:
1. `compojure` is failing but the [verification result](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/246979612#L2538-L2553) is strange: some things pass, others fail.
1. `grails` is also failing but this time [everything fails](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/246979663#L2369-L2379). Side note: `grails` is downloading [embedded Tomcat 7](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/246979663#L2151-L2153) which probably is not needed.
1. `spark` is also failing but the [`Plaintext` is passing](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/246979699#L3057-L3067).

Resin logs should be inspected for clues in each case.

PR #2900, opened 2017-06-26
Update Maven to 3.5.0

zloster:
This is a custom shell script that mimics the behaviour of the `natecarlson PPA` i.e. all of the configurations are in `/usr/share/maven3/` directory and `mvn` is not added to the standart path shell variable. Features are:

- asking Apache servers for closest mirror
- GPG verification of the downloaded files

Notable change in Maven 3.5.0: there are color escape sequences in the terminal.

I've examined the other shell scripts in the TFB repository and this code is not following their style. But it should be relatively easy to be adapted.


zloster:
1. `restexpress` is failing because the `restexpress-mysql-raw` expects `Mongo` to be available: https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/246983561#L3449-L3450. Also note that it is trying to connect to `127.0.0.1` and not to `TFB-database`.
1. `dropwizard` is failing with [java.net.BindException: Address already in use](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/246983544#L4838) which is *very curious*.

Issue #2888, opened 2017-06-22
Java/Jooby JSON test implementation


zloster:
I've found this old issue: #593 and decided to check the other implementations in the Java "world".
And I've found this:
https://github.com/zloster/FrameworkBenchmarks/blob/master/frameworks/Java/jooby/src/main/java/com/techempower/App.java#L62-L63
It is the exact same described behaviour.

PR #2902, opened 2017-06-27
Java/play-2 JSON test must create object instance

zloster:
The test implementations are now using the [embedded Jackson](https://www.playframework.com/documentation/2.4.x/api/java/play/libs/Json.html) in Play 2.4.
Fix for #2888.


nbrady-techempower:
@zloster Thanks for getting to this one too!

PR #2904, opened 2017-06-28
Play2-scala now uses Play 2.6.0

mfirry:

### DIFF
--- a/frameworks/Scala/play2-scala/play2-scala/app/controllers/Application.scala
+++ b/frameworks/Scala/play2-scala/play2-scala/app/controllers/Application.scala
@@ -1,24 +1,24 @@
 package controllers
 
-import javax.inject.Singleton
-import akka.util.CompactByteString
-import play.api.http.HttpEntity.Strict
+import javax.inject.{ Inject, Singleton }
+
 import play.api.mvc._
 import play.api.libs.json.Json
 import play.mvc.Http
 
 @Singleton
-class Application extends Controller {
-  val header = ResponseHeader(OK, Map(Http.HeaderNames.SERVER -> "EXAMPLE"))
+class Application @Inject() (cc: ControllerComponents)
+extends AbstractController(cc) {
+
   implicit val helloWorldWrites = Json.writes[HelloWorld]
 
   def getJsonMessage = Action {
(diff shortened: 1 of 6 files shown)


mfirry:
I know, I should've squashed all those commits. Sorry.

nbrady-techempower:
@mfirry No need to squash. I sqaush with every merge.

This is failing:
```
Setup play2-scala: Oops, cannot start the server.
Setup play2-scala: java.lang.RuntimeException: No application loader is configured. Please configure an application loader either using the play.application.loader configuration property, or by depending on a module that configures one. You can add the Guice support module by adding "libraryDependencies += guice" to your build.sbt.
Setup play2-scala: 	at scala.sys.package$.error(package.scala:27)
Setup play2-scala: 	at play.api.ApplicationLoader$.play$api$ApplicationLoader$$loaderNotFound(ApplicationLoader.scala:43)
Setup play2-scala: 	at play.api.ApplicationLoader$.apply(ApplicationLoader.scala:70)
Setup play2-scala: 	at play.core.server.ProdServerStart$.start(ProdServerStart.scala:46)
Setup play2-scala: 	at play.core.server.ProdServerStart$.main(ProdServerStart.scala:22)
Setup play2-scala: 	at play.core.server.ProdServerStart.main(ProdServerStart.scala)
```

mfirry:
Can't understand what the issue is now

mfirry:
@nbrady-techempower any hint?

nbrady-techempower:
@mfirry Not really sure. Looks like there's some mongo issues? 
`Server play2-scala-reactivemongo: [error] r.c.a.MongoDBSystem - The entire node set is unreachable, is there a network problem?`

mfirry:
I haven't touched that at all though.

nbrady-techempower:
@mfirry I'll rerun the job and see if it was a one-off problem. The other changes we made were making sure all db services were shut off between tests, so if this tests relies on mongo, and mongo isn't in the setup file, it will fail. It would pass before if mongo was in a setup file that ran previously.

nbrady-techempower:
@mfirry Also, looks like launchpad was down. So that might be the source of all of it. Ran again and we'll see what the output is.

PR #2906, opened 2017-06-28
Java/servlet Caching implementation and updates

zloster:
- Updated dependencies
- Code clean up
- Added Caching implementation using [cache2k](https://cache2k.org/)

My notes about the test rules: IMO the test must mandate the usage of non-repeatable random numbers for the IDs. Here I've tried to use the API to get all the cache objects with one operation. But if there are duplicates in the keys list, the resulting list will contain less than the requested number of objects. And currently this is flagged as a warning:

> Accessing URL http://TFB-server:8080/servlet/cached-worlds?queries=:
>   Response (trimmed to 40 bytes): "[{"id":8829,"randomNumber":685}]"
>    WARN for http://TFB-server:8080/servlet/cached-worlds?queries=501
> 
>      JSON array length of 491 != expected length of 500

I suspect that most of the cache implementation will have similar semantic.


zloster:
Travis build for `servlet` will need a restart. The computer where the build was running was [powered off](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/248095179#L2336-L2339).

nbrady-techempower:
@zloster You are right, we can explicitly state that somewhere, however, it is implied by the following rule:

> The request handler must retrieve a **set** of CachedWorld objects, equal in count to the count parameter, from the CachedWorld database table. Note that the CachedWorld database table uses a schema idential to the World table used for Test #3 but is distinct to avoid confusion.

I'll get some wording for the rest of the team and see if we can't get that in the docs.

nbrady-techempower:
Restarted the job

zloster:
More updates:

- Introduce Maven profiles and update shell scripts to reflect that
- Enable only the relevant servlets i.e. when testing PostgreSQL don't deploy MySQL and plaintext/JSON servlets

Please, don't merge yet. I want to update the `README.md`

nbrady-techempower:
@zloster sure, let me know

zloster:
@nbrady-techempower Ready.

 - Documentation updated
 - Removed the duplicated code for DB and Fortune tests for MySQL and PostgreSQL. The Update test is still different: MySQL is using `batch updates`, PostgreSQL in not
 - Also minor clean up in the **Java/Jetty** README.md

TODO: different implementation for single query and multiple queries tests. Currently there are warnings due to usage of the same code.

PR #2919, opened 2017-07-18
Updated weppy framework to v1.1

gi0baro:


frameworks/Python/weppy/requirements-pypy.txt
```diff
@@ -1,4 +1,4 @@
 pg8000==1.10.6
-weppy==1.0.1
+weppy==1.1
 gunicorn==19.4.5
 tornado==4.3
```
Dith3r: This could be also upgraded to latest version of tornado.

frameworks/Python/weppy/requirements-pypy.txt
```diff
@@ -1,4 +1,4 @@
 pg8000==1.10.6
```
Dith3r: Version of 1.11.0 was released

frameworks/Python/weppy/requirements-pypy.txt
```diff
@@ -1,4 +1,4 @@
...
-weppy==1.0.1
+weppy==1.1
 gunicorn==19.4.5
```
Dith3r: Version 19.7.1 was released

gi0baro:
@Dith3r all the version you mentioned for upgrade are the same in the other python frameworks. They should be upgraded on the whole framework set.

Dith3r:
For me if you upgrade version of framework you can also upgrade libraries it's depends on and test if everything works as intend. Anyway my comments are only suggestions.

gi0baro:
@Dith3r yup, my point is that a PR should be made to upgrade versions of the dependencies on every python framework that use them :)

PR #2920, opened 2017-07-23
Updates the Rust compiler to 1.19

JordiPolo:
Updates the Rust compiler to the latest stable version.
It uses LLVM 4.0 and that seems to have some performance implications, it would be useful to see  the results.
@alexcrichton


alexcrichton:
👍

PR #2922, opened 2017-07-24
Add Amber framework.

faustinoaq:
This PR adds Amber framework.

 [Amber](https://www.ambercr.io) is a web application framework written in [Crystal](http://www.crystal-lang.org) inspired by Kemal, Rails, Phoenix and other popular application frameworks.

amber

### DIFF
--- a/frameworks/Crystal/amber/setup.sh
+++ b/frameworks/Crystal/amber/setup.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+fw_depends postgresql crystal
+
+shards install --production
+
+crystal build --release --no-debug src/amber.cr
+
+for i in $(seq 1 $((2 * $(nproc --all)))); do
+  ./amber &
+done
+
+wait
--- a/frameworks/Crystal/amber/setup.sh
+++ b/frameworks/Crystal/amber/setup.sh
@@ -6,7 +6,7 @@ shards install --production
 
 crystal build --release --no-debug src/amber.cr
 
-for i in $(seq 1 $((2 * $(nproc --all)))); do
+for i in $(seq 1 $(nproc --all)); do
   ./amber &
 done
 
--- a/frameworks/Crystal/amber/shard.yml
+++ b/frameworks/Crystal/amber/shard.yml
@@ -9,11 +9,9 @@ license: MIT
 dependencies:
   amber:
     github: Amber-Crystal/amber 
-    version: ~> 0.1.4
 
   granite_orm:
     github: Amber-Crystal/granite-orm
-    version: ~> 0.6.2
 
   pg:
     github: will/crystal-pg
--- a/frameworks/Crystal/amber/shard.yml
+++ b/frameworks/Crystal/amber/shard.yml
@@ -1,0 +1,19 @@
+name: amber
+version: 0.1.0
+
+authors:
+  - Amber <contact@ambercr.io>
+
+license: MIT
+
+dependencies:
+  amber:
+    github: Amber-Crystal/amber 
+    version: ~> 0.1.4
+
+  granite_orm:
+    github: Amber-Crystal/granite-orm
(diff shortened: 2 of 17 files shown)


frameworks/Crystal/amber/shard.yml
```diff
@@ -0,0 +1,19 @@
...
+license: MIT
+
+dependencies:
+  amber:
+    github: Amber-Crystal/amber 
+    version: ~> 0.1.4
+
+  granite_orm:
+    github: Amber-Crystal/granite-orm
+    version: ~> 0.6.2
```
drujensen: @faustinoaq We will want the latest version before this gets merged.

faustinoaq: @drujensen I just see this PR https://github.com/Amber-Crystal/granite-orm/pull/39 :+1: 

Do I need to change version to master? or a new version of granite-orm will be released?

drujensen: @faustinoaq I pushed a new release `0.6.3`.  I recommend removing the version from the shard.yml, perform a `shard update` and then check in the `shard.lock` file so it doesn't get updated.

faustinoaq: @drujensen done :+1:

faustinoaq:
@nbrady-techempower I think this PR is working well, Could Amber be on round 15?

frameworks/Crystal/amber/setup.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+fw_depends postgresql crystal
+
+shards install --production
+
+crystal build --release --no-debug src/amber.cr
+
+for i in $(seq 1 $((2 * $(nproc --all)))); do
```
RX14: I'm not sure launching 2 processes per core will gain anything on the actual benchmark machine. I think it'd be best to actually benchmark this on the target machine. Excess process switches are bad.

drujensen:
@RX14 I found a small performance improvement since there seems to be some io wait time in the other processes, however I agree that `2 *` is excessive.  Adding just a couple more processes above the number of cpu's will suffice.

RX14:
@drujensen How many cores are you testing on locally?

drujensen:
@RX14 I tested on 2 cores but admittedly in a VM.  Maybe not a valid test?

faustinoaq:
Yes, then 1 process per core would be better?, I tested in my computer and
always the best result was when I executed the same amount as process as
cores.

drujensen:
ok, go ahead and remove the `2 *`.  Thanks.

RX14:
I don't think testing on 2 cores is very valid when the actual benchmark setup has 80 cores. At 80 processes, at least 1 of them will likely be running a GC at 1 time, and the GC runs concurrently. This means that i hypothesize that the ideal number of threads will be less than the number of cores. Unfortunately I don't have the benchmark machine to test that on.

drujensen:
@RX14 right, agree.

faustinoaq:
@drujensen Do you will update Kemalyst too?

nbrady-techempower:
Hey guys, let me know when you feel like this is ready to merge.

faustinoaq:
I tested it in my local machine without problems.

Also, all dependencies are updated.

@nbrady-techempower Could Amber be included on round 15?

@drujensen and @RX14, Do you think this PR is ready?

drujensen:
@nbrady-techempower @faustinoaq LGTM.  This is ready to be merged.  Thanks!

nbrady-techempower:
Thanks guys. @drujensen I'll get this in now and it should be included in Preview 3. It's been extremely busy in the office but we should have Preview 2 out any day.

drujensen:
@nbrady-techempower :+1:  Thanks!

PR #2923, opened 2017-07-24
Updates Rust Iron packages to latest version in the current minor and postgresql

JordiPolo:
Lifting the restriction of the last number should bring slightly newer dependencies. Also updated Postgresql. 

Updating further would be ideal but makes the compilation break and I do not know enough about these packages to fix it.


nbrady-techempower:
@JordiPolo This isn't passing. Any clues?

PR #2925, opened 2017-07-25
Remove flags to be like other Crystal setups

faustinoaq:
Other crystal projects aren't using `--no-debug` or `--production` flags.

This PR removes `--no-debug` and `--production` flags to have the same install and build command as other crystal projects in this repository.

See: [Crystal](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Crystal/crystal/setup.sh#L7), [Kemal](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Crystal/kemal/setup-postgres.sh#L7) and [Kemalyst](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Crystal/kemalyst/setup.sh#L7)

### DIFF
--- a/frameworks/Crystal/amber/setup.sh
+++ b/frameworks/Crystal/amber/setup.sh
@@ -2,7 +2,7 @@
 
 fw_depends postgresql crystal
 
-shards install --production
+shards install
 
 crystal build --release src/amber.cr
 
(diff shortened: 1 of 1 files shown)


faustinoaq:
@RX14 Can you approve these changes? Do you think remove these flags is useful?

I forget to remove them while I was coding this Amber test :sweat_smile:

RX14:
It won't make any difference but you might as well merge it.

PR #2926, opened 2017-07-26
disable static file handler

drujensen:
This disables the static file handler for Kemal.  This provides a major performance enhancement and would be a standard configuration for RESTful servers.


drujensen:
@RX14 @sdogruyol please review

@nbrady-techempower not sure if its possible, but this would be great to have in preview 2.

sdogruyol:
This is great improvement, thanks @drujensen 👍

PR #2936, opened 2017-08-15
Add IncludeOS Mana framework (C++ unikernel)

ciarancourtney:
Hi,

I think this is the first unikernel based framework to be added, in its current state it will not generate results as expected, the request count will be zero, as the framework becomes unresponsive soon after the benchmark begins, I'm still working on getting this resolved.

Also I think this is the first test to use docker to execute the framework (the native framework of IncludeOS is bare-metal or KVM hypervisor) as it requires Ubuntu 16.04 to build. The built unikernel is then executed on Ubuntu using QEMU, which is not a typical production scenario at all, the poor results probably reflect that. To fix this later would require booting the built unikernel on KVM (GCE for example)

Also the Travis build will fail at the "Building IncludeOS Mana server" step (CMake file does not exist error), but not locally on vagrant, couldn't figure this out, input welcome.

Feedback welcome!

Thanks

### DIFF
--- a/frameworks/C++/includeos-mana/benchmark_config.json
+++ b/frameworks/C++/includeos-mana/benchmark_config.json
@@ -1,0 +1,24 @@
+{
+  "framework": "includeos-mana",
+  "tests": [{
+    "default": {
+      "setup_file": "setup",
+      "plaintext_url": "/plaintext",
+      "port": 8080,
+      "approach": "Stripped",
+      "classification": "Platform",
+      "database": "None",
+      "framework": "Mana",
+      "language": "C++",
+      "flavor": "None",
+      "orm": "Raw",
+      "platform": "None",
+      "webserver": "None",
+      "os": "Linux",
+      "database_os": "Linux",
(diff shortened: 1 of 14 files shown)


nbrady-techempower:
@ciarancourtney Thanks for both of your recent submissions! Resources are a little tight for the next week or so but we should be able to review these shortly.

frameworks/C++/includeos-mana/benchmark_config.json
```diff
@@ -0,0 +1,24 @@
...
+  "framework": "includeos-mana",
+  "tests": [{
+    "default": {
+      "setup_file": "setup",
+      "plaintext_url": "/plaintext",
+      "port": 8080,
+      "approach": "Stripped",
+      "classification": "Platform",
+      "database": "None",
+      "framework": "Mana",
```
msmith-techempower: This should be `None` if the classification is `Platform` (or if there is a different framework used, list that here)

PR #2938, opened 2017-08-16
Add MirageOS HTTP server (OCaml unikernel)

ciarancourtney:
Hi,

Carrying on from #2936, this PR adds another popular unikernel implementation called [MirageOS](mirage.io)

Tested on my own Travis and the plaintext test passes with results.

There are two workarounds applied to the code that prevent memory leaks and max open fd exceptions respectively, hopefully these are resolved in subsequent releases.

Feedback welcome!

Thanks


frameworks/OCaml/mirageos/setup.sh
```diff
@@ -0,0 +1,48 @@
...
+sudo ip addr add 10.0.0.1/24 dev tap0
+sudo ip link set dev tap0 up
+
+echo "Booting MirageOS Conduit HTTP Server"
+# The qemu command below is the output of '$ solo5-run-virtio -n tap0 ./conduit_server.virtio' with the addition of '-daemonize' so as to detach the qemu process (& detach will not work)
+qemu-system-x86_64 -daemonize -cpu Westmere -m 128 -nodefaults -no-acpi -display none -device virtio-net,netdev=n0 -netdev tap,id=n0,ifname=tap0,script=no,downscript=no -device isa-debug-exit -kernel conduit_server.virtio
+
+# FIXME
+# * Need to change TFB-Server to 10.0.0.2 in /etc/hosts, hostfwd doesn't seem to work with qemu tap network config
+# *
```
msmith-techempower: This entire file needs to be redone. OCaml should be installed in a similar fashion to how other languages are; see [DLang](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/languages/dlang.sh).

Also, we would prefer an installation which does not rely on `apt-get` if possible

frameworks/OCaml/mirageos/README.md
```diff
@@ -0,0 +1,24 @@
...
+Content-Length: 13
+
+Hello, world!
+```
+
+
+### TODO
+
+* Under load the server will crash with a Unix.EINVAL or similar exception, this open issue is 
+detailed [here](https://github.com/mirage/ocaml-cohttp/issues/503)
```
msmith-techempower: This is likely to be a problem in actual benchmarking efforts.

frameworks/OCaml/mirageos/setup.sh
```diff
@@ -0,0 +1,48 @@
...
+sudo ip tuntap add tap0 mode tap
+sudo ip addr add 10.0.0.1/24 dev tap0
+sudo ip link set dev tap0 up
+
+echo "Booting MirageOS Conduit HTTP Server"
+# The qemu command below is the output of '$ solo5-run-virtio -n tap0 ./conduit_server.virtio' with the addition of '-daemonize' so as to detach the qemu process (& detach will not work)
+qemu-system-x86_64 -daemonize -cpu Westmere -m 128 -nodefaults -no-acpi -display none -device virtio-net,netdev=n0 -netdev tap,id=n0,ifname=tap0,script=no,downscript=no -device isa-debug-exit -kernel conduit_server.virtio
+
+# FIXME
+# * Need to change TFB-Server to 10.0.0.2 in /etc/hosts, hostfwd doesn't seem to work with qemu tap network config
```
msmith-techempower: Using qemu may introduce latency for every lookup of `TFB-Server`, but I am unfamiliar.

nbrady-techempower:
@ciarancourtney Going to close this for now since it's getting stale. If you're able to address any of the reviews, or have any questions please feel free to re-open and continue.

PR #2939, opened 2017-08-16
Update .NET Core to 2.0.0

mikeharder:
CC: @dougbu

### DIFF
--- a/frameworks/CSharp/aspnetcore/Benchmarks/Benchmarks.csproj
+++ b/frameworks/CSharp/aspnetcore/Benchmarks/Benchmarks.csproj
@@ -13,17 +13,17 @@
 
   <ItemGroup>
     <PackageReference Include="Dapper" Version="1.50.2" />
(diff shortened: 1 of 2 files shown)


dougbu:
LGTM as long as the checks pass.

Great npgsql/Npgsql.EntityFrameworkCore.PostgreSQL#223 took less than a week to fix.

mikeharder:
@nbrady-techempower: It appears that Travis is running too many tests for this PR.  I believe it should only be running `CSharp/aspnetcore`, but it also ran `CSharp/aspnet` and `C++/ffead-cpp`.

nbrady-techempower:
@mikeharder We've put together quite the hack on the travis side to get around them not having a dynamic build matrix, so it happens every now and then. I usually dive in to the builds to see if the relevant tests pass. You guys are all set. Thanks!

mikeharder:
Fixed in https://github.com/TechEmpower/FrameworkBenchmarks/pull/2939

PR #2946, opened 2017-08-30
Sails can work in cluster

Crusader4Christ:
Sails should work in cluster like any other nodejs framework


nbrady-techempower:
Hi @Crusader4Christ would you mind rebasing this with our current master branch and then reopening. There have been updates to the travis env. This will need to get green check from travis before it can be merged.

Crusader4Christ:
#2999

PR #2953, opened 2017-09-04
kemalyst is now amber

drujensen:
This PR removes kemalyst from the benchmarks since it has merged with amber and is no longer being maintained.


nbrady-techempower:
@drujensen this just needs kemalyst removed from `.travis.yml` and then it should pass and be good to go

drujensen:
@nbrady-techempower ok, removed from `.travis.yml`

PR #2955, opened 2017-09-04
Akka http dependency updates

fedgehog:

PR #2956, opened 2017-09-05
Speed up the app packaging process

fedgehog:
Packaging of any non-trivial scala app with sbt-assembly plugin takes ages. 
To get faster feedback on changes this commit uses sbt-native-packager plugin
to prepare bundle to benchmark.


nbrady-techempower:
See comment on #2955

PR #2959, opened 2017-09-06
Update Kemal to 0.21.0

sdogruyol:
This PR updates Kemal to 0.21.0 and changes the repo to new organization

### DIFF
--- a/frameworks/Crystal/kemal/.gitignore
+++ b/frameworks/Crystal/kemal/.gitignore
@@ -1,4 +1,5 @@
 libs/
+lib/
 .crystal
 .shards
 !shard.lock
--- a/frameworks/Crystal/kemal/shard.lock
+++ b/frameworks/Crystal/kemal/shard.lock
@@ -6,11 +6,11 @@ shards:
 
   kemal:
     github: sdogruyol/kemal
-    version: 0.19.0
+    version: 0.21.0
 
   kilt:
     github: jeromegn/kilt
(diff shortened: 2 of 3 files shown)


faustinoaq:
I think we need to update crystal too. Currently TFB is using crystal 0.22.0

https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/languages/crystal.sh#L7

Also, if we update crystal to 0.23.1 then we need to add `--no-debug` flag to avoid https://github.com/crystal-lang/crystal/issues/4719

nbrady-techempower:
@sdogruyol looks like this change needs to be made to the committed `shard.lock` as well right?

`Setup kemal: Outdated shard.lock (kemal requirements changed). Please run shards update instead.`

faustinoaq:
@nbrady-techempower yeah, you are right, thanks your comment!

faustinoaq:
@sdogruyol you need to update `shard.lock` too, just delete the current `shard lock` and generate a new one. Also @drujensen made a PR to update crystal, so, all should work fine.

faustinoaq:
All crystal test are passing now, Why do Go/aah failed? Is it related to Crystal ?

nbrady-techempower:
@faustinoaq No, just a bug with the dynamic matrix that decides what tests need to be run based on the edits provided. I'll get this in for you guys.

PR #2979, opened 2017-10-02
rust/hyper: enable pipeline flushing strategy

seanmonstar:

### DIFF
--- a/frameworks/Rust/hyper/src/main.rs
+++ b/frameworks/Rust/hyper/src/main.rs
@@ -55,7 +55,9 @@ impl Service for TechEmpower {
 
 fn main() {
     let addr: SocketAddr = "0.0.0.0:8080".parse().unwrap();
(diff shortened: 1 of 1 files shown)


nbrady-techempower:
@seanmonstar looks like this isn't passing. Any clues? Going to close this for now. We've made some adjustments to travis. If you have a chance to poke at this would you mind rebasing off master and reopening? Thanks!

seanmonstar:
Oh shoot, I didn't realize there was a failing CI thing. There was indeed a bug in hyper around the same time I filed this, but new versions have fixed it. Is it possible to retry the build so that it runs with the newer version?

nbrady-techempower:
@seanmonstar hey sorry i missed this. I think it's also important to merge in master because there have been some other rust changes and travis updates that have been merged since. Unfortunately, the travis-ci strategy is to run the PR as it exists before merging, as opposed to after.

PR #2983, opened 2017-10-08
Implemented remaining tests for the Pronghorn server.

caturner81:
Added the remaining tests so Pronghorn implements all current tests.  Passes on TravisCI.


caturner81:
What happened here?  I bumped the version of the Pronghorn library and the Travis-CI build seems to have failed to install Java.  Looking at some of the other recent pull requests this seems to happen sometimes for other tests as well.  Is there something I should do differently, or is there something wrong with the way the Java is being installed in the benchmarks VM currently?

nbrady-techempower:
@caturner81 Probably nothing you need to do. Rerunning the test, we'll see what happens and troubleshoot from there

caturner81:
Looking at the PPA you guys use it seems this issue would have been fixed on October 20th.  I see you're working through pull requests again, may want to trigger another run on this one, it should pass now that the Java install should work again.

PR #2999, opened 2017-10-20
Sails can work in cluster too

Crusader4Christ:
Sails should work in cluster like any other nodejs framework


nbrady-techempower:
Thanks for reopening! Looks good now!

PR #3000, opened 2017-10-20
Sf3 improvments

raziel057:
Update Symfony Framework from 3.3.0 to 3.4.0
Cleanups and small performance improvements (removing loading of some unused bundles).
The specific symfony2-stripped can now be removed.


nbrady-techempower:
@raziel057 I've removed symfony2. If you have some time, would you mind rebasing from master and renaming this project to symfony?

raziel057:
@nbrady-techempower Thanks. I will try to do it this week end.

raziel057:
@nbrady-techempower Is it ok for you to merge?

frameworks/PHP/symfony/app/Resources/views/base.html.twig
```diff
@@ -1,3 +1,4 @@
+{% spaceless %}
```
nicolas-grekas: spaceless should be removed if possible: it adds runtime overhead

fabpot: +1 we don't recommend doing it in real projects.

raziel057: I though it was done only when compiling the template to PHP. Thanks for the tip.

raziel057:
@nicolas-grekas I can't find were is used the ``container.dumper.inline_class_loader`` parameter in Symfony 3.3.4. I have no result when searching ``inline_class_loader`` in vendors. Is it normal?

frameworks/PHP/symfony/app/Resources/views/base.html.twig
raziel057: The issue here is that without ``spaceless`` the resulting page is a bit bigger (41,27ko rather than 40,28 for fortune test). So in the benchmark it can impact the response transmission time. See https://www.techempower.com/benchmarks/#section=code&hw=ph&test=fortune

fabpot: Then, remove whitespace "by hand".

nicolas-grekas:
> parameter in Symfony 3.3.4

I read that as 3.4, where the param applies :)

frameworks/PHP/symfony/app/Resources/views/base.html.twig
```diff
@@ -0,0 +1,7 @@
+<!DOCTYPE html>
+<html>
```
nicolas-grekas: if we really want to reduce the size of the output, this is valid HMTL5:
```
<!DOCTYPE html>
<title>{% block title %}Welcome!{% endblock %}</title>
{% block body %}{% endblock %}
```

frameworks/PHP/symfony/composer.json
```diff
@@ -0,0 +1,72 @@
...
+		"php" : ">=5.5.9",
+		"doctrine/doctrine-bundle" : "^1.6",
+		"doctrine/orm" : "^2.5",
+		"incenteev/composer-parameter-handler" : "^2.0",
+		"sensio/distribution-bundle" : "^5.0.19",
+		"sensio/framework-extra-bundle" : "^3.0.2",
+		"symfony/monolog-bundle" : "^3.1.0",
+		"symfony/polyfill-apcu" : "^1.0",
+		"symfony/swiftmailer-bundle" : "^2.3.10",
+		"symfony/symfony" : "^3.3",
```
nicolas-grekas: would be great to replace y standalone components, the default config adds overhead otherwise

frameworks/PHP/symfony/composer.json
```diff
@@ -0,0 +1,72 @@
...
+		},
+		"files" : [
+			"vendor/symfony/symfony/src/Symfony/Component/VarDumper/Resources/functions/dump.php"
+		]
+	},
+	"require" : {
+		"php" : ">=5.5.9",
+		"doctrine/doctrine-bundle" : "^1.6",
+		"doctrine/orm" : "^2.5",
+		"incenteev/composer-parameter-handler" : "^2.0",
```
nicolas-grekas: not needed now that 3.2 enabled env vars

frameworks/PHP/symfony/composer.json
```diff
@@ -0,0 +1,72 @@
...
+		"sensio/distribution-bundle" : "^5.0.19",
+		"sensio/framework-extra-bundle" : "^3.0.2",
+		"symfony/monolog-bundle" : "^3.1.0",
+		"symfony/polyfill-apcu" : "^1.0",
```
nicolas-grekas: really needed?

raziel057:
For info this PR can be merged. Tests releated to symfony are correct:

Verification Summary

Test | symfony 
--------|----------
fortune | PASS |
plaintext | WARN |
db  | PASS |
update | PASS |
json | PASS |
query | PASS |

Test | symfony-raw
--------|---------------
query | PASS
db | PASS
update | PASS

frameworks/PHP/symfony/composer.json
fabpot: We should really use Flex instead.

frameworks/PHP/symfony/app/Resources/views/base.html.twig
raziel057: Yes but it's not conform to the requirements: 
https://www.techempower.com/benchmarks/#section=code&hw=ph&test=fortune

raziel057:
For information, I just tried to migrate from Symfony 3.4 to Symfony Flex to make some benchmarks (with JMeter) to measure the performance gain. You can find here a comparison between SF3.4.0 vs SF Flex 4.0.0:

plaintext (50 users - 10000 iteration)
-----------------------
Symfony 3.4.0

plaintext-sf3 4

Symfony Flex / 4.0.0

plaintext-sf4 0

json (50 users - 10000 iteration)
-----------------------
Symfony 3.4.0

json-sf3 4

Symfony Flex / 4.0.0

json-sf4 0

fortunes (50 users - 5000 iteration)
-----------------------
Symfony 3.4.0

fortunes-sf3 4

Symfony Flex / 4.0.0

fortunes-sf4 0

Issue #2960, opened 2017-09-06
Amber, Kemal, Kemalyst results in question


drujensen:
Hi Nate,

We are baffled by the results for the 3 crystal projects and hoping you can help shed some light on the 100k RPS results. We were expecting 1.8 to 2.0 million RPS range for these projects.

We have done multiple runs on different sized systems in AWS trying to replicate the slow response times for plaintext and json in preview 2.  @sdogruyol has also ran the benchmarks and can confirm similar findings.

I'm attaching the results on a c4.xlarge 8 core system and all three projects are showing 3x the results than the preview 2.  We have also ran the tests on a 64 core 1.2ghz system with 10x the requests per second (~1 million RPS) than preview 2.

We followed the instructions provided for setting up the linux environments Ubuntu 14.04 and cannot determine what the difference is that would cause the slowness.

We think that the preview 2 is not correctly reflecting the performance for these 3 projects.  It's as if the multiple processes or the reuse port was not properly working.  What we find interesting is the `crystal-raw` did not encounter the same issue and had 2.5m RPS which is in line with our expectations.

We are aware of socket errors when there is heavy load, be we are not seeing this slow down the performance reflected in preview 2.

Any help is solving this mystery is appreciated.  If possible, can you rerun the plaintext for the 3 projects and let us know if you are still seeing 100k RPS results?

Thanks,
Dru

[results.zip](https://github.com/TechEmpower/FrameworkBenchmarks/files/1280006/results.zip)

sdogruyol:
Thanks for this @drujensen :+1: I've also done some small runs and can second this issue, unfortunately I don't have any logs :cry:

nbrady-techempower:
Hi @drujensen! We are investigating some issues now with r15-p2 and are doing some reruns. I'll let you know as soon as I see some results and we will go from there.

nbrady-techempower:
@drujensen @sdogruyol Looks like we found the issue. We've started a new run and amber has completed but it will be some time before we get a full picture for R15-P3. Here are the results:

```
"plaintext": {
      "amber": [
        {
          "latencyAvg": "23.70ms", 
          "latencyMax": "114.14ms", 
          "latencyStdev": "12.06ms", 
          "totalRequests": 1448302, 
          "startTime": 1504802515, 
          "endTime": 1504802530
        }, 
        {
          "latencyAvg": "87.41ms", 
          "latencyMax": "404.84ms", 
[...]

PR #3001, opened 2017-10-20
Setup GC threads on Crystal tests

faustinoaq:
Hi @nbrady-techempower We're having some issues on Kemal & Amber frameworks https://github.com/amberframework/amber/issues/273 & https://github.com/TechEmpower/FrameworkBenchmarks/issues/2960

Latest Crystal release doesn't support parallelism yet, so we are launching processes and using `reuse_port`. However current GC ([bdwgc](https://github.com/ivmai/bdwgc)) is multithreaded.

This PR tries a new GC setting to avoid launching multiples threads across cores competing for CPU time.

Also add some small optimizations (compress html and remove extra variable) and update some dependencies

/cc @drujensen


frameworks/Crystal/amber/setup.sh
```diff
@@ -6,6 +6,10 @@ shards install
 
 crystal build --release --no-debug src/amber.cr
 
+export GC_MARKERS=1
+
+export GC_NPROCS=1
```
RX14: > Note that setting this to 1
                when multiple processors are available will preserve
                correctness, but may lead to really horrible performance,
                since the lock implementation will immediately yield without
first spinning.

> It is safer to adjust GC_MARKERS than GC_NPROCS,
since GC_MARKERS has no impact on the lock implementation.

frameworks/Crystal/amber/src/views/fortune/index.ecr
```diff
@@ -1,12 +1 @@
...
-  </tr>
-  <% fortunes.each do |fortune| %>
-    <tr>
-    <td><%= fortune.id %></td>
-    <td><%= HTML.escape(fortune.message || "") %></td>
-    </tr>
-  <% end %>
-</table>
\ No newline at end of file
+<table><tr><th>id</th><th>message</th></tr><% fortunes.each do |fortune| %><tr><td><%= fortune.id %></td><td><%= HTML.escape(fortune.message || "") %></td></tr><% end %></table>
```
drujensen: @faustinoaq this doesn't seem like a realistic template.  I would rather keep the spaces so if someone comes and looks at this Amber project, they can see what is going on.  

Do you anticipate a significant increase in performance by removing the spaces?

faustinoaq: Oh, I see let me revert this changes

faustinoaq: Done!

faustinoaq:
@drujensen Finally worked! :) I think we're ready to test performance

nbrady-techempower:
Hey guys, I'm going to leave this open until our new Server Central environment is ready to go. So feel free to make additional changes and as soon as we get that all set up, I'll merge in and we'll spit out another preview round. Make as many commits as you'd like; I always squash and merge.

drujensen:
@nbrady-techempower It looks like this is ready to test.  Really interested to see if this makes a difference.  Thanks for your help.

nbrady-techempower:
@drujensen It's on top of my list to get these numbers to you guys. We're waiting on the vendor to finish moving our machines to another server. It's out of our hands at the moment, but as soon as it's all set and we're able to start running tests again you will be among the first to know!

eliasjpr:
@nbrady-techempower keep us in the loop we are interested in these results :)

nbrady-techempower:
Hey guys, we've got our environment set up. I just kicked off a new run with this PR merged in. I should have some results to share for you by Monday.

PR #3002, opened 2017-10-22
Add Japronto

gsakkis:
Submitting [Japronto](https://github.com/squeaky-pl/japronto) to the framework benchmarks; just the plaintext and json tests are implemented. Would love to see this tested for round 15!


nbrady-techempower:
Hey @gsakkis Thanks for the contribution! This should make round 15, as we'll be doing additional preview rounds soon

PR #3006, opened 2017-10-24
Use dynamic server_host in header template

Shad0wPie:
No reason as far as I can see to have the hard coded TFB-server here when we can use the same server host that is used to build the url.

When running this with Pivotal Cloud Foundry, I hit an issue where traffic from the client to the server was being rejected due to the 'Host: TFB-server' entry in the header. This proposed change fixed the problem.

### DIFF
--- a/toolset/benchmark/framework_test.py
+++ b/toolset/benchmark/framework_test.py
@@ -29,7 +29,7 @@ from datetime import datetime
 from datetime import timedelta
 
 class FrameworkTest:
-  headers_template = "-H 'Host: TFB-server' -H 'Accept: {accept}' -H 'Connection: keep-alive'"
+  headers_template = "-H 'Host: {server_host}' -H 'Accept: {accept}' -H 'Connection: keep-alive'"
 
   # Used for test types that require no pipelining or query string params.
   concurrency_template = """
@@ -644,7 +644,7 @@ class FrameworkTest:
   # and DB)
   ############################################################
   def __generate_concurrency_script(self, url, port, accept_header, wrk_command="wrk"):
-    headers = self.headers_template.format(accept=accept_header)
+    headers = self.headers_template.format(server_host=self.benchmarker.server_host, accept=accept_header)
     return self.concurrency_template.format(max_concurrency=max(self.benchmarker.concurrency_levels),
       name=self.name, duration=self.benchmarker.duration,
       levels=" ".join("{}".format(item) for item in self.benchmarker.concurrency_levels),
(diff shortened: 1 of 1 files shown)


nbrady-techempower:
Do you have entries for TFB-server, TFB-client, TFB-database in your hosts file? We'd probably be going the other direction with this for consistency. It shouldn't cause any issues.

http://frameworkbenchmarks.readthedocs.io/en/latest/Development/Installation-Guide/

Shad0wPie:
Hi and thanks for the response!

Yes I tried that, and the installation instructions worked great for a run directly on aws. However due to the nature of Pivotal Cloud Foundry the requests are intercepted and blocked before they ever hit the server (due to the intercepting instance not being able to resolve the header I assume). The workaround that I have been using is to set:
server_host=app.example.com
in the benchmark.cfg file.

I completely understand that you want to TFB-server to be the 99% use case, but as long as server_host=TFB-server the behaviour doesn't change with my change. However it does fix the odd use case where someone needs to set server_host to something else. Why have server_host at all in the config file if we are not allowed to change it? :)

nbrady-techempower:
@Shad0wPie Yeah, I understand. `server_host` is in the config file for backwards compatibility as we were making the change. To be honest, you found one of the places that should have already been changed, allowing us to remove `server_host` from the config altogether.

Shad0wPie:
I see. Do you have any suggestion for how I can resolve my usecase then? To reiterate, the problem is that the requests are blocked by an intermediary where I cannot modify the hosts file.

nbrady-techempower:
@Shad0wPie Maybe I need a little more information because I'm not sure I understand. Do you not control the server that is making the outgoing `wrk` request? Because `TFB-server` would be translated there before it hit some intermediary. Or is it that you're able to install on this intermediary that is acting as the client making the `wrk` request? In that case, i'm not sure we want to make changes for that specific use case.

Shad0wPie:
@nbrady-techempower Absolutely, I'll try to clarify. I have a Pivotal Cloud Foundry deployment, which I do not completely control. I then create 2 aws vms that i completely control, the load client and the database. Finally I deploy the server as a container inside PCF. The server is able to connect to both the database and the client just fine, and install any necessary software. 

The problem is that the client and database instance don't have direct access to the server, the requests need to go through the url app.example.com, where they hit a load balancer, then PCF which routes the request to the correct container. Somewhere in the process of this routing the requests get blocked if the header contains "TFB-server", but they don't if the header is changed to app.example.com. I assume that this is because this intermediary step in PCF can't resolve TFB-server.

Now, do you mean that the "Host: TFB-server" header will be translated by the load client before leaving the instance? Are you 100% sure about that? Because the way I understand it, this header is just sent as a string as is later interpreted on the receiving side. wrk will of course translate TFB-server in the request url (e.g. http://TFB-server:8080/json -> http://app.example.com:8080/json), but I think it leaves the header unmodified. 

I understand that this is an uncommon use case but I don't think it's completely invalid. More and more web apps are being deployed in these containerised environments and I found your tool to be a great way to compare web app performance in different environments :)

nbrady-techempower:
@Shad0wPie Thanks for clarifying your case. Yes, I forgot for a moment that we were talking about headers. This seems like a reasonable solution for the moment. Thanks for being so detailed!

PR #3022, opened 2017-10-29
Update Wildfly to 11.0.0.Final

codylerum:


nbrady-techempower:
Failures probably related to #3040 so I'll get this merged in. Thanks!

PR #3025, opened 2017-10-30
update jooby version to 1.2.1

vkirilov:


nbrady-techempower:
Failures probably related to #3040 merging now. thanks!

PR #3036, opened 2017-11-02
Upgrade play2-java-jpa-hikaricp to Play 2.5

benmccann:


benmccann:
>      Caused Exception in TFB
                 This almost certainly means your return value is incorrect,
                 but also that you have found a bug. Please submit an issue
                 including this message: HTTPConnectionPool(host='TFB-server', port=9000): Read timed out. (read timeout=15)

PR #3037, opened 2017-11-03
the default cache size too low for php frameworks

omarsoufiane:
the default php.ini config is not suitable for php fameworks (symfony/laravel) that open many PHP files, especially on Windows systems.


Tobion:
opcache.interned_strings_buffer and opcache.memory_consumption should also be increased

PR #3038, opened 2017-11-03
Update to final release

pmlopes:
The current release is 3.5.0


nbrady-techempower:
Looks like the failures came from maven which i fixed in #3040 so this is good

PR #3049, opened 2017-11-08
upgrade http4k and fix to plaintext

daviddenton:
If we could get this in for R15Preview3 then that's be great, because there was a bug with the plaintext implementation.

thanks :)


daviddenton:
myself and @michaelhixson both made the same plaintext fix (thanks Michael!). I've resolved the conflict and the new version of http4k should fix the undertow implementation.

thanks :)

PR #3059, opened 2017-11-09
Change ports to 8080 for amber and kemal

marksiemers:
Attempt to address #2960

### DIFF
--- a/frameworks/Crystal/amber/benchmark_config.json
+++ b/frameworks/Crystal/amber/benchmark_config.json
@@ -9,7 +9,7 @@
       "fortune_url": "/fortunes",
       "update_url": "/updates?queries=",
       "plaintext_url": "/plaintext",
-      "port": 3000,
+      "port": 8080,
       "approach": "Realistic",
       "classification": "Fullstack",
       "database": "Postgres",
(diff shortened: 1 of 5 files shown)


marksiemers:
Attempt to address #2960

marksiemers:
@sdogruyol @faustinoaq @drujensen

faustinoaq:
> amber is not passing in that environment and the logs didn't provide an immediate clue as to why. [0]

@marksiemers Maybe we can fix versions on [amber shard.yml](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Crystal/amber/shard.yml#L9) like [Kemal shard.yml](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Crystal/kemal/shard.yml) and [Crystal shard.yml](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Crystal/crystal/shard.yml#L4) are doing.

Maybe it could solve the issue @nbrady-techempower said about amber isn't passing test.

[0]. https://github.com/TechEmpower/FrameworkBenchmarks/issues/2960#issuecomment-343308162

PR #3062, opened 2017-11-12
Add play2-java-jooq-hikaricp

mkurz:
Adds a new benchmark project: play2-java with [jOOQ](https://www.jooq.org/). Will be interesting how its results will look like compared to the jpa and ebean ones.

@lukaseder you might be interested as well :wink:


lukaseder:
OK, I was pinged, so here are a few comments, directly in the code...

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    public CompletionStage<Result> queries(final String queries) {
+        return getRandomWorlds(queryCount(queries)).thenApply(worlds -> ok(worlds.formatJSON(JSON_FORMAT)).as(JSON));
+    }
+
+    public CompletionStage<Result> fortunes() {
+        return CompletableFuture.supplyAsync(() -> {
+            final List<FortuneRecord> fortunes = this.db.withConnection(connection -> {
+                return DSL.using(connection, DIALECT).selectFrom(FORTUNE).fetch();
```
lukaseder: There's some initial setup to be done inside of the `DSL.using()` call. This doesn't matter much in real world situations, but in a benchmark, it certainly biases the result. It would be more fair to store the `DSLContext` reference along with the `Database` reference.

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
```diff
@@ -0,0 +1,121 @@
...
+            });
+            fortunes.add(new FortuneRecord(UInteger.valueOf(0), "Additional fortune added at request time."));
+            Collections.sort(fortunes, (f1, f2) -> f1.getMessage().compareTo(f2.getMessage()));
```
lukaseder: Why is this being done in the client, and not in the database? I mean, this screams "ORDER BY" all over it :)

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
lukaseder: Oh, and I don't know what the benchmark is supposed to do, but here, you're doing a `SELECT *`. Is that really the intention?

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
```diff
@@ -0,0 +1,121 @@
...
+            return ok(views.html.fortunes.render(fortunes));
+        }, dbEc);
+    }
+
+    public CompletionStage<Result> update(final String queries) {
+        return getRandomWorlds(queryCount(queries)).thenApplyAsync(worlds -> {
+            final Random random = ThreadLocalRandom.current();
+            for (final WorldRecord world : worlds) {
+                world.setRandomnumber((random.nextInt(10000) + 1));
+            }
```
lukaseder: Again, what is being benchmarked here? The database access layer, or some random number generator? All of this stuff can lead to bias in a benchmark result - not sure if that's relevant, of course...

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
```diff
@@ -0,0 +1,121 @@
...
+
+            final int batchSize = 25;
+            final int batches = ((worlds.size() / batchSize) + 1);
+            this.db.withConnection(connection -> {
+                final DSLContext sql = DSL.using(connection, DIALECT);
```
lukaseder: Again, better store this as a member, rather than initialising it every time.

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
```diff
@@ -0,0 +1,121 @@
...
+                for ( int i = 0 ; i < batches ; ++i ) {
+                    sql.batchUpdate(worlds.subList(i * batchSize, Math.min((i + 1) * batchSize, worlds.size()))).execute();
```
lukaseder: Why are jOOQ records chunked like this in the batch call? This isn't being done in neither the JPA nor the Ebean versions.

Granted, this call may suffer from a known issue: https://github.com/jOOQ/jOOQ/issues/5962

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        return queryCount;
+    }
+
+    private CompletionStage<org.jooq.Result<WorldRecord>> getRandomWorlds(final int n) {
+        return CompletableFuture.supplyAsync(() -> {
+            final Random random = ThreadLocalRandom.current();
+            org.jooq.Result<WorldRecord> worlds = null;
+            for (int i = 0; i < n; ++i) {
```
lukaseder: If `n` is large, as a SQL person, I might be inclined to create a temp table, insert all the random values in that, and then semi-join that table to the `WORLD` table to get all the rows in one go. Given how this algorithm is implemented here, this is probably a more reasonable approach.

Alternatively, if this test is run against H2, HSQLDB, Oracle, PostgreSQL, or SQL Server, an array parameter or table-valued parameter might be suitable as well for use in an `IN` predicate.

I think the jOOQ approach would favour something like this.

mkurz:
@lukaseder Did you submit your comments? I can't see them.

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
```diff
@@ -0,0 +1,121 @@
...
+                long randomId = random.nextInt(10000) + 1;
+                final org.jooq.Result<WorldRecord> world = this.db.withConnection(connection -> {
+                    return DSL.using(connection, DIALECT).selectFrom(WORLD).where(WORLD.ID.eq(UInteger.valueOf(randomId))).fetch();
```
lukaseder: This is again something specific to benchmarks, and not the real world, but if you run thousands of individual selects like this, you could save some time by generating the SQL string and bind variables only once, storing that in some local variable, and then executing it several times.

JPA implementations can easily cache their trivial CRUD queries, because they are trivial ("fetch by ID"). It would be much more interesting to benchmark more complex dynamic SQL (e.g. jOOQ vs. Criteria API)...

lukaseder:
@mkurz 

image

;-)

mkurz:
@lukaseder Please check out the "Test requirements": https://www.techempower.com/benchmarks/#section=code

This should answer some of your questions.

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
mkurz: This is a test requirement:
> The list of Fortune objects must be sorted by the order of the message field. No ORDER BY clause is permitted in the database query (ordering within the query would be of negligible value anyway since a newly instantiated Fortune is added to the list prior to sorting).

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
mkurz: This is a test requirement.
Also, that is the same code like for the jpa and ebean tests, so it's fair.

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
mkurz: Also in the jpa and ebean version there are batch updates of size 25.

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
mkurz: `IN` is not allowed, please read the test requirements.

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
mkurz: Test requirement says there needs to be a database round trip for each fetch.
Do you think there is a major impact when creating the jOOQ query each time? Can that one be cached on client side?

lukaseder:
Yeah, definitely does. Although I might not find the time to fully validate the approach right now. Most of these benchmarks always cover the super trivial stuff in very trivial ways. I'm not so sure what the result really means to someone planning to write a large application, for instance.

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
lukaseder: That's funky. If done correctly, consistent ordering can profit from merge sorts, which can be done in `O(N)` in ideal cases, e.g. this benchmark case. Not permitting that is kind of an artificial limitation in my opinion.

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
lukaseder: What's the reason for this?

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
lukaseder: Yeah, I've seen it now, but *why not*? If a benchmark coerces a framework into non-idiomatic usage, then the benchmark is pretty meaningless, in my opinion.

mkurz:
The benchmark should be [relastic](https://github.com/TechEmpower/FrameworkBenchmarks/blob/5bc429ea9291fc99f3c2cb566f09753803f34e66/frameworks/Java/play2-java/benchmark_config.json#L78). However I may cache the `DSLContext` and get rid of the `SELECT *`.

Only one thing I thought about: Do you know how expensive it is to add an element to an ArrayList - when the arraylist is full and therefore its internal array needs to be copied and resized? Because as far as I could find out this is what a jOOQ `Result` has to do each time it adds new record (via some listener). Maybe that is something to optimize, by setting the ResultImpl's internal ArraysList initialcapacity to the count of records being added/fetched or use something more efficiant the arraylist? But again, I don't know how efficient Arraylist already is.

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
lukaseder: Again, there isn't a major impact in a *real world* application, but there can be a significant impact in a *benchmark*.

Nevertheless, in this particular comment, I didn't advocate for preventing the database round trip per fetch, but the SQL generation per fetch. You can easily do:

```java
ResultQuery<?> q1 = ...
String sql = q1.getSQL();
List<Object> binds = q1.getBindValues();

ResultQuery q2 = DSL.using(...).resultQuery(sql, binds.toArray());

// and now loop and reuse q2
```

Again, that's not something that is advisable for every day business logic, where the improvement should be non-negligible, but this is a *benchmark*. It measures *mainly* this overhead many many times.

mkurz:
> Most of these benchmarks always cover the super trivial stuff in very trivial ways. I'm not so sure what the result really means to someone planning to write a large application, for instance.

Sure. However I tried to keep the code between the ebean, jpa and jOOQ benchmark as similar as possible. So that might give you a hint of a "base" performance of an ORM. E.g. ebean is always far behind jpa and jOOQ on my machine, and jOOQ is a bit behind jpa. Of course that alone shouldn't influence someone on which ORM to choose, there are many other factors.

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
mkurz: You should talk to the @TechEmpower people to discuss the *why* :smile:

lukaseder:
What does "realistic" mean?

> Only one thing I thought about: Do you know how expensive it is to add an element to an ArrayList - when the arraylist is full and therefore its internal array needs to be copied and resized?

In a benchmark, that sort of stuff usually bubbles to the top. The loops are the most expensive. And in the case of jOOQ, the internal `StringBuilder` will create the most CPU-bound bucket, as its internal `char[]` or `byte[]` (depending on the JDK version) needs to be copied around and resized all the time.

> Because as far as I could find out this is what a jOOQ Result has to do each time it adds new record (via some listener).

If that matters to you, you can always use `ResultQuery.fetchLazy()` and then operate on the `Cursor` lazily (i.e. by keeping an open JDBC `ResultSet`). Perhaps that improves things? Can't judge right now.

> Maybe that is something to optimize, by setting the ResultImpl's internal ArraysList initialcapacity to the count of records being added/fetched or use something more efficiant the arraylist?

That count would be super expensive inside of the database in most real world cases where you don't actually want to calculate such a count. I think we can live with arraylist resizing in this case, but I'm happy to review concrete benchmark results / comparisons that show that such an improvement might be worthwhile.

> Sure. However I tried to keep the code between the ebean, jpa and jOOQ benchmark as similar as possible. 

Yes yes, I realized I'm not criticising your port to jOOQ, but the original approach here. Take the TPC benchmarks for instance: http://www.tpc.org/information/benchmarks.asp. They're really much much more sophisticated, and their results are much more useful. Of course, they're more work too, but why do any work at all, if it's not going to be very meaningful? :)

> So that might give you a hint of a "base" performance of an ORM

Seriously, those "base" performances have contributed to so much damage in our industries, in the past. Because people quickly forget about the "base", and they don't even validate things...

Nonetheless, they help vendors tune their ORMs to look better in "base" performance benchmarks, so still helpful for the vendors ;-)

lukaseder:
Btw: The most important caveats right now are already on the roadmap including new SPIs will allow for bypassing the intermediate `Record` types: https://github.com/jOOQ/jOOQ/issues/6739. This will be useful when it is known that the result will be serialised to JSON. It can be done directly, without storing stuff in a `Record` (internal `Object[]`), a `Result` (internal `ArrayList`)

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
lukaseder: You're right of course!

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
mkurz: No reason, do you think it's better to just put everything into a single batch?

mkurz:
@lukaseder Thanks for all your information. I will create a new pr to address your feedback as soon as I find time.
I am fully aware that these benchmarks might not be ideal. Look at it the other way round: Maybe someone will learn about jOOQ *because* of them :stuck_out_tongue_closed_eyes: :wink: Lukas, I have switched to jOOQ already anyway, we switched over most of our projects from hibernate (hell) the last few month. I may get you a paying SQLServer-user customer in the near future as well :smile:. Also I am not tired to advertise jOOQ to every dev I meet, just recently I gave a talk about Play in Czech Republic expressly mentioning jOOQ in favor of JPA.

frameworks/Java/play2-java/play2-java-jooq-hikaricp/app/controllers/Application.java
lukaseder: I don't know...

lukaseder:
> Maybe someone will learn about jOOQ because of them

Fair enough :)

Thanks again for your work and for this and for your many other contributions!

lukaseder:
@mkurz, I came across these benchmarks again, I think there are some things that would be worth addressing. Can you please review if you have time? https://github.com/TechEmpower/FrameworkBenchmarks/discussions/6673

mkurz:
@lukaseder I put it on my TODO list. The Play benchmarks [need some love anyway](https://github.com/TechEmpower/FrameworkBenchmarks/pull/5486#issuecomment-616778706), however I am not sure when I find time...

PR #3070, opened 2017-11-14
Create the requried directories for saving php session files

michaelhixson:
By setting `session.save_path` to "3;/tmp" in php.ini, we were
effectively breaking session files for all PHP frameworks, because we
were never creating the required subdirectories inside /tmp to hold
those session files.  For whatever reason, PHP doesn't create those
directories on demand, but it does provide a script to create them.
We're now invoking that script.

Since /tmp is cleaned out each time a framework is tested, we must
recreate these directories whenever a framework claims to depend on PHP,
regardless of whether PHP was installed already.

This should allow PHP to create session files once again, which should
prevent frameworks such as phreeze from printing a huge amount of error
messages to the logs.  Theoretically, no PHP framework should need to use
sessions in order to satisfy our test requirements, so they could be more
efficient by avoiding sessions, but that's a problem for people
optimizing those frameworks to solve.  The toolset should provide
non-broken session files to frameworks that want them.

I also added the "when already instealled, copy php.ini and php-fpm.conf"
step from our PHP5 install script to the PHP7 install script.  I don't
actually know why the PHP5 install script does this, but it seems fishy
to do this for one PHP version and not the other.

### DIFF
--- a/toolset/setup/linux/languages/php/php7.sh
+++ b/toolset/setup/linux/languages/php/php7.sh
@@ -1,10 +1,19 @@
 #!/bin/bash
 
-fw_installed php7 && return 0
+RETCODE=$(fw_exists ${IROOT}/php7.installed)
+[ ! "$RETCODE" == 0 ] || { \
+  echo "Moving PHP config files into place";
+  sudo cp $FWROOT/toolset/setup/linux/languages/php/php.ini /usr/local/lib/php.ini
+  sudo cp $FWROOT/toolset/setup/linux/languages/php/php-fpm.conf /usr/local/lib/php-fpm.conf
+  rm -rf /tmp/php_sessions
+  /bin/bash $IROOT/php7/ext/session/mod_files.sh /tmp/php_sessions 3 5
+  source $IROOT/php7.installed
+  return 0; }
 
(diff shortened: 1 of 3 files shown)


toolset/setup/linux/languages/php/php7.sh
```diff
@@ -1,10 +1,19 @@
...
-fw_installed php7 && return 0
+RETCODE=$(fw_exists ${IROOT}/php7.installed)
+[ ! "$RETCODE" == 0 ] || { \
+  echo "Moving PHP config files into place";
+  sudo cp $FWROOT/toolset/setup/linux/languages/php/php.ini /usr/local/lib/php.ini
+  sudo cp $FWROOT/toolset/setup/linux/languages/php/php-fpm.conf /usr/local/lib/php-fpm.conf
+  rm -rf /tmp/php_sessions
+  /bin/bash $IROOT/php7/ext/session/mod_files.sh /tmp/php_sessions 3 5
+  source $IROOT/php7.installed
+  return 0; }
```
msmith-techempower: This right here is a good catch in and of itself... I'm surprised this worked before.

PR #3085, opened 2017-11-21
updating the jawn framework

MTDdk:


zapov:
It would be nice if you also fixed the update test and did the actual DB update ;)

MTDdk:
Wow! That was quite the mistake

Thanks for noticing!

Should be fixed now

nbrady-techempower:
Thanks @zapov That was on our list of things to look into on Monday, as it was leading the pack there by a lot. Appreciate you getting that fix in @MTDdk

PR #3092, opened 2017-11-24
Fix amber dependencies

faustinoaq:
Hi @nbrady-techempower I found amber is failing on round 15 preview 3 because dependency mismatch, this PR fixes amber's dependencies.

failing

Also we have had multiple discussions about Amber and Kemal performance vs `crystal-raw` and some of us concluded crystal's frameworks are slow not because radix algorithm but because they're using proc in some critical places inside the core slowing down high RPS tests like plaintext and json.

We added `Amber::Pipe::Last` to replace current proc in amber framework core, so we hope to see improvement on plaintext and json tests 😅 

I'm not surprised about slow result in crystal database tests, we're still very young and we need a lot of performance improves in database drivers, so we can accept those results for now 😉 

BTW, Thank @nbrady-techempower and @msmith-techempower for all your nice support!

You, friends, helped us a lot 👍

### DIFF
--- a/frameworks/Crystal/amber/benchmark_config.json
+++ b/frameworks/Crystal/amber/benchmark_config.json
@@ -21,7 +21,7 @@
       "webserver": "None",
       "os": "Linux",
       "database_os": "Linux",
-      "display_name": "Amber (MVC, Postgres)",
+      "display_name": "Amber (MVC, PSQL)",
       "notes": "",
       "versus": "kemal"
     }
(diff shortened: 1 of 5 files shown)


faustinoaq:
Passing https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/306637851#L1743 🎉

faustinoaq:
Also I fixed dependencies to a specify version to avoid it happen again.

We didn't specify version before, so `shard.lock` became outdated and then `shards install` failed.

nbrady-techempower:
Great news guys!

PR #3093, opened 2017-11-24
Updated Wt to 4.0.2

RockinRoel:
This updates the Wt framework test to the latest version (4.0.2), released today.

Wt now includes the framework benchmarks as one of its examples, so this will build the implementation that's part of the Wt 4.0.2 source tree. If this is unacceptable, let me know. I can put the example C++ code back in, and do it the old way, compiling the example manually and separate from Wt.

Also, I made it so Wt builds with the latest GCC version available on the test systems, which appears to be version 6, and I made it so that boost, instead of being installed, is built with only the libraries that Wt needs, so it should still be fast enough for Travis, so it doesn't kill the build.

I hope this also fixes the strange performance issues we're seeing Wt having on the test hardware (on our own hardware, Wt does a lot better).

### DIFF
--- a/frameworks/C++/wt/benchmark.cpp
+++ b/frameworks/C++/wt/benchmark.cpp
@@ -1,28 +1,30 @@
 #include <signal.h>
 #include <cstdlib>
+#include <cstring>
 #include <string>
 #include <vector>
 #include <algorithm>
-#include <mutex>
 
-#include <Wt/WServer>
-#include <Wt/WResource>
-#include <Wt/Http/Request>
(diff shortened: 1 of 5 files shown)


nbrady-techempower:
@RockinRoel This sounds like great news and will hopefully fix those performance issues. We will need the tfb implementation in this repo though. Sorry for the trouble.

emweb:
No problem. I've updated this PR so it includes the source code for the examples again.

RockinRoel:
No problem. I've updated this PR so it includes the source code for the examples again.

PR #3094, opened 2017-11-25
Add crystal-radix

marksiemers:


marksiemers:
@drujensen @faustinoaq @elorest @eliasjpr

faustinoaq:
My expectation for this test is very good, I don't think radix would be the high overload here but other things like extra procs or something like that, by example Amber is using [Amber::Pipe::Last](https://github.com/amberframework/amber/blob/639b60b709d476be9b12c378a6e9ea2db557e1e0/src/amber/router/pipe/pipeline.cr#L42) now

PR #3119, opened 2017-12-01
Don't restrict reitit HTTP requests to localhost

michaelhixson:
Without this, the reitit tests fail in the multi-machine benchmark setup,
since the wrk client runs on a different machine than the application
server.


ikitommi:
Thanks!

PR #3128, opened 2017-12-07
Upgrade to Resin 4.0.55, which is compatible with Java 9

michaelhixson:
Of the 11 frameworks that use Resin, this change makes 6 of them
compatible with Java 9.  The remaining 5 are incompatible with Java 9 for
other reasons, which we'll try to address.

I removed the line that modifies the contents of JAVA_HOME because:

 a) There is no jre directory in Java 9, so this line fails.
 b) It's not clear what this line is supposed to accomplish in the first
    place.  I've never had to do this for my own Java + Resin
    installations.

### DIFF
--- a/toolset/setup/linux/webservers/resin/resin.sh
+++ b/toolset/setup/linux/webservers/resin/resin.sh
@@ -4,11 +4,9 @@ fw_depends java
 
 fw_installed resin && return 0
 
-RVER=4.0.53
+RVER=4.0.55
 RESIN=resin-$RVER
 RESIN_HOME=$IROOT/$RESIN
-
-sudo cp -r $JAVA_HOME/include $JAVA_HOME/jre/bin/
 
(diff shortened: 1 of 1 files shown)


toolset/setup/linux/webservers/resin/resin.sh
```diff
@@ -4,12 +4,10 @@ fw_depends java
 
 fw_installed resin && return 0
 
-RVER=4.0.53
+RVER=4.0.55
 RESIN=resin-$RVER
 RESIN_HOME=$IROOT/$RESIN
 
-sudo cp -r $JAVA_HOME/include $JAVA_HOME/jre/bin/
```
msmith-techempower: I **believe** some frameworks (maybe legacy support) relied on stuff in `include` to be on the path, but that can be added (if actually needed) to the path setup of `Java` later.

PR #3143, opened 2017-12-11
Added bechmarks for the play2 netty backend

mkurz:
Also optimzed the mysql flags for all the play2 tests and configured the threadpools to fit the CPU count of the application and database server.


mkurz:
@TechEmpower This can be merged, thanks!

PR #3152, opened 2017-12-20
upgrade snap to lts-10.0

flip111:


flip111:
build fails on php/laravel is it posslbe to rerun somehow?

msmith-techempower:
This has nothing to do with Laravel - we will ignore that failing (though, I'm not even sure why it ran to begin with).

nbrady-techempower:
This PR reverts the changes I just made that tied the dependencies down to specific version numbers. Snap was the only Haskell framework I had working. Updating the resolver is great but lets keep the version numbers.

flip111:
@nbrady-techempower tying versions down with cabal was the old way of doing things. The majority of people started to use stack a while ago, which automatically pins down the package version numbers. You can look up the version numbers in the following way:

1. Go to the webpage of the right resolver version, in this case https://www.stackage.org/lts-10.0
2. type the package name you want to know the version of, example: aeson
3. in the search results click on "package aeson"

You will now end up on the specific page for the aeson version used in lts-10.0 https://www.stackage.org/lts-10.0/package/aeson-1.2.3.0

so the version for this package is 1.2.3.0

Can you try to clone my patch to see if it works on your end too? I would be very surprised if it doesn't since it's ok on my end and by travis too

msmith-techempower:
@nbrady-techempower I remember @ssmith-techempower telling me about stack a while ago, and I believe the idea is to have a sort of 'build version' that is composed by all of the dependency versions. Cabal has indeed been somewhat sidelined by stack, and I believe this approach is appropriate.

flip111:
@nbrady-techempower forgot to add that specifying version in cabal files is not specifying all the dependencies there are because each dependency has it's own dependencies again. If your goal is to know which versions are used you can use the command `stack list-dependencies`

nbrady-techempower:
@flip111 Yes, I understand that. That's true with most dependency managers. See npm for instance. Locking down these top level dependencies still gives you tighter control and decreases the likelihood that something will break. The goal is, if the framework builds and runs once, it should build and run every time.

msmith-techempower:
> The goal is, if the framework builds and runs once, it should build and run every time.

That and providing an at-a-glance reference for the versions being benchmarked.

flip111:
> use the latest version of each that works with each other

This is not how stack lts releases work. Let's say an LTS has been released and you build your project. Now you come back a year later and it's gonna look for "the latest version of each that works with each other" ? No. It's going to use those same versions that were proven to work together one year ago FOR THAT SPECIFIC LTS.

> We would prefer if you locked down the versions in the source

stack users rely on stack to lock down versions

> this means that we don't necessarily know which version was run with a given benchmark

Yes you can run `stack list-dependencies`.
It will show you the same set of packages for the same lts release 

> The goal is, if the framework builds and runs once, it should build and run every time.

It's guaranteed by stack

> providing an at-a-glance reference for the versions being benchmarked.

`stack list-dependencies`

You can find information about this in the stack guide https://docs.haskellstack.org/en/stable/GUIDE

> What makes stack special?
The primary stack design point is reproducible builds. If you run stack build today, you should get the same result running stack build tomorrow. There are some cases that can break that rule (changes in your operating system configuration, for example), but, overall, stack follows this design philosophy closely. To make this a simple process, stack uses curated package sets called snapshots.

> The final field is resolver. This tells stack how to build your package: which GHC version to use, versions of package dependencies, and so on.

If you want to specify specific versions (which i don't want ... but in case you insist)

> Sometimes, you may want to use specific versions of certain packages for your project. To do that you can fix those versions by specifying them in the extra-deps section and then use stack solver to figure out whether it is feasible to use those or what other dependencies are needed as a result.

Sorry to be so strong about this, but i'm not gonna make haskell PR's on this repo if i need to start looking up dependencies manually, this is just dedication to helping haskell frameworks along in these benchmarks without any further profit motive.

nbrady-techempower:
@flip111 If what you're saying is accurate, can you explain what caused these Haskell frameworks to break suddenly?

msmith-techempower:
Based on the gory details about LTS releases above, I approve as-is.

flip111:
> If what you're saying is accurate, can you explain what caused these Haskell frameworks to break suddenly?

I was also very worried about this, but i think it's the same as this issue https://github.com/commercialhaskell/stack/issues/3631 which is very recent and the haskell community is already on it ^___^

> Also, nothing wrong with having a strong opinion here. We are not experts in all languages/frameworks, so we appreciate it when people closer to these languages/frameworks help explain some of these things.

Thanks for understanding.

nbrady-techempower:
Ok we'll get this one in then. Thanks!

PR #3149, opened 2017-12-19
upgrade to lts-10.0 (ghc 8.2.2)

flip111:
I wanted to compare wai performance with the old libraries and compiler on lts-6.3 but i was unable to compile (this is strange because i followed the recommended procedure)

```
Setup wai: http2-1.6.0: configure
Setup wai: aeson-0.11.2.0: copy/register
Setup wai: 
Setup wai: --  While building custom Setup.hs for package http2-1.6.0 using:
[...]

PR #3153, opened 2017-12-20
upgrade servant to lts-10.0

flip111:
don't have bechmarks of lts-6.3 because of https://github.com/TechEmpower/FrameworkBenchmarks/pull/3149#issuecomment-352936343


flip111:
This PR is not good at the moment. I think i rather close this for now, maybe open later.

PR #3159, opened 2017-12-21
Add Rust web framework Actix

fafhrd91:
Add Actix asynchronous web framework for Rust language.

### DIFF
--- a/frameworks/Rust/actix/Cargo.lock
+++ b/frameworks/Rust/actix/Cargo.lock
@@ -1,0 +1,1144 @@
+[[package]]
+name = "actix"
+version = "0.0.1"
+dependencies = [
+ "actix 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "actix-web 0.3.0 (git+https://github.com/actix/actix-web.git)",
+ "askama 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "diesel 1.0.0-beta1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "futures 0.1.17 (registry+https://github.com/rust-lang/crates.io-index)",
+ "http 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "num_cpus 1.7.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rand 0.3.18 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.24 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_derive 1.0.24 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_json 1.0.8 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "actix"
+version = "0.3.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "crossbeam 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "futures 0.1.17 (registry+https://github.com/rust-lang/crates.io-index)",
+ "log 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "skeptic 0.13.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tokio-core 0.1.11 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tokio-io 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "uuid 0.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "actix-web"
+version = "0.3.0"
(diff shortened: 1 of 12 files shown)


frameworks/Rust/actix/Cargo.lock
```diff
@@ -0,0 +1,1144 @@
...
+"checksum vcpkg 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)" = "9e0a7d8bed3178a8fb112199d466eeca9ed09a14ba8ad67718179b4fd5487d0b"
+"checksum version_check 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)" = "6b772017e347561807c1aa192438c5fd74242a670a6cffacc40f2defd1dc069d"
+"checksum void 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)" = "6a02e4885ed3bc0f2de90ea6dd45ebcbb66dacffe03547fadbb0eeae2770887d"
+"checksum walkdir 1.0.7 (registry+https://github.com/rust-lang/crates.io-index)" = "bb08f9e670fab86099470b97cd2b252d6527f0b3cc1401acdb595ffc9dd288ff"
+"checksum winapi 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)" = "167dc9d6949a9b857f3451275e911c3f44255842c1f7a76f33c55103a909087a"
+"checksum winapi-build 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)" = "2d315eee3b34aca4797b2da6b13ed88266e6d612562a0c46390af8299fc699bc"
+"checksum ws2_32-sys 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)" = "d59cefebd0c892fa2dd6de581e937301d8552cb44489cdff035c6187cb63fa5e"
```
msmith-techempower: Please tell me this file is generated.

frameworks/Rust/actix/benchmark_config.json
```diff
@@ -0,0 +1,28 @@
...
+      "query_url": "/queries?q=",
+      "update_url": "/updates?q=",
+      "port": 8080,
+      "approach": "Realistic",
+      "classification": "Micro",
+      "database": "Postgres",
+      "framework": "actix",
+      "language": "Rust",
+      "orm": "Full",
+      "platform": "Rust",
```
msmith-techempower: This should be "None"

frameworks/Rust/actix/src/db.rs
```diff
@@ -0,0 +1,95 @@
...
+    type Item = Vec<models::Fortune>;
+    type Error = io::Error;
+}
+
+impl Handler<TellFortune> for DbExecutor {
+    fn handle(&mut self, _: TellFortune, _: &mut Self::Context) -> Response<Self, TellFortune>
+    {
+        use schema::Fortune::dsl::*;
+
+        match Fortune.load::<models::Fortune>(&self.0) {
```
msmith-techempower: What's with all these `something.0`s? You don't have to change anything, I just want to know what it means.

frameworks/Rust/actix/src/main.rs
```diff
@@ -0,0 +1,203 @@
...
+    HttpServer::new(
+        move || Application::with_state(State{db: addr.clone()})
+            .resource("/json", |r| r.f(json))
+            .resource("/plaintext", |r| r.f(plaintext))
+            .resource("/db", |r| r.route().a(world_row))
+            .resource("/queries", |r| r.route().a(queries))
+            .resource("/fortune", |r| r.route().a(fortune))
+            .resource("/updates", |r| r.route().a(updates)))
+        .backlog(8192)
+        .bind("127.0.0.1:8080").unwrap()
```
msmith-techempower: We employ a host entry for `TFB_SERVER` (or something else) that point to the IP we use in the benchmarks and in the past some implementations have had trouble with `127.0.0.1`; is it possible to bind to something like `0.0.0.0:8080` instead? If not, have you tested with a three machine/vm setup (as is our production benchmark environment) to ensure that the client machine can actually hit this application running on the server?

frameworks/Rust/actix/Cargo.lock
fafhrd91: yes, this is auto generated, but this is same as "requirements.txt" for python project. this file serve purpose of reproducible build (freezing dependencies)

frameworks/Rust/actix/src/db.rs
fafhrd91: in rust you can represent structs in two form. normal form with named attributes and tuple struct.
for tuple struct you address attributes by index. so following code is similar

```rust
pub struct DbExecutor(PgConnection);

fn handle(&mut self, _: RandomWorld, _: &mut Self::Context) -> Response<Self, RandomWorld>
{
        match World::load(&self.0)
        {
            Ok(mut items) => ...
            Err(_) => ...
        }
}
```

vs

```rust
pub struct DbExecutor{
   conn: PgConnection
}

fn handle(&mut self, _: RandomWorld, _: &mut Self::Context) -> Response<Self, RandomWorld>
{
        match World::load(&self.conn)
        {
            Ok(mut items) => ...
            Err(_) => ...
        }
}
```

PR #3165, opened 2018-01-05
Php optimize

joanhey:
Updated to PHP 7.2.1
The MCrypt extension has now been moved out of the core.
Udpated Phalcon and Yaf to work with php 7.2


joanhey:
Installing Mcrypt via pecl. (Not recommended)
Till old frameworks update.

Framework | Version bench | Version without Mcrypt | Actual Version
-------------- | ------------------ | ----------------------------- | ------------------
CakePHP | 2.10.4 | 3.0 | 3.5
Laravel | 4.2 | 5.1 | 5.5

joanhey:
As no one has answered.
I will use the second option

nbrady-techempower:
Ah perfect @joanhey that's what I would have recommended

PR #3167, opened 2018-01-05
implement retries in wildfly since it has to be transactional

codylerum:
Basically we can't turn transactions off so we should retry when we have db contention rather than reporting thousands of errors like we do each round. Wildfly had 4,484 errors in round 14 for the updates test.


codylerum:
Since we are transaction there is a really good chance there there will be db lock contention (only 10,0000 rows and lots of random updates) so lots of our update end up failing. Rather than just reporting those as errors I would prefer to retry the operation a few times before giving up and responding with an error.

msmith-techempower:
Alright, that sounds reasonable to me.

LGTM

PR #3168, opened 2018-01-07
Clone web2py master branch

ilvalle:
Removed specific checkout


nbrady-techempower:
Specific checkouts need to stay to ensure stability in testing. If there's an update that works, that checkout hash can be submitted for verification.

PR #3173, opened 2018-01-09
Added framework cppcms

ekudritski:


frameworks/C++/cppcms/src/main.cpp
```diff
@@ -0,0 +1,56 @@
...
+                "<a href=\"/db\">db</a><br/>"
+                "<a href=\"/queries/20\">queries</a><br/>"
+                "<a href=\"/fortunes\">fortunes</a><br/>"
+                "<a href=\"/updates/20\">updates</a><br/>"
+                "<a href=\"/plaintext\">plaintext</a><br/>"
+                "<a href=\"/cached-worlds/20\">cached-worlds</a><br/>"
+                "</body>"
+                "</html>";
+
+    }
```
msmith-techempower: This function looks like its just for testing; can it be removed?

frameworks/C++/cppcms/src/test_db.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+bool test_db::update_world_object_db(world_object &c) {
+    cppdb::statement st = sql
+            << "UPDATE world SET randomNumber = ? WHERE id = ?"
+            << c.randomNumber << c.id;
+    st.exec();
+    //    if (st.affected() == 0) {
+    //        response().make_error_response(response().internal_server_error, "SQL bad update");
+    //        return false;
+    //    }
```
msmith-techempower: There are also a couple of places with commented out code - let's delete these as well.

toolset/setup/linux/frameworks/cppcms-cppdb.sh
```diff
@@ -0,0 +1,34 @@
...
+fw_depends cppcms
+
+#HOW THAT GOES
+#http://cppcms.com/sql/cppdb/build.html
+
+VERSION=0.3.1
+BACKNAME=cppdb
+CPPCMS_HOME=$IROOT/$BACKNAME-$VERSION
+
+sudo apt-get update -qq
```
msmith-techempower: Unnecessary, this is called as a part of the suite spin-up.

toolset/setup/linux/frameworks/cppcms-cppdb.sh
```diff
@@ -0,0 +1,34 @@
...
+#unixodbc-dev libsqlite3-dev
+sudo apt-get install -y cmake libmysqld-dev libpq-dev
```
msmith-techempower: I believe you can drop this line and instead hoist those libs up to the `fw_depends` line; it should check locally, and if no file exists (`cmake.sh` for example) it will fallback to `apt-get`.

toolset/setup/linux/frameworks/cppcms-cppdb.sh
```diff
@@ -0,0 +1,34 @@
...
+
+cd $BACKNAME-$VERSION
+rm -rf build
+mkdir build
+cd build
+cmake ..
+
+make
+#make test
+sudo make install
```
msmith-techempower: Please take some time to try and get this installed without `sudo` and in the `$IROOT` directory.

toolset/setup/linux/frameworks/cppcms.sh
```diff
@@ -0,0 +1,31 @@
...
+
+#http://cppcms.com/wikipp/en/page/cppcms_1x_build
+
+#note the -rc1 in url
+VERSION=1.1.1
+BACKNAME=cppcms
+CPPCMS_HOME=$IROOT/$BACKNAME-$VERSION
+
+sudo apt-get update -qq
+sudo apt-get install -y cmake libpcre3-dev zlib1g-dev libgcrypt11-dev libicu-dev python
```
msmith-techempower: Same comment here about hoisting to `fw_depends`, and specify which version of `python` is required (e.g. `fw_depends python3`).

toolset/setup/linux/frameworks/cppcms.sh
```diff
@@ -0,0 +1,31 @@
...
+
+cd $BACKNAME-$VERSION
+rm -rf build
+mkdir build
+cd build
+cmake ..
+
+make
+#make test
+sudo make install
```
msmith-techempower: Same comment here about no `sudo` and using `$IROOT`.

See [this](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/setup/linux/webservers/openresty.sh) for an example of how others are doing it.

ekudritski:
Mysql version installs fine. But postgresql does not. 
I dont know how to debug that.

```
Server cppcms: 2018-01-10 17:39:33; cppcms, error: Caught exception [cppdb::posgresql: failed to connect: could not connect to server: Connection refused
Server cppcms: 	Is the server running on host "127.0.0.1" and accepting
Server cppcms: 	TCP/IP connections on port 5432?
Server cppcms: ]

cppcms: setup-nginx-postgresql.sh and framework processes have terminated
```

ekudritski:
It seems that it runs ok in Travis. It does not install PostgreSQL on my local Vagrant.

ekudritski:
Do you need me to de anything else?

PR #3184, opened 2018-01-16
Turn on NUMA & aggressive optimizations and switch back to throughput GC

plokhotnyuk:
It improves throughput results on 5-10% at my notebook, but expecting greater speed up for multi-soket servers.


plokhotnyuk:
Should CI build be restarted? https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/329406949

nbrady-techempower:
It passes for your test, so it's good.

PR #3189, opened 2018-01-17
[WIP] Install resin with gcc-6 in separate directory and make Java/servlet with MySQL to use it

zloster:
I'm curious if the recompilation of the native parts of the Resin server with `gcc 6` will improve performance. I've played a little with the JSON test only. I see small improvements of the average response time compared to the default `gcc 4.8`. The throughput is almost the same.

Could TFB team provide a sample result for the JSON test of this PR and some recent numbers from the `master` branch:
```
vagrant@tfb-all:~$ tfb --mode benchmark --type json --test servlet
```
Of course if the time and resources allow it.

Considerations:
`Resin` is used from several frameworks and I don't want to hit their performance.

### DIFF
--- a/frameworks/Java/servlet/benchmark_config.json
+++ b/frameworks/Java/servlet/benchmark_config.json
@@ -2,7 +2,7 @@
   "framework": "servlet",
   "tests": [{
     "default": {
-      "setup_file": "setup",
+      "setup_file": "setup-resin-gcc-6",
       "json_url": "/servlet/json",
       "plaintext_url": "/servlet/plaintext",
       "port": 8080,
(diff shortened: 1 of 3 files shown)


zloster:
Oracle JDK download is failing: https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/330034213#L1187-L1204

msmith-techempower:
Yeah, I'm not sure why but I noticed that this morning as well.

zloster:
The setup scripts are using `ppa:webupd8team/java`. But they haven't updated the scripts with the latest release of JDK: https://blogs.oracle.com/thejavatutorials/jdk-904%2c-8u161%2c-8u162%2c-7u171%2c-and-6u181-are-released
I only see `8u151-1~webupd8~0`: https://launchpad.net/~webupd8team/+archive/ubuntu/java?field.series_filter=trusty

It seems Oracle have changed some of the download server names: now this is working:
```
wget http://javadl.oracle.com/webapps/download/AutoDL?BundleId=230532_2f38c3b165be4555a1fa6e98c45e0808
```
You could get the URL from here: https://www.java.com/en/download/manual.jsp

PR #3192, opened 2018-01-19
[ci fw-only Java/act]  Act: disable session resolving for the tests

greenlaw110:


greenlaw110:
@nbrady-techempower once again the act test passed while there are other failures in the test.

Question: I have the `[ci fw-only Java/act`] tag in the PR subject, why it still run tests of other projects:?

nbrady-techempower:
@greenlaw110 it needs to be in the commit message, not the PR subject. I don't know how why that caught on, but like travis's `[ci skip]` it looks at the commit message. 

https://github.com/TechEmpower/FrameworkBenchmarks/pull/2495

PR #3194, opened 2018-01-19
Update Gemini

msmith-techempower:
Gemini uses our maven repo now.


frameworks/Java/gemini/.classpath
```diff
@@ -4,29 +4,15 @@
 	<classpathentry kind="lib" path="Docroot/WEB-INF/lib/mail.jar"/>
```
zloster: As far as I remember the `.classpath` file is coming from Eclipse IDE. Are you sure you need it in the source control?

PR #3209, opened 2018-01-26
Update pool parameters for crystal

RX14:
See https://github.com/crystal-lang/crystal-db/issues/77

### DIFF
--- a/frameworks/Crystal/crystal/server.cr
+++ b/frameworks/Crystal/crystal/server.cr
@@ -3,7 +3,7 @@ require "json"
 require "pg"
 require "ecr"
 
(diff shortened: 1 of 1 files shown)


RX14:
Is this likely to make it into the next round? I heard you might release the next preview as a round.

nbrady-techempower:
We've had the opportunity to get a couple more eyes on the project the last few days, so we've fixed several broken frameworks and a few that caused some ports to be locked up. I can't guarantee anything but I think it's likely this will make it in. The commit hash will be released with the round, so you'll know for sure.

RX14:
Either way, i'm interested to see the percentages between the last run before this PR and the next!

PR #3212, opened 2018-01-29
[Dlang/vibed] use mir.random for faster random number generation

Kozzi11:


Kozzi11:
@msmith-techempower should be fixed

Kozzi11:
merge into #3213

PR #3217, opened 2018-01-29
Java/ninja revert transactional change so data persists

### LINKED ISSUE #3216: Ninja-standalone updates test failed to commit data into database
### OS (Please include kernel version)
Linux mint 18 
4.13.0-32-generic #35~16.04.1-Ubuntu SMP
x86_64 x86_64 x86_64

### Expected Behavior
Updates test committed the changed world record into database

### Actual Behavior
The changed world record did not committed into database

### Steps to reproduce behavior
1. Run ninja-standalone on local server with script `setup.sh`
2. send a request to `http://localhost:8080/update`
It will print out a world record including `id` and `randomNumber`
Open mysql client and type `SELECT * FROM hello_world.world where id = <the id printed>`, verify if the randomNumber selected from database is the same as the app returned.

### Other details and logs
The [commit #9627](https://github.com/TechEmpower/FrameworkBenchmarks/commit/96271256ef62936f02f6c114101a4c2a1b00ff5f) changed `@Transactional` into `@ninja.jpa.UnitOfWork`, where the [UnitOfWork source code](https://github.com/ninjaframework/ninja/blob/develop/ninja-core/src/main/java/ninja/jpa/UnitOfWork.java) says:

image

nbrady-techempower:
Thanks @greenlaw110 for catching this

closes #3216

### DIFF
--- a/frameworks/Java/ninja-standalone/src/main/java/controllers/HelloDbController.java
+++ b/frameworks/Java/ninja-standalone/src/main/java/controllers/HelloDbController.java
@@ -10,6 +10,7 @@ import ninja.Results;
 
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
+import com.google.inject.persist.Transactional;
 import ninja.jpa.UnitOfWork;
 import ninja.params.Param;
 
@@ -44,7 +45,7 @@ public class HelloDbController {
         return Results.json().render(worlds);
     }
 
-    @UnitOfWork
+    @Transactional
     public Result update(@Param("queries") Integer queries) {
         if (queries == null || queries < 1) {
             queries = 1;
--- a/frameworks/Java/ninja-standalone/src/main/java/controllers/HelloDbController.java
+++ b/frameworks/Java/ninja-standalone/src/main/java/controllers/HelloDbController.java
@@ -45,7 +45,7 @@ public class HelloDbController {
         return Results.json().render(worlds);
     }
 
-    @Transactional
+    @UnitOfWork
     public Result update(@Param("queries") Integer queries) {
         if (queries == null || queries < 1) {
             queries = 1;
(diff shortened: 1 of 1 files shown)


nbrady-techempower:
Ah, you are correct.. was just reverting back to this.

nbrady-techempower:
This should do the trick.

> Marking a method @Transactional will start a new transaction before the method executes and commit it after the method returns.

Probably not the most performant, but as @zloster pointed out this was previously not following the rules, and then updates weren't persisting.

greenlaw110:
@nbrady-techempower 

I tried the following approach:

> Marking a method `@transactional` will start a new transaction before the method executes and commit it after the method returns.

And here is my gotchas:

1. You must add `@UnitOfWork` annotation to the original controller method:

```java
    @UnitOfWork
    public Result update(@Param("queries") Integer queries) {
        ...
    }
```

2. You must make sure the new method with `@Transactional` be `public`.

3. The `Transactional` must be `com.google.inject.persist.Transactional`

Otherwise the approach won't work, i.e. it won't commit the data into the database

nbrady-techempower:
I did all those things except making the method that commits the transaction `public`. I'm surprised that's needed to actually persist the data, but ill update. Thanks guys!

zloster:
@nbrady-techempower This restriction comes from Guice. Note the sentence after the sample code: https://github.com/google/guice/wiki/Transactions#method-level-transactional
Another note: I don't know how Guice is implementing the AOP stuff i.e. using JDK proxy objects or CGLIB instrumentation. I see the `cglib` dependency in their pom.xml but I don't know the details. 
Depending on the details the calls within an object will not trigger the AOP stuff connected to the methods - for example with the default Spring AOP configuration using JDK proxies and run-time weaving. Some details here: https://docs.spring.io/spring/docs/current/spring-framework-reference/core.html#aop-pointcuts-designators, scroll down to the first green block.

> Due to the proxy-based nature of Spring’s AOP framework, **calls within the target object** are by definition not intercepted.

> For JDK proxies, only public interface method calls on the proxy can be intercepted.

> With CGLIB, public and protected method calls on the proxy will be intercepted, and even package-visible methods if necessary.

PR #3221, opened 2018-01-30
Update framework info, better json handling

fafhrd91:
Update db bench commits transaction for each entry in update batch.

### DIFF
--- a/frameworks/Rust/actix/src/db.rs
+++ b/frameworks/Rust/actix/src/db.rs
@@ -57,7 +57,7 @@ impl Handler<UpdateWorld> for DbExecutor {
     fn handle(&mut self, msg: UpdateWorld, _: &mut Self::Context) -> io::Result<()> {
         use schema::World::dsl::*;
 
-        for world in msg.0 {
+        for world in &msg.0 {
             let _ = diesel::update(World)
                 .filter(id.eq(world.id))
                 .set(randomnumber.eq(world.randomnumber))
--- a/frameworks/Rust/actix/Cargo.toml
+++ b/frameworks/Rust/actix/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "actix"
-version = "0.1.0"
+version = "0.1.1"
 build = "build.rs"
 
 [dependencies]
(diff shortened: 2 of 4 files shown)


frameworks/Rust/actix/src/db.rs
```diff
@@ -57,12 +58,15 @@ impl Handler<UpdateWorld> for DbExecutor {
     fn handle(&mut self, msg: UpdateWorld, _: &mut Self::Context) -> io::Result<()> {
         use schema::World::dsl::*;
 
-        for world in msg.0 {
-            let _ = diesel::update(World)
-                .filter(id.eq(world.id))
-                .set(randomnumber.eq(world.randomnumber))
-                .execute(&self.0);
-        }
+        let _ = self.0.transaction::<_, Error, _>(|| {
```
msmith-techempower: I don't think this is allowed. To me, this reads like all those updates are going to be batched into a single transaction, and that is against the rules.

fafhrd91: as I understand batch updates is acceptable . section 5, point 9

https://frameworkbenchmarks.readthedocs.io/en/latest/Project-Information/Framework-Tests/#database-updates

```
9. Use of batch updates is acceptable but not required. To be clear: batches are not permissible for selecting/reading the rows, but batches are acceptable for writing the updates.
```

I see some code commits once, for example https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Python/flask/app.py#L159

fafhrd91: here is another example https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Python/pyramid/frameworkbenchmarks/views.py#L86

nbrady-techempower: Batch updates are acceptable, not transactions for batch updating. see rule 10 http://frameworkbenchmarks.readthedocs.io/en/latest/Project-Information/Framework-Tests/#database-updates

fafhrd91: that is the reason for my initial implementation :)

but actix results seems very low for "updates" bench, compared to other benchmarks. I see only reason that other frameworks commits once per batch updates. 
I know python pretty well, two examples I linked commit once for sure, should those results be banned then?

nbrady-techempower: There's a difference between batch processing and transactions. The frameworks you linked to are not using transactions, but a single batch update. This is allowed. If you can do that without using transactions, you're allowed to do that.

fafhrd91: ok, I see. let me change pr

fafhrd91: transaction is removed

PR #3225, opened 2018-01-30
Revert "Install resin with gcc-6 in separate directory (#3189)"

michaelhixson:
This reverts commit d967f73f8e22a01d8c25d6cb7dd530d9c4e10820.

This line was breaking the normal resin installation, despite the wishes
communicated in the comment:

```
# We need separate folder so not to clash with the normal resin folder
mv resin-$RVER resingcc6-$RVER
```

(resin-$RVER is where normal resin was installed, so after this line of
 code ran, frameworks that depended on normal resin later in the same run
 would fail.)

We could fix that by naming the .tar.gz file in resingcc6.sh to something
else, but there are more important issues:

1. The original PR #3189 asked us to test the change and provide sample
   results.  It's not clear they wanted us to merge the change in as is.

2. The way gcc-6 is installed, it's not sandboxed, so actually resin.sh
   and resingcc6.sh are installing exactly the same build of resin.
   gcc-6 happens to already be installed from other frameworks and
   previous runs, so we've already been testing resin with gcc-6
   apparently, and this alternative installation is not necessary.


michaelhixson:
@zloster FYI it looks like we were already using gcc-6 when installing resin, and the resingcc6 installer had an issue that caused problems for "normal" resin, so in this PR I'm removing resingcc6.

zloster:
I see the problem. Thank you for the very detailed explanation.

PR #3229, opened 2018-01-31
Update crystal kemal amber

faustinoaq:
Hi @nbrady-techempower 🙋‍♂️  this PR updates Crystal, Kemal and Amber to their latest versions.

Crystal: @RX14 ^

Kemal: @sdogruyol @crisward ^

Amber: @eliasjpr @drujensen @elorest @marksiemers @robacarp ^


RX14:
No, why are we encrypting configuration? This sucks, remove this.

faustinoaq:
> No, why are we encrypting configuration? This sucks, remove this.

@RX14 That's only for amber 😅 

Commonly amber `production` environment uses `.production.enc` file

My bad, I will use `config/application.cr` instead 👍

sdogruyol:
Thank you @faustinoaq :pray:

frameworks/Crystal/amber/setup.sh
```diff
@@ -10,6 +10,8 @@ export GC_MARKERS=1
 
 export AMBER_ENV=production
 
+export DATABASE_URL=postgres://benchmarkdbuser:benchmarkdbpass@TFB-database/hello_world
```
msmith-techempower: Does this not need the port specified?

frameworks/Crystal/amber/shard.lock
```diff
@@ -48,31 +48,27 @@ shards:
...
-    version: 0.1.1
-
   shell-table:
     github: jwaldrip/shell-table.cr
     version: 0.9.2
 
   slang:
     github: jeromegn/slang
-    version: 1.7.0
+    commit: 777e2fc3e63daf5cfe9bd4475257905903b78542
```
drujensen: why is this one not versioned?

faustinoaq: @drujensen because this 👉 https://github.com/amberframework/amber/pull/536 fix was after `v0.6.4`

frameworks/Crystal/amber/setup.sh
faustinoaq: @msmith-techempower We didn't specify database port before but we specify amber port here 👉 https://github.com/TechEmpower/FrameworkBenchmarks/pull/3229/files#diff-2a38afc8380af63cc2e9aea0778f0d6fR53 

@drujensen I'm using `DATABASE_URL` to avoid using encrypted settings (_This is a production environment_), another alternative is using an initializer with this:

```
Granite::ORM.settings.database_url = Amber.settings.database_url
```

WDYT?

faustinoaq: @msmith-techempower Done! 👍

marksiemers:
I think the `GC_MARKERS=1` actually reduced the performance of crystal-raw - at least for the plaintext results.

It may be better in a different PR, but we may think about removing that. I don't think it helped Kemal or Amber.

faustinoaq:
@marksiemers you're right, about a 14% less performance on `plaintext` test. However other test like `json` test increased about a 16 %. So I don't know if `GC_MARKERS=1` was good or bad 😅 

screenshot_20180131_125450

The logic was:

If crystal uses a multithreading GC per process and we launching many processes then GC is causing a big overload.

So, How to solve that overload?

Possible answer: using `GC_MARKERS=1` to set the number of marker threads to 1.


@RX14 WDYT?

faustinoaq:
@msmith-techempower Travis is passing now 😄 🎉

faustinoaq:
@msmith-techempower  Wait, **Do not merge it yet!!!**, 

@drujensen @eliasjpr Amber is printing a lot of logs:

```
Server amber: 01:27:26 Server     | (INFO) SELECT world.id, world.randomNumber FROM world WHERE id=$1 LIMIT 1: 4226
Server amber: 01:27:26 Server     | (INFO) SELECT world.id, world.randomNumber FROM world WHERE id=$1 LIMIT 1: 9486
Server amber: 01:27:26 Server     | (INFO) SELECT world.id, world.randomNumber FROM world WHERE id=$1 LIMIT 1: 3726
Server amber: 01:27:26 Server     | (INFO) SELECT world.id, world.randomNumber FROM world WHERE id=$1 LIMIT 1: 4135
Server amber: 01:27:26 Server     | (INFO) SELECT world.id, world.randomNumber FROM world WHERE id=$1 LIMIT 1: 5238
Server amber: 01:27:26 Server     | (INFO) SELECT world.id, world.randomNumber FROM world WHERE id=$1 LIMIT 1: 8444
```

Why?

faustinoaq:
I think amber/granite logs are fixed now, just waiting for Travis 🤞

faustinoaq:
@msmith-techempower Travis is passing now 💯 I think this is ready to merge, just waiting for Travis to complete 🎉 

screenshot_20180201_105704

nbrady-techempower:
Travis is good, it's skipping the rest of the tests.

PR #3244, opened 2018-02-02
[ci fw-only Java/servlet] Fix cache_query implementation to return the requested number of results

zloster:
I'm really trying to use the `mass retrieve` operation of the cache implementation. I need a non-repeated keys for this.
The explanation:
1. Generate all the numbers from 1 to 10000 when starting;
1. Shuffle them;
1. Store the result in a custom `ArrayList` which allows the reads of indexes past the end to read data from the begging;
1. When a request comes choose a random start element and read the requested number of keys - wrapping around at the end.

### DIFF
--- a/frameworks/Java/servlet/src/main/java/hello/Cache2kPostgresServlet.java
+++ b/frameworks/Java/servlet/src/main/java/hello/Cache2kPostgresServlet.java
@@ -7,7 +7,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
@@ -60,8 +59,8 @@ public class Cache2kPostgresServlet extends HttpServlet {
 			IOException {
 		final int count = Common.normalise(req.getParameter("queries"));
 		final int start = ThreadLocalRandom.current().nextInt(DB_ROWS);
-		Set<Integer> keys = IntStream.range(start, start + count)
-				.mapToObj(i -> randomNumbers.get(i)).collect(Collectors.toSet());
+		List<Integer> keys = IntStream.range(start, start + count)
+				.mapToObj(i -> randomNumbers.get(i)).collect(Collectors.toList());
 
 		// Set content type to JSON
 		res.setHeader(Common.HEADER_CONTENT_TYPE, Common.CONTENT_TYPE_JSON);
--- a/frameworks/Java/servlet/src/main/java/hello/Cache2kPostgresServlet.java
+++ b/frameworks/Java/servlet/src/main/java/hello/Cache2kPostgresServlet.java
@@ -1,14 +1,24 @@
 package hello;
 
-import java.io.*;
-import java.sql.*;
-import java.util.*;
-import java.util.concurrent.*;
+import java.io.IOException;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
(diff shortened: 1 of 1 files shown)


nbrady-techempower:
@zloster let me know when you're ready here

zloster:
@nbrady-techempower This is ready but I've just read again the rules (specifically rules 5 and 6) and this change is violating the later: there is no **separate** `cache` process but I'm using **single** `cache query operation` to extract the data from the cache.
In short my understanding is: if an in-process cache is used than there should be N cache query operations to retrieve a N number of objects.
My thoughts:
1) I don't understand the motivation about such a restriction;
2) If I want to use a single cache operation for retrieval than I will need a way to get N number of non-repeating keys. The naive approach will be some sort of looping with checking what we get so far and requesting random numbers until the requested number of non-repeating keys is gathered. But this is increasing the noise of the test. The other approach is the random shuffling as it is done in this PR. Each HTTP request for cache data will depend **only** on one random number - the starting index in the shuffled array.

So my suggestion is: what about modifying the requirements of the test to mandate the use of shuffled array for the random keys. This should decrease the noise from the test and will unify the code for test implementations which are not using in-process caches and opt for extracting data with one operation.

nbrady-techempower:
@zloster I'm not sure I understand why your 2nd point is naive. It's really no different than the queries or updates test. For example, in Gemini we do exactly that for the cache test.

```
for (int i = 0; i < queries; i++)
{
     worlds[i] = store.get(CachedWorld.class, random.nextInt(DB_ROWS) + 1);
}
```

I think the point here is to try and keep it as close to the query test as possible, to see the difference between "turning on" your framework's caching mechanism.

For the "separate cache" point, I can see why that is strange. I can talk with the team on Monday about that and see how we want to proceed. My guess is we'll want to stick with `N` retrieval operations.

I'll leave this open and get back to you.

frameworks/Java/servlet/src/main/java/hello/Cache2kPostgresServlet.java
```diff
@@ -29,39 +39,51 @@
 	@Override
 	public void init(ServletConfig config) throws ServletException {
 		super.init(config);
+		randomNumbers = new CircularList<>(shuffledNumbers(DB_ROWS));
```
msmith-techempower: This is circumventing work meant to occur in the request.

frameworks/Java/servlet/src/main/java/hello/Cache2kPostgresServlet.java
```diff
@@ -29,39 +39,51 @@
...
 	}
 
 	@Override
 	protected void doGet(HttpServletRequest req, HttpServletResponse res) throws ServletException,
 			IOException {
 		final int count = Common.normalise(req.getParameter("queries"));
-		final Random random = ThreadLocalRandom.current();
+		final int start = ThreadLocalRandom.current().nextInt(DB_ROWS);
+		List<Integer> keys = IntStream.range(start, start + count)
+				.mapToObj(i -> randomNumbers.get(i)).collect(Collectors.toList());
```
msmith-techempower: While `start` will indeed be random on all requests, the order of the queried ids will never change after startup.

This must be changed to loop `count` times and perform `nextInt` and `get` each time through the loop to be merged.

zloster: @msmith-techempower I've already noted that the current implementation is [not following the rules](https://github.com/TechEmpower/FrameworkBenchmarks/pull/3244#issuecomment-362849176) for the test `cache query` test.
@nbrady-techempower left it open for some additional feedback of my thoughts.

PR #3245, opened 2018-02-02
Update python and some frameworks.

methane:


nbrady-techempower:
@methane looks like you removed tornado, a dependency for the `flask-pypy2` test which is causing it to fail

nbrady-techempower:
@methane looks like `flask-pypy2` tests are still failing and also see #3267

PR #3256, opened 2018-02-05
use ikitommi/immutant-web

ikitommi:
use `ikitommi/immutant-web`, less noise, much better perf.

### DIFF
--- a/frameworks/Clojure/reitit/README.md
+++ b/frameworks/Clojure/reitit/README.md
@@ -11,9 +11,9 @@ This is the [Reitit](https://github.com/metosin/reitit) portion of a [benchmarki
 The dependencies are documented in [project.clj](hello/project.clj),
 but the main ones are:
 
-* [Immutant 2.1.9](http://immutant.org/)
-* [Clojure 1.9.0-RC1](http://clojure.org/)
-* [metosin/jsonista "0.1.0-20171106.055332-4"](https://github.com/metosin/jsonista)
+* [Clojure 1.9.0](http://clojure.org/)
+* [ikitommi/immutant-web "3.0.0-alpha1"](https://github.com/ikitommi/immutant) - a performant fork of [Immutant]((http://immutant.org/))
+* [metosin/jsonista "0.1.1"](https://github.com/metosin/jsonista)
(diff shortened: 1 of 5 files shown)


nbrady-techempower:
Hi @ikitommi - 
Would you be able to take a look at the updates test for Closure while you're in there? It is either not persisting updates to the database or returning a response before the updates were committed.

ikitommi:
"test for Closure", I'm not sure what that means. I guess it doesn't relate to this PR? If there is some other clojure benchmark with a problem, I can try to find the persons from Clojure Slack.

cheers,

Tommi

nbrady-techempower:
Thanks tommi, it was taken care of already

ikitommi:
Hi. Sadly, this doesn't seem to be part of the valentine / round15. Is there any change of getting feedback on the perf?

ikitommi:
About to release the library next week, would have loved to get real numbers on this.

msmith-techempower:
https://tfb-status.techempower.com/ you can see live numbers as the benchmarks are continuously run

ikitommi:
Thanks for the info. I'll parse the results when they arrive (can see from test meta-data when the changed have been applied). I guess the official site & numbers will update next time in far future?

ikitommi:
It seems that the live data is not running the latest code, https://tfb-status.techempower.com/results/a1110174-a3af-4c2e-a4e1-bfcdca8bd8eb. Any guess when this patch would get a result?

cjnething:
Hi @ikitommi it looks like [this run](https://tfb-status.techempower.com/results/0cf97d5a-86b1-4ebe-b53a-485a6532dae9) has your code. Let me know if that doesn't look right!

ikitommi:
Thanks. It seems that 306/444 have failed on that run, including reitit. Coudn't figure out why: https://tfb-status.techempower.com/unzip/results.2018-03-04-21-34-49-221.zip/reitit/out.txt

cjnething:
@ikitommi We're in the middle of switching the frameworks over to Docker and it looks like this framework hasn't been switched over yet, hence the Docker errors.

PR #3257, opened 2018-02-05
Include the git commit id and repository url in results.json

michaelhixson:
We already generate a commit_id.txt file when we do a full continuous
run, but not when we run tests manually.  With this change, now we always
record the commit id, and it's in results.json, which is a more
convenient home.


fafhrd91:
vagrant tests doesnt work after this commit. FrameworkBecnhmarks in vagrant doesnt contain.git directory

michaelhixson:
@fafhrd91 Thanks, I'll fix that up.

PR #3264, opened 2018-02-07
Optimize db benchmarks for Actix framework

fafhrd91:


frameworks/Rust/actix/Cargo.toml
```diff
@@ -25,7 +25,7 @@ num_cpus = "1.0"
 futures = "0.1"
 http = "0.1"
 actix = "^0.4.3"
-actix-web = "=0.3.1"
+actix-web = "0.3"
```
nbrady-techempower: Just curious, but why are you regressing here?

fafhrd91: it just "any available version 0.3.x", it it will use current version 0.3.3

fafhrd91: here is link to travis job, https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/338621337#L2183

nbrady-techempower: What's the difference between "^0.4" and "0.4" is that still getting you "0.4.*"? Where possible, we'd like to tie down packages, even to their patch version. The reason for this is, when we run the benchmark rounds and tag the commit, we want to be able to replicate the runs at any time. "0.3" could be "0.3.22" in some future replication, and could create radically different results (or even break the tests - sadly non-breaking semantic versioning is a utopia I only dream of).

nbrady-techempower: In fact, we should probably be adding the `cargo.lock` for all rust frameworks

fafhrd91: "^" mean lowest allowed version. i.e I can say "^0.3.2" - any version starting "0.3.2", "0.3" is any "0.3.x" version. to pin specific version use "=0.3.2" - only 0.3.2. 

Cargo.lock file contains specific version, it solves reproducible builds problem

nbrady-techempower: Fantastic, as we just had this issue with a few rust frameworks breaking randomly, let's go with a cargo.lock here.

fafhrd91: do you want me to commit Cargo.lock for actix in this PR?

nbrady-techempower: That would be great @fafhrd91 !

fafhrd91: added

fafhrd91:
could you guys merge this before next test run? 

btw are you planing to do another test run?

nbrady-techempower:
We've already captured R15 results and are preparing things for release, so this will not make it. However, you can see how it performs here: https://tfb-status.techempower.com/

fafhrd91:
ok, thanks

frameworks/Rust/actix/src/main.rs
```diff
@@ -48,32 +47,26 @@ fn world_row(req: HttpRequest<State>) -> Box<Future<Item=HttpResponse, Error=Err
 fn queries(req: HttpRequest<State>) -> Box<Future<Item=HttpResponse, Error=Error>> {
     // get queries parameter
     let q = if let Some(q) = req.query().get("q") {
-        q.parse::<usize>().ok().unwrap_or(1)
+        q.parse::<u16>().ok().unwrap_or(1)
     } else {
         1
     };
```
nbrady-techempower: I've been playing around with rust a bit. These 5 lines could be:

`let q = req.query().get("q").unwrap_or("1").parse::<u16>().unwrap_or(1);`

Happens in updates too in this file, which is still using `<usize>`

Not a nit or anything, just enjoying messing around with the language.

fafhrd91: well, if this ```.unwrap_or("1")``` happen, then you would parse default string.

equivalent would be:
```rust
let q = req.query().get("q").unwrap_or(1).map(|v| v.parse::<u16>().unwrap_or(1));
```

fafhrd91:
```rust
let q: u16 = req.query().get("q").map(|v| v.parse().unwrap_or(1)).unwrap_or(1);
```

PR #3266, opened 2018-02-07
Don't crash in benchmarker.py when there is no local git repository

michaelhixson:
see https://github.com/TechEmpower/FrameworkBenchmarks/pull/3257#issuecomment-363851440


nbrady-techempower:
Tested locally

PR #3269, opened 2018-02-08
[ci fw-only Java/act]  Act Updates test - now batch update in single transaction

greenlaw110:


greenlaw110:
@nbrady-techempower there are failure tests in the CI. But Java/act tests passed. Can you please merge this PR? thank ... Green

PR #3270, opened 2018-02-08
Updated Nginx and PHP

joanhey:
Disabled debug to PHP for better performance.


nbrady-techempower:
@joanhey can you merge this into #3265 so we can get a single travis run with the nginx update and your changes?

joanhey:
OK
They are different things. I will merge now

nbrady-techempower:
Yeah it's fine. The nginx update probably could have gone there, and the php updates here.

PR #3279, opened 2018-02-08
Feat/add basic fanta

trezm:


trezm:
Did this prematurely! Sorry for the false alarm!

PR #3282, opened 2018-02-09
web2py updates fix

### LINKED ISSUE #3262: Frameworks Failing for not passing Update Test Verification
From #3210 & #2027 we are now verifying to make sure that database entries have changed after the Updates test runs. There are still a few framework tests that need to be addressed. If I pinged you here, it's because you might be able to help, or you were the core contributor for these tests. Note that these are failures and the results of their update tests will no longer be captured.

ffead-cpp
ffead-cpp-apache-mongo
~~macchiato~~
~~koa-postgres~~
~~sailsjs~~
~~sailsjs-postgres~~
~~play2-scala-reactivemongo~~
~~play2-scala-reactivemongo-netty~~
~~web2py-optimized~~

@sumeetchhetri @methane @mkurz @ilvalle 

Some things to look for that I've already fixed in other tests are:

Using the queries handler that doesn't update the db at all
Not having the proper database name in the `benchmark_config.json`: MySQL, Postgres or Mongodb
Returning a response before the updates are done writing to the db

abastardi:
Fixes #3262 for web2py-optimized.

### DIFF
--- a/frameworks/Python/web2py/app/standard/modules/controller.py
+++ b/frameworks/Python/web2py/app/standard/modules/controller.py
@@ -40,8 +40,6 @@ def updates():
         world['randomNumber'] = newNumber
         worlds.append(world)
         update_world(wid, newNumber)
-    if current.optimized:
-        db.flush_world_updates() # Batch updates.
     return jsonOut.dumps(worlds)
 
 def fortune():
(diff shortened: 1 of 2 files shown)


nbrady-techempower:
Thanks @abastardi but this was fixed in the PR right before yours. #3281

PR #3284, opened 2018-02-10
Fix ffead-cpp(-apache-mongo) updates test implementation

michaelhixson:


nbrady-techempower:
merging this, fyi @sumeetchhetri

PR #3287, opened 2018-02-13
python3: Install Python 3.6.4 from deadsnakes APT PPA (PGO optimized)

ciarancourtney:


nbrady-techempower:
With work being done in the docker branch for the python tests on ubuntu 16, im going to close this for now. Can be re-opened / discussed if it's still needed there.

PR #3289, opened 2018-02-13
Rust/iron add cached queries test

nbrady-techempower:
Added the cached query test for iron, optimized the build process similar to other Rust frameworks, and used `TFB-database`


fafhrd91:
iron is dead. As of iron/iron@9e5bccb Iron was marked inactive. Do you want to maintain benchmark for it?

PR #3291, opened 2018-02-15
Update and reconfigure PHP stack

joanhey:
# Update
PHP 7.2.1 -> 7.2.2 (Before only changed no debug compilation)
Phalcon 3.2.4 -> 3.3.1
HHVM 3.21 -> 3.24 (Fixed crash with utf-8 and mysql. See https://hhvm.com/blog/2018/02/07/hhvm-3.24.2.htmll)

# PHP now passes the tests better
Need more work to separate db and queries tests files. And create new files for PHP7 only

# Change ngix.conf
Was from the i7 benchmarks

`worker_processes 8;` (Change to 'auto', 1 per CPU)
`worker_connections 1024;` (Change to 16384, it's not necessary so high, only for the vagrant box)
`server_tokens off;` (Add to show only 'server: nginx' without version)

Before 8*1024 = 8,192 max connections for nginx.
Now some benchmarks use up to 16,384 concurrent connections.
When arrive near to the max connections, nginx return this:
`[alert] 27050#0: 1024 worker_connections are not enough`
And these connections simply never arrive to php.

# Change php-fpm.conf
Also from the i7 benchmark. The problem of php-fpm, is that there aren't a lot information about it.

## listen.backlog = 256 -> 4096
The PHP FastCGI Process Manager is limited to 256 open sockets, when the server is configured to `net.core.somaxconn = 65535`.
Perhaps is for that than php-raw, stop the max responses at 256 concurrent connections in the data table.
See https://linux.die.net/man/2/listen

## error_log = /dev/stderr
Send the log to stderr and it will show in the out.txt

PHP-FPM log is not only for errors, so you will watch messages like:
`NOTICE: ready to handle connections, ...`

But also the errors than help to adjust it.
`WARNING: [pool www] seems busy (you may need to increase pm.start_servers, or pm.min/max_spare_servers)...`

## Double child processes for the pool

pm.max_children = 512 -> 1024
pm.start_servers = 256 -> 512
pm.min_spare_servers = 16 -> 50
pm.max_spare_servers = 256 -> 512

## More things to do
pm = dynamic
Dynamic is for shared hosting, to better share the resources of the servers at peaks.
With one application per server should use `static`, but the vagrant box would not support it.

Worker processes are not multi-threaded, the same that in nginx, nodejs, ....
Now only use 1 pool, with multi CPU computers it will use more pools. Ideally a pool for CPU.

This configuration is more than conservative, for the 'server central'.
Need a bash script to adjust the config of pools and children to the server. When I have time I will create it

Remember to enable keepalive and fastcgi_keep_conn, if you use upstream in nginx. See #3265.

Add KumbiaPHP framework.

These benchmarks are not a competition between frameworks, they are a great help tool to improve all frameworks.
Thank you for this great benchmark.


msmith-techempower:
I haven't reviewed this, but to everyone who will ever open a PR, use this one as your model.

frameworks/PHP/kumbiaphp/benchmark_config.json
```diff
@@ -0,0 +1,51 @@
...
+            "json_url": "/json",
+            "db_url": "/db",
+            "query_url": "/db/queries/",
+            "fortune_url": "/fortune",
+            "update_url": "/db/update/",
+            "port": 8080,
+            "approach": "Realistic",
+            "classification": "Fullstack",
+            "database": "MySQL",
+            "framework": "None",
```
nbrady-techempower: kumbia here?

toolset/setup/linux/languages/hhvm.sh
```diff
@@ -12,10 +12,7 @@ fi
 
 fw_get http://dl.hhvm.com/conf/hhvm.gpg.key | sudo apt-key add -
 
-# hhvm 3.24 causes most of our hhvm-using test implementations to segmentation
-# fault when connecting to the mysql database (the crash happens right as they
-# try to specify utf8 as the charset).  hhvm 3.21 doesn't have this problem.
-echo deb [arch=amd64] http://dl.hhvm.com/ubuntu `lsb_release -sc`-lts-3.21 main | sudo tee /etc/apt/sources.list.d/hhvm.list
```
nbrady-techempower: You're reverting a fix we needed to make here. We will have to test locally but probably can't accept this change.

michaelhixson: Apparently that bug in hhvm 3.24 that we were working around is fixed in 3.24.2: https://hhvm.com/blog/2018/02/07/hhvm-3.24.2.html

nbrady-techempower: Oh good!

frameworks/PHP/kumbiaphp/setup.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+fw_depends mysql php7 nginx composer
+
+sed -i 's|localhost|'"${DBHOST}"'|g' bench/app/config/databases.php
```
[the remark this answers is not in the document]
joanhey: The raw connection use `TFB-database` and the ORM `DBHOST`, just for try both.
For speed, the best bet is an IP, since it avoid a name lookup. 
But it is insignificant, since all the frameworks use persistent connections.

It would be nice if all the frameworks used only one and avoid controversy. Add it to general requirements in https://www.techempower.com/benchmarks/#section=code

`DBHOST` -> 200 https://github.com/TechEmpower/FrameworkBenchmarks/search?utf8=%E2%9C%93&q=DBHOST&type=
`TFB-database` -> 96 https://github.com/TechEmpower/FrameworkBenchmarks/search?utf8=%E2%9C%93&q=TFB-database&type=

toolset/setup/linux/languages/hhvm.sh
joanhey: Yes, I wrote:
HHVM 3.21 -> 3.24 (Fixed crash with utf-8 and mysql)
I will add the link too.

I tested the benchmarks locally without problems, before submitting the commit.

frameworks/PHP/kumbiaphp/benchmark_config.json
```diff
@@ -0,0 +1,51 @@
+{
+    "framework": "kumbiaphp",
+    "tests": [{
+        "default": {
+            "setup_file": "setup",
+            "plaintext_url": "/",
+            "json_url": "/json",
+            "db_url": "/db",
+            "query_url": "/db/queries/",
```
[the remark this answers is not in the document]
joanhey: Where are the requirements for the cached_queries ?

Personally, I don't believe in cached queries, it is always better to cache the output, even with different output formats.

frameworks/PHP/kumbiaphp/benchmark_config.json
joanhey: 👍 :)

joanhey:
I will send some issues now that is fresh, before I don't rememer it.

frameworks/PHP/kumbiaphp/benchmark_config.json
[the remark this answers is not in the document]
nbrady-techempower: http://frameworkbenchmarks.readthedocs.io/en/latest/Project-Information/Framework-Tests/

frameworks/PHP/kumbiaphp/setup.sh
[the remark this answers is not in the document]
michaelhixson: `DBHOST` was used first before `TFB-database` was a thing, and now every time I am fixing some other issue in a framework and I see it using `DBHOST`, I change it to use `TFB-database` instead because the code is always simpler that way.  We're not interested in measuring the difference between the two in this project.  Use `TFB-database`.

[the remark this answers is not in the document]
joanhey: :+1:

joanhey:
Restart the build in travis. Now that there are free jobs.
https://travis-ci.org/joanhey/FrameworkBenchmarks  passed
Thank you

PR #3292, opened 2018-02-15
Dockerify

msmith-techempower:
Switches the suite to using Docker. Initially this breaks everything except `gemini` and `gemini-mysql` which were used as proofs of concept.


frameworks/Java/gemini/gemini-mysql.dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM resin:latest
```
michaelhixson: I'm reading in the [Docker docs](https://docs.docker.com/engine/reference/builder/#from) that

> Each `FROM` instruction clears any state created by previous instructions.

Does that mean `FROM` is basically single inheritance?  If yes, then how would I declare multiple dependencies?

frameworks/Java/gemini/gemini-mysql.dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+RUN apt-get install -qqy -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" \
+    ant
+
+ADD Docroot/ /gemini/Docroot
+ADD Source/ /gemini/Source
+ADD build.xml /gemini/
+ADD ivy.xml /gemini/
+ADD ivysettings.xml /gemini/
```
michaelhixson: This is copying files in TFB into the docker container, right?  Would something like `ADD ./ impl` also work, if the goal was to copy everything in the framework's directory into the container?  I'm thinking it would be nice to be able to copy and paste this part from framework to framework without worrying about individual files like we are here.

frameworks/Java/gemini/gemini-mysql.dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+RUN cd /gemini/Docroot/WEB-INF; mv gemini-mysql.conf GeminiHello.conf;
+
+RUN cd /gemini; mkdir -p Docroot/WEB-INF/classes; mkdir -p Docroot/WEB-INF/lib; ant resolve; ant compile
+
+EXPOSE 8080
+
+CMD ["resinctl", "-conf", "/gemini/Docroot/WEB-INF/resin.xml", "console"]
```
michaelhixson: What is our general guideline for when to use `RUN` or `CMD`?  Is it:  Use `RUN` for every bash shell command involved in setup except the last one, and use `CMD` for the last bash command that launches the server?

frameworks/Java/gemini/gemini-mysql.dockerfile
msmith-techempower: Yes, and you cannot; instead, you chain the inheritance. I have not yet seen a need for multiple inheritance, but if we have one we can probably just tackle it via having some redundant installs in docker images (not great, but not terrible).

frameworks/Java/gemini/gemini-mysql.dockerfile
msmith-techempower: That should work. I was trying to keep things clean for my POC.

frameworks/Java/gemini/gemini-mysql.dockerfile
msmith-techempower: `RUN` is only used for setup; it will not actually run and persist a process. See [this comment](https://github.com/TechEmpower/FrameworkBenchmarks/pull/3292/files#diff-f4a364bcc831a5afd517148bd6ce7415R31) for an example.

`CMD` is what is executed when we start the container. This version does not execute a shell, but it could have just as well been `CMD resin -conf /gemini/Docroot/WEB-INF/resin.xml console` and that would have invoked `/bin/sh resin -conf ...` (the reaper will clean this either way. For instances where we need to run multiple processes (like PHP using `php-fpm` and `nginx`) we should be able to get away with `CMD php-fpm ... && nginx`

msmith-techempower:
Woo, `gemini` and `gemini-mysql` passed in Travis-CI (`gemini-postgres` failed because I didn't implement it yet).

LGTM and start building out the rest!

PR #3298, opened 2018-02-16
Docker - Implement postgresql

msmith-techempower:
Also implemented `gemini-postgres` as a proof of concept.


sebastienros:
Any reason not to use the official docker images for mysql and postgres ?

msmith-techempower:
@sebastienros For this working version, we are aiming for parity with the current toolset setup, and we don't use the official docker images for mysql/postgres/etc for that.

We can explore moving to using those when this effort is further along.

Issue #3296, opened 2018-02-16
Dockerify all the things


msmith-techempower:
# Abstract
Move the suite away from the current methodology of using scripting files to handle dependencies, installation into the `$IROOT` directory for sandboxing, and instead use [Docker CE](https://www.google.com/search?q=docker&ie=utf-8&oe=utf-8&client=firefox-b-1-ab).

# Reasoning
1. Installing software to `$IROOT` is hard (much harder than `apt-get`) and prone to difficult-to-identify errors
2. Many dependencies use `apt-get` already for various reasons, and migrating away from globally installed software is hard (much harder than allowing `apt-get`)
3. Docker provides a much cleaner reusable sandbox environment that has been fire-tested much more-so than our in-house implementation
4. In testing, Docker does not appear to incur much (if any) performance penalty

Ongoing work is in [this branch](https://github.com/msmith-techempower/FrameworkBenchmarks/tree/dockerify).

# Action Items (in no particular order)
1. - [x] Create base `tfb` dockerfile with as minimal software installed as possible #3292
2. - [x] Implement base dockerfile to have an entrypoint of `TFBReapear` so that everything is run through him #3292
3. - [x] Create POC using only `gemini` and the dockerfile(s) required to build and run in console mode #3292
4. - [x] Create POC suite implementation using the dockerfile(s) require to build and run `gemini` and perform a verification #3292
5. - [x] Implement stable mechanism for finding running docker containers and stopping them #3292
6. - [x] Implement stable mechanism for removing docker images for a specific test after the container is stopped (there is no reason to keep a `gemini` image after we have run the `gemini` test) #3292
7. - [x] Create docker folder hierarchy that would allow for similar grouping to our current dependency hierarchy (`docker/languages`, `docker/frameworks`, etc) #3316
8. - [x] Change `benchmark_config.json` to specify a dockerfile instead of a `setup.sh` #3292
9. - [x] Implement database mechanisms currently handled by `fw_depends` (which is being deprecated and removed with the docker work) #3292
10. - [x] Implement client mechanisms currently handled by `fw_depends` (installing `wrk` on the remote client, etc) #3292
11. - [x] Update vagrant install to call `tfb --init --quiet` to initialize the software on the database and client machine (localhost) #3292
12. - [x] Fix docker build process to split the output to the console as well as the log (see `framework_test.py` for an example of this; this is needed to see output in Travis-CI, for example) #3292
[...]

PR #3301, opened 2018-02-16
Better logging error checking

msmith-techempower:
The suite will now log the docker stuff better (stdout and the log file for the test) and exit when something fails.

Addresses [#14](https://github.com/TechEmpower/FrameworkBenchmarks/issues/3296)

### DIFF
--- a/toolset/benchmark/framework_test.py
+++ b/toolset/benchmark/framework_test.py
@@ -278,6 +278,9 @@ class FrameworkTest:
               tee_output(prefix, line)
           except setup_util.EndOfStream:
             break
+      if p.returncode != 0:
+        tee_output(prefix, "Docker build failed; terminating\n")
+        return 1
       p = subprocess.Popen(["docker", "build", "-f", test_docker_file, "-t", "tfb-test-%s" % self.name, self.directory],
           stdout=subprocess.PIPE,
           stderr=subprocess.STDOUT)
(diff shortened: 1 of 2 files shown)


msmith-techempower:
Hmm, this Travis-CI run didn't pass because it looks like it only built `tfb.dockerfile` before allowing the `gemini.dockerfile` to try and build (read; meaning that neither `java.dockerfile` or `resin.dockerfile` were built).

toolset/benchmark/framework_test.py
```diff
@@ -278,18 +212,24 @@ def tee_output(prefix, line):
...
-          stderr=subprocess.STDOUT)
-      nbsr = setup_util.NonBlockingStreamReader(p.stdout)
-      while (p.poll() is None):
-        for i in xrange(10):
-          try:
-            line = nbsr.readline(0.05)
-            if line:
-              tee_output(prefix, line)
-          except setup_util.EndOfStream:
-            break
```
msmith-techempower: All the above was done inside the loop that builds the dependencies... this should NEVER have worked and yet we got here somehow. 😕

msmith-techempower: I get it - it tries to build the test's dockerfile every time in the loop, and we never checked the success of this until this PR where it returns and exits upon failure. It worked before because it would fail every time until all the dependencies had successfully built. 🤣

PR #3304, opened 2018-02-19
C tests for docker

nbrady-techempower:
Dockerizing the C tests

- [x] duda
- [x] facil.io
- [x] h2o
- [x] libreactor
- [x] octane
- [x] onion


frameworks/C/h2o/h2o.dockerfile
```diff
@@ -0,0 +1,46 @@
...
+
+RUN git clone https://github.com/x86-64/mustache-c.git && \
+    cd mustache-c && \
+    git checkout 55dafd1e95adaca90ea50efb9a8573786514c85a && \
+    CFLAGS="-O3 -flto -march=native" ./configure --prefix=/root/mustache-c && \
+    make -j "$(nproc)" install
+
+ENV LD_LIBRARY_PATH=/root/mustache-c/lib:${LD_LIBRARY_PATH}
+
+### Install yaji
```
volyrique: It's actually YAJL, not YAJI.

nbrady-techempower: Good catch, thanks!

frameworks/C/h2o/h2o.dockerfile
```diff
@@ -0,0 +1,59 @@
...
+
+ENV H2O_VERSION="2.2.4"
+ENV H2O_ARCHIVE="v${H2O_VERSION}.tar.gz"
+ENV H2O_BUILD_DIR="h2o-${H2O_VERSION}"
+
+RUN wget https://github.com/h2o/h2o/archive/${H2O_ARCHIVE} && \
+    tar xvf ${H2O_ARCHIVE} && \
+    cd $H2O_BUILD_DIR && \
+    cmake -DCMAKE_INSTALL_PREFIX="/h2o" -DCMAKE_C_FLAGS="-flto -march=native" \
+          -DCMAKE_AR=/usr/bin/gcc-ar -DCMAKE_RANLIB=/usr/bin/gcc-ranlib -DWITH_MRUBY=on && \
```
volyrique: "-DWITH_MRUBY=on" should not be necessary, and omitting it might decrease the build time a bit. If I understand correctly, h2o_mruby is going to get its own Docker file, so it will be completely independent. Currently this is not the case - there is only one H2O setup script, shared by both the C and the Ruby implementations.

frameworks/C/h2o/h2o.dockerfile
```diff
@@ -0,0 +1,59 @@
+FROM ruby-2.4:latest
```
volyrique: See below.

frameworks/C/h2o/h2o.dockerfile
```diff
@@ -0,0 +1,59 @@
...
+
+COPY ./ ./
+
+RUN ls
+
+RUN apt install -yqq cmake automake libuv1-dev checkinstall autoconf pkg-config libtool python-sphinx libcunit1-dev nettle-dev libyaml-dev
```
volyrique: Why do we need libuv1-dev?

frameworks/C/h2o/h2o.dockerfile
```diff
@@ -0,0 +1,59 @@
...
+
+RUN wget https://github.com/lloyd/yajl/archive/${YAJL_ARCHIVE} && \
+    tar xvf ${YAJL_ARCHIVE} && \
+    cd ${YAJL_BUILD_DIR} && \
+    ./configure -p /yajl && \
+    make -j "$(nproc)" install
+
+ENV LD_LIBRARY_PATH=/yajl/lib:${LD_LIBRARY_PATH}
+
+### Install wslay
```
volyrique: Why is this necessary? Currently the scripts do not require it.

frameworks/C/h2o/setup.sh
```diff
@@ -1,12 +1,12 @@
 #!/bin/bash
 
-fw_depends postgresql h2o mustache-c yajl
+ls
```
volyrique: Is this a temporary debugging thing?

nbrady-techempower: Yes @volyrique this wasn't ready

frameworks/C/onion/benchmark_config.json
```diff
@@ -2,26 +2,7 @@
...
-      "platform": "None",
-      "webserver": "None",
-      "os": "Linux",
-      "database_os": "Linux",
-      "display_name": "onion",
-      "notes": "",
-      "versus": "onion"
-    },
-    "raw": {
-      "setup_file": "setup",
```
msmith-techempower: Would rather have both tests here; what is the reasoning behind combining them?

nbrady-techempower:
@volyrique Feel free to take a look at this at it's current state. It's passing now. I was getting a `wlay` error earlier that led me to that installation, but it was probably because I was missing something else. That's been fixed. As far as things like `-DWITH_MRUBY=on` you can go ahead and update that where you'd like. My goal here is simply to get all the frameworks running in docker containers in ubuntu 16.04, with no intentions of making changes to the original setup beyond making sure it works in this new environment. As far as things being broken out into even more containers, that will eventually be the case yes. Right now for instance, h2o is the only framework that uses `mustache-c` and `yajl` and with single inheritance one would have to extend the other, which is just strange. So we're working towards something better there. Thanks for your input.

volyrique:
@nbrady-techempower Thanks for the explanation. I'll have a look once things are in a more or less stable state. I am looking forward to the results for Ubuntu 16.04!

Issue #3295, opened 2018-02-16
Add the test running in out.txt and terminal


joanhey:
# In out.txt
To know when and where the problems begin.
And the time of dstat, to not look the stats blindly.

### Expected Behavior
```
--------------------------------------------------------------------------------
Started plaintext  concurrency 256
2018/02/06 19:10:00 | 1517943552.195 Dstat time
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
Started plaintext  concurrency 512
2018/02/06 19:10:20 | 1517943552.195 Dstat time
--------------------------------------------------------------------------------
Server php: 2018/02/06 19:10:00 [alert] 32170#0: 1024 worker_connections are not enough
Server php: 2018/02/06 19:10:00 [alert] 32170#0: 1024 worker_connections are not enough
Server php: 2018/02/06 19:10:00 [alert] 32170#0: 1024 worker_connections are not enough
Server php: 2018/02/06 19:10:00 [alert] 32170#0: 1024 worker_connections are not enough
Server php: 2018/02/06 19:10:00 [alert] 32170#0: 1024 worker_connections are not enough
--------------------------------------------------------------------------------
Started plaintext  concurrency 1024
2018/02/06 19:10:30 | 1517943552.195 Dstat time
--------------------------------------------------------------------------------
Server php: 2018/02/06 19:10:00 [alert] 32170#0: 1024 worker_connections are not enough
Server php: 2018/02/06 19:10:00 [alert] 32170#0: 1024 worker_connections are not enough
Server php: 2018/02/06 19:10:00 [alert] 32170#0: 1024 worker_connections are not enough
Server php: 2018/02/06 19:10:00 [alert] 32170#0: 1024 worker_connections are not enough
Server php: 2018/02/06 19:10:00 [alert] 32170#0: 1024 worker_connections are not enough
Server php: 2018/02/06 19:10:00 [alert] 32170#0: 1024 worker_connections are not enough
Server php: 2018/02/06 19:10:00 [alert] 32170#0: 1024 worker_connections are not enough
Server php: 2018/02/06 19:10:00 [alert] 32170#0: 1024 worker_connections are not enough
Server php: 2018/02/06 19:10:00 [alert] 32170#0: 1024 worker_connections are not enough
Server php: 2018/02/06 19:10:00 [alert] 32170#0: 1024 worker_connections are not enough
[...]

PR #3308, opened 2018-02-19
[ci fw-only Kotlin/http4k] Fix JSON structure to stop warnings

zloster:
Addresses a minor point in #3295. And indeed it was annoying :)


nbrady-techempower:
Thanks for this @zloster ! Yeah looks like http4k isn't in the .travis.yml file for some reason. There's a PR open for an upgrade and I'll have it added there. Thanks again for all your contributions!

PR #3311, opened 2018-02-20
Convert the undertow tests to docker

michaelhixson:
The databases don't work for me locally, but I'm told they are supposed to work, so let's see what Travis says.


frameworks/Java/undertow/undertow-base.dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM java:latest
```
RX14: One thing I'd comment on about the implementation of docker in FWBM so far is that I'd much prefer that the images were named `techempower/java` instead of just `java`. There already exists a `java:latest` image in the docker hub, which I can imagine is confusing to implementers. You also might accidentally end up pulling the docker hub java image and changing the build environment.

toolset/setup/linux/docker/java.dockerfile
```diff
@@ -1,12 +1,13 @@
...
+ENV JAVA_HOME=/java/jdk-9.0.4
+ENV PATH="${JAVA_HOME}/bin:${PATH}"
 
-# https://bugs.launchpad.net/ubuntu/+source/ca-certificates-java/+bug/1396760
-RUN /var/lib/dpkg/info/ca-certificates-java.postinst configure
-
-ENV JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
-ENV PATH="/usr/lib/jvm/java-8-openjdk-amd64/bin:${PATH}"
\ No newline at end of file
+RUN mkdir /maven
```
RX14: What about gradle, sbt, ant? Surely these should be seperate docker images.

frameworks/Java/undertow/undertow-base.dockerfile
msmith-techempower: It is on the [roadmap](https://github.com/TechEmpower/FrameworkBenchmarks/issues/3296)

toolset/setup/linux/docker/java.dockerfile
nbrady-techempower: This is just one test. Separate docker images for _most_ things is the plan. You're getting way ahead of us. Long road ahead. :)

frameworks/Java/undertow/undertow-base.dockerfile
RX14: Where?

toolset/setup/linux/docker/java.dockerfile
michaelhixson: I think RX14 is saying I should have created a `maven.dockerfile` that is `FROM java:latest` (or `FROM tfb/java`) instead of cramming Maven in with the base Java installation, and then `undertow-base.dockerfile` would be `FROM maven:latest` instead of `FROM java:latest`.  And that's different from saying I need to add gradle, sbt, and ant installations right now.

It's not clear to me that is better, so I think it can wait.

This single-inheritance system of dependencies we've created using `FROM` is really unsatisfying.  What does the hierarchy look like for a project that needs both Resin and Maven, for example?  How about another project that uses those two tools but needs Java 8 instead of 9?  (These combinations do exist in TFB right now.)

Maybe I should be building the jar in a Maven container, pulling that jar out, then executing it in a base Java container?  I don't know enough about Docker to know if that's possible / makes sense.

nbrady-techempower: I agree @michaelhixson I feel the same way. I've discussed this with @msmith-techempower too, as it was one of the first concerns that jumped out at me. I don't know of a better way other than creating different combinations as the need arises, and that is definitely unsatisfying. Right now, I don't think there are a ton of frameworks that would need this.

I think a compromise could be installing maven, ant, gradle, etc within the java-8 and java-9 dockerfiles. Does it make sense to separate these things out? Does having an image built with gradle on it negatively affect java frameworks that don't use gradle? Other than the first installation time?

michaelhixson: If we keep with the direction of this PR, then yes, the plan would be to install all Java build tools along with the base Java installation.  The code would be copy and pasted in two places, since we'd have Java 8 and Java 9 installers.

By "all Java build tools" maybe I only mean Maven and SBT though.  For Gradle, apparently the convention is that everyone includes their own Gradle executable in their implementation.  I don't see anyone using Ant or Ivy besides Gemini, which already has a working dockerfile.

I don't think the overhead of including Maven and SBT in the base Java dockerfile is worth worrying about.  Both installations consist of downloading a zip (8 MB and 57 MB respectively), unzipping it, and adding something to `PATH`.

PR #3319, opened 2018-02-22
Servlet 3.1 with Async I/O implementation

zloster:
This adds implementation of Servlet 3.1 with Async I/O for `Plaintext` and `JSON` tests.
Includes setup of:

- Tomcat 9.0.5
- OpenSSL 1.0.2n
- Apache APR (Apache Portable Runtime) 1.6.3
- Tomcat native adapter which comes bundled with the given version of Tomcat
- Default `server.xml` with Tomcat `NIO2` connector enabled.

I've played with the Apache Tomcat Native connector and the performance was slightly worse than the NIO2 connector. But my benchmark setup is very limited. The nice thing is that the native connector allows using OpenSSL for the HTTPS traffic. So this could come handy for the upcoming SSL tests.


zloster:
I'll move this source under the `Java/servlet` directory and create another PR.

zloster:
I don't have an idea why [the first installation attempt for Tomcat fails](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/346037581#L2900-L2901) and [the second succeeds](https://api.travis-ci.org/v3/job/346037581/log.txt). Note that the second is a link to the full log because the output is too big for Travis.

zloster:
@msmith-techempower AFAIR the requirement for `&` at the end was relaxed some time ago (see `gemini` and `undertow` scripts for example).

So back to the current problem. Here I have two scripts:

- [servlet3-async](https://github.com/zloster/FrameworkBenchmarks/blob/issue-20/frameworks/Java/servlet/servlet3-async.sh)
- [servlet3-sync](https://github.com/zloster/FrameworkBenchmarks/blob/issue-20/frameworks/Java/servlet/servlet3-sync.sh)

They are exactly the same with [the only difference](https://github.com/zloster/FrameworkBenchmarks/blob/issue-20/frameworks/Java/servlet/servlet3-sync.sh#L6) of the Maven profile in the `servlet3-sync.sh`.
The TravisCI log (https://api.travis-ci.org/v3/job/346037581/log.txt) shows full success for the `servlet-sync-tomcat` (search for this string in the log) benchmark profile/configuration. The log for the `servlet-async-tomcat` is giving [the following error](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/346037581#L2901):
`Setup servlet-async-tomcat: gpg: keyserver receive failed: bad URI`

When I run this in a local Vagrant environment and cleaned user profile directory (i.e. removed the `~.gnupg` directory and `tomcat.installed` file) I don't get this error.

My guess is that this is some transient problem, so could you restart this build or try it locally?

nbrady-techempower:
Restarted

zloster:
@nbrady-techempower Thank you. It is not exactly a transient error but it seems it's quite a corner case - maybe the fresh OS install is helping to hit the problem. Explicitly adding a keyserver URL solves the problem.

Issue #29, opened 2013-03-29
PHP is not a Framework or a Platform


philsturgeon:
This statement is good, but doesn't cover PHP:

> We're using the word "framework" loosely to refer to platforms, micro-frameworks, and full-stack frameworks.

PHP is none of those things, it's a language just like Ruby or Python, and it is not a framework for the same reason Ruby, Python, JavaScript or Java are not frameworks.

PHP is not a platform either, and only becomes one platform when you utilise Apache or other web software such as Nginx. As you are testing with apache list "php" as "apache-php", just like you have listed ruby as "rack-ruby".

Beyond that, PHP 5.4 is the current stable, so use that. Its much faster than PHP 5.3. It would be like using Ruby 1.8.x, which is considerably slower than 1.9.x.

Also, nobody in their right minds is going to put their site live without APC or Zend Optimizer, so enable one of those instead.

Finally, [React](http://reactphp.org/) in there for a NodeJS-style setup. The author of this is very adamant it's not a framework, but since this project is already fragrantly disregarding widely accepted definitions of these terms it fits right in with the other systems you've been benchmarking.

PR #3320, opened 2018-02-22
Implement supporting multiple `FROM` directives

msmith-techempower:
Addresses [#29](https://github.com/TechEmpower/FrameworkBenchmarks/issues/3296)

Also accidentally address [#31](https://github.com/TechEmpower/FrameworkBenchmarks/issues/3296)


frameworks/C/h2o/h2o.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ruby-2.4:latest
+tfb/ruby-2.4:latest
```
RX14: Typo?

msmith-techempower: Yes

RX14: Just making sure you knew :)

Issue #26, opened 2013-03-29
Go - Unfair example


StarpTech:
Go is a concurrent system programming language the strength is its ability to produce scalable and reliable software. In your example you set the MAXPROCESS config high but don´t use gorotines to make it parallel/concurrent that´s like feature cutting and that´s no scenario where the language can show his nature of working.

PR #3321, opened 2018-02-22
Docker Multiple Container Support + Dart start/stream as examples

nbrady-techempower:
The goal here is to allow a `benchmark_config.json` to have an additional parameter called `docker_files` where a user can specify an array of __additional__ dockerfiles from the framework test root that must run in conjunction with the `<test name>.dockerfile`

#### framework_test.py

Added the docker-py module and used the low level api to first build all the dependencies of all of the test docker files. Then it builds all of the test docker files. Once all are built, we use docker-py to run all of the containers and detach them. This is where we used to check that the port was bound or that the docker containers exited. There are a few reasons we can no longer do it this way. However, I believe we've stumbled upon a much easier solution. If the images are already built, the only thing that happens during the run is the final `CMD` line (and in some cases not even that). This should only be the server startup command, or the start of another service (see nginx in the dart/stream example). This means that, with no errors once we start running the containers, the 60 second wait time we normally apply should suffice and no other hackery is needed.

#### benchmarker.py

Right now, just a quick change to `__stop_test` using docker-py. We stop all containers, ignoring the database container, and then remove all images tagged with `tfb/test/*`

#### travis.yml

Removing the deprecated group as tests are now being installed in `ubuntu:16.04` containers.

### Framework tests

- [x] dart
- [x] redstone
- [x] start __(will prove out multi container support)__ passes locally
- [x] stream __(will prove out multi container support)__ passes locally
- [x] urweb


toolset/benchmark/framework_test.py
```diff
@@ -195,139 +196,147 @@ def tee_output(prefix, line):
...
+              stderr=subprocess.STDOUT)
+        nbsr = setup_util.NonBlockingStreamReader(p.stdout,
+          "%s: framework processes have terminated" % self.name)
 
-      if (travis_timeout - datetime.now()).total_seconds() < 0:
-        sys.stdout.write(prefix + 'Printing so Travis-CI does not time out\n')
-        sys.stdout.write(prefix + "Status: Poll: %s, Port %s bound: %s, Time Left: %s\n" % (
-          p.poll(), self.port, self.benchmarker.is_port_bound(self.port), time_remaining))
-        sys.stdout.flush()
+        # Set a limit on total execution time of setup.sh
```
msmith-techempower: There's no such thing as `setup.sh` anymore...

nbrady-techempower: this is git diffing being git diffing, i did not "add" that line

toolset/setup/linux/prerequisites.sh
```diff
@@ -69,6 +69,9 @@ sudo apt-get -qqy install -o Dpkg::Options::="--force-confdef" -o Dpkg::Options:
 sudo groupadd docker
 sudo usermod -aG docker $USER
 
+sudo pip install docker
+pip show docker
+
```
msmith-techempower: lol@alldis

nbrady-techempower: image

toolset/benchmark/benchmarker.py
```diff
@@ -714,22 +709,16 @@ def __stop_database(self, database_container_id, out):
...
-                    subprocess.check_call(["docker", "image", "prune", "-f"])
+        client = docker.from_env()
+        # Stop all the containers
+        for container in client.containers.list():
+            if container.status == "running" and container.id != database_container_id:
+              container.stop()
+        # Remove only the tfb/test images
+        for image in client.images.list():
+            if len(image.tags) > 0 and image.tags[0].startswith("tfb/test/"):
+                client.images.remove(image.id, force=True)
```
msmith-techempower: This block also needs `client.images.prune()` to remove all the unused interim images.

toolset/benchmark/framework_test.py
```diff
@@ -205,13 +205,15 @@ def tee_output(prefix, line):
 
     client = docker.APIClient(base_url='unix://var/run/docker.sock')
 
+    prev_line = "\n"
```
msmith-techempower: `os.linesep` please

toolset/benchmark/framework_test.py
```diff
@@ -205,13 +205,15 @@ def tee_output(prefix, line):
...
     def handle_build_output(line):
-        if line.startswith('{"stream":'):
-          line = json.loads(line)
-          line = line[line.keys()[0]].encode('utf-8')
-          if not line.endswith('\n'):
-            line = line + '\n'
+      if line.startswith('{"stream":'):
+        line = json.loads(line)
+        line = line[line.keys()[0]].encode('utf-8')
+        if prev_line.endswith('\n'):
```
msmith-techempower: `os.linesep` please

frameworks/Dart/start/start-nginx.dockerfile
```diff
@@ -2,4 +2,6 @@ FROM tfb/nginx:latest
 
 COPY ./ ./
 
-CMD ["./start-nginx.sh"]
+RUN ./start-nginx.sh
+
+CMD ["nginx", "-c", "/nginx.conf", "-g", "daemon off;"]
```
msmith-techempower: Nit: I don't think you need the semi-colon

toolset/benchmark/framework_test.py
```diff
@@ -241,7 +239,7 @@ def handle_build_output(line):
 
         # Build the dependency image
         try:
-          for line in client.build(
+          for line in docker.APIClient(base_url='unix://var/run/docker.sock').build(
```
msmith-techempower: Ugh... like... just ugh... windows will never happen at this rate.

frameworks/Dart/start/start-nginx.dockerfile
nbrady-techempower: You need the semi-colon
`Server start: nginx: [emerg] unexpected end of parameter, expecting ";" in command line`

frameworks/Dart/stream/stream-nginx.dockerfile
```diff
@@ -2,6 +2,7 @@ FROM tfb/nginx:latest
 
 COPY ./ ./
 
+RUN chmod a+rwx nginx-conf.sh
```
msmith-techempower: Hey man... don't add. Now we won't know if this line fixes it or all my insanity in the `nginx.conf`

msmith-techempower:
Addresses [#26](https://github.com/TechEmpower/FrameworkBenchmarks/issues/3296) and [#30](https://github.com/TechEmpower/FrameworkBenchmarks/issues/3296)

PR #3323, opened 2018-02-23
Mutliple docker containers for a single test

nbrady-techempower:
The goal here is to allow a `benchmark_config.json` to have an additional parameter called `docker_files` where a user can specify an array of __additional__ dockerfiles from the framework test root that must run in conjunction with the `<test name>.dockerfile`

#### WIP

@msmith-techempower this is currently working _except_ that in the array case all dockerfiles are being run and bound with `self.port` [here](https://github.com/TechEmpower/FrameworkBenchmarks/compare/docker...nbrady-techempower:docker-multi?expand=1#diff-104b04ab89e124380ed9be9c86341479R292). Binding the same port didn't seem to matter in my tests, and I'm wondering if we even need to do that with the `--network` param. My suspicion is we don't need to do it, but I won't be testing it tonight.

### DIFF
--- a/toolset/benchmark/framework_test.py
+++ b/toolset/benchmark/framework_test.py
@@ -196,21 +196,15 @@ class FrameworkTest:
     # Build the Docker images
     ##########################
 
-    # Grab a list of all the docker files we need with
-    # the following priority:
-    # 1) An array of files in docker_files
-    # 2) A single file in docker_files
-    # 3) If none of the above, default to <test name>.dockerfile
+    # Build the test docker file based on the test name
+    # then build any additional docker files specified in the benchmark_config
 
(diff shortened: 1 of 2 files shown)


toolset/benchmark/framework_test.py
```diff
@@ -876,7 +870,7 @@ def __init__(self, name, directory, benchmarker, runTests, args):
     self.display_name = ""
     self.notes = ""
     self.versus = ""
-    self.docker_file = None
+    self.docker_files = None
```
msmith-techempower: My name is Michael Smith, an admitted dreameater, and I support this message.

nbrady-techempower:
Closing this for #3321

PR #3328, opened 2018-02-23
Optimal thread count for aspnetcore plaintext

sebastienros:
The previous logic would result `1` thread on Azure where `2` is actually the one to use.
Now using an explicit value, and also failing in case the environment changes and needs to be updated.

### DIFF
--- a/frameworks/CSharp/aspnetcore/setup-plaintext.sh
+++ b/frameworks/CSharp/aspnetcore/setup-plaintext.sh
@@ -12,8 +12,8 @@ fi
 
 if [ -z "$threadCount" ]
 then
-    echo "Invalid thread count, aborting"
-    exit 1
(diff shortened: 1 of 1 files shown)


msmith-techempower:
I'm cancelling the Travis-CI run as your tests passed (and I want Travis to work on my stuff now). LGTM!

sebastienros:
Thanks, I was wondering if we could just put a flag in the PR comments to trigger specific frameworks when needed.

nbrady-techempower:
@sebastienros We have a crude travis-diffing tool that is _supposed_ to detect where the file change was and what frameworks it _might_ affect. It doesn't always work. Another trick is putting `[ci fw-only CSharp/aspnetcore]` in each commit message to override the default functionality. When the docker branch is done this will change a little and I'll update documentation accordingly.

PR #3331, opened 2018-02-23
I need an explanation for why this is required

msmith-techempower:
I... I just don't know anymore.

mysql just didn't work on ubuntu 16 with docker unless I did this nonsense.


nbrady-techempower:
need this, thanks.

no answers for you.

Issue #3324, opened 2018-02-23
TravisCI - Build matrix exceeds 200 jobs


zloster:
Currently in the `master` branch there are [exactly 200 build jobs](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/344929209) (scroll to the bottom). TravisCI refuses to build the project when new framework is added. So currently adding a new framework will halt builds executions.

Short term mitigation (NOT fix) is some clean up of the framework projects. In the Java directory there are several candidates that could be merged in single framework/build job:

- [x] `comsat-servlet` and `comsat-webactors`
- [x] `grizzly-bm` and `grizzly-jersey`
- [x] `undertow-jersey-c3p0` and `undertow-jersey-hikaricp`

I'll start with the `grizzly`-based projects because I have some knowledge about them. Each merge will free one build job slot.

For a long term solution - I don't know.

cjnething:
Interesting, looks like other projects are having this issue too https://github.com/aschn/drf-tracking/issues/97

joanhey:
And remember that a job on travis-ci.org is limited by 50 minutes.
https://docs.travis-ci.com/user/customizing-the-build/#Build-Timeouts

zloster:
There should not be a problem with the build timeout at least in the Java directory. Check here where a rebuild of the Java frameworks happens: https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/341133956.

The longest builds are:
1. Java/act - ~29 min. - 2 database configurations with 3 ORM frameworks plus MongoDB variant
1. Java/play2-java - ~21 min.
1. Java/dropwizard - ~11 min.
1. Java/undertow - ~10 min.
1. Java/servlet - ~9 min.

nbrady-techempower:
If push comes to shove, we may group some of the lesser worked on languages together in travis, like Erlang and Elixir. This would also speed up full runs because shared dependencies won't have to be reinstalled.

zloster:
Now there are 3 (three) free jobs in the `master` branch and a flexible solution in the `docker` branch. Added bonus: a little less code in the source repository and a little faster full runs.
Thank you all.

@nbrady-techempower Feel free to close this.

nbrady-techempower:
Thanks again for your help here!

PR #3332, opened 2018-02-24
Travis and --test-lang option

nbrady-techempower:
Travis now limits execution to 200 jobs in the matrix. If you go over 200 jobs, travis will not start any of the jobs. Also, travis time limit is 50 mins. So, a potential solution is to group jobs by language. This may not be suitable for all languages as the execution time may be exceeded if there are too many tests. However, I think in a lot of cases this will also speed up runs because of docker caching in a single job. Travis diffing logic would have to change some to avoid full runs, but it will anyway for the travis docker branch.

Total execution time separately for the C languages I've grouped here is just over 60 minutes. Curious to see how long one job would take here.

See #3324 

If we do move forward with something like this, i'd like to see a `--test-lang C` in the toolset as well, and we can do - "TESTLANG=C" in the build jobs for cleaner display.


RX14:
Isn't the startup time of each travis VM the most time consuming factor here (given that almost all builds are partial)? Surely it'd be far faster to have a single travis job and simply execute the few tests that change in order?

Unfortunately that breaks down when you adjust some core infrastructure and have to rebuild everything... Hmm

RX14:
Perhaps the optimal solution is to randomly partition the languages into 4 VMs (given only 4 jobs in the matrix can run in parallel) which gets you max parallelism when everything needs to be redone, but reduces the overhead of (in the common case) starting 26 new VMs just to not use 25 of them.

nbrady-techempower:
@RX14 You cant partition a full run into 4 VMs because you have a max execution time of 50 minutes.

RX14:
@nbrady-techempower oh, that makes sense. Quite a difficult binpacking problem so I guess it's far easier to just be less efficient. This PR is an improvement at least.

nbrady-techempower:
So this worked out well. I added the `--test-lang` option and made travis decipher between `TESTDIR` and `TESTLANG` environment variables. All of the `C` tests ran in about half the time. Doing this for a few languages will free up job spaces and generate quicker full runs. Probably best to do the ones that are worked on the least to start, Elixir, Erlang, etc. I'll have to modify the diffing tool as well so that only the framework tests that changed inside `TESTLANG` run but I think this is 👍 

ping @msmith-techempower

PR #3333, opened 2018-02-25
Merge undertow-jersey-c3p0 and undertow-jersey-hikaricp

zloster:
For details see #3324.
Also the project directory is renamed to `undertow-jersey` and the `benchmark_config.json` is adjusted slightly.

The changes:

- Turned off Plaintext and JSON test for the `hikaricp` variant.
- There are 2 (two) Maven profiles now in the pom.xml. One for the `c3p0` dependency which the default, the other for the `hikaricp` dependency.
- The only addition is the Maven resource filtering for the `hibernate.cfg.xml` file. For each DB connection pool there is a separate `hibernate.cfg.xml`. The `hikaricp` configuration is in sub-folder of the main resource folder and there is a small filtering logic to manage the copy of the configuration file to the output directory. The copy is explicitly allowed for the one file and explicitly disabled for the other.
- Changed DB URL to use `TFB-database`.
- Commented code for runtime manipulation of the DB connection string.


nbrady-techempower:
This looks good @zloster thanks for doing this. Let me know when you're ready here.

zloster:
@nbrady-techempower I'm ready here. The build dependency updates will happen some other time.

PR #3338, opened 2018-02-27
Merge comsat-servlet and comsat-webactors

zloster:
For details see #3324.
Also the project directory is renamed to `comsat` and the `benchmark_config.json` is adjusted slightly.

The changes are:

- Only one `setup-generic.sh` build script.
- All dependencies are common now. My quick attempt to make `gradle` work similarly to the Maven profiles was unsuccessful - I'm able to import dependencies conditionally but the conditional compile was not that easy. I don't know if this will cause problems.


zloster:
`comsat` project [is OK](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/346859335#L1991-L2006).
`undertow-jersey` has failed due to [build timeout](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/346859359#L1623-L1642). There was some problem with the MySQL installation and the installation never finished.

nbrady-techempower:
👍 Thanks @zloster !

PR #3354, opened 2018-03-02
Rust/may-minihttp add may-minihttp framework

Xudong-Huang:
add may-minihttp framework, currently not include database related test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -181,6 +181,7 @@ env:
     # - "TESTDIR=Rust/iron"
     # - "TESTDIR=Rust/nickel"
     # - "TESTDIR=Rust/hyper"
+    # - "TESTDIR=Rust/may-minihttp"
     # - "TESTDIR=Rust/tokio-minihttp"
     # - "TESTDIR=Rust/rouille"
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -181,6 +181,7 @@ env:
     # - "TESTDIR=Rust/iron"
     # - "TESTDIR=Rust/nickel"
     # - "TESTDIR=Rust/hyper"
+    # - "TESTDIR=Rust/may-minihttp"
```
zloster: This line should NOT be commented. There is a revert of some accidental merge of the Docker branch into master (https://github.com/TechEmpower/FrameworkBenchmarks/pull/3357) several commits later.
@Xudong-Huang @nbrady-techempower

nbrady-techempower: This is in #3358 now

Xudong-Huang: Hope someone would fix it, I just see all rust related test is disabled currently, I think there is a reason.

nbrady-techempower: @Xudong-Huang it is fixed in the docker branch. We're moving toward docker for all test implementations.

PR #3359, opened 2018-03-04
Update facil.io version and add docker file

boazsegev:
The main relevant difference in the updated facil.io version is the memory allocator (I wrote a custom memory allocator that should minimize lock contention).

It's possible to test the new version without the multi-threaded memory allocator by using:

```sh
export CFLAGS="${CFLAGS} -DFIO_FORCE_MALLOC"
```

I also added a dockerfile, as I understand that this is where things are headed...

...however, I didn’t know how to test the dockerfile. Any guidance would be appreciated (I especially wonder about the use of `ENV CC`).

### DIFF
--- a/frameworks/C/facil.io/bench_app.c
+++ b/frameworks/C/facil.io/bench_app.c
@@ -1,9 +1,9 @@
 /*
-This is an incomplete example meant to be used as a benchmark application for
-the TechEmpower Framework Benchmarks. See:
+This is a short implementation fo the TechEmpower Framework Benchmarks. See:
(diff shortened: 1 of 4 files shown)


nbrady-techempower:
@boazsegev You'll want to look at the docker branch. As a matter of fact, since all the C frameworks have already been moved to docker, if you could do this update there it would be more helpful. 

The next round will come from the changes on that branch. Testing is done the same way, `tfb --mode verify --test facil.io`

The dockerfile is the name of the test, you'll notice that in that branch. So I'm going to close this here and please re-open to target the docker branch. If you're having trouble just ping me in a new PR and I can help you from there!

PR #3367, opened 2018-03-07
Use pug 2.0.0 in express instead of a release candidate version

michaelhixson:
This fixes the express fortunes tests for me locally.


nbrady-techempower:
this fixes the pug fortune tests, but mitol is still failing. will check that out.

PR #3371, opened 2018-03-07
Port aspnetcore tests to docker

michaelhixson:
I removed the os:Windows tests since we have no plans to support Windows
anytime soon.


sebastienros:
🎉

PR #3373, opened 2018-03-07
Port clojure tests to docker

michaelhixson:


msmith-techempower:
I want to :+1: 3a69eef but I don't think I can do it any other way than this.

PR #3375, opened 2018-03-08
Move all dockering to docker_helper

msmith-techempower:
Just got a little excited and we merged the bigger branch before I finished this bit of the work.


msmith-techempower:
I cancelled the Travis run; this should only require local testing.

PR #3380, opened 2018-03-09
Misc toolset improvements

michaelhixson:
1. Fix the uploads to tfb-status.
2. Get rid of the current_benchmark resume feature because it's been harmful not helpful in the continuous benchmarking environments.
3. Fail faster on frameworks that don't have dockerfiles.

### DIFF
--- a/toolset/benchmark/framework_test.py
+++ b/toolset/benchmark/framework_test.py
@@ -65,11 +65,11 @@ class FrameworkTest:
                 raise Exception(
                     "docker_files in benchmark_config.json must be an array")
 
-        docker_helper.build(self.benchmarker_config, [self.name], out)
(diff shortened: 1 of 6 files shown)


michaelhixson:
For what it's worth, I tested these changes in the Citrine environment and they worked as expected.

toolset/benchmark/framework_test.py
```diff
@@ -65,11 +65,11 @@ def start(self, out):
                 raise Exception(
                     "docker_files in benchmark_config.json must be an array")
 
-        docker_helper.build(self.benchmarker_config, [self.name], out)
+        result = docker_helper.build(self.benchmarker_config, [self.name], out)
+        if result != 0:
+            return result
```
nbrady-techempower: Nice. Fixes failing builds continuing on to test!

PR #3387, opened 2018-03-09
Improve error handling for database:none (as opposed to database:None)

michaelhixson:


toolset/benchmark/benchmarker.py
```diff
@@ -482,6 +482,7 @@ def __run_test(self, test):
                 docker_helper.stop(self.config, database_container_id, test,
                                    out)
             except (OSError, IOError, subprocess.CalledProcessError):
+                traceback.print_exc()
                 self.results.write_intermediate(
                     test.name, "<setup.py> raised an exception")
```
msmith-techempower: `setup.py` has been wrong for a very long time. We should just drop that but leave the string otherwise.

michaelhixson: You're saying it should be this?
```python
                self.results.write_intermediate(
                    test.name, "raised an exception")
```

msmith-techempower: Yeah, or something similar; it's just that `setup.py` is misleading because that's not a thing anymore.

PR #3391, opened 2018-03-10
Java/grizzly-jersey fix

zloster:
It was using the same code as here: https://github.com/TechEmpower/FrameworkBenchmarks/pull/3333/commits/7fbc1c8fed8e2b715c6eb354dddb8f3ddcdabcf2
Changes:

- Use `TFB-server`
- Comment out the search and replace on the JDBC URL

### DIFF
--- a/frameworks/Java/grizzly-bm/grizzly-jersey/src/main/java/hello/SessionFactoryProvider.java
+++ b/frameworks/Java/grizzly-bm/grizzly-jersey/src/main/java/hello/SessionFactoryProvider.java
@@ -22,11 +22,11 @@ public class SessionFactoryProvider
 
   private static SessionFactory createSessionFactory(ResourceConfig rc) {
     Configuration configuration = new Configuration().configure();
-    String url = configuration.getProperty("hibernate.hikari.dataSource.url");
-    url = url.replace(
-        "//localhost:3306/",
-        "//" + rc.getProperty("dbhost") + ":" + rc.getProperty("dbport") + "/");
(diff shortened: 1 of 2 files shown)


nbrady-techempower:
@zloster would you mind targeting the docker branch with this, and with any future changes for Java stuff? that's where all the continuous runs are coming from from here on.

zloster:
@nbrady-techempower Java/grizzly-jersey was migrated to Docker infrastructure very recently but is not updated in the Docker branch with the changes from #3324 (in master).
That's the reason I've created it against the master branch.

nbrady-techempower:
@zloster Ah that makes sense, thanks. I'll ping @michaelhixson since he's been handing the java stuff and see how he wants to proceed. Thanks! :)

michaelhixson:
The change itself looks fine.  However, I agree this should be done against the docker branch.

The changes from #3324 were merged into the docker branch, but I re-separated the grizzly test implementations there afterwards.  (1) With the new `TESTLANG` directive, I think we'll be able to avoid any job limit restrictions.  (2) I would like to avoid squishing implementations together when they don't share any code, especially when they don't use the same framework.  (`grizzly-jersey` probably has more in common with `undertow-jersey` than the no-framework grizzly implementation.)

zloster:
@michaelhixson These considerations should have been added in #3324 - it could save time and effort of each one of us.

michaelhixson:
@zloster Yeah, sorry about that.

PR #3392, opened 2018-03-10
facil.io: use the `FIO_DEDICATED_SYSTEM` implementation (from master)

boazsegev:
I hope this isn’t rude of me, but instead of releasing a facil.io version, I am updating this app to test the effects of the `FIO_DEDICATED_SYSTEM` flag introduced in [the facil.io master branch](https://github.com/boazsegev/facil.io/blob/2fb10a91248eb359f1aaec3f46ecc0c63b043afc/lib/facil/core/facil.c#L192-L204).

facil.io assumes processes are for concurrency and threads are for
parallelism (protection against slow/blocking user code).

However, when deploying facil.io on dedicated systems, this isn’t always the best design choice. Hence, the `FIO_DEDICATED_SYSTEM` adds concurrent thread execution into the mix.

I’m not sure if this flag should be enabled on Citrin, but this should demonstrate this flag’s effects.


nbrady-techempower:
Not rude of you at all! Thanks for doing this on the docker branch!

boazsegev:
Thanks!

I fixed the ENV directive in the docketfile (my attempts at math broke the implementation for some reason).

Following the fix, the C tests pass. However, I will need to find a way to automate the calculation.

nbrady-techempower:
@boazsegev sadly Docker doesn't have a great way for handling environment variables with `ENV` and a calculation, however, in this case since you're doing it on the command line, don't pass the command as JSON. Instead do something like this:

`CMD /facil_app/tmp/demo -p 8080 -db "TFB-database"-w 4 -t $((${CPU_COUNT} * 3))`

We give you the `CPU_COUNT` env variable.

boazsegev:
@nbrady-techempower thanks for the tip 👍🏻

I updated the docker file :-)

I like this mathematical approach. I'm planning to make the it available as part of the framework (i.e., on the next release, `-t -4` will create cores/4 threads per worker)

By the way, the last test didn't fail on my account, something with `D/vibed`.

boazsegev:
The failed test was unrelated to this PR.

This time the JavaScript tests failed and the Java tests timed out... I didn't touch these...
("I didn't do it, nobody saw me do it, no way you can prove anything!" - [B. Simpson](https://youtu.be/WTbgsoHDc24))

PR #3396, opened 2018-03-12
Fix actix build hanging in docker

michaelhixson:


fafhrd91:
i think it is better to add `thread::sleep` before starting `DbExecutor`, here https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Rust/actix/src/main.rs#L136

fafhrd91:
"FATAL: the database system is starting up" looks like timing issue with postgresql startup.

maybe use something like this before start `DbExecutor`?

```rust
while PgConnection::establish(db_url).is_err() {
    thread::sleep(Duration::from_millis(1000)
}
```

michaelhixson:
Thanks, I'll try out your suggestions tomorrow.

The real fix for the "FATAL: the database system is starting up" problems should probably go in our [PostgreSQL installer](https://github.com/TechEmpower/FrameworkBenchmarks/blob/32dd06a4a323f9784e059efd421c45643dde5a45/toolset/setup/docker/databases/postgres/postgres.dockerfile#L51).  Unfortunately I couldn't figure out what that fix was in a reasonable amount of time, and I don't plan to spend any more time on that.

fafhrd91:
now actix is failing on fortune test. it passes in master, did anything changed in docker?

fafhrd91:
why did you disable actix again? it passed test

michaelhixson:
@fafhrd91 Because the test will fail randomly because of a problem in our toolset.

When we create the PostgreSQL database, we create two fortune tables: "fortune" (lowercase "f") and "Fortune" (uppercase "F").  But something goes wrong some of the time and the uppercase Fortune table does not get created.   When I run it locally, this happens about half the time.  I don't want to introduce a new test in Travis that fails half the time.

The reason I merged this PR anyway is that before, because of the way we wrote the actix dockerfiles, these tests would hang and cause a 6 hour delay in every continuous benchmarking run.  (The continuous runner doesn't care about `.travis.yml`.  If it sees a correctly-named dockerfile, it will try to run it.)  I still want to fix the problem in our toolset with PostgreSQL, because actix and other frameworks may sometimes fail the fortunes test until that's fixed, and then I'll re-enable the actix tests in Travis, but I want to fix the 6 hour delay right now.

fafhrd91:
thanks for clarification!

do you want me to change table name in actix code?

michaelhixson:
You can if you want.  You could re-enable actix in .travis.yml too then.

PR #3404, opened 2018-03-13
Sleep more intelligently

msmith-techempower:
1. Wait until the database is accepting connections before returning from `start_database`
2. Wait until the application server is accepting requests before verifying instead of sleeping a configurable amount of time
3. After both above checks, then check that the proper number of containers are running

### DIFF
--- a/toolset/benchmark/benchmarker.py
+++ b/toolset/benchmark/benchmarker.py
@@ -399,16 +399,17 @@ class Benchmarker:
                 slept = 0
                 max_sleep = 60
                 while not test.is_running() and slept < max_sleep:
+                    if not docker_helper.successfully_running_containers(
+                            test.get_docker_files(), database_container_id,
+                            out):
+                        tee_output(
+                            out,
+                            "ERROR: One or more expected docker container exited early"
+                            + os.linesep)
+                        return sys.exit(1)
+
                     time.sleep(1)
                     slept += 1
-
-                if not docker_helper.successfully_running_containers(
-                        test.get_docker_files(), database_container_id, out):
-                    tee_output(
-                        out,
-                        "ERROR: One or more expected docker container exited early"
-                        + os.linesep)
-                    return sys.exit(1)
 
                 # Debug mode blocks execution here until ctrl+c
                 if self.config.mode == "debug":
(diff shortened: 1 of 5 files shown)


toolset/benchmark/benchmarker.py
```diff
@@ -396,20 +396,31 @@ def __run_test(self, test):
...
-                logging.info("Sleeping %s seconds to ensure framework is ready"
-                             % self.config.sleep)
-                time.sleep(self.config.sleep)
+                slept = 0
+                max_sleep = 60
+                while not test.is_running() and slept < max_sleep:
+                    time.sleep(1)
+                    slept += 1
+
+                if not docker_helper.successfully_running_containers(
```
michaelhixson: If this `if not docker_helper.successfully_running_containers` block was moved into the `while not test.is_running()` block, would that give us early failure when a container dies early?

msmith-techempower: Probably; I'll do that.

toolset/utils/docker_helper.py
```diff
@@ -162,6 +164,15 @@ def watch_container(container):
             print(e)
             return 1
 
+    return 0
+
+
+def successfully_running_containers(docker_files, database_container_id, out):
```
michaelhixson:
```python
if (running_container_length != expected_length):
```

I'm wondering about the value of `!=` here as opposed to `<`.  Recent runs in the Citrine environment were failing all frameworks that used a database with errors like this:

```
Running Containers (id, name):
b5a97fbe02, <Image: 'tfb/test/undertow-mysql:latest'>
Excepted 1 running containers; saw 2
ERROR: Problem starting undertow-mysql
```

The theory was that a container was sticking around from a previous test?  If nothing has changed to make that situation less likely, then I think `<` would be better than `!=` if only for the sake of the Citrine environment.

msmith-techempower: Specifically, if we do not have exactly the expected number of running containers then things have gone wrong and we want to error on that.

I expect containers to be properly stopped and that if we're expecting 1 running container, then we will only see 1 running container; if that is not true, we need to address that directly.

PR #3408, opened 2018-03-13
Master into docker march 13

michaelhixson:
Do not squash commits when merging this PR, please.

A plain merge of master into docker no longer works, probably because of the accidental merge of master into docker and the revert afterwards.  So I cherry-picked all these commits instead.  Most contained conflicts that I had to resolve... I think is why it's showing my name next to each commit.


michaelhixson:
Travis says C/octane is failing, but we don't think it's related to the changes in this PR.

PR #3409, opened 2018-03-14

2vg:
add Nim's web framework "mofuw", currently include plaintext


nbrady-techempower:
@2vg Thanks for submitting a new test!

2 Things:

- You must include the test implementation within the repo. Presumably this is the src of `mofuw/tests/techempower`

- Would you mind taking a look at the docker branch, and submitting this test to target that branch? `jester` is already done there so that may help you. Since we don't have the docs ready, if it's hard to see what's going on there, you can submit it back to master and I will convert it. 

Let me know if you need any help! Thanks!

PR #3413, opened 2018-03-14
Fixed a bug on multi-env

msmith-techempower:
Set noninteractive on base docker image
Make `tfb` an "executable" on PATH instead of a function


nbrady-techempower:
If you really want to skip ci, you should probably do `[ci skip]`

PR #3414, opened 2018-03-15
use lower case table names

fafhrd91:


fafhrd91:
this is for docker tests

nbrady-techempower:
Sorry you had to go through that trouble. We should have that fix in shortly, but this works great too. Thanks!

PR #3420, opened 2018-03-16
Prevent fortunes output being accumulated across tests by the verifier

michaelhixson:
It makes sense to me why it was doing what it did - it was accumulating
the fortunes output into a static class-level variable - and it makes
sense why this change fixes that problem.  What doesn't make sense is
why this problem is new.  How did it ever work?

### DIFF
--- a/toolset/benchmark/fortune_html_parser.py
+++ b/toolset/benchmark/fortune_html_parser.py
@@ -4,7 +4,10 @@ from difflib import unified_diff
 
 
 class FortuneHTMLParser(HTMLParser):
-    body = []
+
+    def __init__(self):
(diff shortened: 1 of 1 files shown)


msmith-techempower:
I suspect that the reason it used to work (and forgive me, I’m on my phone so it’s hard to reference the code base) is because the tests were individually run in sub processes that exited on error or completion, so that class was lost after the exit. I changed it recently to not so this (because it made no sense to me anymore) and just missed this weird corner case.

nbrady-techempower:
Dug into a test with mutliple fortunes and they're passing as expected now.

PR #3421, opened 2018-03-17
use latest aiohttp release

fafhrd91:


nbrady-techempower:
Fails the second fortune test but is related to #3420

PR #3422, opened 2018-03-17
Docker ruby tests

nbrady-techempower:
This is all the ruby tests except for `Ruby/h2o_mruby` which I haven't attacked yet and the `Ruby/ngx_mruby` test which was a few years out of date and not working well. I also removed all the `thin` permutations. Those were failing under heavy load, and looking at the [Round 15 results](https://www.techempower.com/benchmarks/#section=data-r15&hw=ph&test=update) they were producing mostly errors.

Edit: Added `h2o_mruby`


nbrady-techempower:
A lot of the Postgres tests are failing due to the second Fortune table issue we're having. Will try and get that resolved here.

PR #3423, opened 2018-03-18

2vg:
add Nim's web framework "mofuw". currently include plaintext tests


2vg:
why error ?
fork and run travis, but no error

2vg:
gave up…

hlaaftana:
Test failures were from an unrelated C++ framework, this should be opened again

PR #3425, opened 2018-03-19
Docker postgres fix

nbrady-techempower:
Stopping the service in the same `RUN` command ensures that the transaction has been written to the data directory


msmith-techempower:
Travis didn't test this, but w/e.

PR #3427, opened 2018-03-19
Bump Finagle benchmarks

vkostyukov:
This PR bumps Finagle-based benchmarks to new versions:

 - Finch was bumped to 0.18
 - Finatra was bumped to 18.3
 - Finagle was bumped to 18.3

It also introduces new JVM options for Finch and Finagle benchmarks (inspired by what Finatra uses) as well as migrates all of the above to the Netty 4 version.

### DIFF
--- a/frameworks/Scala/finagle/build.sbt
+++ b/frameworks/Scala/finagle/build.sbt
@@ -2,11 +2,11 @@ name := "finagle"
 
 scalaVersion := "2.11.12"
 
-version := "17.11.0"
+version := "18.3.0"
 
 com.github.retronym.SbtOneJar.oneJarSettings
 
 libraryDependencies ++= Seq(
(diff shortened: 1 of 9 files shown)


nbrady-techempower:
@vkostyukov since the Scala frameworks are already done, I'd like to put this into the `docker` branch instead of merging this into master. It will reintroduce setup files that we don't want when we merge master into docker. If you can do that, that would be great. If not, I'm just going to leave this open until I can get around to it. (A few days probably) Thanks as always for your contributions!

vkostyukov:
Hey @nbrady-techempower! I'm happy to re-open this PR against the `docker` branch.

vkostyukov:
@nbrady-techempower The new PR (against `docker` branch) is here: #3431

PR #3430, opened 2018-03-20
Print all the lines from build

msmith-techempower:
See title


msmith-techempower:
Closing in favor of #3432

PR #3432, opened 2018-03-20
Docker logging multiple build lines

nbrady-techempower:
So I did a few things here:

 - Instead of using ordered arguments I used `kwargs`. This made everywhere we're doing logging a lot more readable.
 - I combined `header()` into `log()` by allowing a `border` and `border_bottom` option.
 - I added a color option for logging. This makes coloring output a lot easier. It will handle styling before and after, and it will do it after any `prefix`. It also colors the borders.
 - I've highlighted the build steps from the docker build process. This will help out a ton when trudging back through the logs to see which step caused a problem.

I think this makes logging a bit easier and gives us the ability to bake in additional options down the road worrying about ordering args, etc.

### DIFF
--- a/toolset/benchmark/test_types/framework_test_type.py
+++ b/toolset/benchmark/test_types/framework_test_type.py
@@ -8,7 +8,7 @@ import pymongo
 import traceback
 
 from colorama import Fore
-from toolset.utils.output_helper import log, log_error
+from toolset.utils.output_helper import log
 
 
 class FrameworkTestType:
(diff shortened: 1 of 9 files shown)


toolset/benchmark/test_types/framework_test_type.py
```diff
@@ -85,7 +86,7 @@ def request_headers_and_body(self, url):
         Downloads a URL and returns the HTTP response headers
         and body content as a tuple
         '''
-        log("Accessing URL {!s}:".format(url))
+        log("{!s}Accessing URL {!s}:{!s}".format(Fore.CYAN, url, Fore.RESET))
```
msmith-techempower: This doesn't even use `url` anymore...

nbrady-techempower:
image
image

I've also dimmed the prefix text, which will show up better on terminals that handle ansi better

toolset/utils/output_helper.py
```diff
@@ -8,80 +10,50 @@
...
-        result += "%s%s" % (os.linesep, bottomheader)
-    log(result + os.linesep, None, log_file, True, quiet)
-
-
-def log(log_text=None,
-        prefix=None,
-        log_file=None,
-        allow_newlines=False,
-        quiet=False):
+def log(log_text=None, **kwargs):
```
msmith-techempower: Please no `kwargs`. If you're just going to call them out later (`kwargs.get('color','')` for example), then just specify them here.

PR #3428, opened 2018-03-20
add Gotham web framework for Rust

fafhrd91:
Gotham web framework for Rust https://gotham.rs/

PR #3443, opened 2018-03-22
remove Rust/gotham; undo #3428

hobofan:
One of the maintainers of Gotham @bradleybeddoes requested the removal of the test from the benchmark in a [Reddit comment](https://www.reddit.com/r/rust/comments/86a716/rust_is_now_top15_in_techempower_db_query/dw3nfli/).


fafhrd91:
Take down notice :)

bhauer:
Understood, but thanks for the contribution anyway @fafhrd91!  We'll be happy to receive a new implementation when the maintainers are ready.

bradleybeddoes:
@hobofan thank you for submitting this whilst I was unable to get to my laptop, very much appreciated.

@bhauer / @nbrady-techempower thanks for quick response here. Is there any process available to help us ensure this doesn't happen again? We intend to submit to TechEmpower in the future but we'd like the implementation to be created, driven and maintained by our community, not by a third party.

bhauer:
@bradleybeddoes There is no formal process for that, but we'll try to keep it in mind if we see another submission of a Gotham implementation that isn't clearly from the maintainers such as yourself.

Our project as a whole is supportive of third-parties submitting implementations, but we understand when maintainers want to make sure the implementation for their framework accurately represents the framework's capabilities, best practices for its use, recommended deployment strategy, and so on.  So we handle each on a case-by-case basis.

We will look forward to the official implementation submission!

bradleybeddoes:
@bhauer thanks very much for your reply, what you've outlined sounds like a great way forward.

I too look forward to Gotham getting to a maturity point where we can have an official implementation on the TechEmpower benchmarks.

In the interim if you have any queries relating to Gotham please don't hesitate to get in touch with either @smangelsdorf or myself.

PR #3445, opened 2018-03-22
Cutelyst 2 on docker

dantti:


nbrady-techempower:
hi @dantti you just need to uncomment the line in `.travis.yml` so we can see this run :)

PR #3447, opened 2018-03-22
H2O: Adjust the number of database connections

volyrique:
This is the first step to tune the number of database connections, especially for the Citrine environment.


volyrique:
Unfortunately, Travis ran the tests for all C frameworks, and facil.io and Octane failed; H2O passed everything.

nbrady-techempower:
Thanks @volyrique we'll look into those 2 today

PR #3451, opened 2018-03-23
Port a few PHP tests to docker

michaelhixson:
There are both php5 and php7 tests here, so hopefully that makes porting
the rest easier.  I guess the other big category (missing here) is HHVM.

For some reason I had to use "sleep infinity" as the last line of these
bash scripts instead of "wait", which I had used for the same purpose
for other frameworks.  I don't know why that is.

The zend test prints out a lot of errors but it still passes verification
locally.


nbrady-techempower:
adding `-g "daemon off;"` to the nginx line allows you to remove sleep infinity and wait

This also worked locally:

```

michaelhixson:
Thanks, that's much nicer.  I changed all four frameworks' dockerfiles to be like that.

nbrady-techempower:
cakephp is failing, but going to merge this in to continue work on php while we investigate

PR #3452, opened 2018-03-23
ubuntu:18.04

nbrady-techempower:
This isn't meant to be merged anytime soon, but I'm curious how close we are to using ubuntu/bionic as the base when it's fully released. My guess is this will explode.


nbrady-techempower:
Too many packages not ready yet, as expected

PR #3455, opened 2018-03-23
hyper: revert tokio upgrade (while bugs are addressed)

seanmonstar:


fafhrd91:
would you mind sharing what bugs?

seanmonstar:
@fafhrd91 I've seen a few bugs in the threadpool, but that might actually be related to the futures 0.2 integration, uncertain.

Moreso, it's really just about marketing. The benchmarks are faster with the previous reactor implementation, and while the new reactor/threadpool are probably a better choice in real world situations, people look at the benchmarks are decide from there if hyper is fast or not.

I believe with a little bit of work, the new reactor's performance can improve, and I can switch this back over.

PR #3462, opened 2018-03-26
Update tornado

Dith3r:
Update of Tornado and Motor


nbrady-techempower:
Hi @Dith3r  - since tornado is already ported on the docker branch, we'd like to get this update on there instead of master. If you get a chance to target this on the docker branch, that'd be great. If not, I can do it in the next couple of days.

PR #3466, opened 2018-03-27
Limit the size of log files generated by the toolset

michaelhixson:
When a framework generates an enormous log file it's an indication that
that framework has a problem we need to fix.  But the failure mode of
"our continuous benchmarking environment goes down because of a full
disk and needs manual intervention and we can't restart it until we fix
the framework" is not really acceptable.


michaelhixson:
I don't think the travis failures are related to my change.

PR #3473, opened 2018-03-29
Misc cleanup and upgrades in Scala frameworks

michaelhixson:
- Upgrade to Scala 2.12.5
- Upgrade to sbt 1.1.2
- Use sbt assembly plugin instead of onejar plugin, since the onejar
  plugin was preventing upgrades to sbt 1.x
- Delete scruffy, whose github project page no longer exists, it was
  probably abandoned
- Explicitly set working directory in scala framework dockerfiles rather
  than inheriting it from the sbt dockerfiles (which were inheriting it
  from java dockerfiles, so all the work was being done in /java?)
- The main Scala website says to use Java 8, so don't bother having an
  alternative sbt(-not-java8) dockerfile.
- Remove spray because according to its GitHub project page, it was
  abandoned and people should use akka instead.
- Remove s-server because it doesn't have a release that's compatible
  with the newest version of Scala, and the framework hasn't been worked
  on in about a year, and it probably never had many users.  We could
  leave it there working with an old Scala version, but what would happen
  is we'd run into this again down the road next time we want to get
  Scala or Java updated.  The upside to including this framework isn't
  enough to justify that use of time.


frameworks/Scala/play2-scala/play2-scala-anorm-netty.dockerfile
```diff
@@ -1,7 +1,9 @@
 FROM tfb/sbt:latest
 
-COPY ./play2-scala-anorm ./
+ADD ./play2-scala-anorm-netty /play_app
```
nbrady-techempower: I think this is just supposed to be `./play2-scala-anorm`

frameworks/Scala/play2-scala/play2-scala-slick-netty.dockerfile
```diff
@@ -1,7 +1,9 @@
 FROM tfb/sbt:latest
 
-COPY ./play2-scala-slick ./
+ADD ./play2-scala-slick /play_app
+WORKDIR /play_app
 
 RUN sbt stage
 
-CMD target/universal/stage/bin/play2-scala-slick -Dplay.server.provider=play.core.server.NettyServerProvider
+CMD target/universal/stage/bin/play2-scala-slick
```
nbrady-techempower: just missing the `\` here at eol

michaelhixson: Thanks for debugging.  I admit I got impatient and didn't actually run the play2 tests locally.

PR #3474, opened 2018-03-29
remove $CPU_COUNT

nbrady-techempower:
This removes the use of `$CPU_COUNT` and replaces it with `$(nproc)`. This also attempts to move all commands that rely on that number to `CMD` such that pre-built images will use a configuration that matches the current host environment.


frameworks/Dart/stream/stream.dockerfile
```diff
@@ -14,6 +14,6 @@ RUN pub upgrade
 
 RUN chmod -R 777 /stream
 
-RUN ./nginx-conf.sh
-
-CMD  ./start-servers.sh && sleep 20 && nginx -c /stream/nginx.conf -g "daemon off;"
+CMD ./nginx-conf.sh && \
+    ./start-servers.sh && \
+    sleep 20 && nginx -c /start/nginx.conf -g "daemon off;"
```
msmith-techempower: This should probably be `/stream/nginx.conf` right?

nbrady-techempower:
dart/stream passes with last commit. there seems to be a connectivity issue with travis atm but the other changed tests pass in the previous run.

PR #3481, opened 2018-04-03
Attempt to fix Lua hostname issues

nbrady-techempower:


msmith-techempower:
[Something](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/361445202#L533) is still trying to connect to `tfb-database` instead of the resolved IP.

PR #3488, opened 2018-04-03
Add support for SwiftNIO (default tests)

0xpablo:
Hello!
This PR introduces SwiftNIO plaintext and JSON tests. I also updated the swift.dockerfile to use the stable release of Swift 4.1 (needed by the way for vapor 3). So I imagine you will have to publish a new swift image to the docker hub.
I also updated swift.sh that I guess is still used for vagrant.

Finally, I would like to tag @Lukasa (a SwiftNIO contributor) to see if himself or someone on the team is able to review the [implementation](https://github.com/TechEmpower/FrameworkBenchmarks/compare/docker...0xpablo:docker?expand=1#diff-126c870aaeb792fb0bbd1ac0d834430c) in case there are some tuning or performance optimizations that I will very happily address.
I tried to follow the guidelines and you will see that the implementation is similar to [netty's benchmark](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/netty/src/main/java/hello/HelloServerHandler.java).

Thank you all in advance!

### DIFF
--- a/frameworks/Swift/swift-nio/Sources/swift-nio-tfb-default/main.swift
+++ b/frameworks/Swift/swift-nio/Sources/swift-nio-tfb-default/main.swift
@@ -1,0 +1,126 @@
+
+import Foundation
+import NIO
+import NIOHTTP1
+
+struct JSONTestResponse: Encodable {
+    let message = "Hello, World!"
+}
+
+enum Constants {
+    static let httpVersion = HTTPVersion(major: 1, minor: 1)
+    static let serverName = "SwiftNIO"
+
+    static let plainTextResponse: StaticString = "Hello, World!"
+    static let plainTextResponseLength = plainTextResponse.count
+    static let plainTextResponseLengthString = String(plainTextResponseLength)
+
+    static let jsonResponseLength = try! JSONEncoder().encode(JSONTestResponse()).count
+    static let jsonResponseLengthString = String(jsonResponseLength)
+}
+
+private final class HTTPHandler: ChannelInboundHandler {
+
+    public typealias InboundIn = HTTPServerRequestPart
+    public typealias OutboundOut = HTTPServerResponsePart
+
+    let dateFormatter = RFC1123DateFormatter()
+    let jsonEncoder = JSONEncoder()
+
+    var plaintextBuffer: ByteBuffer!
+    var jsonBuffer: ByteBuffer!
+
+    init() {
+        let allocator = ByteBufferAllocator()
+
+        plaintextBuffer = allocator.buffer(capacity: Constants.plainTextResponseLength)
+        plaintextBuffer.write(staticString: Constants.plainTextResponse)
+
+        jsonBuffer = allocator.buffer(capacity: Constants.jsonResponseLength)
+    }
+
+    func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
+        let reqPart = self.unwrapInboundIn(data)
+
+        switch reqPart {
+        case .head(let request):
+            switch request.uri {
+            case "/plaintext":
+                processPlaintext(ctx: ctx)
+            case "/json":
+                processJSON(ctx: ctx)
+            default:
+                _ = ctx.close()
+            }
+        case .body:
+            break
+        case .end:
+            _ = ctx.write(self.wrapOutboundOut(.end(nil)))
+        }
+    }
+
+    func channelReadComplete(ctx: ChannelHandlerContext) {
+        ctx.flush()
+        ctx.fireChannelReadComplete()
+    }
+
+    private func processPlaintext(ctx: ChannelHandlerContext) {
+        let responseHead = plainTextResponseHead(contentLength: Constants.plainTextResponseLengthString)
+        ctx.write(self.wrapOutboundOut(.head(responseHead)), promise: nil)
+        ctx.write(self.wrapOutboundOut(.body(.byteBuffer(plaintextBuffer))), promise: nil)
+    }
+
+    private func processJSON(ctx: ChannelHandlerContext) {
+        let responseHead = jsonResponseHead(contentLength: Constants.jsonResponseLengthString)
+        ctx.write(self.wrapOutboundOut(.head(responseHead)), promise: nil)
+
+        let responseData = try! jsonEncoder.encode(JSONTestResponse())
+        jsonBuffer.clear()
+        jsonBuffer.write(bytes: responseData)
+        ctx.write(self.wrapOutboundOut(.body(.byteBuffer(jsonBuffer))), promise: nil)
+    }
+
+    private func jsonResponseHead(contentLength: String) -> HTTPResponseHead {
+        return responseHead(contentType: "application/json", contentLength: contentLength)
+    }
+
+    private func plainTextResponseHead(contentLength: String) -> HTTPResponseHead {
+        return responseHead(contentType: "text/plain", contentLength: contentLength)
+    }
+
+    private func responseHead(contentType: String, contentLength: String) -> HTTPResponseHead {
+        var headers = HTTPHeaders()
+        headers.replaceOrAdd(name: "content-type", value: contentType)
+        headers.replaceOrAdd(name: "content-length", value: contentLength)
+        headers.replaceOrAdd(name: "server", value: Constants.serverName)
+        headers.replaceOrAdd(name: "date", value: dateFormatter.getDate())
+
+        return HTTPResponseHead(version: Constants.httpVersion,
+                                status: .ok,
+                                headers: headers)
+    }
+}
+
+let group = MultiThreadedEventLoopGroup(numThreads: System.coreCount)
+let bootstrap = ServerBootstrap(group: group)
+    .serverChannelOption(ChannelOptions.backlog, value: 8192)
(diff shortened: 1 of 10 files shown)


frameworks/Swift/swift-nio/Sources/swift-nio-tfb-default/main.swift
```diff
@@ -0,0 +1,114 @@
...
+
+	private func processJSON(ctx: ChannelHandlerContext) {
+		let responseHead = jsonResponseHead(contentLength: Constants.jsonResponseLengthString)
+		ctx.write(self.wrapOutboundOut(.head(responseHead)), promise: nil)
+
+		var bodyBuffer = ctx.channel.allocator.buffer(capacity: Constants.jsonResponseLength)
+		let responseData = try! jsonEncoder.encode(JSONTestResponse())
+		bodyBuffer.write(bytes: responseData)
+		ctx.write(self.wrapOutboundOut(.body(.byteBuffer(bodyBuffer))), promise: nil)
+	}
```
Lukasa: Depends on whether you consider this "cheating", but given that both the plaintext and JSON responses are static, they could be computed at `handlerAdded` time and stored in byte buffers on the handler. That will avoid many repeated allocations (in calls to `allocator.buffer`) and also avoid repeated copies.

Lukasa: To be clear, I mean this:

```swift
var cachedJSONResponse: ByteBuffer?
var cachedPlaintextResponse: ByteBuffer?

func handlerAdded(ctx: ChannelHandlerContext) {
    self.cachedJSONResponse = ctx.channel.allocator.buffer(capacity: Constants.jsonResponseLength)
    self.cachedJSONResponse.write(bytes: try! jsonEncoder.encode(JSONTestResponse()))
    self.cachedPlaintextResponse = ctx.channel.allocator.buffer(capacity: Constants.plainTextResponseLength)
    self.cachedPlaintextResponse.write(staticString: Constants.plainTextResponse)
}
```

then using those two `self` properties directly.

frameworks/Swift/swift-nio/Sources/swift-nio-tfb-default/main.swift
```diff
@@ -0,0 +1,114 @@
...
+	private func responseHead(contentType: String, contentLength: String) -> HTTPResponseHead {
+		var headers = HTTPHeaders()
+		headers.replaceOrAdd(name: "content-type", value: contentType)
+		headers.replaceOrAdd(name: "content-length", value: contentLength)
+		headers.replaceOrAdd(name: "server", value: Constants.serverName)
+		headers.replaceOrAdd(name: "date", value: dateFormatter.getDate())
+
+		return HTTPResponseHead(version: Constants.httpVersion,
+								status: .ok,
+								headers: headers)
```
Lukasa: The indentation here is a bit weird.

frameworks/Swift/swift-nio/Sources/swift-nio-tfb-default/main.swift
```diff
@@ -0,0 +1,114 @@
...
+				processPlaintext(ctx: ctx)
+			case "/json":
+				processJSON(ctx: ctx)
+			default:
+				_ = ctx.close()
+			}
+		case .body:
+			break
+		case .end:
+			_ = ctx.writeAndFlush(self.wrapOutboundOut(.end(nil)), promise: nil)
```
Lukasa: You'll perform better if you just call `write` here, and instead call `flush` in `channelReadComplete`. That will shrink the number of flushes you do, improving performance.

frameworks/Swift/swift-nio/Sources/swift-nio-tfb-default/main.swift
```diff
@@ -0,0 +1,114 @@
...
+
+		return HTTPResponseHead(version: Constants.httpVersion,
+								status: .ok,
+								headers: headers)
+	}
+}
+
+let group = MultiThreadedEventLoopGroup(numThreads: System.coreCount)
+let bootstrap = ServerBootstrap(group: group)
+	.serverChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEPORT), value: 1)
```
Lukasa: I don't think you're getting anything from setting `SO_REUSEPORT` here: you only ever bind this socket once.

frameworks/Swift/swift-nio/Sources/swift-nio-tfb-default/main.swift
```diff
@@ -0,0 +1,114 @@
...
+	.serverChannelOption(ChannelOptions.backlog, value: 8192)
+	.serverChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)
+
+	.childChannelInitializer { channel in
+		channel.pipeline.add(handler: BackPressureHandler()).then {
+			channel.pipeline.configureHTTPServerPipeline().then {
```
Lukasa: Given that this is a pipelining-aware setup, you *could* remove some of the safeguards here. In particular, you could call `configureHTTPServerPipeline(withPipeliningAssistance: false)`, which would get you a nice speedup. The `BackPressureHandler` is also probably strictly optional.

frameworks/Swift/swift-nio/Sources/swift-nio-tfb-default/main.swift
Lukasa: Ah, just realised the JSON one at least should live-compute the JSON. You could still improve performance by just keeping a scratch buffer lying around and writing the JSON into that, but that's up to you.

0xpablo:
Thanks a lot for your tips @Lukasa and @normanmaurer ! 
I updated the PR with the optimizations that you have pointed. I hope everything looks better now 🙂

I also benchmarked every group of optimizations individually to see the impact of each one of them, and then I did a final benchmark with all of them combined.
I generally saw improvements in the tests so that is great 👍 although my environment is not great (I run the tests on a Mac using docker), so we will see more decent benchmarks after the branch is integrated and this runs on CI (I read somewhere it takes about 90hrs to do a full run). 

In case you want to take a look to the benchmarks, here is [this zip](https://github.com/TechEmpower/FrameworkBenchmarks/files/1876375/Benchmarks.zip) with them (you'll see the baseline, pipeline optimization, buffer optimization, flush optimization and combined optimizations). Two files per benchmark are interesting: `results.json` and `raw.txt`, where you will be able to see info like this:
```
Running 5s test @ http://tfb-server:8080/plaintext
  8 threads and 8 connections
    Latency   311.56us  374.05us  23.90ms   98.00%
    Req/Sec     3.16k   611.31     5.15k    83.25%
  Latency Distribution
     50%  275.00us
     75%  349.00us
     90%  457.00us
     99%    0.88ms
  127710 requests in 5.10s, 16.20MB read
Requests/sec:  25041.21
Transfer/sec:      3.18MB
```

If you see something I messed up in the updated review please let me know. Thank you again for your time!

frameworks/Swift/swift-nio/benchmark_config.json
```diff
@@ -0,0 +1,24 @@
...
+  "framework": "swift-nio",
+  "tests": [{
+    "default": {
+      "json_url": "/json",
+      "plaintext_url": "/plaintext",
+      "port": 8080,
+      "approach": "Realistic",
+      "classification": "Platform",
+      "database": "None",
+      "framework": "SwiftNIO",
```
msmith-techempower: The code doesn't look like it is using a framework, to me (see examples like Rails where a framework library sits between the language built-ins and the application code).

I think this should be `"None"`

frameworks/Swift/swift-nio/benchmark_config.json
```diff
@@ -0,0 +1,24 @@
...
+      "language": "Swift",
+      "flavor": "None",
+      "orm": "Raw",
+      "platform": "SwiftNIO",
```
msmith-techempower: Similarly, it feels to me like we are using language built-ins rather than a platform for development (see examples like `Gemini` where we use the `Servlet` platform which sits between the language and application code, providing an interface for serving requests that the framework must implement).

I think this should be `"None"`

frameworks/Swift/swift-nio/benchmark_config.json
0xpablo: You are right @msmith-techempower ! Thank you. I used the same config as the netty benchmarks, so maybe those need to get fixed as well. I have updated the PR.

0xpablo:
Hello @msmith-techempower, I reviewed your changes, changed the base branch to master as I noticed you merged docker into master (congrats btw 🎉) and rebased. If you need further changes let me know.
Regarding uploading the updated Swift docker image, I don't know if I missed something, maybe `swift-nio.dockerfile` needs a version bump after that, but that needs to be done after uploading the new Swift image (due to CI)

Thanks!

msmith-techempower:
@0xpablo Hi! Thanks for the grats.

I do not believe you need to do anything else, but for coverage, would you mind adding some whitespace or something to your `README.md` and push to trigger a Travis-CI run?

0xpablo:
@msmith-techempower done!

PR #3489, opened 2018-04-04
[ci skip] Try fixing a couple of issues in continuous benchmarking

michaelhixson:
I think the tfb.wrk image needs to be republished in order for the `query.sh` fix to do anything.  Not sure how to test that one in the Citrine environment.

I think the `docker_helper.py` fix should be testable.  I'm pointing Citrine at my branch (where this PR is from) and running it now.  Will remove the "Do Not Merge" tag if it looks good.

### DIFF
--- a/toolset/utils/docker_helper.py
+++ b/toolset/utils/docker_helper.py
@@ -279,6 +279,13 @@ def stop(benchmarker_config=None,
     client.containers.prune()
     client.networks.prune()
     client.volumes.prune()
+    if benchmarker_config.server_docker_host != benchmarker_config.database_docker_host:
+        db_client = docker.DockerClient(
+            base_url=benchmarker_config.database_docker_host)
+        db_client.images.prune()
+        db_client.containers.prune()
+        db_client.networks.prune()
+        db_client.volumes.prune()
 
 
 def find(path, pattern):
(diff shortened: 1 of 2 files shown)


michaelhixson:
@msmith-techempower this fixes those "tfb-database" conflict errors we were seeing earlier today.  We're still getting no RPS numbers from the Query and Updates tests, but I'm hopeful that publishing tfb.wrk with this new `query.sh` would resolve that.

toolset/utils/docker_helper.py
```diff
@@ -279,6 +279,13 @@ def stop(benchmarker_config=None,
     client.containers.prune()
     client.networks.prune()
     client.volumes.prune()
+    if benchmarker_config.server_docker_host != benchmarker_config.database_docker_host:
+        db_client = docker.DockerClient(
+            base_url=benchmarker_config.database_docker_host)
+        db_client.images.prune()
+        db_client.containers.prune()
+        db_client.networks.prune()
+        db_client.volumes.prune()
```
msmith-techempower: Another thing I'm curious to test is line 277 above this change. Instead of opening `db_client`, I just pass the container originally created from a `db_client` down to here, and then call `database_container.stop()`. I wonder if it is clever enough to use the original client rather than one freshly created. It may be best to just err on the side of caution and move those three lines (including the comment) into this block, and instead call `db_client.containers.stop(database_container.id)`

msmith-techempower:
> Will remove the "Do Not Merge" tag if it looks good.

Ah okay, well must be good. I'll merge and republish the `wrk` image.

PR #3492, opened 2018-04-04
Put the docker branch into the master branch

michaelhixson:
This is what I did:  https://stackoverflow.com/a/2862938/359008


nbrady-techempower:
Note to self: don't squash

msmith-techempower:
Addresses most of #3296

PR #3494, opened 2018-04-05
pulling official images from dockerhub

nbrady-techempower:
This allows us to begin the process of moving away from `techempower/` dependencies, but for now gives us both.

This also makes sure that during the `--build` only process that non-techmpower images that don't exist are pulled in for building.

I've updated all the node frameworks to pull directly from the [official node repo](https://hub.docker.com/r/_/node/)

I would also suggest that for now we limit this to official images only. More info about official images here: https://docs.docker.com/docker-hub/official_repos/

Edit: Removed the node mitol permutations as they don't work with current version of node. Also added the `[ci lang-only` option


toolset/utils/docker_helper.py
```diff
@@ -509,8 +509,8 @@ def __build_dependencies(benchmarker_config,
             log_prefix = dep + ": "
             pulled = False
 
-            # Do not pull images if we are building specifically
-            if not benchmarker_config.build:
+            # Do not pull techempower/ images if we are building specifically
+            if not benchmarker_config.build and not dep.startswith('techempower/'):
```
msmith-techempower: This reads so oddly; I would prefer `and 'techempower/' not in dep`

PR #3500, opened 2018-04-06
Update facil.io version

boazsegev:
Please update the facil.io version.

The 0.6.2 facil.io version includes many fixes and changes, including
security concerns for the shutdown / hot-restart feature and and issue
where large response data might not get sent to the clients (causing
connections to hang on larger payloads).

The fixes might effect performance (for better or worst).

Thanks!

P.S.

Congratulations on a major milestone in the docker transition 👍🏻


msmith-techempower:
Thanks on the congrats! We are still iterating it, but it should be getting better and better over the coming days.

PR #3515, opened 2018-04-08

2vg:
add: Nim's web framework "mofuw".
I pray that no problems will occur this time lol.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -114,6 +114,7 @@ env:
      - "TESTDIR=Lua/octopus"
      - "TESTDIR=Lua/openresty"
      - "TESTDIR=Nim/jester"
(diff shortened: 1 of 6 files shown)


2vg:
I do not know the cause, but I pass the first test, but the same problem as before is happening the second time.
Do not merge this pull request yet as it builds in debug mode to investigate the problem.

2vg:
... I understood the cause
The time update function for the Date header is executed for after 1 second the server is started, but the problem occurred because the benchmark was executed earlier than updating the first time at time.
I fixed it.

2vg:
yeee, all ok. It can be merged !!!

nbrady-techempower:
Awesome!

PR #3516, opened 2018-04-08
Quality of life improvements for docker

RX14:
Adds a script to automatically set up the docker network and call the docker container. The sigterm issue is solved by using `exec` in the bash script to replace the shell process with the docker process, instead of spawning it underneath. No solution is provided for windows yet, except for using vagrant.

I couldn't reproduce the sigterm issues on my laptop, it'd be useful if you could test this script to verify i've solved the problem. I've also simplified the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,36 +22,20 @@ To get started developing you'll need to install [docker](https://docs.docker.co
 
         $ git clone https://github.com/TechEmpower/FrameworkBenchmarks.git
 
-2. Create the TFB Docker virtual network
+2. Run a test.
 
-        $ docker network create tfb
+        $ ./tfb --mode verify --test gemini
 
-3. Run a test.
+### Explanation of the `./tfb` script
 
-        $ docker run -it --network=tfb -v /var/run/docker.sock:/var/run/docker.sock --mount type=bind,source=[ABS PATH TO THIS DIR],target=/FrameworkBenchmarks techempower/tfb --mode verify --test gemini
-
-### Explanation of the run script
-
-That run script is pretty wordy, but each and every flag is required. Unfortunately, because of the way that Docker runs processes, you **cannot** put this inside of a shell script without breaking how `ctrl+c` and `SIGTERM` work (the shell script would receive the signal, do nothing with the underlying python suite running, and exit, orphaning the toolset to continue running).
+The run script is pretty wordy, but each and every flag is required. If you are using windows, either adapt the docker command at the end of the `./tfb` shell script (replacing `${SCRIPT_ROOT}` with `/c/path/to/FrameworkBenchmarks`), or use vagrant.
 
 - `-it` tells docker to run this in 'interactive' mode and simulate a TTY, so that `ctrl+c` is propagated.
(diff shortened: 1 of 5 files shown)


nbrady-techempower:
@RX14 I pasted this into my vagrant box and set the alias and all works fine. The only thing you'd have to do with vagrant is get the `tfb` file onto the box (or just create is in `boostrap.sh`) Currently we link the toolset and frameworks directories and create the results directory here https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/deployment/vagrant/core.rb#L58-L60

RX14:
Why not sync the whole repo?

README.md
```diff
@@ -28,36 +28,9 @@ Git on Windows will, by default, automatically convert line endings from `lf` to
...
-You may not want to call step 4 from above every time. You can add an `alias` to your `~/.bash_aliases` file to shorten it since it will not change once configured:
-
-`$ alias tfb="docker run -it --network=tfb -v /var/run/docker.sock:/var/run/docker.sock --mount type=bind,source=[ABS PATH TO THIS DIR],target=/FrameworkBenchmarks techempower/tfb"`
-
-`$ source ~/.bash_aliases`
-
-Now you can run the toolset via `tfb`:
-
-`$ tfb --mode verify --test gemini`
+        $ ./tfb --mode verify --test gemini
```
msmith-techempower: I am fine with this change, but I would like to keep the explanation of the `docker run` command.

README.md
```diff
@@ -28,36 +28,9 @@ Git on Windows will, by default, automatically convert line endings from `lf` to
...
 See [this writeup](https://help.github.com/articles/dealing-with-line-endings/) for more information.
 
-2. Create the TFB Docker virtual network
+2. Run a test.
 
-        $ docker network create tfb
-
-3. Run a test.
-
-        $ docker run -it --network=tfb -v /var/run/docker.sock:/var/run/docker.sock --mount type=bind,source=[ABS PATH TO THIS DIR],target=/FrameworkBenchmarks techempower/tfb --mode verify --test gemini
```
msmith-techempower: Also, since this doesn't work on Windows at all, we need to keep the `docker run` command in the instructions.

README.md
RX14: OK, shall we paste the explanation of the docker run command into the `tfb` script?

nbrady-techempower:
@RX14 there were legacy reasons why we we're syncing the whole repo. It's probably fine now. Give it a shot

README.md
RX14: OK, makes sense. I'll be less heavy-handed on the readme :)

The opther option of course is to always direct windows users at vagrant.

msmith-techempower:
The legacy reason was "$IROOT and syncing doesn't play well with others," but as that it no longer a thing, the vagrant setup can sync everything, I believe.

RX14:
Updated. Deleting `.gitattributes` in this PR is because it was impossible to use git in any fashion without this commit. See https://stackoverflow.com/questions/21122094/git-line-endings-cant-stash-reset-and-now-cant-rebase-over-spurious-line-en.

PR #3525, opened 2018-04-10
[ci skip] Avoid hiding an IOError that is probably happening in Citrine

michaelhixson:


toolset/utils/results_helper.py
```diff
@@ -175,6 +176,7 @@ def parse_test(self, framework_test, test_type):
 
             return results
         except IOError:
+            traceback.print_exc()
```
msmith-techempower: I don't know the difference, but we call `traceback.format_exc()` everywhere; let's stick with that.

michaelhixson: That returns a string instead of printing it anywhere.

msmith-techempower: Yeah, I meant we should log it:
```
tb = traceback.format_exc()
log("ERROR: We don't know how we got here, but it was an error [TODO: fix me]")
log(tb)
```

michaelhixson: Thinking about this more, how about I remove the `try` and `except IOError`, and let this function throw?  Its callers aren't prepared to handle the `None` return value anyway, so why not let them deal with an exception.

msmith-techempower: Are the callers prepared to handle an exception?

michaelhixson: In a sense, yes.  The IOError would get logged instead of the error resulting from using `None`.

PR #3528, opened 2018-04-10
Create .gitattributes

### LINKED ISSUE #3527: Use .gitattributes instead of requiring autocrlf=false
Per #3504, it's necessary to specify `git config --global core.autocrlf false` on Windows. This is a global setting that's the opposite of GitHub's recommendations.

It would be better to use `.gitattributes` to set this per-repo, as per these docs: https://help.github.com/articles/dealing-with-line-endings/#per-repository-settings

This would then work out of the box on all operating systems without requiring anyone to change global git config.

jongalloway:
Set autocrlf off so cloning repo works on Windows. 

Fixes #3527

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text eol=lf


RX14:
See https://stackoverflow.com/questions/21122094/git-line-endings-cant-stash-reset-and-now-cant-rebase-over-spurious-line-en

RX14:
The ideal way to fix this would be to add *only* the filetypes which are affected by line-endings to `.gitattributes`, not `*`. Then run `git rm -r . --cached -q && git reset --hard` inside the repo.

Issue #3478, opened 2018-03-30
I think the config "pm.max_children = 1024" would affect the correctness of the PHP's related test results.


eechen:
If the test server's CPU has 40 cores, 
I think the "max_children" should adjust to 40 or 60.

In other words, the number of PHP-FPM processes , 
should be one and a half times as more as CPU cores.

Because it can reduce CPU context switch significantly in TechEmpower's tests.
Just like Go's "runtime.GOMAXPROCS(runtime.NumCPU())".

pm = static (TechEmpower's test use "dynamic")
pm.max_children = 60 (TechEmpower's test use "1024")

Thanks.

joanhey:
Hi @eechen 
I changed the pm.max_children in #3291 where the changes are explained.

In this case was exactly because PHP-FPM was sending this error:
`WARNING: [pool www] seems busy (you may need to increase pm.start_servers, or pm.min/max_spare_servers)...`

It's only in the plain text benchmark where the max concurrency is 16,384.
But for the other tests, the benchmark changed the concurrency from 256 to 512. So I changed the `pm.start_servers = 256` to 512 and the rest according to this change.

I think you're confusing **Master process** and **Child processes**. Like nginx, nodjes, go, ... use.
For example nginx, they recommend one or one and half Master process(worker_processes) per CPU.
And by default they have 512 maximum child processes (worker_connections) by master process.
The worker_connections is more dependent of your application: memory, cpu or IO. If your application need 10 Mb per child, you need to calculate to not use more than your server memory.

The theory is good but I prefer the real data, and more with great performance servers. I can only test with my 8 cores PC.
But unfortunately :sob: I have not seen the results yet from pr #3265 and #3291 in the continuous benchmarking for adjusting better the php stack. But the change to use Docker is finishing and We will soon see the results.

eechen:
In PHP's tests, the interprocess communication of Nginx and PHP-FPM consumes significant CPU.
You can look this test, PHP runs with Swoole(PECL), without Nginx, is faster than Go's HTTP server.
[](url)https://github.com/swoole/swoole-src/issues/1401
Because Swoole this PECL has a built-in HTTP server like Node.js.

eechen:
I also do the test JSON serialization on my laptop (Ubuntu 14.04 with i5-3230M).
```
<?php
$server = new swoole_http_server('0.0.0.0', 8080);
$server->set(array(
	'worker_num' => 4, // i5-3230M with 4 cores
));
[...]

PR #3535, opened 2018-04-10
PHP/* Remove TE base dependencies

nbrady-techempower:
Removed the TE dependencies and added a swoole test per #3478 

I'm going to leave any optimizing to others

ping @joanhey @eechen


frameworks/PHP/phalcon/phalcon-micro.dockerfile
```diff
@@ -18,7 +18,7 @@ ENV PATH /composer:${PATH}
 ADD ./ /phalcon
 WORKDIR /phalcon
 
-RUN apt install -yqq php7.2-phalcon php7.2-dev
+RUN apt install -yqq wget php7.2-phalcon php7.2-dev
```
joanhey: Here it's installing php7.2-phalcon.
And the next lines download and build again phalcon

joanhey:
When you merge this PR I will optimize.
Now the php docker doesn't exist.
Best way to change php.ini and php-fpm.conf with docker ?
Copy the files for every framework ?
Another sugestion ?

nbrady-techempower:
@joanhey sadly that seems to be the best way atm.. i guess it will give us a chance to test different php.ini optimizations with different frameworks but for now i dont see a better option

PR #3536, opened 2018-04-10
[ci fw-only Java/gemini Java/servlet3] gemini, servlet3: remove TE deps

michaelhixson:


msmith-techempower:
@nbrady-techempower I was typing a review note when you merged this.

`servlet3` does not appear in `.travis.yml`

nbrady-techempower:
ah good catch

PR #3548, opened 2018-04-12
Add Redkale Framework

redkale:

### DIFF
--- a/frameworks/Java/redkale/pom.xml
+++ b/frameworks/Java/redkale/pom.xml
@@ -83,9 +83,6 @@
 								  <mainClass>${main.class}</mainClass>
 								</transformer>
 							  </transformers>
-							<artifactSet>
-							</artifactSet>
-							<outputFile>${project.build.directory}/${project.artifactId}-${project.version}.jar</outputFile>
 						</configuration>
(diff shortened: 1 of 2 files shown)


michaelhixson:
We recently changed this project to use Docker, so each framework needs dockerfiles instead of those `setup.sh` scripts.

Try adding two new files to `frameworks/Java/redkale` named `redkale.dockerfile` and `redkale-postgres.dockerfile`.  Put the following text in each one:

```
FROM maven:3.5.3-jdk-8
WORKDIR /redkale
COPY src src
COPY conf conf
COPY pom.xml pom.xml
RUN mvn package -q
CMD ["java", "-DAPP_HOME=./", "-jar", "target/redkale-benchmark-0.0.1.jar"]
```

I didn't test that so it might not be exactly right, but it should be close.

michaelhixson:
You'll also need to add another line to our `.travis.yml` to tell it about your framework, otherwise travis won't automatically test it.  Throw it in there after rapidoid:

```
     - "TESTDIR=Java/rapidoid"
     - "TESTDIR=Java/redkale"
     - "TESTDIR=Java/restexpress"
```

PR #3557, opened 2018-04-12
Remove C++/luna per maintainer's recommendation

nbrady-techempower:


nbrady-techempower:
@DEGoodmanWilson Awesome! Thanks for the quick response!

PR #3561, opened 2018-04-13
Add Dart Aqueduct framework benchmark

indiealexh:
Add the Dart [Aqueduct](https://pub.dartlang.org/packages/aqueduct) framework to the TechEmpower Benchmark


indiealexh:
Travis failure happened in code I have not edited.

frameworks/Dart/aqueduct/server.dart
```diff
@@ -0,0 +1,227 @@
...
+
+const int _min_query_count = 1;
+const int _max_query_count = 500;
+
+Future main() async {
+  try {
+    var app = new Application<DartAqueductBenchmarkSink>();
+    var config = new ApplicationConfiguration()
+      ..port = 8080
+      ..configurationFilePath = "config.yaml";
```
michaelhixson: It looks like there is a `config.yaml` and a `postgresql.yaml` with the same contents.  Is only `config.yaml` used?

indiealexh: Oh. Yes, only the config.yaml is used. I'll correct this.

frameworks/Dart/aqueduct/aqueduct.dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM google/dart:1.24
+
+COPY ./ ./
+
+RUN pub upgrade
```
michaelhixson: I see that our other Dart tests have this `pub upgrade` line as well, but I'm curious of your opinion - Is this something we should be doing?  Is it possible for this to upgrade a dependency such that it pulls in breaking changes?

indiealexh: As long as the Developer properly constains their pubspec.yaml then pub upgrade should only apply bug fix versions and not breaking changes.

A fresh pub get would do virtually the same thing without a pub.lock existing.

michaelhixson: Got it, thanks.

maurobotta:
Which version of Aqueduct is using in last 2020 test Round 19 ?

https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Dart/aqueduct

official last:
https://github.com/stablekernel/aqueduct    is 3.2.0 ( 4.0 in beta )

PR #3569, opened 2018-04-13
[ci fw-only C++/cutelyst] remove base dockerfile in cutelyst

michaelhixson:
I'm testing the theory that our toolset and docker will be smart enough to re-use the shared parts of these dockerfiles.  If the cutelyst build times out in travis, then the answer is no.  If the build takes about 11 minutes (which I think was the build time before this change), then the answer is yes.


nbrady-techempower:
So. Good.

dantti:
Isn't this worse due the duplication of commands?

msmith-techempower:
"Worse" in terms of running every test, but better when we get to leverage the docker cache from run to run.

dantti:
how will it behave if I change a pkg to install in one test and the others aren't equal

msmith-techempower:
That one test will be updated and the others will not be. It will be on contributors first, then us on QA'ing pull requests to ensure that those sorts of situations don't slip through.

michaelhixson:
@dantti We felt the downside of copy and pasting was worth the upside of having "vanilla" dockerfiles that don't require special interpretation/parsing by our toolset.  (Recognizing these `*-base` dependencies and building them was behavior we had to add to our toolset.)

An even better solution in cutelyst's case might be to pull the shared installation steps out into an `install.sh` file.  I'm not volunteering to do that myself, though.

dantti:
and what about a bash.sh too? Does docker recognizes if I change the install.sh file?

msmith-techempower:
Docker will recognize changes made to files upon which the image relies. If you have a `script.sh` and update a single line, Docker will rebuild from the line that relies upon that script to make a new container image.

PR #3574, opened 2018-04-15
Toolset cleanup & remove TE database dependencies

nbrady-techempower:
We no longer have any tests that use local dependencies. Everything pulls from a dockerhub dependency.

  - Remove `__build_dependencies`
  - Remove `__gather_dependencies`
  - Remove build arguments (`TFB_DATABASE, MAX_CONCURRENCY`)
  - Move `wrk` and `database` up and remove `toolset/setup/docker` and `toolset/setup`

  - Remove the techempower database dependencies and build the database images locally. Right now this will only minimally affect travis runs especially since the 3 database images are pulling from `ubuntu:16.04` instead of the official database images (some of the official images had some quirks I haven't worked out yet.) For continuous benchmarking, this will have virtually no time impact.

    - We'll be able to see some of the time differences between the `toolset cleanup` commit and the `build database dockerfiles locally` commit

 - Yes I hoorayed my own PR

 - Grouped more tests together with `TESTLANG` from #3575 bringing down the total time impact of building the database tests locally

Edit: JavaScript, which is using all 3 databases, went from 13:52 to 15:10. And this is with using the `ubuntu:16.04` base and doing `apt updates`. I'd say this is a win, especially when we can pull from the official bases for the database.


nbrady-techempower:
@msmith-techempower FYI, [aspnet-mono](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/366910193) has been failing but it's not being reported by travis as failed. It looks like `__run_test` is returning `False` when this happens but nothing is being done with the return value.

toolset/utils/docker_helper.py
```diff
@@ -51,21 +50,12 @@ def build(benchmarker_config, test_names, build_log_dir=os.devnull):
     Builds the dependency chain as well as the test implementation docker images
     for the given tests.
     '''
-    docker_buildargs = {
-        'MAX_CONCURRENCY': str(max(benchmarker_config.concurrency_levels)),
```
msmith-techempower: Is this no longer needed because we're just leveraging `nproc` now?

toolset/utils/docker_helper.py
```diff
@@ -81,7 +71,6 @@ def build(benchmarker_config, test_names, build_log_dir=os.devnull):
                             dockerfile=test_docker_file,
                             tag="techempower/tfb.test.%s" %
                             test_docker_file.replace(".dockerfile", ""),
-                            buildargs=docker_buildargs,
                             forcerm=True):
```
msmith-techempower: Do we want `forcerm` anymore?

toolset/run-tests.py
```diff
@@ -235,7 +235,9 @@ def main(argv=None):
 
     else:
         benchmarker = Benchmarker(config, results)
-        benchmarker.run()
+        all_passed = benchmarker.run()
+        if config.mode == "verify":
+            return 0 if all_passed else 1
```
msmith-techempower: Why not just return `all_passed`?

toolset/benchmark/benchmarker.py
```diff
@@ -39,14 +39,16 @@ def run(self):
         # Get a list of all known  tests that we can run.
         all_tests = gather_remaining_tests(self.config, self.results)
 
+        all_passed = True
```
msmith-techempower: Ugh, change this to something like `any_failed` and reverse the binary logic so that you can just bubble the return value up. That line of `return 0 if all_passed else 1` is so gross.

msmith-techempower:
```
finch: [info] Packaging /finch/target/scala-2.12/techempower-benchmarks-finch-assembly-0.19.0.jar ...
finch: [info] Done packaging.
...
finch: Error: Unable to access jarfile target/scala-2.12/techempower-benchmarks-finch-assembly-1.0.jar
```

michaelhixson:
Regarding the scala travis failures:  I think the `CMD` lines of those dockerfiles need to be updated to align with the version number changes in this commit https://github.com/TechEmpower/FrameworkBenchmarks/commit/2bb5629eb45bd5767537ef822a4a53ce4385dc53

nbrady-techempower:
I'll open a separate PR for the scala tests

PR #3576, opened 2018-04-15
Add wizzardo-http framework

wizzardo:


frameworks/Java/wizzardo-http/wizzardo-http-base.dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM techempower/gradle-java8:0.1
```
michaelhixson: Please use `FROM gradle:4.6.0-jdk8` instead, assuming that works.

Very recently we rewrote all our dockerfiles to avoid `techempower/*` dependencies and to favor "official" or widely-used dependencies from docker hub instead.  We realized that we don't want the responsibility of publishing and maintaining our own set of artifacts if we can avoid it.

Here's another Java framework that depends on gradle as an example:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/d018a0df30e6e3ed92d390915dd61c5fabd4f1b9/frameworks/Java/jawn/jawn.dockerfile

frameworks/Java/wizzardo-http/wizzardo-http-postgres.dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM techempower/wizzardo-http-base:0.1
```
michaelhixson: You'll have to replace this line with the full contents of `wizzardo-http-base.dockerfile`.  We no longer support these relative path base dockerfiles.

The ability to find and extend these framework-specific base files was something we had to build into the TFB toolset ourselves.  As far as we know, it's not supported by docker out of the box.  We became uncomfortable with the idea that our dockerfiles (at least the ones that used this feature) were not "real" dockerfiles that could be trivially built with a single `docker build` command -- it was almost as if we introduced a new keyword or syntax.

Without these framework-specific base files we have a lot more copy-and-pasted code (between test variants in each framework), but we think that is worth the upside of having vanilla dockerfiles.  Fortunately docker's built-in caching is able to see when builds share copy-and-pasted steps and it doesn't need to repeat them, and our toolset now takes advantage of that caching (by not deleting build artifacts between tests like it used to).

wizzardo:
sorry, I used wrong framework as example, fixed

michaelhixson:
Thanks!  It's interesting that one of your framework's responses in the updates test in travis included `{"id":5271,"randomNumber":10001}` since the highest possible random number is supposed to be 10000.  Actually, I don't see how it's possible that your code produced that response.  Any ideas?  I'm referring to line 1524 here: https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/367401712

wizzardo:
Interesting.. I checked in my postgres (9.6.6), `(random()*10000 + 1) :: integer = 10001` sometimes returns `true`, so the issue is in `toolset/databases/postgres/create-postgres.sql`

michaelhixson:
Oh I see.  I misread the logs and thought the 10001 came from the updates test, where your Java code is producing all the random numbers, so I was really confused.  But it came from the queries test.  So I need to go fix `create-postgres.sql`.  Thanks for looking into that.

PR #3585, opened 2018-04-17
check dockerhub for latest during build

nbrady-techempower:


msmith-techempower:
Hard to tell if the `aspnetcore` issues stem from your change to using the `pull` flag. They all claimed to have started, but then timed out before responding to the client.

sebastienros:
Interesting, since yesterday afternoon the images started running on port 80 instead of the expected 8080 on my side, but using `master`, I will look at the logs here.

sebastienros:
Oh, and apparently we released 2.0.7 around the same time, which concerns me as this could mean there is a regression ... it's very improbable but I will still triple check

PR #3590, opened 2018-04-17
Fix randomnumber in postgres

nbrady-techempower:
postgres's `random()` is inclusive, so we were creating `randomnumber`s with value of `10001`

https://www.postgresql.org/docs/9.5/static/functions-math.html

### DIFF
--- a/toolset/databases/postgres/create-postgres.sql
+++ b/toolset/databases/postgres/create-postgres.sql
@@ -8,7 +8,7 @@ CREATE TABLE  World (
 GRANT SELECT, UPDATE ON World to benchmarkdbuser;
 
 INSERT INTO World (id, randomnumber)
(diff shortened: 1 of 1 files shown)


nbrady-techempower:
Postgres tests have passed, skipping the full run to get other PRs tested

michaelhixson:
I think the issue was not that `random()` is inclusive of the upper endpoint, but that we were rounding the result to the nearest integer rather than ignoring the fraction part.  As the code is your initial change, `random() * 10000`, I believe it will be half as likely to produce 10000 as any other number.  That doesn't actually matter, but could you use `floor(random() * 10000 + 1)` like [this tutorial](http://www.postgresqltutorial.com/postgresql-random-range/) suggests just so we don't perpetuate slightly wrong random number generation? :)

PR #3591, opened 2018-04-17
Fixing command line parameters for aspnetcore

sebastienros:
`urls` is the documented way to configure the ports. `server-urls` should not have worked since 2.0.


sebastienros:
Waiting for the CI to pass, then we ensure it fixes #3585

PR #3596, opened 2018-04-17
[ci skip] Fix wrk's Host and Accept headers

michaelhixson:
Assuming this change looks good, we'll need to publish a new `tfb.wrk`
image.

The variable substitution for these header values wasn't happening
because those parts of the script were single quoted.  So we were
sending literally "Accept: $accept" as a header, for example.  I noticed
this because dropwizard was responding with 100% errors when benchmarked
with wrk.  Turns out that every response was "406 Not Acceptable" because
of the malformed Accept header.

I tested this locally by building the `tfb.wrk` image like this:

```
docker build -t techempower/tfb.wrk -f toolset/wrk/wrk.dockerfile toolset/wrk
```


msmith-techempower:
[Pushed to Dockerhub](https://hub.docker.com/r/techempower/tfb.wrk/tags/)

PR #3600, opened 2018-04-18
add redkale source

redkale:


msmith-techempower:
This is unnecessary. Having a publish jar on maven is sufficient; we only want your test implementation code here.

PR #3601, opened 2018-04-18
add redkale source

redkale:


msmith-techempower:
Repeat of #3600. Closing

PR #3602, opened 2018-04-18
add redkale source

redkale:


msmith-techempower:
Do **not** open another PR like this.

PR #3609, opened 2018-04-19
Build wrk and only once at the start of all tests

nbrady-techempower:
This does a couple of things:

 - this is a fork of #3605 so includes the changes there 
    - Remove duplicate build code
    - Properly build display errors and exit appropriately
 - this allows us to make changes to `wrk` locally without pushing to dockerhub and allowing to easily troubleshoot/iterate
 - Instead of attempting to pull `wrk` between every test request, this builds `tfb.wrk` once and then just runs it during the `test_client_connection` calls
 - pulled from `buildpack-deps:xenial` and wrk build time is much faster

Again, this won't have any affect on citrine or local dev after the first test.

Edit: closing #3605

### DIFF
--- a/toolset/utils/docker_helper.py
+++ b/toolset/utils/docker_helper.py
@@ -79,7 +79,10 @@ def build(benchmarker_config, test_names, build_log_dir=os.devnull):
                     if token.startswith('{"stream":'):
                         token = json.loads(token)
                         token = token[token.keys()[0]].encode('utf-8')
-                    buffer += token
+                        buffer += token
+                    elif token.startswith('{"errorDetail":'):
+                        token = json.loads(token)
+                        raise Exception(token['errorDetail']['message'])
                     while "\n" in buffer:
                         index = buffer.index("\n")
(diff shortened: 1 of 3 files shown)


toolset/utils/docker_helper.py
```diff
@@ -44,6 +44,55 @@ def clean(benchmarker_config):
                 client.images.remove(image.id, force=True)
     client.images.prune()
 
+def __build(base_url, path, build_log_file, log_prefix, dockerfile, tag):
+    '''
+    Builds the dependency chain as well as the test implementation docker images
+    for the given tests.
+    '''
```
msmith-techempower: There is no such thing as a 'dependency chain' anymore.

nbrady-techempower: Ah yeah, that was a copy/pasta comment.

msmith-techempower: I figured. Change it with a `[ci skip]` and you got my approval.

PR #3610, opened 2018-04-19
add redkale source

redkale:


nbrady-techempower:
@redkale Please don't open Pull requests like this. We don't want source in the repo. 

请不要这样做。我们不希望存储库中的源代码。

PR #3614, opened 2018-04-19
Disable docker logging

msmith-techempower:
We use stdout/stderr for logging and we don't want Docker filling up the disk with logging about which we do not care.

### DIFF
--- a/toolset/utils/docker_helper.py
+++ b/toolset/utils/docker_helper.py
@@ -44,6 +44,7 @@ def clean(benchmarker_config):
                 client.images.remove(image.id, force=True)
     client.images.prune()
 
+
 def __build(base_url, path, build_log_file, log_prefix, dockerfile, tag):
     '''
     Builds docker containers using docker-py low-level api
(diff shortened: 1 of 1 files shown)


RX14:
Can you even get the stdout/stderr without a log driver? I'm pretty sure you can't if you put it in the background.

msmith-techempower:
I tested this locally and still got the stdout/stderr. This has to do with `docker run`'s logging external to those two, and it was filling up our disks on prod. Hoping this resolves.

RX14:
If the containers get deleted, the logs get deleted. So looks like you're leaking containers somewhere.

RX14:
But if it works, that's great. I'm just surprised.

I guess the python driver has different terminology to the commandline. If I use `--log-driver none` on the commandline, I can only `docker attach`, not run `docker logs`. I guess the former does the streaming thing and the latter doesn't.

msmith-techempower:
> I guess the python driver has different terminology to the commandline. If I use --log-driver none on the commandline

Correct, and this update uses the sdk to get that same result.

> If the containers get deleted, the logs get deleted. So looks like you're leaking containers somewhere.

We are intentionally leaving images (and there's a good chance the containers remain after they are stopped) after being benchmarked so that we can leverage the Docker cache from run to run.

The hope is we can continue that, but simply tell the Docker daemon to not log anything.

RX14:
> We are intentionally leaving images (and there's a good chance the containers remain after they are stopped) after being benchmarked so that we can leverage the Docker cache from run to run.

Leaving images around is great, leaving containers around is always the wrong thing. You should probably use the equivalent to the `--rm` commandline option.

msmith-techempower:
See #3618

PR #3617, opened 2018-04-19
Remove pulling wrk from travis script

nbrady-techempower:
Pulling from the techempower repo for wrk was still happening in the travis setup.


nbrady-techempower:
This fixes the error from travis after the wrk repo was deleted, getting this in fast

PR #3622, opened 2018-04-19
Modernize the spring test implementation

michaelhixson:


rcollette:
I really wonder why this wasn't done as a new test, so that there was both a full orm and a micro orm representation for spring, similar aspcore-mvc-ef and aspcore-mvc-dap.

michaelhixson:
What motivated this change was me reading comments online that amounted to, "The Spring code is so out of date that the results are useless."  I noticed the version of Spring we were using was a couple of years old and a major version behind, so I figured it would be best to start from scratch, and this is the code I came up with.  The change from full orm to micro orm was incidental; I probably didn't even notice the old code was using full orm.

I'm not opposed to adding a second variant of this that uses a full orm.  I'm also not opposed to replacing this with a full orm approach, if people who use Spring mostly favor the full orm.

PR #3629, opened 2018-04-20
Log container build times

nbrady-techempower:
image


nbrady-techempower:
more comprehensive logging coming in place of this

PR #3630, opened 2018-04-20
TimeLogger class and other toolset changes

nbrady-techempower:
### benchmarker.py

  - Move signal stop process here for quicker exit
    - ctrl-c won't try and stop docker containers if we haven't tried starting any yet
  - Move __run_list_metadata to metadata.py
  - Remove port binding check
    - This should never be an issue between the docker init process and forcefully killing docker containers

### benchmakr_config.py

  - Add more directory roots to the config instead of joining everywhere in code

### run-tests.py

  - No longer need for global
  - While making some dev efforts with the toolset, if a fatal error occurred I had to manually shutdown the docker containers. This tries one last time to stop the docker containers if an error bubbles all the way up
  - Removed `--build` flag
  
### docker_helper.py

  - Initialize docker_helper in the benchmarker application
    - Avoids re-initializing connections in every method
  - More forcefully stop docker containers and prevent "no container" errors if already killed
  - fixed redundant search for containers to stop on same machine set up
  - Prune client, server, and database on stop
  - Remove gather_tests call / only sending one test to build at a time

### metadata.py

  - take advantage of gather_languages in gather_tests removing duplicate code
  - Add a couple of additional helper methods
  - created a static var to easily update supported databases
  - Allow benchmark_config values that would be "None" to be missing
  - Added a new schema key `required_with` for keys dependent on other keys
    - We were forcing orm to be specified even if database is None, etc
	- See JavaScript/express for working example

### framework_test_type.py

  - There was an awkward self.out setup that wasn't used by anything but fortune diffing. Removed this and used the response body for diffing.
  
### scaffolding.py

  - Added database option back in using supported_dbs from Metadata
  - Skip orm prompt if database is none
  
### time_logger.py

  - This will log the time it takes for a test to build, verify, and benchmark as well as keep running totals to be displayed after each test. An example of a full bechmark running `tfb --test nodejs nodejs-mongodb` with cached images in vagrant:

image

### audit.py

  Started a simple auditing tool `tfb --audit` that we can use to verify consistency across all frameworks.
  
  - For now this just checks for a `source_code` and `README.md` file but we can extend it to do more complicated things for individual languages (like maybe parse java dockerfiles for jvm options) and add a travis test for it to error on any warnings
  - Now includes an audit for each framework before the verification summary
  
image
  
### Removed Unused Files

  - ordered_set.py
  - unbuffered.py

### Bug fixes

  - [x] when a default test has the wrong key, the test name `default` is displayed without a hint to which benchmark_config
        `benchmark_config.json for test default is invalid, please amend by adding the following required keys: [orm]`

### toolset changes while this is open
  
  - [x] #3633 - catch more errors
  - [x] #3631 - Avoid waiting on a container that died
  
  
### TODO

  - [x] Do we care about --build anymore? (I decided we do not. A full verify run accomplishes the same thing now that we don't remove test containers. And after logging times, verifying is minimal.)
  - [x] Test Scaffolding
  - [x] Remove all duplicate config calls
  - [x] Refactor `gather_tests` - no more "old, hacky" comments
  - [ ] Add toolset tests for travis
  - [x] Allow `orm` to be `None` when `database` is `None`
  - [x] Allow `None` key/value pairs to be left out of `benchmakr_config.json`


toolset/utils/docker_helper.py
```diff
@@ -10,98 +9,104 @@
...
+
+    def __init__(self, benchmarker = None):
+        self.benchmarker = benchmarker
+
+        self.client = docker.DockerClient(
+            base_url=self.benchmarker.config.client_docker_host)
+        self.server = docker.DockerClient(
+            base_url=self.benchmarker.config.server_docker_host)
+        self.database = docker.DockerClient(
+            base_url=self.benchmarker.config.database_docker_host)
```
msmith-techempower: I have a **slight** concern here that the client may close a connection if idle for too long. May not come up in practice, but maybe dig into the API to see if there is a timeout (or something similar).

toolset/utils/docker_helper.py
```diff
@@ -10,98 +9,104 @@
...
+                if image_tag != 'techempower/tfb' and 'techempower' in image_tag:
+                    self.server.images.remove(image.id, force=True)
+        self.server.images.prune()
+
+        self.database.images.prune()
+        for image in self.database.images.list():
+            if len(image.tags) > 0:
+                # 'techempower/tfb.test.gemini:0.1' -> 'techempower/tfb.test.gemini'
+                image_tag = image.tags[0].split(':')[0]
+                if image_tag != 'techempower/tfb':
```
msmith-techempower: I think that this still needs to check for the `'techempower' in image_tag` in the case of single-machine setups (otherwise, we may be removing images on the host machine unrelated to the project, and that should be avoided.

toolset/utils/docker_helper.py
```diff
@@ -124,272 +129,238 @@ def build(benchmarker_config, test_names, build_log_dir=os.devnull):
...
+                    file=run_log)
+                log(tb, prefix=log_prefix, file=run_log)
 
-        sysctl = {'net.core.somaxconn': 65535}
+        return container
+
+    @staticmethod
+    def kill(container):
+        try:
+            container.kill()
```
msmith-techempower: Why is `container.kill` instead of `container.stop`?

toolset/utils/docker_helper.py
nbrady-techempower: I'm not sure this actually opens the connection until you start sending requests to the daemon

toolset/utils/docker_helper.py
nbrady-techempower: stop sends `SIGTERM` to the container and then after a grace period sends `SIGKILL`, this _always_ leads to a long wait on ctrl-c and errors because it always ends up sending `SIGKILL`. `container.kill` sends `SIGKILL` immediately. `ctrl-c` works instantly now.

msmith-techempower: Okay, that's fine for `ctrl-c`, but the normal stop should still be `stop`, right?

toolset/utils/docker_helper.py
nbrady-techempower: Agreed. I hadn't yet figured out why we weren't doing it [here](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/utils/docker_helper.py#L43) but will update.

toolset/utils/benchmark_config.py
```diff
@@ -74,8 +74,13 @@ def __init__(self, args):
...
 
-        # Remember root directory
-        self.fwroot = os.getenv('FWROOT')
+        # Remember directories
+        self.fw_root = os.getenv('FWROOT')
+        self.db_root = os.path.join(self.fw_root, "toolset", "databases")
+        self.lang_root = os.path.join(self.fw_root, "frameworks")
+        self.results_root = os.path.join(self.fw_root, "results")
+        self.wrk_root = os.path.join(self.fw_root, "toolset", "wrk")
+        self.scaffold_root = os.path.join(self.fw_root, "toolset", "scaffolding")
```
msmith-techempower: Okay, honestly liking this more and more.

msmith-techempower:
I am going to merge this in; let's make a new issue and pull request for adding toolset tests.

PR #3631, opened 2018-04-20
[ci skip] Avoid waiting on a container that died

michaelhixson:
This is happening for me locally with mofuw.  Its CMD line fails:

```
/bin/sh: 1: ./start-servers.sh: not found
```

and then the toolset waits around for a long time for it to be running.
With the change, it detects that it has failed quickly.

### DIFF
--- a/toolset/benchmark/benchmarker.py
+++ b/toolset/benchmark/benchmarker.py
@@ -158,6 +158,8 @@ class Benchmarker:
             max_sleep = 60
             accepting_requests = False
             while not accepting_requests and slept < max_sleep:
+                if not docker_helper.server_container_exists(self.config, container.id):
+                    break
                 accepting_requests = test.is_accepting_requests()
                 time.sleep(1)
                 slept += 1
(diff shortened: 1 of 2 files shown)


nbrady-techempower:
There's a potential issue here that I found last week which was causing all of those start/stop errors on citrine. We are sometimes losing a race where we tell the docker daemon to start a container in detached mode and immediately check its availability.

michaelhixson:
@msmith-techempower Did you see Nate's comment?  Do you remember what that original fix for that race was?

msmith-techempower:
@michaelhixson Yeah, it was #3574 

Basically, we found rare instances where the call to start the container daemonized would return control to the toolset and we would query the docker sdk for the running container before it was actually running. This check is unnecessary now and it was removed.

michaelhixson:
Ok, so just to be clear, I'm not reintroducing the race that Nate fixed.  That's because our original check (which Nate removed) used `client.containers.list()`, which only returns **running** containers.  My new check uses `client.containers.get(id_or_name)`, which will return the container even if it isn't running yet.

msmith-techempower:
Ah yes, this does appear to reintroduce a similar mechanism, but with your safeguard.

PR #3632, opened 2018-04-20
Remove unnecessary bash script in mofuw setup

michaelhixson:


frameworks/Nim/mofuw/mofuw.dockerfile
```diff
@@ -14,17 +14,16 @@ RUN apt update -yqq &&                                                  \
...
-COPY ./ /app
-
-RUN chmod a+wrx start-servers.sh
-
-CMD ./start-servers.sh
\ No newline at end of file
+WORKDIR /mofuw_app
+COPY techempower.nim techempower.nim
+RUN nim c -d:release --threads:on techempower.nim
+RUN find /
```
msmith-techempower: Surely, this line is unnecessary.

frameworks/Nim/mofuw/mofuw.dockerfile
```diff
@@ -14,17 +14,15 @@ RUN apt update -yqq &&                                                  \
     ./koch nimble &&                                                    \
     ./koch tools
 
-ENV PATH=$PATH:/nim/nim-devel/bin:/root/.nimble/bin
+ENV PATH $PATH:/nim/nim-devel/bin:/root/.nimble/bin
 
+WORKDIR /mofu_framework
```
2vg: Is this WOKDIR necessary?

michaelhixson: As far as I know, there's no functional difference.  The intent is to make it clear where the framework is being installed, and to show that where/how it's installed does not depend on a working directory inherited from earlier steps in the dockerfile.

PR #3637, opened 2018-04-22
Try a less vebose apt and composer

joanhey:
Now that the tests are joined by language, it's too verbose for check the fails in travis.
I will try in the php dir, and check the result.

apt  STDOUT  > /dev/null
STDERROR will still show

Composer --quiet
It will only show errors.

Later, perhaps will be better to build the docker containers quietly.


joanhey:
In travis-ci

Before:
Line 10000: phalcon-micro (installing from apt)

Now:
Line 10000: workerman (finishing the tests)

joanhey:
Now:
Line 10000: yii2 (installing)

We can check the raw log in travis. But it is **almost impossible to read**.

joanhey:
The travis build seems blocked.
Pass the tests in my fork:
https://travis-ci.org/joanhey/FrameworkBenchmarks

PR #3639, opened 2018-04-23
new: Swift/kitura

djones6:
This is a [Kitura](https://github.com/IBM-Swift/Kitura)-based implementation, running on Swift 4.1, with variants for Postgres and Mongodb.


djones6:
So I had a kitura entry earlier, but `.travis.yml` recently changed to specify `"TESTLANG=Swift"` which I assumed meant that all frameworks under Swift would be tested.  Is there something else I need to add?

msmith-techempower:
Sorry, this is on me; I completely forgot about the `testlang` change. I'll restart the switch tests and we'll go from there.

PR #3641, opened 2018-04-23
[ci skip] Upgrade to wrk 4.1.0

michaelhixson:
This is the most recent version of wrk, released in January 2018.  There are no particular new features or fixes we want out of this release compared to the last one, which was from 2015.  I'm upgrading just because.

Looking at the commit history, the changes are mostly dependency updates. https://github.com/wg/wrk/commits/master


msmith-techempower:
> I'm upgrading just because

Not a very inspiring reason.

PR #3643, opened 2018-04-23
Make HTTP header verification more strict

michaelhixson:
It has long been a *requirement* for responses to include these headers:
 - Server
 - Date
 - Content-Type
 - either Content-Length or Transfer-Encoding

Now we are enforcing that requirement instead of emitting a warning.

Frameworks missing these headers will no longer be benchmarked (until we
fix them).  I bet there is at least one framework that will fail because
of this.  I'm sure I've seen warnings about a missing Server header
before.

Also, do not emit a warning for frameworks that include ";charset=utf-8"
on plaintext and json responses.  Many frameworks do this by default and
that behavior isn't a problem that we want anyone to fix.


michaelhixson:
We should let travis run the full test suite for this PR.  If there are only a couple of new failures, I will probably try to fix them within this PR.

toolset/benchmark/test_types/verifications.py
```diff
@@ -41,58 +41,34 @@ def verify_headers(headers, url, should_be='json'):
...
     if all(v.lower() not in headers
            for v in ('Content-Length', 'Transfer-Encoding')):
         problems.append((
-            'warn',
+            'fail',
             'Required response size header missing, please include either "Content-Length" or "Transfer-Encoding"',
             url))
 
-    content_type = headers.get('Content-Type', None)
+    content_type = headers.get('Content-Type')
```
michaelhixson: TODO(me): Figure out if I introduced an exception here when Content-Type is missing.

michaelhixson:
Here's where we ended up with this:

* It is now a failure to omit the `Content-Type`, `Server`, `Date`, or one of `Content-Length`/`Transfer-Encoding`.
* It is now a failure to have an incorrect `Content-Type` for a given test, such as `text/html` for the plaintext test.
* It is now a failure to **not** have `charset=utf-8` in the `Content-Type` of the fortunes response.
* We no longer warn about including `charset=utf-8` in the plaintext and JSON responses (which is unnecessary given the test requirements).
* New warning:  including a `Date` header that does not appear to be in RFC 1123 format.  Including multiple `Date` headers in the response will also trigger this warning.
* I fixed as many of the new failures and warnings as I could, but I couldn't fix all of them.

Frameworks with new warnings about date header format (these will still be benchmarked):
* `octane` - has a slightly different date format, but that format might be valid according to an older RFC, but discouraged?  Didn't dive into it too much.
* `ninja-standalone` - sends two `Date` headers in each response.

Frameworks with new failures (these will no longer be benchmarked as long as they're failing):
* `cppcms-direct` - missing `Content-Length` or `Transfer-Encoding`
* `ffead-cpp-apache-mongo` and `ffead-cpp-apache-postgresql` - present but malformed/junk `Content-Type` headers
* `elli` - missing `Date` headers.  There exists an [elli_date](https://github.com/elli-lib/elli_date) mod for exactly this purpose, but I couldn't figure out how to use it.
* `http4k-apache` - missing `Content-Length` or `Transfer-Encoding`

michaelhixson:
I'm pinging the contributors for the frameworks with new failures, just in case they have time to help us fix things.

I don't know who to ping about Erlang/elli.

@ekudritski You noted the `Content-Length`/`Transfer-Encoding` warning in cppcms's README, but now its a failure.  Here's the log: https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/371391895

@sumeetchhetri The variants of ffead that use apache have weird `Content-Type` headers like `Content-Type: p�`.  Here's the log: https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/371391896

@daviddenton The `http4k-apache` variant is missing `Content-Length` or `Transfer-Encoding`.  Could it be an issue with the `http4k-server-apache` module itself (rather than an issue with the test implementation in our repo)?  Here's the log: https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/371391954

daviddenton:
Both of those methods are missing because they are indeed stripped out by the http4k-server-apache implementation. Since I didn’t write that module, I’ve no idea this is happening, but I’ll endeavour to find out and fix it.

sumeetchhetri:
Fixed with PR https://github.com/TechEmpower/FrameworkBenchmarks/pull/3717

PR #3647, opened 2018-04-25
Installs databases quietly

joanhey:
In the php dir

**Now:**
First `VERIFYING DB` test in the line 791.

**Before clean the verbosity:**
First `VERIFYING DB` test in the line 2300.

More lines for the tests and less noise.


joanhey:
One job failed but it is not related to this commit.
Also pass the test in my fork.

PR #3648, opened 2018-04-25
Clean nginx.conf

joanhey:
### Now use the `www-data` user.
More secure and production ready.

### Remove `use epoll;`
Nginx will by default use the most efficient method.
So it will work in any OS and avoid problems.
If we see a regression, we will change it again.

### Performance change

**Remove extra try_files**
This is for an old security bug. And only work if php-fpm and web server run in the same system.
The correct way is change the `path_info` config in php.ini (I don't know why not come so in the php.ini for php-fpm)
Also all the people send de path_info from nginx directly. And the bug was resolved in php 5.3.9.
This change remove extra `stat()`s in each request.

Later I will add more performance changes.
But one for each run. So all the people can check the differences.
In a small server is insignificant, but in _citrine_ we can see easily the differences.


msmith-techempower:
@joanhey Is this ready to merge and you will open a new pull request at a later time, or do you want to leave this open and continue it as a work-in-progress?

joanhey:
It' ready
Thanks

PR #3650, opened 2018-04-25
Build databases up front

nbrady-techempower:
This builds all the databases required by any tests found with the `tfb` command up front and then runs them at test time.

### DIFF
--- a/toolset/utils/docker_helper.py
+++ b/toolset/utils/docker_helper.py
@@ -210,7 +210,11 @@ class DockerHelper:
     @staticmethod
     def __stop_container(container):
         try:
+            client = container.client
             container.kill()
+            while container.id in map(
+                    lambda x: x.id, client.containers.list()):
+                pass
         except:
             # container has already been killed
             pass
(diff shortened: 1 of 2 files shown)


toolset/utils/docker_helper.py
```diff
@@ -210,7 +210,11 @@ def watch_container(docker_container, docker_file):
     @staticmethod
     def __stop_container(container):
         try:
+            client = container.client
             container.kill()
+            while container.id in map(
+                    lambda x: x.id, client.containers.list()):
+                pass
```
msmith-techempower: Wait, does this work? I thought that we had a problem using `list` in that it would still have stopped containers in it.

nbrady-techempower: I think the problem we had before was the opposite. We were checking list too fast for a started container.

nbrady-techempower: In other words, we were requesting that the docker daemon start a container, and then checking the list immediately to see that the container had started, and sometimes it wasn't in the list yet. 

This works in that, if the container is not in the list now, it's definitely stopped.

PR #3657, opened 2018-04-27
Swoole include all tests + using SWOOLE_BASE mode for performance

Wulfklaue:
The Swoole author pointed out that using SWOOLE_BASE has higher performance.

https://github.com/swoole/swoole-src/issues/1588


Wulfklaue:
Updated the pull with all the required tests.

PR #3658, opened 2018-04-27
Use persistent db connections

joanhey:
The rest of the php frameworks have the same problem from #3656 

But there isn't any information in their configs, for select persistent.
Will be necessary to check each manual, for select it.

I don't have enough time for check all the frameworks.
Help wanted :)


joanhey:
The jobs failed are not from this commit.

PR #3661, opened 2018-04-27
Updated PHP benchmark_config.json with Swoole APIs

Wulfklaue:
Needed to finalize #3657

### DIFF
--- a/frameworks/PHP/php/benchmark_config.json
+++ b/frameworks/PHP/php/benchmark_config.json
@@ -87,6 +87,11 @@
     },
     "swoole": {
       "json_url": "/json",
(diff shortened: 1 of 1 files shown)


michaelhixson:
Could you cherry pick this commit into #3657 and close this PR, or vice versa?  We need the two sets of changes together in one PR in order to get a meaningful test result from travis.

Also, in `benchmark_config.json`, under swoole you'll need to change `"database": "None",` to `"database": "MySQL",`.  Our toolset looks at that field when running tests so that it knows which database to start up.  If the database is "None", it doesn't start up any database.

Wulfklaue:
Merged in other commit. Closing...

PR #3665, opened 2018-04-28
update http4k to fix apache server problem (content-length)

daviddenton:


daviddenton:
@michaelhixson this should fix the problem with the apache server not returning the proper headers

PR #3668, opened 2018-04-28
Opcache performance change

joanhey:
Will fail in frameworks than create compiled files for views, config, ....
So first check which frameworks can benefit from this change.


joanhey:
Perfect :+1: 
Looks like that in the last version that problem has been solved

PR #3673, opened 2018-04-29
updated docker file for wizzardo-http

wizzardo:


michaelhixson:
In general we don't want the build step for test implementations to be in the `CMD` because it defeats docker's built-in caching, which makes testing take longer.

I imagine the purpose of your change is exactly that - you don't want the build artifact to be cached.  You expect the dependencies to be changing, specifically the wizzardo-http dependency which is currently a `SNAPSHOT` version, and they'll change independently of any changes in the FrameworkBenchmarks repository itself, yet you still want our benchmarking environments to be aware of those changes.  That's not an approach we want to encourage in general for test implementations.  It's at odds with our desire to have reproducible results.  I realize that other test implementations are doing this (IIRC the Go tests do this a lot) but we don't want to move any farther in that direction.

To that end, I'd also like this implementation's dependency on your framework to become a non-`SNAPSHOT` version whenever that becomes possible, and then I don't think the `--refresh-dependencies` would be necessary.  But I understand if that's not possible right now.

So I'm closing this PR.  I hope the reasoning makes sense.  Feel free to argue if you think I've missed anything.

PR #3676, opened 2018-04-30
Add amphp/http-server (json and plaintext cases)

kelunik:
The template says the PR should be opened against the `docker` branch, but there's no such branch anymore. I guess it's fine to open against `master` in that case. If not, just respond and I'll change as necessary.


michaelhixson:
Oops, our PR template is outdated.  I'll update it.  Targeting `master` is correct, thanks.

PR #3681, opened 2018-05-02
Simplify the docker file

joanhey:
Install composer with apt-get

### DIFF
--- a/frameworks/PHP/cakephp/cakephp.dockerfile
+++ b/frameworks/PHP/cakephp/cakephp.dockerfile
@@ -7,15 +7,7 @@ RUN LC_ALL=C.UTF-8 add-apt-repository ppa:ondrej/php
 RUN apt-get update -yqq  > /dev/null
 RUN apt-get install -yqq nginx git unzip php5.6 php5.6-common php5.6-cli php5.6-fpm php5.6-mysql php5.6-xml php5.6-mbstring php5.6-mcrypt  > /dev/null
 
-RUN mkdir /composer
-WORKDIR /composer
-
-ENV PATH /composer:${PATH}
-
-RUN php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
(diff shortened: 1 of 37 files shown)


frameworks/PHP/cakephp/cakephp.dockerfile
```diff
@@ -7,15 +7,7 @@ RUN LC_ALL=C.UTF-8 add-apt-repository ppa:ondrej/php
...
-RUN mkdir /composer
-WORKDIR /composer
-
-ENV PATH /composer:${PATH}
-
-RUN php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
-RUN php -r "if (hash_file('SHA384', 'composer-setup.php') === '544e09ee996cdf60ece3804abc52599c22b1f40f4323403c44d44fdfdd586475ca9813a858088ffbc1f233e9b180f061') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"
-RUN php composer-setup.php
-RUN php -r "unlink('composer-setup.php');"
+RUN apt-get install -yqq composer > /dev/null
```
michaelhixson: Just FYI, for consistency we're now trying to use `apt` instead of `apt-get` everywhere, like in this commit: https://github.com/TechEmpower/FrameworkBenchmarks/commit/a2de73455678f38861e6cf1cfca5b46db43c1fb0

I see that `apt-get` has crept back into the PHP dockerfiles elsewhere, not just in the lines you're changing here.  I will probably do a global search and replace for `apt-get -> apt` later this week after merging this.

nbrady-techempower:
@joanhey The docker documentation has probably had that line for a while.

From: https://www.linux.com/blog/difference-between-apt-vs-apt-get-explained

> It became common to see apt install package instead of the usual apt-get install package. Eventually, many other distributions followed Ubuntu’s footsteps and started to encourage users to use apt instead of apt-get.

`apt` looks like the way forward and while I was writing all these docker files it was just less bytes / less typing. I'm not super concerned about the single line warning but all I care about right now is consistency.

Honestly, it doesn't matter to me which we use so long as every `apt` command looks the same.

michaelhixson:
Merging this as is.  We can do the apt/apt-get cleanup separately since it touches more than what this PR touches.

PR #3685, opened 2018-05-02
Log Less

msmith-techempower:
Gemini got hit pretty bad in the last few plaintext benchmarks on Citrine, and we suspect that it was due to removing the resin.xml which configured logging.

This commit attempts to turn off access logging entirely, although in prior builds, the `access-log` directive was not required; may remove in a future commit.


msmith-techempower:
@michaelhixson Did I miss any `resin` impls?

michaelhixson:
You got all of them.

PR #3688, opened 2018-05-02
try to avoid launching do-nothing travis jobs

michaelhixson:
I'm experimenting with travis to see if we can speed up build times using a combination of build stages and `travis_terminate`.  It might not work out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,8 @@ jobs:
         - FOO=foo
       script:
         - echo "tests to run = [$RUN_TESTS]"
-        - export TRAVIS_TEST_RESULT=1
-        - travis_terminate 0
+        - export TRAVIS_TEST_RESULT=0
+        - travis_terminate 1
 
 stages:
   - quicktest
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,8 @@ jobs:
       env:
         - FOO=foo
       script:
-        - echo 'In the quicktest stage:' && echo $RUN_TESTS
+        - echo "tests to run = [$RUN_TESTS]"
+        - travis_terminate 0
 
 stages:
   - quicktest
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ jobs:
       env:
         - FOO=foo
       script:
-        - echo "In the quicktest stage: ${RUN_TESTS}"
+        - echo 'In the quicktest stage:' && echo $RUN_TESTS
 
 stages:
   - quicktest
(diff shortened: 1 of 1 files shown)


nbrady-techempower:
Ooooh I'm excited for this. I saw this in beta but couldn't quite get it working myself. What I wanted to do was try and run the git diffing tool to determine what tests needed to be run before any of the stages, then use the new conditional builds to test if the current test was in a global ENV var set outside the stages of tests that needed to be run.

michaelhixson:
Well, what I tried did not work.

I didn't try sharing an environment variable between stages.  I assumed from the description of the feature that it can only access environment variables defined in .travis.yml, not ones you set dynamically.

(This is hard to explain in writing.  Maybe it'll be better to explain next time we meet in person.)

What I was tried to do was:  Have Stage 1 consist of one job that runs the differ and looks at how many tests should be run.  If it thinks it can run all of those tests without timing out in one job, go ahead and run them in the current job, then cancel Stage 2.  Otherwise, if it thinks it will time out, don't run any tests in this job and proceed to Stage 2.

If that worked, since many of our PRs only affect 1 framework at a time, they would only launch the one job in Stage 1.  That job would show them the test results for their framework, and the big matrix of jobs in Stage 2 would never start.

The problem is I can only get it to cancel Stage 2 if it fails Stage 1.  I can't make it cancel Stage 2 if it thinks Stage 1 is a success (that is, when it thinks you changed one framework and that framework passed its test).

To avoid Stage 2, we must give Stage 1 a red X, and the build result has to **appear** to be a failure.  We could have it appear to fail but actually run the framework test and print the result, and then we always read the log to figure out whether it was really a success or really a failure.  I thought that would be too annoying - we like the green check marks.

nbrady-techempower:
> I didn't try sharing an environment variable between stages. I assumed from the description of the feature that it can only access environment variables defined in .travis.yml, not ones you set dynamically.

I thought I saw somewhere that you could do something like:

```
env:
  global:
   TESTS_TO_RUN=$()

# Build stages here
```

I don't think we can run scripts in `$()` but I thought it might be possible to parse `git diff` in one bash line such that `TESTS_TO_RUN` might look like `Java/Gemini JavaScript/nodejs` etc. 

Then a job might look like this:
```
jobs:
  include:
   if "Java/gemini" IN ENV(TESTS_TO_RUN):
   env: TESTDIR=Java/gemini

# Toolset scripts here
```

You probably got further along than I did testing different scenarios. I hadn't played with it much.

PR #3689, opened 2018-05-03
Updating to Cutelyst 2.2.1

dantti:

### DIFF
--- a/frameworks/C++/cutelyst/README.md
+++ b/frameworks/C++/cutelyst/README.md
@@ -1,0 +1,13 @@
+# Cutelyst 
+
+Cutelyst is a C++ Full Stack Web Framework built with Qt.
+
+Inspired on the Perl Catalyst Framework, it integrages with templating
+engines such as Grantlee (DJango syntax), ClearSilver. Many
+ready to use plugins to send templated emails, validade user input,
+CSRF protection, authentication and session increase the productivity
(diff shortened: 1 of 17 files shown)


nbrady-techempower:
@dantti I'd like to get this in before I get #3692 in so I can fix any merge conflicts myself. Before I do, would you consider changing to `FROM buildpack-deps:xenial` This will give you some preinstalled tools to make your build faster, and we use it for other tests so it will be cached. I think it's a better base than ubuntu:16.04. For reference here's what I just did with ulib: https://github.com/TechEmpower/FrameworkBenchmarks/pull/3692/files#diff-eb25de16844f32f6f80221c5a6f00acc

Here are all the packages that are pre-installed:
https://github.com/docker-library/buildpack-deps/blob/d7da72aaf3bb93fecf5fcb7c6ff154cb0c55d1d1/xenial/Dockerfile

https://github.com/docker-library/buildpack-deps/blob/b0fc01aa5e3aed6820d8fed6f3301e0542fbeb36/xenial/curl/Dockerfile

https://github.com/docker-library/buildpack-deps/blob/2da658b9a1b91fa61d63ffad2ea52685cac6c702/xenial/scm/Dockerfile

Thanks!

nbrady-techempower:
ok ill get to this!

PR #3695, opened 2018-05-04
D/Kiss: initial submit for benchmark test

Heromyth:

### DIFF
--- a/frameworks/D/kiss/benchmark_config.json
+++ b/frameworks/D/kiss/benchmark_config.json
@@ -1,0 +1,43 @@
+{
+  "framework": "kiss",
+  "tests": [{
+    "default": {
+      "json_url": "/json",
+      "plaintext_url": "/plaintext",
+      "port": 8080,
+      "approach": "Realistic",
+      "classification": "Platform",
+      "database": "None",
+      "framework": "Kiss",
+      "language": "D",
+      "flavor": "DLang2",
+      "orm": "Raw",
+      "platform": "Kiss",
+      "webserver": "None",
+      "os": "Linux",
+      "database_os": "Linux",
+      "display_name": "Kiss",
+      "notes": "",
+      "versus": "Kiss"
+  },
+    "ldc": {
+      "json_url": "/json",
+      "plaintext_url": "/plaintext",
+      "port": 8080,
+      "approach": "Realistic",
+      "classification": "Platform",
+      "database": "None",
+      "framework": "Kiss",
+      "language": "D",
+      "flavor": "DLang2",
+      "orm": "Raw",
+      "platform": "Kiss",
+      "webserver": "None",
+      "os": "Linux",
+      "database_os": "Linux",
+      "display_name": "Kiss",
+      "notes": "",
+      "versus": "Kiss"
+    }
+  }]
+}
--- a/frameworks/D/kiss/kiss.dockerfile
+++ b/frameworks/D/kiss/kiss.dockerfile
@@ -1,4 +1,4 @@
-FROM dlanguage/ldc:1.9.0
+FROM dlanguage/ldc:1.7.0
 
 ADD ./ /kiss
 WORKDIR /kiss
--- a/frameworks/D/kiss/kiss.dockerfile
+++ b/frameworks/D/kiss/kiss.dockerfile
@@ -1,0 +1,9 @@
+FROM dlanguage/ldc:1.9.0
+
+ADD ./ /kiss
+WORKDIR /kiss
+
+RUN dub upgrade --verbose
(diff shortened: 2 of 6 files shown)


frameworks/D/kiss/kiss.dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM dlanguage/ldc:1.9.0
```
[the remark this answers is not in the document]
Heromyth: It updated to use LDC 1.7.0. 
Thanks for your reply.

frameworks/D/kiss/benchmark_config.json
```diff
@@ -0,0 +1,43 @@
+{
+  "framework": "kiss",
+  "tests": [{
+    "default": {
+      "json_url": "/json",
+      "plaintext_url": "/plaintext",
+      "port": 8080,
+      "approach": "Realistic",
```
michaelhixson: FYI, since kiss does not appear to be a web application framework and since it doesn't appear to provide any functionality related to HTTP, we're going to change this to `"approach": "Stripped"`.  It will be hidden on the results website by default.

frameworks/D/kiss/source/app.d
```diff
@@ -0,0 +1,197 @@
...
+		else
+		{
+			badRequest(client);
+		}
+
+	}
+
+	private void respondPlaintext(TcpStream client, bool keepAlive)
+	{
+		string writeData = "HTTP/1.1 200 OK\r\nContent-Length: 13\r\nConnection: Keep-Alive\r\nContent-Type: text/plain\r\nServer: Kiss/0.3\r\nDate: Wed, 17 Apr 2013 12:00:00 GMT\r\n\r\nHello, World!";
```
michaelhixson: This is a violation of the plaintext test requirements, which state:

> This test is not intended to exercise the allocation of memory or instantiation of objects. Therefore it is acceptable but not required to re-use a single buffer for the response text (`Hello, World`). However, the response must be fully composed from the response text and response headers within the scope of each request and **it is not acceptable to store the entire payload of the response, or an unnaturally large subset of the response, headers inclusive, as a pre-rendered buffer**. "Buffer" here refers to a byte array, byte buffer, character array, character buffer, string, or string-like data structure. The spirit of the test is to require the construction of the HTTP response as is typically done by a framework or platform via concatenation of strings or similar. For example, pre-rendering a buffer with `HTTP/1.1 200 OK<cr>Content-length: 15<cr>Server: Example<cr>` would not be acceptable.

To fix this, please make at least these changes:

1. Dynamically render the `Date` header value.
2. Dynamically render the `Content-Length` header value as you do already in the JSON test implementation.
3. Dynamically concatenate each HTTP header to the response separately.  For this one, I suppose it's sufficient to use a StringBuilder or whatever method of string concatenation is common in D code, just as long as the compiler isn't smart enough to optimize away the concatenation.

frameworks/D/kiss/source/app.d
```diff
@@ -0,0 +1,197 @@
...
+		});
+	}
+
+	private void respondJson(TcpStream client, bool keepAlive)
+	{
+		JSONValue js;
+		js["message"] = "Hello, World!";
+		string content = js.toString();
+
+		string writeData = "HTTP/1.1 200 OK\r\nConnection: Keep-Alive\r\nContent-Type: application/json\r\nContent-Length: " ~ to!string(
```
michaelhixson: Please apply the fixes for `Date` and HTTP header concatenation here as well.

frameworks/D/kiss/source/app.d
```diff
@@ -0,0 +1,197 @@
...
+			debug writeln("sent bytes: ", size, "  content: ", cast(string) writeData);
+			if (!keepAlive)
+				client.close();
+		});
+	}
+
+	private void badRequest(TcpStream client)
+	{
+		// string writeData = "HTTP/1.1 404 Not Found\r\nServer: Kiss/0.3\r\nConnection: close\r\n\r\n";
+		string writeData = `HTTP/1.1 404 Not Found
```
michaelhixson: This one is fine to leave as-is since it's not part of the test requirements.  Out of curiosity though:  are the correct line separators being used?

Heromyth: The message is same as the default 404 response from Nginx.

frameworks/D/kiss/benchmark_config.json
Heromyth: That's it. The [Collie](https://github.com/huntlabs/collie) is base on Kiss.

frameworks/D/kiss/source/app.d
Heromyth: We will update these as your suggestion.

PR #3699, opened 2018-05-04
Remove some large files and images

nbrady-techempower:
⬆️


nbrady-techempower:
PHP errors related to #3702

PR #3708, opened 2018-05-08
Yii2 db persistent connection

joanhey:


joanhey:
The fail it isn't with this change.

Looks like a problem with hhvm again. Also fail with the json test.

michaelhixson:
Confirmed that yii2-hhvm and hhvm are both failing for me locally on master without this change.  Sigh.  I'll see if I can fix those up.

michaelhixson:
The hhvm failures should be fixed with a toolset change: https://github.com/TechEmpower/FrameworkBenchmarks/pull/3709

PR #3709, opened 2018-05-08
Don't consider a server up after 5xx

michaelhixson:
Without this change, we consider a server ready for verification if it
responds to a curl command with any HTTP response, even if that response
indicates a failure like 502 Bad Gateway.  (This particular response can
happen with frameworks that use nginx, when nginx has started but the
back end server isn't ready yet.)  With this change, we wait for a
successful response.

I added the fw-only tags because these two frameworks are currently
failing on master due to this issue.  They should pass now with this
change.

### DIFF
--- a/toolset/utils/docker_helper.py
+++ b/toolset/utils/docker_helper.py
@@ -336,7 +336,7 @@ class DockerHelper:
         try:
             self.client.containers.run(
                 'techempower/tfb.wrk',
-                'curl --max-time 5 %s' % url,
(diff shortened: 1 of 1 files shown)


michaelhixson:
I misspelled the tag for travis, but maybe that's for the best.

michaelhixson:
I don't think any of the failures in that travis build are related to this, so I will go ahead with the merge.  hhvm and yii2-hhvm are now succeeding where they were failing before.

PR #3716, opened 2018-05-10
Add php-raw7-tcp

joanhey:
To check the differences with tcp sockets and unix sockets.

Normally unix sockets are faster, but tcp sockets scale better.
We will try to resolve a typical question.

Ping @nbrady-techempower 
I need to change from dynamic to static children, and add more children as the latency is still very low.
No problem with citrine or azure.
But it's a problem in the vagrant box, max 512 children.

I know that other frameworks change the processes number with the number of cores.
Could you pass me any example?
Thanks

### DIFF
--- a/frameworks/PHP/php/benchmark_config.json
+++ b/frameworks/PHP/php/benchmark_config.json
@@ -43,6 +43,27 @@
       "notes": "",
       "versus": "php"
     },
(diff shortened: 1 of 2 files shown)


joanhey:
I will add another PR for the static change.
This one is ready to merge.

PR #3717, opened 2018-05-10
add core dump support, base image/multi command support

sumeetchhetri:
Fixed ffead-cpp nginx/httpd tests
Added cached query tests for redis/memcached
Added support for single docker file with multi command variations to reduce build times
Added support to enable core dumps in containers to debug issues in framework
Latest code from  https://github.com/sumeetchhetri/ffead-cpp
Fixes issue with Mongodb writes as well as incorrect content-type headers in apache tests
Fix for issue 3643


toolset/databases/mongodb/create.js
```diff
@@ -7,6 +7,14 @@ for (var i = 1; i <= 10000; i++) {
 db.world.createIndex({_id: 1})
 db.world.createIndex({id: 1})
 
+db.cachedworld.drop()
+for (var i = 1; i <= 10000; i++) {
+  db.cachedworld.save( { _id: i, id: i, randomNumber: (Math.floor(Math.random() * 10000) + 1) })
```
michaelhixson: There should not be a cachedworld table in any of the databases.  Please undo the changes to mongodb/create.js, mysql/create.sql, and postgres/create-postgres.sql.

toolset/run-tests.py
```diff
@@ -189,6 +189,17 @@ def main(argv=None):
         default=None,
         help='The network mode to run docker in')
 
+    parser.add_argument(
+        '--local-dev-mode',
+        action='store_true',
+        default=False,
+        help='server and database both run locally, very easy during development, building containers takes huge aount of time every single time if there are changes in framework source code')
```
michaelhixson: We don't want to support `--local-dev-mode`, but there is a decent alternative for running the server locally.  You can pass in the `--server-host` argument with a hostname or IP address that resolves to your workstation, just something other than localhost/127.0.0.1.  If you do that and you temporarily replace your dockerfile with something like this:
```
FROM buildpack-deps:xenial
CMD sleep infinity
```
then you can skip the server build times and hit your local server instance.

I don't think the same approach will work for the database, but you shouldn't have to build any of the databases more than once.  I think the changes in this PR to the databases was maybe coming from confusion about the cached query test requirements - it's supposed to be backed by the same worlds table as the other tests.

toolset/run-tests.py
```diff
@@ -189,6 +189,17 @@ def main(argv=None):
...
+    parser.add_argument(
+        '--enable-core-dumps',
+        action='store_true',
+        default=False,
+        help='sometimes we may need core dumps from the conatiner')
```
michaelhixson: This sounds interesting.  Where/how do we look at the core dumps?

toolset/databases/mongodb/create.js
sumeetchhetri: Please read the docs at http://frameworkbenchmarks.readthedocs.io/en/latest/Project-Information/Framework-Tests/#caching, 
`The request handler must retrieve a set of CachedWorld objects, equal in count to the count parameter, from the CachedWorld database table. Note that the CachedWorld database table uses a schema idential to the World table used for Test #3 but is distinct to avoid confusion.`
Hence I added these new tables to all the schemas

toolset/run-tests.py
sumeetchhetri: You could probably get inside the container with privileged mode on and then install gdb once in and then look at the dumps there or could also copy that back to the host and analyze there.

frameworks/C++/ffead-cpp/benchmark_config.json
```diff
@@ -1,95 +1,311 @@
...
-      "webserver": "apache",
-      "os": "Linux",
-      "database_os": "Linux",
-      "display_name": "ffead-cpp-apache-postgresql",
-      "notes": "",
-      "versus": ""
-    }
-  }]
+	"framework": "ffead-cpp",
+	"tests": [{
```
michaelhixson: It looks like the travis build for ffead hit its 50 minute timeout.  We could work around this by splitting ffead into multiple jobs, but we don't want to do that.  50 minutes is more time than we want to spend building or testing any single framework.

ffead needs to fit within that 50 minute window somehow.  Here are a few ideas for how you could make that happen:
* Consider combining the variants that only implement the cached query test type into the other variants.  For example, can `redis` be combined with `default`?
* Check if you can make better use of docker's build cache by reorganizing the dockerfiles to share larger common prefixes.  For example, if two dockerfiles differ only in one line, try moving that line as far down in both dockerfiles as possible.
* Consider simply removing some of these variants.  For example, consider removing all of the `apache` variants or all of the `nginx` variants, and consider only targeting one database rather than all three, and either redis or memcached but not both.  As the framework's maintainer, you're in a better position than most people to decide which of these variants is best, and readers who find out about your framework by way of TFB may appreciate you being decisive -- "run my framework like this" -- rather than being shown several options.  Consolidating/removing options tends to make these test implementations much less complicated, which makes the frameworks look less imposing and easier to use.

toolset/databases/mongodb/create.js
michaelhixson: Oh interesting!  Our fault - that part of the docs is wrong/outdated.  We've decided these tests should use the same World table as the other tests.  There are already implementations of the cached queries test that do this.

We'll update the docs.  Sorry for the confusion.

frameworks/C++/ffead-cpp/benchmark_config.json
sumeetchhetri: Docker cache works for a specific dockerfile and not across multiple dockerfiles. To resolve this issue I have added support for a single docker file with multiple command variations, these commands are defined in the specific test benchmark config files along with the base image name to use. The idea is to build a whole/complete base image for the framework with all bells/whistles and then based on the entrypoint/command specified, change the framework behavior accordingly. Also I would not want to remove any of these tests from the benchmark as the whole point of having this project is to provide various combinations of frameworks/servers/databases to identify better performing ones. I also have plans to add a lighthttpd module down the line, I want ffead-cpp to be the one that supports any server backend, any database any cache, my idea with ffead-cpp is to make it super simple to develop with C++ without worrying too much about the internals. So with this approach I can sneak in more combinations of the framework in the future as well. Do let me know what you think of this approach?

sumeetchhetri: If you look at the python code that builds/runs a docker image, you will see that every test flavor is named and this same name is used as the container name, so in short a Dockerfile currently in the framework is equal to a named container. The Docker cache works on the Dockerfile and hence this does not work for multiple dockerfiles correctly. In order for Docker cache to work correctly the child container should be spawned from the same base image (same named container). Btw I already tried doing what you suggested and it did not work out, what I did in short was have 13 different Dockerfiles all similar except their last lines (the CMD). So yes i have tried your suggested approach and then I zeroed in on the this change that you see here. This approach as you said is easier as it entertains the framework maintainers and authors to keep the number of artifacts to a minimum and yes less duplication. The spearate Dockerfile approach takes almost 18*13 minutes of time to run which is way higher that your stipulated 50 min travis runtime.

michaelhixson: > Btw I already tried doing what you suggested and it did not work out, what I did in short was have 13 different Dockerfiles all similar except their last lines (the CMD). So yes i have tried your suggested approach and then I zeroed in on the this change that you see here.

What you're claiming here conflicts with what I observe locally and in the travis builds.

Could you demonstrate that I'm wrong by making the change I requested, which will trigger another travis build?

sumeetchhetri: Have raised another PR with the changes and it works as you expected.

msmith-techempower:
Closing as this has not been updated for over a month; please re-open or create a new one when you have addressed the issues.

PR #3718, opened 2018-05-10
Silex-raw db persistent connection

joanhey:


joanhey:
Fail codeigniter-hhvm, not related with this commit.

PR #3729, opened 2018-05-12
Add benchmarks for chi + gojay

dcu:


msmith-techempower:
Appears to be duplicate of #3734

PR #3737, opened 2018-05-13
Modified shebang to use bash rather than plain sh

programingjd:
`tfb` doesn't work on wsl with the `#!/bin/sh` shebang, `bash` needs to be specified instead.

### DIFF
--- a/tfb
+++ b/tfb
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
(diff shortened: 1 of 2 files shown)


tfb
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
msmith-techempower: Does this actually matter?

PR #3738, opened 2018-05-13
Updated Readme to note that on Windows, Docker with Windows containers and LCOW cannot be used, and Linux containers should be used instead.

programingjd:
On Windows, Docker can be used with Linux containers, or with Windows containers and LCOW. However, LCOW doesn't recognize the path to` /var/run/docker.sock`. That means that on Windows, users should switch Docker to Linux containers mode or it won't work.
I have updated the README and added a note warning about this.


nbrady-techempower:
This is in #3737. I haven't tested any of this. Will check it out on Monday.

PR #3742, opened 2018-05-16
tfb - add core dump support, base image/multi command support | ffead-cpp fixed all failing tests and added cached tests (Separate dockerfiles)

sumeetchhetri:


sumeetchhetri:
@michaelhixson So yes you were right, i'm proven wrong. I concede defeat, can you please get either of the PR's merged accordingly.

michaelhixson:
Thanks for testing this out.  What I was really after was separating your improvements to ffead from your new toolset features.  Can this become the PR for your ffead changes only -- can you undo the changes to the toolset here?

I'm ready to accept your ffead changes as-is but I'm not ready to do the same with the toolset changes.  Those need closer review from me and/or my colleagues, and I won't be able to put in the necessary time for that right away.

On the toolset side - If you're willing to wait a little bit for review and you're willing to possibly tweak the new features based on our feedback, you could leave the other PR open.  After this PR gets merged in, the diff in that PR should automatically shrink down to only the new toolset features plus a demo of how to use them with ffead.  If you're feeling like you've had enough and don't want to see the other PR through, you could close it.  I see this PR as the more important one.  We could probably figure out how to add the same toolset features from scratch later if we wanted, but we have no hope of doing the same with your improvements to ffead.

sumeetchhetri:
Done, I have left the other pr open just in-case we need the toolset changes.

PR #3747, opened 2018-05-17
Adding Platform Fortune

sebastienros:
/cc @benaadams @davidfowl


frameworks/CSharp/aspnetcore/PlatformBenchmarks/Startup.cs
```diff
@@ -2,15 +2,51 @@
...
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Npgsql;
+using MySql.Data.MySqlClient;
 
 namespace PlatformBenchmarks
 {
     public class Startup
     {
+        public Startup(IHostingEnvironment hostingEnv)
```
davidfowl: Why are we using Startup at all? Why not configure things in Program.cs?

sebastienros:
4 hours before the next CI run
crossing fingers it gets in before

bhauer:
@sebastienros We're ready to merge this before the next continuous run starts, but @davidfowl has not yet approved. Let us know if you would like it merged without that approval.

PR #3751, opened 2018-05-17
H2O: Final reduction of the number of database connections

volyrique:


sebastienros:
No way, they are already too fast, it's unfair 😄

michaelhixson:
@volyrique Just curious...  Why is h2o's fortunes implementation so fast?  :)

Do you think it's anything that other frameworks can learn from?  When I read through `fortune.c`, it seems like it's doing the same general sequence of things as most other frameworks but with lower-level APIs.  Maybe that's all it is - every little step of the implementation is faster?

volyrique:
Currently the best explanation I have is that the database connection pools are properly sized (well, it is still a work in progress). Otherwise the implementation uses pretty standard techniques:
- fully asynchronous database commands - the program never blocks unless absolutely necessary
- prepared statements - created once when the database connection is established
- minimizing the number of system calls - libh2o uses `writev()`, so all fortunes are sent with one system call
- avoiding memory allocations and copies as much as possible
- extensive usage of the memory pool functionality provided by libh2o - every HTTP request has an associated memory pool, and all allocations belonging to the pool are freed together in one go after the request is completed; note that it is not possible to override the memory routines used by the database client library, libpq, for example, so not every allocation is a pooled one

By design, the application has thread-local database connection pools, so that the amount of shared mutable state (database connections being an example of it) is minimized, which has performance and correctness implications (less lock contention and lower probability of race conditions, for example). Furthermore, one of the major decisions I have made is to use the same configuration for all test types (I have noticed that some implementations have special configurations just for the database tests), and although I haven't done much experiments (too many values to try), it seems that the best value for the number of worker threads is the same as the number of logical processors in the system, i.e. 28 (which is the expected situation for a workload that does not touch the disk). As a result, I can vary the number of database connections only in multiples of 28.

I don't think the performance of h2o is exceptional. When the runs in the Citrine environment started, h2o was fast, but not that much: It was not even in the top 10, as visible in [run 8de314d3-b970-4295-a3ca-3682f3905a3b](https://tfb-status.techempower.com/results/8de314d3-b970-4295-a3ca-3682f3905a3b), for instance. However, at that point it was using only 1 database connection per thread (28 in total), which was a good value for the ServerCentral environment, but obviously not that good for Citrine. Previously I noticed that the Postgres variants of ULib usually performed similarly, but it seems that they have not been tuned for the current environment.

I have been experimenting a lot with the pool size, and with this change I will have data for 1-6 and 8 connections per thread (again, multiply by 28 for the total number). It seems that for all tests except the updates one (which is special) the application reaches saturation around 4. I will have to wait for a run with this change, but the updates test probably peaks at 2 connections per thread, and gets worse quickly with higher values. So, I will probably push a final change that is going to be a compromise between the performance on the updates test and the rest.

The other frameworks that have a global connection pool might be able to tweak that further, and obtain even better performance, assuming that they got the basics right.

bhauer:
@volyrique Thanks for the thorough write-up.  It's a great demonstration of the amazing talent and generosity of the contributors to this project.

I do not want to diminish the achievement, but as a separate matter, I'd like a sanity check that observes the network traffic using Wireshark or similar to confirm the behavior is what we want and expect.  A field of very high-performance frameworks appear to converge to approximately 340K RPS on Fortunes and then a step function jumps h2o up to nearly 500K.  I feel some additional due diligence is warranted, but I would be happy to have it all check out just fine.

volyrique:
Sure, no problem, is that additional verification something that I would have to do? BTW, h2o was also performing at around 340k RPS on the fortunes test with 2 database connections per thread (refer to [run 089634a5-833f-4a8b-abc3-b694244913a5](https://tfb-status.techempower.com/results/089634a5-833f-4a8b-abc3-b694244913a5), for example).

bhauer:
@volyrique Not required, but if you have time and are interested, it would be appreciated. I'd like some sanity checking, but I wasn't sure who would have the time and when. For the time being, however, we'll just trust the results.

volyrique:
All right, in order to obtain a reasonable amount of data, I ran a reduced configuration - 2 worker processes, each with 1 thread (which is more or less equivalent to 1 process with 2 threads) that handles 2 database connections in its pool, so 4 database connections in total. The necessary modifications to the `frameworks/C/h2o/start-servers.sh` script are the following:
- set `NUM_WORKERS` to `2`
- set all occurences of `USE_PROCESSES` to `true`
- set all occurences of `DB_CONN` to `2`

The Web application was started in the standard way with Docker and the following parameters:
`tfb --mode debug --test h2o`
I captured the database communication with tcpdump because the toolset validation functionality doesn't really cover it:
`tcpdump -w fortunes.pcap tcp port 5432`

I used the same wrk command line as in the continuous benchmarking setup, except that I changed the duration to 1 s instead of 15 s (again, to keep the amount of data reasonable), and the server address to 127.0.0.1 (I was running wrk in the server container):
```
root@ed68eaaf334f:/# wrk -H 'Host: 10.0.0.1' -H 'Accept: application/json,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 1 -c 512 --timeout 8 -t 28 "http://127.0.0.1:8080/fortunes"
Running 1s test @ http://127.0.0.1:8080/fortunes
  28 threads and 512 connections
    Latency    36.97ms   17.28ms  84.20ms   78.68%
    Req/Sec   482.45    123.84     1.36k    90.75%
  Latency Distribution
     50%   32.15ms
     75%   41.42ms
     90%   68.70ms
     99%   82.28ms
  14087 requests in 1.10s, 18.77MB read
  Non-2xx or 3xx responses: 4
Requests/sec:  12821.33
Transfer/sec:     17.09MB
```
By design the Web application established the database connections as soon as it started (and before I ran tcpdump), so I killed manually the 4 PostgreSQL processes handling the connections to force the program to repeat the whole process of connecting to the database and to make that appear in the packet capture. This was the reason there were 4 failed requests.

The packet capture can be found [here](https://github.com/TechEmpower/FrameworkBenchmarks/files/2019079/fortunes.zip). You can inspect it with Wireshark, but the interesting bits are the packets with sizes 44 and 939, which are the commands from the client and the responses from the database respectively:
```
tcpdump -r fortunes.pcap 2>/dev/null | grep "length 44" | wc
  14589  306369 2440564
tcpdump -r fortunes.pcap 2>/dev/null | grep "length 939" | wc
  14585  306285 2459962
```
wrk reports 14087 requests, and if we presume around 512 (the concurrency parameter) more being in progress when the time has run out, the numbers add up. There are 4 less responses from the database because of the initial failed requests I mentioned.

I hope that this is sufficient as a sanity check.

bhauer:
I didn't thank you for this detail earlier, @volyrique.  I just saw you mention it from the Google Group.  Thanks for digging in and verifying things and for sharing tips with the rest of us!

PR #3752, opened 2018-05-17
Update actix to latest stable version

fafhrd91:
* upgrade actix to latest stable version
* fix update db, pg benchmark
* rename main files to main_*


frameworks/Rust/actix/src/main.rs
```diff
@@ -2,161 +2,53 @@ extern crate actix;
...
-        q.parse::<usize>().ok().unwrap_or(1)
-    } else {
-        1
+fn json(req: HttpRequest) -> HttpResponse {
+    let message = Message {
+        message: "Hello, World!",
     };
-    let q = cmp::min(500, cmp::max(1, q));
+    let mut body = BytesMut::with_capacity(SIZE);
+    serde_json::to_writer(Writer(&mut body), &message).unwrap();
```
pickfire: Is there a reason to not use `serde_json::to_string(&message).unwrap()` here?

fafhrd91: It will allocate String, then you’ll need to convert it to Bytes. In this code serde writes to Bytes directly

pickfire: I wonder if zero-cost abstractions of rust helped in this such that they know we want the bytes after conversion to string and if it could skip the process directly. Nevertheless, it seems weird that serde have `to_string` but not `to_bytes`. Let me create an issue there.

fafhrd91: You can convert String to Vec, which is zero cost. Serde can use io::Write and this code uses it

pickfire: Yeah, but that works but not ergonomic. I wonder if `serde_json::to_string('string').as_bytes()` zero cost?

fafhrd91: it is not in sense it is allocates.

PR #3766, opened 2018-05-20
Non-blocking Coroutine Mysql, fix plaintext

Wulfklaue:
As the default PHP database connection is blocking, using the Swoole integrated Coroutine solution, will prevent it from blocking the server.

Also potential fixing the bad plaintext results.


Wulfklaue:
Fortune test will fail as the Mysql query is not being sorted correctly. Will finish the fix tomorrow night.

PR #3767, opened 2018-05-21
New actix-raw benchmarks

fafhrd91:
* Introduce new actix-raw benchmarks
* update actix-web version


frameworks/Rust/actix/benchmark_config.json
```diff
@@ -19,6 +19,44 @@
...
+      "classification": "Micro",
+      "database": "Postgres",
+      "framework": "actix",
+      "language": "Rust",
+      "orm": "Raw",
+      "platform": "None",
+      "webserver": "actix-web",
+      "os": "Linux",
+      "database_os": "Linux",
+      "display_name": "actix-raw",
```
sebastienros: From my experience these values are useless, as this is what is computed automatically.
What I would suggest is to use a friendly name here, like `Actix, Raw`.

frameworks/Rust/actix/benchmark_config.json
```diff
@@ -19,6 +19,44 @@
       "notes": "",
       "versus": ""
     },
+    "raw": {
+      "fortune_url": "/fortune",
+      "json_url": "/json",
+      "plaintext_url": "/plaintext",
```
sebastienros: What is the difference between json and plaintext in the default profile, and this one?

fafhrd91: Other profiles uses high level api, this profile uses low-level api. It is similar to aspcore

sebastienros: Ok, I was wondering if it was intentional to have other json and plaintext scenario, so it is, thanks. In aspnet we use 'raw' for the db access, it confused me ;)

frameworks/Rust/actix/benchmark_config.json
sebastienros: I wish TE would recommend to use more friendly names in `display_name`, instead of just cloning the handles that are displayed in the chart.

frameworks/Rust/actix/benchmark_config.json
```diff
@@ -19,6 +19,44 @@
...
+      "port": 8080,
+      "approach": "Realistic",
+      "classification": "Micro",
```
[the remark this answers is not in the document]
fafhrd91: you right

fafhrd91:
@nbrady-techempower hey, do I need to change anything. It would be super if you could merge it before next test run

PR #3768, opened 2018-05-21
Update Nancy Benchmark

benaadams:
Also added .NET Core variants in addition to Mono

/cc @khellang @thecodejunkie


patricka2:
.NET Core enable "LogLevel" default to "Error" by default even by clear "LogLevel" in appsettings.json (Like now) in both production and development mode.
Is not it better to set "LogLevel" default to "Error" instant "Warring" to disable logging like this:
eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee

PR #3779, opened 2018-05-24
ASP.NET Core - Platform - SingleQuery

benaadams:
/cc @sebastienros

### DIFF
--- a/frameworks/CSharp/aspnetcore/PlatformBenchmarks/BenchmarkApplication.cs
+++ b/frameworks/CSharp/aspnetcore/PlatformBenchmarks/BenchmarkApplication.cs
@@ -42,6 +42,7 @@ namespace PlatformBenchmarks
             public readonly static AsciiString Plaintext = "/plaintext";
             public readonly static AsciiString Json = "/json";
             public readonly static AsciiString Fortunes = "/fortunes";
+            public readonly static AsciiString SingleQuery = "/db";
         }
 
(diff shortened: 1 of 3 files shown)


frameworks/CSharp/aspnetcore/PlatformBenchmarks/BenchmarkApplication.cs
```diff
@@ -63,6 +64,10 @@ public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> targe
                 {
                     requestType = RequestType.Fortunes;
                 }
+                else if (Paths.SingleQuery.Length <= path.Length && path.StartsWith(Paths.SingleQuery))
```
sebastienros: Should we order these length tests from the shortest to the biggest paths?

benaadams: Done

fafhrd91: Do you guys see performance difference compared to simple string comparison?

benaadams: The Length test is a simple memory test, the StartsWith is a non-inlining call - it probably won't be a huge factor, but why make the call for everything its not?

PR #3780, opened 2018-05-24
 update netty to version 4.1.25 and freeze dependency versions 

lneves:
update netty to version 4.1.25 and freeze dependency versions


michaelhixson:
@nbrady-techempower Do you know why Travis thinks it needed to test so many frameworks in that second run?  Is it because of the merge commit?

nbrady-techempower:
@michaelhixson The differ looks at the current state of the pr branch vs the target branch. The most recent commit is still outdated vs master.

PR #3781, opened 2018-05-25
Add "ASP.NET Core, Middleware, Utf8Json"

kiyoaki:
[Utf8Json](https://github.com/neuecc/Utf8Json) is JSON serializer 5x - 7x faster than JSON.NET.

8 integer field object benchmark(lower is better)")


sebastienros:
We haven't measure any major performance gain outside of Platform tests with TechEmpower scenarios.
But let me try to benchmark this branch on Citrine before we merge it.

frameworks/CSharp/aspnetcore/benchmark_config.json
```diff
@@ -75,6 +75,25 @@
       "notes": "",
       "versus": "aspcore"
     },
+    "mw-utf8json": {
+      "plaintext_url": "/plaintext",
```
benaadams: Shouldn't have `plaintext_url` only `json_url` would be no variation between `aspcore-mw` and `aspcore-mw-utf8json`; so add no value (as its an alternative json rather than plaintext test being added).

Also will probably fail as the Plaintext scenario is not added in the docker

sebastienros:
Measure +10% over the JSON.NET variant.

LGTM

tmds:
@sebastienros some devs will use what you pick here as a guideline for their application. I think it's best there is some clarity why you are choosing a specific library. With json.net that was obviously the de facto lib. Now it's less clear.

tmds:
No reason a different lib can't participate. Json.net is also 3rd party oss. Just pointing out your picking one of multiple options and you may want to motivate that choice.

kiyoaki:
I chose in view of serialization performance because here is a framework benchmarks repository.

sebastienros:
> your picking one of multiple options

@tmds I am not sure you understand that we are already using Json.net in most benchmarks. This PR is only __adding__ another scenario with utf8json so users can compare the perf between the two. And this one explicitly because it's the fastest on. NET.

tmds:
@sebastienros thank you for the clarification. I thought you had replaced json.net with Utf8Json completely for the json benchmark.

benaadams:
https://www.techempower.com/benchmarks/#section=test&runid=60fab9eb-a5ad-49cb-aefe-24ad0f377122&hw=ph&test=json&l=hra0hp

image

PR #3793, opened 2018-05-28
Peachpie update .NET runtime, Server GC

benaadams:
Tune Peachpie .NET enviorment
* Server GC: `Microsoft.NET.Sdk` -> `Microsoft.NET.Sdk.Web`
* netcoreapp2.0 -> netcoreapp2.1
* Double threadpool mins
* Turn off ResponseBuffering, DefaultFiles, StaticFiles (not required for these tests)
* Update peachpie.dockerfile build process to match aspcore's


benaadams:
Locally seeing a x2.8 - x5.4 increase in throughput with these changes depending on test (though much higher for `query`)

jakubmisek:
Nice! @roberthusak please merge. I would also change output encoding from UTF8 to ASCII

roberthusak:
@benaadams Wow, that look great, thanks a lot! It's interesting that turning off response buffering actually helped. On my local network it seemed that not having the ```Content-Length``` header slowed the processing about 10x in ```plaintext``` and 2x-4x in the other tests. It looked as if it was more complicated to end the connection without it. But maybe I missed the bigger picture back then. Don't you know how this could have been caused?

I gladly accept the PR, now it's up to TechEmpower to merge it.

benaadams:
>  it seemed that not having the Content-Length header slowed the processing...

Easy enough, added a commit to add it to `plaintext`, `json`, `query`, `db`, `update` but not `fortune` as would make it non-idiomatic php (would have to first write the html to a string to get the length, add content-length header, then echo the string). So `fortune` will still be chunked.

> It looked as if it was more complicated to end the connection without it.

Kestrel will automatically chunk the request if you start writing to the body without first setting the Content-Length header, then add the chunked finish when your request finishes; so it should all be taken care of. If you sent the Content-Length header then it won't chunk the output; however it is important that the length does agree with length of data written.

> Don't you know how this could have been caused?

Might also be from the load generator/parsing end (e.g. wrk) as it needs to do more work to workout when the request ends by putting together the chunk lengths - so the load generator machine needs to not be under powered?

roberthusak:
> Easy enough, added a commit to add it to ```plaintext```, ```json```, ```query```, ```db```, ```update``` but not ```fortune``` as would make it non-idiomatic php...

OK, seems good. Thanks.

> ... so the load generator machine needs to not be under powered?

Yeah, that might have happened - I used ApacheBench from my laptop to generate the load.

PR #3794, opened 2018-05-29
add: json tests "mofuw", and some update and edit.

2vg:
added a json test to mofuw.
Also, deleted some unnecessary code.

By the way, I feel sad because mofuw seems to not start up properly at tfb's benchmark ; - ;


2vg:
fixed the problem, so can merge it.
```plain
before -> after
"text/plain;" -> "text/plain"
"application/json;" -> "application/json"
```

PR #3797, opened 2018-05-29
Cutelyst 2 4 0

dantti:


dantti:
This PR is important for pre-fork benchmarks as previous cutelyst-wsgi CPU affinity logic was binding each process to CPU 0

PR #3801, opened 2018-05-30
Move swoole implementation to its own directory

michaelhixson:
Swoole seems to be a framework and it doesn't share any code with the
PHP/php directory, which is for "raw", framework-less PHP.


michaelhixson:
So you don't wonder where your Swoole code went, FYI @Wulfklaue I'm moving it to its own directory.

PR #3802, opened 2018-05-30
ASP.Net Core add platform dataupdates

benaadams:
Bringing aspnetcore platform in line with Mono improvements

/cc @sebastienros


sebastienros:
Should be top 10 in updates after this PR.

sebastienros:
This is failing because of Nancy

benaadams:
`The specified framework 'Microsoft.AspNetCore.App', version '2.1.0-rc1-final' was not found.`

Is what you get for specifying a version and then you making a release :)

benaadams:
@nbrady-techempower is it possible to get this merged before next run? 

It also fixes the `nancy-netcore` framework which will be otherwise be broken due to a docker image update

PR #3806, opened 2018-05-31
Actix: unify and optimize platform benchmarks

fafhrd91:
* unify actix-raw and actix-raw-db
* optimize memory usage for platform benchmarks


fafhrd91:
@nbrady-techempower is it possible to get this merged before next rin?

PR #3807, opened 2018-05-31
[ci skip] Add SQL Server (on linux) as a database type

### LINKED ISSUE #3786: Add SqlServer on Linux DB?
Could a SqlServer DB be added?

Based on docker `FROM microsoft/mssql-server-linux:latest`

michaelhixson:
I added [ci skip] because this requires a new techempower/tfb image, so
the travis build would fail I think.

I don't know whether this will work out because of licensing.  See the
places where I put "ACCEPT_EULA=Y" in the dockerfiles.  I don't know what
EULA that refers to and it's possible we won't be able to accept it.
I put this together anyway just to see if it's technically possible (it
is) and to get review feedback.

I added a variant of gemini that uses SQL Server and passes all tests.

I did not measure performance at all or do any performance tuning on the
database side.

Fixes #3786

### DIFF
--- a/toolset/databases/sqlserver/create.sql
+++ b/toolset/databases/sqlserver/create.sql
@@ -1,0 +1,52 @@
+create database hello_world;
+go
+use hello_world;
+go
+create table world (
+  id int not null identity(1,1) primary key,
+  randomnumber int not null
+);
+go
+set nocount on;
+go
+declare @i int = 0;
+while @i < 10000
+begin
+  insert into world (randomnumber) values (1 + cast(rand(checksum(newid())) * 10000 as int));
+  set @i = @i + 1;
+end;
+go
+create table fortune (
+  id int not null identity(1,1) primary key,
+  message nvarchar(2048) not null
+);
+go
+insert into fortune (message) values
+(N'fortune: No such file or directory'),
+(N'A computer scientist is someone who fixes things that aren''t broken.'),
+(N'After enough decimal places, nobody gives a damn.'),
+(N'A bad random number generator: 1, 1, 1, 1, 1, 4.33e+67, 1, 1, 1'),
+(N'A computer program does what you tell it to do, not what you want it to do.'),
+(N'Emacs is a nice operating system, but I prefer UNIX. — Tom Christaensen'),
+(N'Any program that runs right is obsolete.'),
+(N'A list is only as strong as its weakest link. — Donald Knuth'),
+(N'Feature: A bug with seniority.'),
+(N'Computers make very fast, very accurate mistakes.'),
+(N'<script>alert("This should not be displayed in a browser alert box.");</script>'),
+(N'フレームワークのベンチマーク');
+go
+set nocount off;
+go
+exec sp_configure 'contained database authentication', 1
+go
+reconfigure
+go
+alter database hello_world set containment = partial;
+go
+create login benchmarkdbuser with password = 'benchmarkdbpass', check_policy = off;
+go
+create user benchmarkdbuser for login benchmarkdbuser;
+go
+grant select, update on world to benchmarkdbuser;
+grant select on fortune to benchmarkdbuser;
+go
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,20 @@
-FROM ubuntu:16.04
+FROM buildpack-deps:xenial
 
-# One -q produces output suitable for logging (mostly hides
-# progress indicators)
-RUN apt update -yqq
+RUN apt update -yqq && apt install -yqq software-properties-common apt-transport-https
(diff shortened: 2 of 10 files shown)


benaadams:
Any reason for installing SqlServer via `apt install` verses using SqlServer docker container? https://hub.docker.com/r/microsoft/mssql-server-linux/

Just curious

> I don't know whether this will work out because of licensing. See the places where I put "ACCEPT_EULA=Y" in the dockerfiles.  I don't know what EULA that refers to and it's possible we won't be able to accept it.

Going from the SqlServer docker hub docs https://hub.docker.com/r/microsoft/mssql-server-linux/

> > By passing the value "Y" to the environment variable "ACCEPT_EULA", you are expressing that you have a valid and existing license for the edition and version of SQL Server that you intend to use. You also agree that your use of SQL Server software running in a Docker container image will be governed by the terms of your SQL Server license.

Is also a short EULA linked on the docker page (link at top)

If you don't specify a `MSSQL_PID` then you are installing `Developer` edition

> > `MSSQL_PID=<your_product_id | edition_name> (default: Developer)`

Info on developer edition from docker page

> > SQL Server Developer edition lets developers build any kind of application on top of SQL Server. It includes all the functionality of Enterprise edition, but is licensed for use as a development and test system, not as a production server. 

Going from the [SQL Server pricing page](https://www.microsoft.com/en-us/sql-server/sql-server-2017-pricing#CP_StickyNav_1)

> **Developer edition**
> Full-featured version of SQL Server software that allows developers to cost-effectively build, test, and demonstrate applications based on SQL Server software.
>
> Open no level price (US$) - **Free**


**Please note**, this is not legal advice :)

toolset/databases/sqlserver/create.sql
```diff
@@ -0,0 +1,52 @@
+create database hello_world;
+go
+use hello_world;
+go
+create table world (
+  id int not null identity(1,1) primary key,
```
benaadams: memory optimised? (using instructions from https://docs.microsoft.com/en-us/sql/relational-databases/in-memory-oltp/demonstration-performance-improvement-of-in-memory-oltp?view=sql-server-2017)
```sql
create database hello_world;
go
alter database hello_world add filegroup hello_world_mod contains memory_optimized_data;
alter database hello_world add file
 (name = hello_world_dir, filename= '/sqlserver/hello_world_dir')
 to filegroup hello_world_mod;
go
use hello_world;
go
create table world (
  id int not null identity(1,1) primary key nonclustered hash with (bucket_count = 70000),
  randomNumber int not null
) with (memory_optimized = on, durability = schema_and_data);
go
```

toolset/databases/sqlserver/create.sql
```diff
@@ -0,0 +1,52 @@
...
+go
+declare @i int = 0;
+while @i < 10000
+begin
+  insert into world (randomnumber) values (1 + cast(rand(checksum(newid())) * 10000 as int));
+  set @i = @i + 1;
+end;
+go
+create table fortune (
+  id int not null identity(1,1) primary key,
```
benaadams: memory optimised?
```sql
create table fortune (
  id int not null identity(1,1) primary key nonclustered hash with (bucket_count = 100),
  message nvarchar(2048) not null
) with (memory_optimized = on, durability = schema_and_data);
go
```

toolset/databases/sqlserver/create.sql
benaadams: This could also be `durability = schema_only`; but then it doesn't persist changes to disk and they are lost across restarts - so that seems more niche than general purpose (e.g. cache/temp data)

michaelhixson:
@benaadams Thanks for the comments, by the way.  I've been pulled onto another project so  there might be some delay before I get back to this, but when I do I'll try out your suggestions.

benaadams:
@michaelhixson no probs; as its a new introduction, am sure authors would want a several iterations of CI before the figures showing up in an official round 😄 (and I assume will be a while before round17)

toolset/databases/sqlserver/create.sql
benaadams: Probably should also be simple recovery mode, so:
```sql
create database hello_world;
go
alter database hello_world set recovery simple; 
go
alter database hello_world add filegroup hello_world_mod contains memory_optimized_data;
alter database hello_world add file
 (name = hello_world_dir, filename= '/sqlserver/hello_world_dir')
 to filegroup hello_world_mod;
go
use hello_world;
go
create table world (
  id int not null identity(1,1) primary key nonclustered hash with (bucket_count = 70000),
  randomNumber int not null
) with (memory_optimized = on, durability = schema_and_data);
go
```

michaelhixson:
I'm closing this PR because of the licensing issues.  I'm not confident those issues will ever resolve in our favor.

PR #3808, opened 2018-05-31
[DO-NOT-MERGE] Request for comments

redknightlois:
Implementation of the full test suite on both MVC and Middleware configurations.
Docker image for a database server running on Windows with an embedded developer license (supports up to 9 cores with 36Gb of memory).

cc @bhauer

Still missing:
- Setting up the docker instances for RavenDB on Linux (toolset directory)
- Run TechEmpower in verify mode.
cc @aviviadi @ayende @gregolsky


msmith-techempower:
I do not see the Dockerfile for the database server mentioned in the original post. What am I missing?

redknightlois:
Hi Mike,

For now only running on windows: 

https://github.com/Corvalius/FrameworkBenchmarks/blob/647f0db808a7b363ef2949254144054585c69d6c/toolset/databases/ravendb-win/raven-win.dockerfile

msmith-techempower:
AHHHH, thank you. I am just blind as a bat \>_<

frameworks/CSharp/aspnetcore/Benchmarks/Controllers/MultipleQueriesController.cs
```diff
@@ -11,6 +12,16 @@ namespace Benchmarks.Controllers
     [Route("mvc/queries")]
     public class MultipleQueriesController : Controller
     {
+        private readonly RavenDb _db;
+
+        public MultipleQueriesController()
+        { }
+
+        public MultipleQueriesController(RavenDb provider)
```
sebastienros: Can't you implement `IDb` and use the same pattern as the other calls?

redknightlois: Tried, but the model has to be different. RavenDB does not support integer keys among other small different things. We can build a dedicated controller if you are worried about the impact that extra instance variable may have on the other controllers.

sebastienros: Right, that would be better. Also I can measure the impact it has on other scenarios (it can't have any) and get preliminary results with our infrastructure so you can iterate more quickly, based on your branches.

redknightlois: @sebastienros Done.

frameworks/CSharp/aspnetcore/Benchmarks/Controllers/FortunesController.cs
```diff
@@ -11,6 +11,16 @@ namespace Benchmarks.Controllers
     [Route("mvc/fortunes")]
     public class FortunesController : Controller
     {
+        private readonly RavenDb _db;
+
+        public FortunesController()
+        { }
+
+        public FortunesController(RavenDb provider)
```
ayende: Shouldn't this be used via `HttpContext.RequestServices.GetRequiredService`, like the others?

frameworks/CSharp/aspnetcore/Benchmarks/Data/RavenDb.cs
```diff
@@ -0,0 +1,135 @@
...
+        public RavenDb(IRandom random, IOptions<AppSettings> appSettings)
+        {
+            _random = random;
+            _connectionString = appSettings.Value.ConnectionString;
+
+            _identifiers = new string[10001];
+            for (var i = 0; i < 10001; i++)
+                _identifiers[i] = $"{_random.Next(1, 10001)}";
+
+            // We need to use different databases because we id's are integer and RavenDB does not have tables, it has collections
```
ayende: Why not use a id prefix, instead?

frameworks/CSharp/aspnetcore/Benchmarks/Data/RavenDb.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        public async Task<List<FortuneRaven>> LoadFortunesRows()
+        {
+            using (var session = _fortuneStore.OpenAsyncSession())
+            {
+                var list = new List<FortuneRaven>();
+
+                // TODO: We need to have an static index and stream it completely. 
+                var query = session.Advanced
+                                   .AsyncRawQuery<FortuneRaven>("from Fortune order by Id");
```
ayende: I think you can drop the `order by Id` here, this is the natural sort order when querying by collection only.

frameworks/CSharp/aspnetcore/Benchmarks/Middleware/MultipleUpdatesRavenMiddleware.cs
```diff
@@ -0,0 +1,63 @@
...
+            _db = provider;
+        }
+
+        public async Task Invoke(HttpContext httpContext)
+        {
+            if (httpContext.Request.Path.StartsWithSegments(_path, StringComparison.Ordinal))
+            {
+                var count = MiddlewareHelpers.GetMultipleQueriesQueryCount(httpContext);
+
+                var rows = await _db.LoadMultipleUpdatesRows(count);
```
ayende: Is there specific format required? We can use the streaming option to pass the raw results directly to the client, bypassing any work on the server side.

frameworks/CSharp/aspnetcore/Benchmarks/Middleware/SingleQueryRavenMiddleware.cs
```diff
@@ -0,0 +1,61 @@
...
+            _db = provider;
+        }
+
+        public async Task Invoke(HttpContext httpContext)
+        {
+            if (httpContext.Request.Path.StartsWithSegments(_path, StringComparison.Ordinal))
+            {               
+                var row = await _db.LoadSingleQueryRow();
+
+                var result = JsonConvert.SerializeObject(row, _jsonSettings);
```
ayende: Same here, we can skip this cost by piping the results directly

frameworks/CSharp/aspnetcore/Benchmarks/Data/RavenDb.cs
redknightlois: I was under the impression that it would be lexicographic instead. Gonna check that.

frameworks/CSharp/aspnetcore/Benchmarks/Middleware/MultipleUpdatesRavenMiddleware.cs
redknightlois: Yes, format is specified. There are even ordering issues AFAIK for streaming. Multiple updates is actually a Bulk Insert, haven't code it like that yet, so no streaming.

ayende: Might be better to see if we can do that directly via a query and stream it, yen.

frameworks/CSharp/aspnetcore/Benchmarks/Data/RavenDb.cs
ayende: It is lexical anyway

frameworks/CSharp/aspnetcore/Benchmarks/Controllers/FortunesController.cs
redknightlois: Not really, unless we want to introduce a scope factory.

frameworks/CSharp/aspnetcore/Benchmarks/Data/RavenDb.cs
redknightlois: Yes, you are right there... I may be able to simplify it further

redknightlois:
Hi guys, anything that we may know about this?

bhauer:
@redknightlois Sorry, we've been distracted trying to get Round 16 wrapped up. We'll try to follow up soon!

redknightlois:
@sebastienros Done. Last commit would give you a linux dockerfile

michaelhixson:
@redknightlois What's the state of this PR from your perspective?  Are you looking for guidance from TechEmpower to get the toolset actually using the new database?

I have an open PR where I added a new database implementation (SQL Server on linux) to the toolset, where a framework is using the new database and passing tests.  See: https://github.com/TechEmpower/FrameworkBenchmarks/pull/3807

See especially the changes I had to make in the `toolset/benchmark` and `toolset/utils` directories.  We would need similar changes here to make the toolset recognize RavenDB.

I also had to change our main root `Dockerfile` to make the necessary Python package available to the toolset so that it was able to talk to the new database.  If you need to make changes in that file, which is the source for the `techempower/tfb` image, then Travis isn't going to be able to test it successfully because Travis pulls the image from here instead of building it: https://hub.docker.com/r/techempower/tfb/ 

You could rebuild the `techempower/tfb` image and test it locally though.  That's what I did for my PR.

redknightlois:
> What's the state of this PR from your perspective? Are you looking for guidance from TechEmpower to get the toolset actually using the new database?
We are setting the groundwork to make it part of the TechEmpower framework. What I am looking for is guidance on both how to integrate everything and also on the benchmark specifically to know for sure if we are not tripping on something that could either give us an unfair advantage or disadvantage against the rest. 

I will look into #3807 to replicate the changes that we would need for it. If I have any question I will ask here.

msmith-techempower:
@michaelhixson Does this require a patch from us? I am a bit confused on the status of this pull request.

redknightlois:
@msmith-techempower I didn't continue it yet. I expect to have what I need to change from #3807, in any way if you prefer, you can close the Request for comments PR and I will create a new one when we continue with this.

PR #3811, opened 2018-06-01
Updated to v2.2.5 included mruby-json

proyb6:
Performant and well maintained C libraries than mruby-iijson which Yannick found an increase in 2-30x throughputs in small JSON test.

Manually update the dependency is preferred as a replacement.
https://github.com/mattn/mruby-json

https://github.com/h2o/h2o/pull/1668


proyb6:
@nbrady-techempower is it possible to get this PR merged?

PR #3815, opened 2018-06-02
Comparison with None performed with equality operators.

MatheusRV:


nbrady-techempower:
thanks for all the idiomatic fixes @MatheusRV - in the future, one PR would be great if you could, but definitely appreciate it :)

PR #3818, opened 2018-06-03
Feature/added crystal raze

CaDs:
Hello.

Added a missing framework for crystal, [Raze](https://github.com/samueleaton/raze)


CaDs:
@nbrady-techempower thanks a lot, you are right.

Looks like it was being ignored. I added it.

PR #3827, opened 2018-06-04
Update README.md

MatheusRV:


nbrady-techempower:
issuestats always has a hard time loading badges, so it often makes the readme look poor. so im going to hold off on this for now.

MatheusRV:
at this PR i remove all issuestats an link with shields.io, that have a nice time response.

nbrady-techempower:
this is what i see when i load your changes.. it varies, but:

image

PR #3838, opened 2018-06-06
Added Swoole Postgres

Wulfklaue:
We can now dual test mysql and postgres results.


Wulfklaue:
When the test is completed, the following code can be merged. Temporary disabled the postgres test until the main dev can fix a few issues. The code itself can be merged already as it fixes a few items with the mysql test. Thanks...

PR #3839, opened 2018-06-06
Add cached_query test for Ur/Web

achlipala:
I'm taking the lazy way out and letting Travis figure out if I set this up correctly!  I only ran tests manually with `wrk` locally, to check that this setup basically makes sense.

And, yes, the Ur/Web compiler really does have a magic command-line argument to add caching automatically!  I hope I set up these new dockerfiles to trigger that feature correctly (only for the new test).


frameworks/Ur/urweb/benchmark_config.json
```diff
@@ -39,6 +39,36 @@
...
+      "os": "Linux",
+      "database_os": "Linux"
+    },
+    "mysql-cache": {
+      "setup_file": "setup-mysql",
+      "cached_query_url": "/queries/",
+      "port": 8080,
+      "approach": "Realistic",
+      "classification": "Fullstack",
+      "database": "Postgres",
```
nbrady-techempower: Changing this to MySQL should get you there :)

achlipala:
Thanks for the suggestion!  Everything seems to be working properly now, though it looks like Travis happens to have gotten stuck on something unrelated to my changes.

achlipala:
Thanks!  Tagging @vizziv, the main author of the Ur/Web caching functionality, who might be interested to watch for comparison results against other frameworks.

P.S.: Our caching support is significantly fancier than this test lets us show off.  For instance, for the Fortunes test, we'd quickly converge to serving the full page contents from a monolithic cache constructed on the first hit.  It would be cool to have expanded tests of this kind at some point!

PR #3840, opened 2018-06-07
Update to Wildfly 13.0.0.Final

codylerum:


michaelhixson:
The Travis build failed because the `Server` header is no longer being included in the responses.  We require that header in all of our tests.  Do you know how to make Wildfly 13 add the `Server` header to its responses?

codylerum:
Guessing it was removed as the default. I don't know the best way to get it enabled with the maven build. Maybe @stuartwdouglas knows?

michaelhixson:
Here is something that works.  In `wildfly-ee7.dockerfile`, you could add these two lines before the `CMD` line:

```
RUN sed -i '/subsystem xmlns="urn:jboss:domain:undertow:/a <filters><response-header name="server-header" header-name="Server" header-value="WildFly"\/><\/filters>' target/wildfly-13.0.0.Final/standalone/configuration/standalone.xml
RUN sed -i '/host name="default-host"/a <filter-ref name="server-header"\/>' target/wildfly-13.0.0.Final/standalone/configuration/standalone.xml
```

This is editing the WildFly application's configuration file, which defaults to `standalone.xml`.  Within that file, it's searching for the Undertow subsystem settings and declaring a new `response-header` filter.  Then it's enabling that new filter in the `<host>` part of the Undertow settings.  (The `/a` in the `sed` commands means it's inserting a new line after the matching string.)

It's a bit brittle to changes in the source `standalone.xml` file, so I'm not sure it will survive the next WildFly upgrade.  Maybe it's still better than copying that entire `standalone.xml` file into this project and maintaining our own copy.

codylerum:
@michaelhixson I have this reworked to use the wildfly-maven-plugin, but we are hitting

https://issues.jboss.org/browse/WFMP-94 due to a change in JDK 8. It runs fine on 1.8.0_144 but was broken by a change shortly after. Apparently they have a fix but just waiting for a new release. I'll follow up once there is a new release for wildfly-maven-plugin

codylerum:
@michaelhixson I made some changes to wrk around the plugin issue but I keep getting a networking error when starting up within docker. Any ideas? Works locally.

michaelhixson:
I can't try this locally right this second, but where you currently have these as your last three lines in the dockerfile:

```
CMD mvn clean install -q -P create-enviroment,benchmark
CMD target/wildfly-13.0.0.Final/bin/jboss-cli.sh --connect --commands='/subsystem=undertow/configuration=filter/response-header=server-header:add(header-name="Server", header-value="wfly")'
CMD target/wildfly-13.0.0.Final/bin/jboss-cli.sh --connect --commands='/subsystem=undertow/server=default-server/host=default-host/filter-ref=server-header:add()' 
```

Try changing all that to this:

```
RUN mvn clean install -q -P create-enviroment,benchmark
RUN target/wildfly-13.0.0.Final/bin/jboss-cli.sh --connect --commands='/subsystem=undertow/configuration=filter/response-header=server-header:add(header-name="Server", header-value="wfly")'
RUN target/wildfly-13.0.0.Final/bin/jboss-cli.sh --connect --commands='/subsystem=undertow/server=default-server/host=default-host/filter-ref=server-header:add()' 
CMD target/wildfly-13.0.0.Final/bin/standalone.sh -b 0.0.0.0
```

(Use `RUN` for steps that are supposed to happen during the build process and then re-add the `CMD` for starting up the HTTP server.)

I am assuming those `jboss-cli.sh` commands are doing the same thing as the `sed` commands I suggested earlier but in a more elegant way.  That is, they're changing the application's config file, and that's it.

codylerum:
I'll give that a spin. The cli commands are editing the configuration of the running server so those should be less brittle.

There is no need for the `CMD target/wildfly-13.0.0.Final/bin/standalone.sh -b 0.0.0.0` as the `RUN mvn clean install -q -P create-enviroment,benchmark` command leaves a running server with the application deployed.

nbrady-techempower:
@codylerum `RUN` will never leave any application or service running. It does its work in an intermediate container, finishes, and then starts the next `RUN` command. You get the side effects to the file system, but no applications or services are up and running. 

The last line should be the `CMD` which is the one that should be running the service you need and there should only be one `CMD`

codylerum:
@michaelhixson OK. I think I have things about sorted out, how do I got about getting wget? Doesn't seem like apt-get works in this docker enviroment

nbrady-techempower:
You need to do an `apt update` before any `apt install`

codylerum:
@nbrady-techempower @michaelhixson 

OK I'm close, but what needs to happen here is I need to start the server with 

`target/wildfly-13.0.0.Final/bin/standalone.sh -b 0.0.0.0 &`

and the execute the `target/wildfly-13.0.0.Final/bin/jboss-cli.sh --connect` commands

But the server wildfly server has to be started before the jboss-cl.sh commands can execute. Seems like the server is shut down after the end of the `RUN target/wildfly-13.0.0.Final/bin/standalone.sh -b 0.0.0.0 &` so I'm not sure the right way to handle that with docker.

michaelhixson:
The general approach you're trying with the .sh script should work.  I was going to suggest separating multiple commands within the `CMD` with `&` and `&&`, but the script may be easier to maintain.

You may need to add something like `sleep infinity` at the end of the script so that it never exits.  I think if the `CMD` exits then the container shuts down.

codylerum:
This should be good to merge now

michaelhixson:
Thank you for seeing it through.

PR #3841, opened 2018-06-07
Optimize test script

twose:
Strong type and PSR code style.
Reuse prepared statement to Improve performance.


nbrady-techempower:
@Wulfklaue would you mind reviewing

twose:
@Wulfklaue Thank you very much, due to the time pressure, the code about statement isn't elegant, the rest of the work will depend on you 😀

PR #3844, opened 2018-06-07
Add falcon and tidy up Ruby rack configurations

ioquatix:
This seems to work well for me on my testing platform, except have some issues with db concurrency test.

It seems like for higher concurrency level, MySQL is not configured to accept large number of incoming connections. Rack doesn't have load balancing so it fails.

What's the best solution for this? Should I change Rack to use persistent connections?


ioquatix:
Do you mind telling me why they are failing?

https://tfb-status.techempower.com/results/bebca46e-899f-4958-8052-d5f5f5c81eb2

Is it due to the large volume of requests and the MySQL persistent connections?

PR #3848, opened 2018-06-08
Minor optimisations to Elixir/Phoenix implementation

michalmuskala:
* Completely disable as much logging as we can
* Remove unneccessary middlewares
* Switch to a different, faster, json encoder

### DIFF
--- a/frameworks/Elixir/phoenix/web/controllers/page_controller.ex
+++ b/frameworks/Elixir/phoenix/web/controllers/page_controller.ex
@@ -6,20 +6,20 @@ defmodule Hello.PageController do
   def index(conn, _params) do
     conn
     |> put_resp_content_type("application/json", nil)
-    |> send_resp(200, Poison.encode!(%{"TE Benchmarks\n" => "Started"}))
+    |> send_resp(200, Jason.encode_to_iodata!(%{"TE Benchmarks\n" => "Started"}))
   end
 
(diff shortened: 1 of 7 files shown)


frameworks/Elixir/phoenix/web/controllers/page_controller.ex
```diff
@@ -60,16 +60,17 @@ defmodule Hello.PageController do
...
+    data = for _ <- 1..q do
       id = :rand.uniform(10000)
       num = :rand.uniform(10000)
       w = Repo.get(World, id)
-      changeset = World.changeset(w, %{randomnumber: num})
-      Repo.update(changeset)
-      %{id: id, randomnumber: num}
-    end)))
+      changeset = Ecto.Changeset.change(w, randomnumber: num)
+      Repo.update!(changeset)
```
leifg: Can't this be optimized by a single update query?

```
from w in World, where: w.id == ^id,
update: [set: [randomnumber: ^num]]
```

fafhrd91: this is benchmark requirement

michalmuskala: It's very likely there are more optimisation opportunities, but those have to be explored with some more careful profiling.

PR #3849, opened 2018-06-08
Update the Play Framework projects to more recent version

KoenDG:
Update the configurations of the play framework to their current latest version: 2.6.15

Also update the dockerfiles to use a more recent base, allowing use of Scala 2.12.6 and sbt 1.1.5

Apart from that, I noticed some docker files providing extra JVM options. Considering these are all using the same framework, I feel either all dockerfiles should use it or non of them. Otherwise, the comparison between them is off.

Lastly, I added some code to the dockerfile to properly disable the http server that's not being used in the build that will be done by the dockerfile.
For instance: when running netty server, it's perfectly fine to disable the AkkaHttpPlugin. And vice versa. This prevents their jars from being put on the classpath.

### DIFF
--- a/frameworks/Java/play2-java/play2-java-ebean-hikaricp-netty.dockerfile
+++ b/frameworks/Java/play2-java/play2-java-ebean-hikaricp-netty.dockerfile
@@ -1,5 +1,8 @@
-FROM hseeberger/scala-sbt:8u151-2.12.5-1.1.2
+FROM hseeberger/scala-sbt:8u171_2.12.6_1.1.5
 WORKDIR /play2
 COPY play2-java-ebean-hikaricp .
+
+RUN sed -i 's/.enablePlugins(PlayJava, PlayEbean, PlayNettyServer)/.enablePlugins(PlayJava, PlayEbean, PlayNettyServer).disablePlugins(PlayAkkaHttpServer)/g' build.sbt
+
 RUN sbt stage
-CMD ["target/universal/stage/bin/play2-java-ebean-hikaricp", "-Dplay.server.provider=play.core.server.NettyServerProvider"]
+CMD ["target/universal/stage/bin/play2-java-ebean-hikaricp", "-Dplay.server.provider=play.core.server.NettyServerProvider", "-J-server", "-J-Xms1g", "-J-Xmx1g", "-J-XX:NewSize=512m", "-J-XX:+UseG1GC", "-J-XX:MaxGCPauseMillis=30", "-J-XX:-UseBiasedLocking", "-J-XX:+AlwaysPreTouch"]
--- a/frameworks/Java/play2-java/play2-java-ebean-hikaricp/conf/application.conf
+++ b/frameworks/Java/play2-java/play2-java-ebean-hikaricp/conf/application.conf
@@ -19,6 +19,9 @@ play.server {
   netty {
     transport = "native"
 
+    # Whether the Netty wire should be logged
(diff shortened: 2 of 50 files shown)


frameworks/Java/play2-java/play2-java-ebean-hikaricp-netty.dockerfile
```diff
@@ -1,5 +1,8 @@
-FROM hseeberger/scala-sbt:8u151-2.12.5-1.1.2
+FROM hseeberger/scala-sbt:8u171_2.12.6_1.1.5
 WORKDIR /play2
 COPY play2-java-ebean-hikaricp .
+
+RUN sed -i 's/.enablePlugins(PlayJava, PlayEbean, PlayNettyServer)/.enablePlugins(PlayJava, PlayEbean, PlayNettyServer).disablePlugins(PlayAkkaHttpServer)/g' build.sbt
+
 RUN sbt stage
-CMD ["target/universal/stage/bin/play2-java-ebean-hikaricp", "-Dplay.server.provider=play.core.server.NettyServerProvider"]
+CMD ["target/universal/stage/bin/play2-java-ebean-hikaricp", "-Dplay.server.provider=play.core.server.NettyServerProvider", "-J-server", "-J-Xms1g", "-J-Xmx1g", "-J-XX:NewSize=512m", "-J-XX:+UseG1GC", "-J-XX:MaxGCPauseMillis=30", "-J-XX:-UseBiasedLocking", "-J-XX:+AlwaysPreTouch"]
```
KoenDG: I found this being used on 2 of the scala versions and nowhere else. So I added it everywhere. For comparison's sake, all configurations for the same framework should use the same parameters as much as possible.

frameworks/Java/play2-java/play2-java-ebean-hikaricp-netty.dockerfile
```diff
@@ -1,5 +1,8 @@
-FROM hseeberger/scala-sbt:8u151-2.12.5-1.1.2
+FROM hseeberger/scala-sbt:8u171_2.12.6_1.1.5
```
KoenDG: Most recent version

frameworks/Java/play2-java/play2-java-ebean-hikaricp/conf/application.conf
```diff
@@ -19,6 +19,9 @@ play.server {
   netty {
     transport = "native"
 
+    # Whether the Netty wire should be logged
+    log.wire = false
+
```
KoenDG: Added this to all `application.conf` files. Disabling a logging setting, as it's not needed. And then 2 settings which, according to the documentation, should have better performance.

frameworks/Java/play2-java/play2-java-ebean-hikaricp-netty.dockerfile
```diff
@@ -1,5 +1,8 @@
...
 WORKDIR /play2
 COPY play2-java-ebean-hikaricp .
+
+RUN sed -i 's/.enablePlugins(PlayJava, PlayEbean, PlayNettyServer)/.enablePlugins(PlayJava, PlayEbean, PlayNettyServer).disablePlugins(PlayAkkaHttpServer)/g' build.sbt
```
KoenDG: Disabling server that isn't actually going to run, this is the netty dockerfile, netty is going to act as the http server here, not akka.

PR #3859, opened 2018-06-12
FSharp giraffe

dv00d00:
Added FSharp language with giraffe implementation of plaintext, json, and jsonutf8 endpoints


frameworks/FSharp/giraffe/src/App/Program.fs
```diff
@@ -0,0 +1,72 @@
...
+    let filter (l : LogLevel) = l.Equals LogLevel.Error
+    builder.AddFilter(filter).AddConsole().AddDebug() |> ignore
+
+[<EntryPoint>]
+let main _ =    
+    WebHostBuilder()
+        .UseKestrel()        
+        .Configure(Action<IApplicationBuilder> configureApp)
+        .ConfigureServices(configureServices)
+        .ConfigureLogging(configureLogging)
```
benaadams: Don't configure logging
Rule 3.
> All test implementations must disable all disk logging. For many reasons, we expect all tests will run without writing logs to disk. Most importantly, the volume of requests is sufficiently high to fill up disks even with only a single line written to disk per request. Please disable all forms of disk logging. We recommend but do not require disabling console logging as well.

Every other framework disables logging completely so having it enabled will be a disadvatage

frameworks/FSharp/giraffe/src/App/Program.fs
```diff
@@ -0,0 +1,72 @@
...
+let errorHandler (ex : Exception) (logger : ILogger) =
+    logger.LogError(EventId(), ex, "An unhandled exception has occurred while executing the request.")
+    clearResponse >=> setStatusCode 500 >=> text ex.Message
+
+// ---------------------------------
+// Config and Main
+// ---------------------------------
+
+let configureApp (app : IApplicationBuilder) =
+    let env = app.ApplicationServices.GetService<IHostingEnvironment>()
```
benaadams: Remove both error handlers (for same reason as logging)

frameworks/FSharp/giraffe/src/App/web.config
```diff
@@ -0,0 +1,9 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
benaadams: Don't need this file, is used when behind IIS

frameworks/FSharp/giraffe/src/App/Program.fs
```diff
@@ -0,0 +1,72 @@
...
+type JsonMessage = { message : string }
+
+// ---------------------------------
+// Web app
+// ---------------------------------
+
+let jsonUtf8 (data:obj) : HttpHandler =
+    fun (next : HttpFunc) (ctx : HttpContext) ->
+        let bytes = Utf8Json.JsonSerializer.Serialize(data)
+        ctx.SetContentType "application/json"
```
benaadams: Additionally set the content length (bytes.Length); otherwise it will go as chunked encoding which is slightly larger and a little more involved.

frameworks/FSharp/giraffe/src/App/App.fsproj
```diff
@@ -0,0 +1,42 @@
...
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="2.1.0" />
+    <PackageReference Include="Giraffe" Version="1.1.*" />
+    <PackageReference Include="Utf8Json" Version="1.3.7" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Include="Program.fs" />
+  </ItemGroup>
+
+  <ItemGroup>
```
benaadams: Can remove this ItemGroup (don't have to though)

frameworks/FSharp/giraffe/src/App/Program.fs
```diff
@@ -0,0 +1,72 @@
...
+        ctx.WriteBytesAsync bytes
+
+let webApp =
+    choose [
+        GET >=>
+            choose [
+                route "/plaintext" >=> text "Hello, World!"
+                route "/json" >=> json { message = "Hello, World!" }
+                route "/jsonutf8" >=> jsonUtf8 { message = "Hello, World!" }
```
benaadams: Will need to add a second dockerfile for this second json route (that's identical to first) `giraffe-utf8json.dockerfile`

Then add a new entry in `benchmark_config.json` for the json
```
"utf8json": {
    "json_url": "/jsonutf8",
   ...
```

frameworks/FSharp/giraffe/benchmark_config.json
```diff
@@ -0,0 +1,24 @@
+{
+  "framework": "giraffe",
+  "tests": [{
+    "default": {
+      "plaintext_url": "/plaintext",
+      "json_url": "/json",
+      "port": 8080,
+      "approach": "Realistic",
+      "classification": "Full",
```
benaadams: Needs to be `Fullstack`

frameworks/FSharp/giraffe/src/App/App.fsproj
```diff
@@ -0,0 +1,42 @@
...
+    <PackageReference Include="Microsoft.AspNetCore.Cors" Version="2.1.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="2.1.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Diagnostics" Version="2.1.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel" Version="2.1.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Server.IISIntegration" Version="2.1.0" />
+    <PackageReference Include="Microsoft.AspNetCore.StaticFiles" Version="2.1.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="2.1.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="2.1.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.Cookies" Version="2.1.0" />
```
benaadams: Some of these PackageReferences can be trimmed

frameworks/FSharp/giraffe/src/App/App.sln
```diff
@@ -0,0 +1,25 @@
+﻿
```
benaadams: Don't need the sln file; up to you though

frameworks/FSharp/giraffe/src/App/Program.fs
```diff
@@ -0,0 +1,72 @@
+module App.App
+
+open System
+open Microsoft.AspNetCore.Builder
+open Microsoft.AspNetCore.Hosting
+open Microsoft.Extensions.Logging
```
benaadams: Some of these can be trimmed

dv00d00:
I will need to check if docker actually able to run the app. It died on my work machine after latest windows update

frameworks/FSharp/giraffe/benchmark_config.json
```diff
@@ -0,0 +1,47 @@
...
+        "webserver": "Kestrel",
+        "os": "Linux",
+        "database_os": "Linux",
+        "display_name": "Giraffe",
+        "notes": "",
+        "versus": "aspcore"
+      }
+    },
+    {
+      "jsonutf8": {
```
benaadams: Call it `utf8json` to match the library name?

Then it will appear in the listings as `giraffe-utf8json` as per https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/CSharp/aspnetcore/benchmark_config.json#L82

Url can remain the same

frameworks/FSharp/giraffe/benchmark_config.json
```diff
@@ -0,0 +1,47 @@
...
+        "plaintext_url": "/plaintext",
```
benaadams: nit: Don't need plaintext entry for this one as nothing is different for plaintext, so no reason to test it twice

frameworks/FSharp/giraffe/src/App/Program.fs
```diff
@@ -0,0 +1,36 @@
+module App.App
+
+open Microsoft.AspNetCore.Hosting
+open Microsoft.AspNetCore.Http
+open Giraffe
+
+[<CLIMutable>]
+type JsonMessage = { message : string }
+
+let jsonutf8 (data:obj) : HttpHandler =
```
benaadams: Should this be `data: JsonMessage`?

Also is `type JsonMessage` the equivalent of a C# struct type? (rather than a class type) 

The generic Utf8Json Serialize overload for will work better for the struct type (don't change it for the regular one as Json.Net would box it to obj anyway...)

benaadams:
Some nits, but beautiful!

frameworks/FSharp/giraffe/src/App/Program.fs
dv00d00: It should be a class
Will add struct for utf8 case + make it generic

benaadams:
> I will need to check if docker actually able to run the app. It died on my work machine after latest windows update

I run it in a linux VM, seems easiest as the tests set up a bunch of docker containers.

You can verify it locally with
```
sudo ./tfb --test=giraffe --mode=verify
```
And locally benchmark (e.g. plaintext) with
```
sudo ./tfb --test=giraffe --type=plaintext
```

Are also docs on the benchmarks https://frameworkbenchmarks.readthedocs.io/

frameworks/FSharp/giraffe/giraffe-jsonutf8.dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM microsoft/dotnet:2.1-sdk-stretch AS build
```
benaadams: need to rename this docker file to match name change 
`giraffe-jsonutf8.dockerfile` -> `giraffe-utf8json.dockerfile`

benaadams:
Don't forget to remove WIP from the title :)

frameworks/FSharp/giraffe/giraffe-jsonutf8.dockerfile
dv00d00: oops

dv00d00:
Out of scope of this PR, but if, let's say, I'll port asp.net core mvc app to F#, will it make any sense?

benaadams:
>  I'll port asp.net core mvc app to F#, will it make any sense?

Maybe? Can add it as a `mvc` branch in `benchmark_config.json` with it own dockerfile as per `utf8json`; may want to have it as a separate fsprog.

The aspnetcore version is a little complicated as it has every permutation in it; then uses command line params to determine what to set up so you may want to go simplier, unless you want to switch for all the databases, different orms and middleware vs mvc.

One item to note
```csharp
route "/plaintext" >=> text "Hello, World!"
```
The aspnetcore middleware pre-converts the UTF16 string to a utf8 byte array in a static and then writes that:

https://github.com/TechEmpower/FrameworkBenchmarks/blob/72496137b0558c48208462b91bf1a159aea1fd63/frameworks/CSharp/aspnetcore/Benchmarks/Middleware/PlaintextMiddleware.cs#L37-L41

Rather than doing the conversion for each request. Isn't necessarily idiomatic, but is valid.

benaadams:
@dv00d00 might want to signal when you are complete with the PR, so the techempower peeps know when to merge

dv00d00:
will update it shortly, going to split stock and custom implementations

benaadams:
Welcome .NET Player 2 😃 

image


/cc @nbrady-techempower

frameworks/FSharp/giraffe/.gitignore
```diff
@@ -0,0 +1,38 @@
...
+*.swp
+*~
+.build/
+.testPublish/
+launchSettings.json
+BenchmarkDotNet.Artifacts/
+BDN.Generated/
+binaries/
+global.json
+*.sln
```
msmith-techempower: With the Docker implementation, `.gitignore` should not be needed.

benaadams: Makes it hard to edit the project in Visual Studio and use git without it (as VS creates lots of support files)

msmith-techempower: Fair enough

isaacabraham:
What's the CLIMutable needed for?

frameworks/FSharp/giraffe/src/App/App.fsproj
```diff
@@ -0,0 +1,27 @@
...
+    <DebugType>portable</DebugType>
+    <AssemblyName>App</AssemblyName>
+    <OutputType>Exe</OutputType>
+    <EnableDefaultContentItems>false</EnableDefaultContentItems>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="2.1.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel" Version="2.1.0" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="2.1.0" />
```
forki: Can we get rid of DI?

frameworks/FSharp/giraffe/src/App/Program.fs
```diff
@@ -0,0 +1,62 @@
...
+    | Custom -> GET >=> choose Routes.custom
+
+[<EntryPoint>]
+let main args = 
+    let implementation = 
+        match args with
+        | [| "stock" |]  -> Stock
+        | _ -> Custom
+
+    printfn "Running with %A implementation" implementation
```
forki: Printf is usually very slow. I'm not sure if it's part of the benchmark, but it should not be used in hot paths.

forki: Console.WriteLine would do

frameworks/FSharp/giraffe/src/App/Program.fs
```diff
@@ -0,0 +1,62 @@
...
+module Routes =
+
+    let stock : HttpHandler list = 
+        [ route "/plaintext" >=> text "Hello, World!"
+          route "/json" >=> json { JsonMessage.message = "Hello, World!" } ]
+
+    let custom : HttpHandler list = 
+        let inline contentLength (x:int32) = new System.Nullable<int64>( int64 x )
+
+        let json data : HttpHandler =
```
forki: I think we should use the giraffe helpers here.

frameworks/FSharp/giraffe/src/App/Program.fs
```diff
@@ -0,0 +1,62 @@
...
+            let bytes = Utf8Json.JsonSerializer.Serialize(data)
+            fun _ ctx ->
+                ctx.Response.ContentLength <- contentLength ( bytes.Length )
+                ctx.Response.ContentType <- "application/json"
+                ctx.Response.StatusCode <- 200
+                ctx.WriteBytesAsync bytes
+
+        let bytes = System.Text.Encoding.UTF8.GetBytes "Hello, World!"
+        let text : HttpHandler =
```
forki: I think we should use giraffe helpers here.

frameworks/FSharp/giraffe/src/App/Program.fs
```diff
@@ -0,0 +1,62 @@
+module App.App
+
+open Microsoft.AspNetCore.Hosting
+open Giraffe
+
+[<CLIMutable>]
```
forki: Mhm. Not 100 percent sure, but usually we don't need cli mutable. I don't have it anywhere in any of my giraffe apps

forki:
Very cool stuff. I added some smaller comments. Maybe you want to move it a bit more in direction of idiomatic giraffe.

frameworks/FSharp/giraffe/src/App/Program.fs
benaadams: Helpers are already used in stock? (idiomatic) 

Static conversions are used for custom plaintext, and utf8json lib (benchmark optimized)

Both varieties will show up in the benchmarks

frameworks/FSharp/giraffe/src/App/Program.fs
benaadams: Helpers are used for this type in stock (idiomatic)

frameworks/FSharp/giraffe/src/App/Program.fs
forki: Ah didn't see that. Thanks

forki:
I also recommend that you you try to use giraffe's tokenrouter if you didn't try it yet. It should be much faster than the "old" suave style routing

dv00d00:
thanks for the review @forki 

pritnfn is called once per app startup, so should not cause any issues.
As for TokenRouter, it does look good, but I think we can't use prerelease packages

forki:
AFAIK token router is released. @dustinmoris? @isaacabraham

Dmitry Kushnir <notifications@github.com> schrieb am Mi., 13. Juni 2018,
10:51:

gerardtoconnor:
On [NuGet ](https://www.nuget.org/packages/Giraffe.TokenRouter/) it's saying it's a beta pre-release, unfortunately :( . I don't release many NuGet packages, do we remove pre-release state, by removing "beta" from project version?

Given this is a performance test we should probably avoid CE, any other fancy syntax and any unneeded runtime ops like filtering GET, sticking to explicit perf code as no-one references these benchmarks for building systems, just getting perf tips?

forki:
Please send PRs if you see improvements.


Gerard <notifications@github.com> schrieb am Mi., 13. Juni 2018, 13:37:

gerardtoconnor:
Did review in [RFC Fsharp giraffe fortunes](https://github.com/TechEmpower/FrameworkBenchmarks/pull/3863#pullrequestreview-129365479), looks good for the most part, not much more to do without changing router or implementing more explicit json writer. 👍

PR #3861, opened 2018-06-15
Add h2o.cr (Crystal bindings to h2o)

S-YOU:
Add h2o.cr (Crystal bindings to h2o)
Tests: Plaintext only (for now)
Maintainer: Me

### DIFF
--- a/frameworks/Crystal/h2o.cr/README.md
+++ b/frameworks/Crystal/h2o.cr/README.md
@@ -1,0 +1,7 @@
+# h2o.cr
+
+Crystal bindings to h2o
+
+### Plaintext
+
+h2o.cr currently only support plaintext requests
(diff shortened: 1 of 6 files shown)


faustinoaq:
@S-YOU Thank you so much! :sparkles: :smile:

michaelhixson:
We need the source code of test implementations to live in this repository.  It looks like the source currently lives here instead:  https://github.com/S-YOU/h2o.cr/blob/master/examples/h2o_evloop_hello.cr

Could you copy the implementation into this repository and change the dockerfile so that it uses that copy instead of the external copy?

S-YOU:
Thanks @michaelhixson, I have copied the file to this repository. Please take a look.

michaelhixson:
Great, thanks!

PR #3863, opened 2018-06-15
Fsharp giraffe fortunes

dv00d00:
Added fortunes test via default and custom HTML rendering. 
Orm is Dapper.

Implemented custom renderer to write HTML into memory stream.

Html renderers comparison

   Method |      Mean |     Error |    StdDev | Scaled |   Gen 0 | Allocated |
--------- |----------:|----------:|----------:|-------:|--------:|----------:|
 Standard | 105.23 us | 1.5500 us | 2.1729 us |   1.00 | 68.3594 | 105.15 KB |
   Custom |  11.70 us | 0.0697 us | 0.1044 us |   0.11 |  7.1411 |  10.99 KB |


dv00d00:
would appreciate a review @benaadams @dustinmoris @isaacabraham @gerardtoconnor @isaacabraham

frameworks/FSharp/giraffe/benchmark_config.json
```diff
@@ -37,27 +38,28 @@
...
     {
-      "utf8plaintext": {
+      "stripped": {
+        "fortune_url": "/fortunes",
         "plaintext_url": "/plaintext",
         "port": 8080,
-        "approach": "Realistic",
-        "classification": "fullstack",
-        "database": "None",
+        "approach": "Stripped",
```
benaadams: Its still `Realistic` and likely to become the defacto method once `Utf8String` becomes mainline https://github.com/dotnet/corefxlab/issues/2350

frameworks/FSharp/giraffe/src/Tests/BenchmarkTests.fsproj
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
benaadams: Benchmarks in the benchmarks? 😄

frameworks/FSharp/giraffe/benchmark_config.json
```diff
@@ -37,27 +38,28 @@
...
         "database_os": "Linux",
-        "display_name": "Giraffe",
+        "display_name": "Giraffe, utf8json",
         "notes": "",
         "versus": "aspcore"
       }
     },
     {
-      "utf8plaintext": {
+      "stripped": {
```
benaadams: `utf8direct`?

frameworks/FSharp/giraffe/benchmark_config.json
benaadams: In recent run its 1,347,868.6 rps vs 1,220,305.2 rps for `giraffe-utf8plaintext` vs `giraffe`

frameworks/FSharp/giraffe/src/App/StatefullRendering.fs
```diff
@@ -0,0 +1,60 @@
...
+        do name  |> writeEndElement target
+
+    let rec writeHtmlNode (target: StreamWriter) (node : XmlNode)  =
+        match node with
+        | EncodedText text -> do target |> add' (WebUtility.HtmlEncode text)
+        | RawText text -> do target |> add' text
+        | ParentNode (e, nodes) -> do writeParentNode target e nodes
+        | VoidElement (n, attrs) -> do writeStartElement target n attrs
+    
+    let renderHtmlToStream (encoding:Encoding) node =
```
gerardtoconnor: Why is there a separate memory stream being created rather then using the asp.net core buffers in response? The Memory stream is allocating a resizing byte array internally so writing into it, only to later flush into the response is just waste. we should look to write directly into the response body stream, example of a similar implementation to above is here [ByteViewEngine](https://github.com/gerardtoconnor/ViewEngineTests/blob/master/ByteViewEngine.fs) and outperforms most due to writing directly body stream. Is there a important reason MS is being used I have overlooked?

frameworks/FSharp/giraffe/src/App/Stock.fs
```diff
@@ -0,0 +1,42 @@
...
+        let extra' = Seq.singleton extra
+
+        fun (_ : HttpFunc) (ctx : HttpContext) ->
+            task {
+                use conn = new NpgsqlConnection(ConnectionString)
+                let! data = conn.QueryAsync<Fortune>("SELECT id, message FROM fortune")
+
+                let view = 
+                    data 
+                    |> Seq.append extra'
```
gerardtoconnor: Using Seq module operations are poor efficiency as involve a lot of temp structures, the Custom implementation of adding extra is better (using lists), why not used here? Is it to be most "intuitive" F# implementation?

frameworks/FSharp/giraffe/src/App/Stock.fs
```diff
@@ -0,0 +1,42 @@
...
+                let bytes = 
+                    view 
+                    |> GiraffeViewEngine.renderHtmlDocument 
+                    |> Encoding.UTF8.GetBytes
+
+                ctx.SetContentType "text/html;charset=utf-8"
+                return! ctx.WriteBytesAsync bytes
+            }
+
+    GET >=> choose [
```
gerardtoconnor: The "GET" Method filter is more correct but adding unneeded bloat to the pipeline when it is not a test condition is just wasted resource, I would recommend leaving the method filter out to squeeze tiny bit more perf.

frameworks/FSharp/giraffe/src/App/Stock.fs
```diff
@@ -0,0 +1,42 @@
...
+        route "/plaintext" >=> text "Hello, World!" 
+        route "/json" >=> json { JsonMessage.message = "Hello, World!" }
+        route "/fortunes" >=> fortunes
```
gerardtoconnor: The standard router is testing each route, one by one till it finds a match so this is inefficient for routes latter in the list like "/fortunes". Would be better to use TokenRouter (needs to be pushed out of beta to be used I believe), or as this is custom, create a custom routing function with a match/Dictionary on path as no complicated parsing required.

frameworks/FSharp/giraffe/src/App/Custom.fs
```diff
@@ -0,0 +1,81 @@
...
+
+    let text' (msg:string): HttpHandler = 
+        let bytes = System.Text.Encoding.UTF8.GetBytes(msg)
+        fun _ ctx ->
+            ctx.Response.ContentLength <- contentLength bytes.Length
+            ctx.Response.ContentType <- "text/plain"
+            ctx.Response.StatusCode <- 200
+            task {
+                do! ctx.Response.Body.WriteAsync(bytes, 0, bytes.Length)
+                return none
```
gerardtoconnor: as mentioned above ^^: I am guessing you are not using "text" helper function to allow caching of known string bytes? Although "none" can be returned and will work, this is generally wrong usage as the "choice" functions will continue to try match routes etc until it gets "Some ctx", therefore, to prevent testing of further routes etc, best to return "Some ctx"

frameworks/FSharp/giraffe/src/App/Custom.fs
```diff
@@ -0,0 +1,81 @@
...
+
+    let json' data : HttpHandler =
+        let bytes = Utf8Json.JsonSerializer.Serialize(data)
+        fun _ ctx -> 
+            ctx.Response.ContentLength <- contentLength bytes.Length
+            ctx.Response.ContentType <- "application/json"
+            ctx.Response.StatusCode <- 200
+            task {
+                do! ctx.Response.Body.WriteAsync(bytes, 0, bytes.Length)
+                return none
```
gerardtoconnor: I am guessing you are not using "json" helper function to allow caching of known serialized data bytes? Although "none" can be returned and will work, this is generally wrong usage as the "choice" functions will continue to try match routes etc until it gets "Some ctx", therefore, to prevent testing of further routes etc, best to return "Some ctx"

frameworks/FSharp/giraffe/src/App/Custom.fs
```diff
@@ -0,0 +1,81 @@
...
+            }
+  
+    let fortunes' : HttpHandler = 
+        let extra = {id = 0; message = "Additional fortune added at request time."}
+        let encoding = new UTF8Encoding(false)
+
+        fun (_ : HttpFunc) (ctx : HttpContext) ->
+            
+                let conn = new NpgsqlConnection(ConnectionString)
+                ctx.Response.RegisterForDispose conn
```
gerardtoconnor: Out of curiosity, why is this used rather then "task { use conn = new … }" for disposable?

frameworks/FSharp/giraffe/src/App/StatefullRendering.fs
```diff
@@ -0,0 +1,60 @@
...
+    let inline private add' (str:string) (target: StreamWriter) =
+        target.Write str
+
+    let private closingBracket = ">"
+
+    let private writeStartElement target elemName (attributes : XmlAttribute array) : unit =
+
+        match attributes with
+        | [||] -> 
+            do target
```
gerardtoconnor: why are the so many explicit "do" statements used in this module? is there an error if not explicitly included? usually "do" is only required for iteration body statements and constructor initialization code?

frameworks/FSharp/giraffe/src/App/Custom.fs
dv00d00: My reasoning was task build will create a closure for every bind internally, so I am trying to reduce this to the max.

frameworks/FSharp/giraffe/src/App/Stock.fs
dv00d00: This is probably how I would write it without thinking of perf, will replace with list bit

frameworks/FSharp/giraffe/src/App/Stock.fs
dv00d00: thx for the spot, it should go

frameworks/FSharp/giraffe/src/App/Stock.fs
dv00d00: Will definitely take a look into it, thanks.

frameworks/FSharp/giraffe/src/App/StatefullRendering.fs
dv00d00: Writing to the http out would mean I need to use task api for every write, which assuming recursive walk would result in more allocations for tasks and binds than memory stream where I could write without the need of async in the first place.

The other bit is that I am not entirely sure about the consequences of writing to the http out stream without knowing the content length upfront, I think kestrell would fallback into chunked content encoding, @benaadams might give a clue here.

gerardtoconnor: This is why I provided the example of my ByteViewEngine, it does not use async api and therefore only makes small compromise on blocking. The writing of the small chunks will be hotpath into the kestrel buffers with no blocking until the buffer fills and needs to flush, the buffer is big enough and pages small enough that there will be minimal buffer flushes, on the flushes, unlike a DB roundtrip or file read request, there is no waiting on the return so minimal blocking. The whole operation can be a task, but it is performant on a single sync thread. 

On the content length, not sure of the upfront content length omission, that is a little trickier but doable if needed. 

We can pool & reuse the memory streams, .SetLength(0) will reset it once done, a more complex concise but painful way ... If we pre-compute & Cache the template length with no variables at start, we can iterate the variables to calculate their additional size each time before writing out to the buffer. the double iteration wont add much more time as the array (List) will already be CPU cached from prior run and was just loaded anyway from DB call, importantly no extra big MS array allocs.

Either way, upfront content length setting is not simple so if anyone has insight into the chunked content encoding, would save us wasting time on any complex work arounds.

gerardtoconnor: Thanks @benaadams , this is helpful, fully agree, more often then not, DB calls/driver are the bottleneck and can make these other micro optimisations immaterial.

I think we should try avoid the chunking in principal, (despite being thin layer) it's bit lazy/cheat when pooling a MemoryStream isn't too difficult. I prefer the pooled MS over SB given there is wasted CPU & memory on converting to UTF-16 string, when the ultimate output is UTF-8, best to write & store as UTF-8 as the c# middleware test inefficiently is having to iterate & convert the SB string (utf-16) twice to first count the bytes (UTF-8), and then write the string (utf-16) to UTF-8.

The idea around pre-computing the base template size and then at runtime only calculating the incremental GetByteCount Sum of variables is probably a level of runtime complexity that will be slower then simply pooling buffers so can be ignored.

@dv00d00 do you think MemoryStream pooling way to go then, means you can leave current templating engine as is, just getting/returning MS from pool? It's probably reasonable to pre-allocate pool on start up (50 MemoryStreams of size 1024 bytes?).

benaadams: > fully agree, more often then not, DB calls/driver are the bottleneck and can make these other micro optimisations immaterial.

aspnetcore still can output ~3.12Gb/s; which is 30% of network capacity; h20 is at ~4.46Gb/s or 45%; assuming the Db transfer is similar they are both 60% - 90% of the network. Possibly for the next round that's bumping up to 100Gb/s network, so that may all change.

Also aspnetcore platform -> aspnetcore mvc on the same db driver is a 64% drop in perf; so its probably worth tweaking that last little bit out 😉

gerardtoconnor: > Also aspnetcore platform -> aspnetcore mvc on the same db driver is a 64% drop in perf; so its probably worth tweaking that last little bit out 😉

That's what we did with Giraffe, MVC being the last little bit out 😉. 

With Giraffe we will continue to micro optimise don't worry, in addition to utilizing all the biggest wins in aspnetcore, it's best of both worlds for us, you're doing all the heavy lifting!

benaadams: Results for current (merged) version of giraffe are available https://twitter.com/ben_a_adams/status/1008406784212656130

dv00d00:
Thanks for the review guys, will continue to work on it during next week.
I am a bit interested in memory stream pooling as well.

I've benchamrked pooled impl as in https://www.nuget.org/packages/Microsoft.IO.RecyclableMemoryStream/

The results are:

|             Method |      Mean |     Error |    StdDev | Scaled |  Gen 0 | Allocated |
|------------------- |----------:|----------:|----------:|-------:|-------:|----------:|
|    NewMemoryStream |  9.211 us | 0.0491 us | 0.0459 us |   1.00 | 3.5858 |  11.06 KB |
| PooledMemoryStream | 11.287 us | 0.0776 us | 0.0725 us |   1.23 | 2.5177 |   7.78 KB |

benaadams:
Could use a [`ThreadStatic` cache](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/CSharp/aspnetcore/Benchmarks/Data/StringBuilderCache.cs) with a rent and return? (Is how I'd do it in C#); the `RecyclableMemoryStream` is more for large content streams; whereas this data never gets big.

gerardtoconnor:
Updates look good.

Thanks for doing initial benchmark, I'm not too surprised, there is usually a large tipping point where these pools pay off. The high Thread lock/syncing on this is a bottleneck so Ben's suggestion is good, we need frequent single small buffers per thread rather then massive irregular ones.

Seems The standard benchmarkdotnet run will have less allocs & GC then TechEmpower , we can also tweak the pool configuration for blocksize etc. When full aspnetcore stack is running with it's memory footprint, there is less free to alloc before GC so in micro benchmarks like this, only minimal GC needed, will be higher in full application test.

I actually remembered, one of my test view engines might work better in this case [TemplateViewEngine](https://github.com/gerardtoconnor/ViewEngineTests/blob/master/TemplateViewEngine.fs), as I could tweak it to do my original complex suggestion of counting the pre-complied elements and using a GetByteCount summation on the model and add it to the pre-computed model length? bit of a radical one but just throwing out there, all static elements of template are built into a rendering tree of byte arrays, with model rendered into the gaps writing direct to body.

dv00d00:
Ported @benaadams cache 
Results:

|Method |     Mean |     Error |    StdDev |   Median | Scaled | ScaledSD |  Gen 0 | Allocated |
|---------------- |---------:|----------:|----------:|---------:|-------:|---------:|-------:|----------:|
| NewMemoryStream | 11.91 us | 0.2376 us | 0.4404 us | 11.69 us |   1.00 |     0.00 | 7.0038 |  10.78 KB |
|      CustomPool | 11.54 us | 0.2295 us | 0.4366 us | 11.49 us |   0.97 |     0.05 | 4.7760 |   7.34 KB |
|          MSPool | 14.55 us | 0.2907 us | 0.8295 us | 14.26 us |   1.22 |     0.08 | 5.0507 |   7.78 KB |

win win,  will adopt it

dv00d00:
@gerardtoconnor I would really like to try out yours TemplateViews engine. Is it published on nuget?

dv00d00:
I think it is ready.
Would appreciate a review by TechEmpower team.

Also not entirely sure about "stripped" approach for custom fortunes test, would "micro" classification be enough to justify custom html rendering? Rest of the code does not feel like something too wild?

benaadams:
> Would appreciate a review by TechEmpower team.
>
> Also not entirely sure about "stripped" approach for custom fortunes test, would "micro" classification be enough to justify custom html rendering? Rest of the code does not feel like something too wild?

/cc @msmith-techempower @nbrady-techempower

nbrady-techempower:
@dv00d00 I think "stripped" approach makes sense here. Micro classification is saying you're providing just enough middleware or whatever to make the implementation work. Stripped approach is saying you're writing a lot more custom code specific for this use case, either because the framework doesn't provide the utilities or you're testing out different methods closer to the metal that you might want to incorporate into the framework. What you have now looks right to me. @msmith-techempower anything to add?

gerardtoconnor:
@dv00d00 I have not got that TemplateViewEngine on nuget, it was only a POC, without the pre-counting of the model bytes to set the contentlength before so as is, it'll chunk so not much use, will have a look at tweaking it to be able to count & set content length.

For now I think your adapted MS pool from Ben's design is a great solution.

frameworks/FSharp/giraffe/benchmark_config.json
```diff
@@ -5,19 +5,20 @@
...
         "language": "F#",
-        "orm": "Raw",
+        "orm": "micro",
         "platform": ".NET",
         "flavor": "CoreCLR",
         "webserver": "Kestrel",
         "os": "Linux",
         "database_os": "Linux",
-        "display_name": "Giraffe",
+        "display_name": "Giraffe, Dapper",
```
msmith-techempower: Not a big fan of this, but since we don't have a key for the name of your database approach/library, I guess it will do.

PR #3870, opened 2018-06-19
Mojolicious on async rails. 

amf3tam1n:
Just replaced blocking code with non-blocking and add some Promises magic.


msmith-techempower:
@amf3tam1n Your pull request should be included in the latest run: https://tfb-status.techempower.com/results/034f1554-2e9f-4959-8dfb-8ee21cad8a14

PR #3873, opened 2018-06-20
Ulib: enable db pipeline

stefanocasazza:


michaelhixson:
Hey @stefanocasazza, sorry for the slow turnaround.

I have a hard time understanding the code here...  Would you mind explaining to me what these changes do?

The reason I ask is, based on how recent conversations on the TFB mailing list went, it's become apparent to me that different people use very different definitions of "pipelining".  Some people define it such that they don't see any difference between pipelining and batch queries, which is a problem because we allow one and not the other.

stefanocasazza:
Hi,
I make many changes:
I grouped the code as SQL vs NoSQL, some optimization for PostGres, etc...
Follow your [decision](https://groups.google.com/d/msg/framework-benchmarks/A78HrYsz4AQ/_zneyvo2CwAJ) I applied a patch to the frontend for PostGres to make async request for server to be pipelined for the query and update test

michaelhixson:
It looks like this implementation is executing batch queries in the multi-query test, which we do not consider to be the same thing as pipelining and which we still disallow.  The reason I think it's doing batch queries is that when I enable the query log in postgres and I hit the implementation's multi-query endpoint, postgres tells me that each of the queries is being executed within the same virtual transaction.

If you think I'm wrong about what your implementation is doing, let me know.  If it's simply a matter of us defining "pipelining" differently, I'm happy to continue a discussion here with you to get that clarified.  I suggest reading the rest of my messages in that TFB mailing list thread first if you haven't already.

There are a lot of changes in this PR that seem unrelated to batching or pipelining.  I'd be happy to merge those, but I don't want to merge the ulib-postgres implementation as it is right now because of the batching issue.  Perhaps it makes sense to undo that one change - disable what I call "batching" and what you call "pipelining" - so we can your other changes in?

stefanocasazza:
I am very confused. Can you indicate literally in your postgres log the difference with what are you expect?
Anyway, I removed the discussed parts

michaelhixson:
Thanks.  I'll merge this in momentarily.

To look at the postgres query logs, I made this change in the toolset locally: https://github.com/michaelhixson/FrameworkBenchmarks/commit/3513992ac0467d029ecadf5068d50163b5c88b7c

Then I ran the toolset like this:
```
tfb --mode verify --type query --test ulib-postgres
```

Then I looked at the console output, which included TFB's normal output plus the postgres query logs.  Here's one part of that output from a couple of days ago, before your most recent change:

```
Accessing URL http://tfb-server:8080/query?queries=2:
2018-07-03 22:03:56 GMT 5b3bf2cb.11 4/5 0 LOG:  execute SQL_A368A700: SELECT randomNumber FROM World WHERE id = $1
2018-07-03 22:03:56 GMT 5b3bf2cb.11 4/5 0 DETAIL:  parameters: $1 = '5159'
2018-07-03 22:03:56 GMT 5b3bf2cb.11 4/5 0 LOG:  execute SQL_A368A700: SELECT randomNumber FROM World WHERE id = $1
2018-07-03 22:03:56 GMT 5b3bf2cb.11 4/5 0 DETAIL:  parameters: $1 = '6337'
```

As I understand it, the `4/5` part is the virtual transaction id.  It stuck out to me that the same transaction id was used for both queries.  That's the same thing I saw in another framework when I implemented batch queries (which are disallowed).  That's *not* the same thing I see when I run vertx-postgres.  With vertx-postgres, each query has a distinct transaction id.

Here's that same part with your most recent change:
```
Accessing URL http://tfb-server:8080/query?queries=2:
2018-07-05 22:31:00 GMT 5b3e9c23.32 8/4 0 LOG:  execute SQL_47C5A700: SELECT randomNumber FROM World WHERE id = $1
2018-07-05 22:31:00 GMT 5b3e9c23.32 8/4 0 DETAIL:  parameters: $1 = '5920'
2018-07-05 22:31:00 GMT 5b3e9c23.32 8/5 0 LOG:  execute SQL_47C5A700: SELECT randomNumber FROM World WHERE id = $1
2018-07-05 22:31:00 GMT 5b3e9c23.32 8/5 0 DETAIL:  parameters: $1 = '4121'
```

This time, each query has its own distinct virtual transaction id: `8/4` and `8/5`.

That's what I was looking at.  It was not a formal analysis.  Having a distinct virtual transaction id per query isn't actually a requirement.  But I was seeing those shared transaction ids, I was noticing how you commented about the query and updates test only (as if pipelining could only be useful in those test types and not in db or fortunes), and there's what little I could piece together from reading this code...  I looked at all of that and concluded you were probably doing batch queries.

If you think I'm wrong about what your code was doing, let me know.

If you want to explore what we mean by "pipelining" further and you've already read through all of my posts in that mailing list thread, you could try doing what another contributor did and monitor vertx-postgres with Wireshark.  If you give it another shot and are wondering if your implementation qualifies, here are a few qualities you could look for:

* Pipelining should improve performance in the single-query test.
* If the multi-query test requirements changed such that each SQL query executed in the scope of a single HTTP request depended on the results of the previous query -- if the the randomnumber returned from query 1 had to be the id used in query 2, and the randomnumber returned from query 2 had to be the id for query 3, and so on -- pipelining should be just as effective under those requirements as it is now.
* Pipelining should not alter query semantics by artificially grouping unrelated queries together in units such as transactions.  For example, if the application were to send a nonsense query like `SELECT foo bar` to the database once per second, that should not affect the outcome of logically unrelated, good queries happening concurrently.

PR #3886, opened 2018-06-23
Add basic benchmarks for Gotham/Rust

whitfin:
As one of the new maintainers of Gotham, here is a simple benchmark for Gotham. Previously @bradleybeddoes objected to inclusion due to the benchmarks being provided from another party, but seeing as we're providing our own I think it's fine. Certainly myself and @nyarly are all for it.

This is a simple benchmark, but I think I might need some help debugging an issue with the LOC counter - it errors out with a seemingly vague Python crash.

### DIFF
--- a/frameworks/Rust/gotham/Cargo.toml
+++ b/frameworks/Rust/gotham/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "gotham_techempower"
+version = "0.1.0"
+authors = ["Isaac Whitfield <iw@whitfin.io>"]
+
+[dependencies]
+gotham = "0.2"
+hyper = "0.11"
+mime = "0.3"
+serde = "1.0"
+serde_derive = "1.0"
+serde_json = "1.0"
+
+[profile.release]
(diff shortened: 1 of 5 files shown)


cjnething:
Hi @whitfin, I think you're missing the addition to travis.yml. Also not required but would you mind adding a README ([example here](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Ruby/rack))? [Here](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/scaffolding/README.md) is a list of all the steps necessary to add a new framework in case I missed something. Thanks!

whitfin:
Hi @cjnething! 

I think .travis.yml is okay because it's testing all of the Rust language? I'll definitely add a README - no problemo. I'll run down that list to check I have everything. Same applies to #3888.

cjnething:
HI @whitfin, yes you're right Rust isn't one of the languages that divides it out in the .travis.yml file, so ignore that! Thanks!

PR #3888, opened 2018-06-24
Add Thruster benchmarks for Rust

whitfin:
Cross referencing with https://github.com/trezm/Thruster/issues/40. 

This shouldn't be merged until Thruster 0.4.5 is released an can successfully run on stable Rust.

Edit: Now ready to be merged!

### DIFF
--- a/frameworks/Rust/thruster/Cargo.toml
+++ b/frameworks/Rust/thruster/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "thruster_techempower"
+version = "0.1.0"
+authors = ["Isaac Whitfield <iw@whitfin.io>"]
+
+[dependencies]
+futures = "0.1"
+serde = "1.0"
+serde_derive = "1.0"
+serde_json = "1.0"
+thruster = { git = "https://github.com/trezm/Thruster" }
+
+[profile.release]
+codegen-units = 1
+opt-level = 3
+lto = true
+
--- a/frameworks/Rust/thruster/Cargo.toml
+++ b/frameworks/Rust/thruster/Cargo.toml
@@ -8,7 +8,7 @@ futures = "0.1"
 serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
-thruster = { git = "https://github.com/trezm/Thruster" }
+thruster = "0.4"
 
 [profile.release]
 codegen-units = 1
--- a/frameworks/Rust/thruster/thruster.dockerfile
+++ b/frameworks/Rust/thruster/thruster.dockerfile
@@ -1,0 +1,11 @@
+FROM rust:1.26
+
+WORKDIR /gotham
+COPY ./src ./src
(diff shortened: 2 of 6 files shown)


frameworks/Rust/thruster/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+name = "thruster_techempower"
+version = "0.1.0"
+authors = ["Isaac Whitfield <iw@whitfin.io>"]
+
+[dependencies]
+futures = "0.1"
+serde = "1.0"
+serde_derive = "1.0"
+serde_json = "1.0"
+thruster = { git = "https://github.com/trezm/Thruster" }
```
trezm: Let's lock this down to a version -- 4.4 for now, just so that when master receives a breaking change we don't break these benchmarks :)

frameworks/Rust/thruster/thruster.dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM rust:1.26
+
+WORKDIR /gotham
```
trezm: Should be `/thruster` I think

frameworks/Rust/thruster/thruster.dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM rust:1.26
```
trezm: let's bump this to `rust:1.27` for the latest and greatest!

frameworks/Rust/thruster/thruster.dockerfile
whitfin: Changed!

frameworks/Rust/thruster/Cargo.toml
whitfin: @trezm I'm going to update this but for the time being only master supports compilation on stable - I'll pin to 0.4.5 once released.

frameworks/Rust/thruster/thruster.dockerfile
whitfin: I chose 1.26 to stay in sync with the other images to avoid compiler differences - I believe they're all bumped at once before a round is run.

nbrady-techempower: @whitfin FYI, we don't necessarily do that bump ourselves. If you'd like, you can bump it in all rust dockerfiles and assuming it doesn't break any of them, we'll merge that in. You don't need to keep it in sync though.

whitfin: @nbrady-techempower awesome - I'll probably file a separate one to bump all of them.

frameworks/Rust/thruster/Cargo.toml
trezm: Just release 0.4.5 if you want to bump this

whitfin:
@trezm thanks for the nudge, bumped to 0.4.5 and this is ready to merge!

PR #3891, opened 2018-06-26
[ci fw-only Java/t-io] update t-io to 3.0.6

tywo45:


tywo45:
@nbrady-techempower, I was only modified t-io version from 3.0.4 to 3.0.6, Can you please merge this PR? Thanks!

PR #3895, opened 2018-06-27
Submission of OfficeFloor (updated to remove tests and source_files)

sagenschneider:
Adding OfficeFloor

 - OfficeFloor is the default implementation
 - OfficeFloor-micro uses bare bones of the framework (but shows flexibility in using different implementations and not forced use of certain libraries)
 - OfficeFloor-raw provides raw HTTP functionality
 - OfficeFloor-netty uses Netty for HTTP handling
 - OfficeFloor-rapidoid uses Rapidoid for HTTP handling (both this and Netty shows flexibility of OfficeFloor)

Note: the woof acronym comes from WoOF (Web on OfficeFloor)

Also, made a slight adjustment to ./tfb script so that can run as part of local CI runs ( -t can not be used for headless runs, so added line to test whether tty is available).  I hope this is ok :)


michaelhixson:
@msmith-techempower Do you have time to review the change to the `tfb` script?  https://github.com/TechEmpower/FrameworkBenchmarks/pull/3895/files#diff-be2f60aaf604c1135e78fcab62ae286c

msmith-techempower:
@michaelhixson Sorry, I thought I added a comment on that the other day.

tfb
```diff
@@ -101,4 +101,5 @@ if ! docker network inspect tfb >/dev/null 2>&1; then
   docker network create tfb >/dev/null
 fi
 
-exec docker run -it --rm --network tfb -v /var/run/docker.sock:/var/run/docker.sock -v ${SCRIPT_ROOT}:/FrameworkBenchmarks techempower/tfb "${@}"
+test -t 1 && USE_TTY="-t"
+exec docker run -i ${USE_TTY} --rm --network tfb -v /var/run/docker.sock:/var/run/docker.sock -v ${SCRIPT_ROOT}:/FrameworkBenchmarks techempower/tfb "${@}"
```
msmith-techempower: This looks fine to me. Basically checks to see if TTY is available, and if so, use the flag which we were using before anyway. For us, nothing changes, for people not using TTY, they are spared the `-t` flag.

frameworks/Java/officefloor/src/woof_benchmark/src/test/java/net/officefloor/benchmark/TestSuite.java
```diff
@@ -0,0 +1,31 @@
...
+import org.junit.runners.Suite;
+import org.junit.runners.Suite.SuiteClasses;
+
+/**
+ * Tests.
+ */
+@RunWith(Suite.class)
+@SuiteClasses({ JsonTest.class, DbTest.class, QueriesTest.class, FortunesTest.class, UpdateTest.class,
+		PlaintextTest.class })
+public class TestSuite {
```
michaelhixson: Thank you for taking the care to make sure the implementation works correctly.  We don't want to include any functional tests in this repository, though.  Could you remove all of these functional tests from the PR?

If you want to maintain those tests on your own outside of this repository, that's fine.  You could add a `.gitignore` to the `frameworks/Java/officefloor` directory with directives to ignore the test files.  And if you want to leave any test-related surefire configuration or stuff like that in the `pom.xml`s (so that you can still run the tests locally without maintaining a separate `pom.xml`) that's also fine.

From our point of view, the TFB toolset's own built-in verification executed by Travis CI are all the functional tests we need.  When frameworks include their own tests, that becomes an additional maintenance burden for us.  We'd rather not maintain any code that isn't essential for the benchmarks.

(The same reasoning applies for startup scripts/configs for running the framework without running the TFB toolset.  We already have a way to run frameworks -- using the toolset -- and we don't want to maintain/support any other ways of running frameworks.  I don't see any files like this in this PR though.)

frameworks/Java/officefloor/source_files
```diff
@@ -0,0 +1,4 @@
+src/woof_benchmark/src/main/java/net/officefloor/benchmark/Fortune.java
+src/woof_benchmark/src/main/java/net/officefloor/benchmark/FortunesLogic.java
+src/woof_benchmark/src/main/java/net/officefloor/benchmark/Logic.java
+src/woof_benchmark/src/main/java/net/officefloor/benchmark/World.java
```
michaelhixson: Remove this `source_files` file, please.  We used to use files like this to help us count the lines of code for each framework, but we don't do that any longer.  We fixed our README to not ask people to create these files anymore: https://github.com/TechEmpower/FrameworkBenchmarks/commit/865a28ba19c2e36374ed664aeedef893df2fce9b

PR #3901, opened 2018-06-30
[ci fw-only Java/act] update act to 1.8.8-RC12-SNAPSHOT

greenlaw110:
try fix fortune mustache performance issue


greenlaw110:
@cjnething fixed! thanks for the reminder!

PR #3904, opened 2018-07-02
H2O: Optimize the cached queries test

volyrique:
The main idea is to use several caches behind an interface that makes them appear as a single one. Every cache operation chooses one of them based on the key hash value, and while all subcaches are multithreaded, the interface in front of them does not modify anything, so it does not contain any critical sections. As a result, if several threads access the cache, contention will be minimized, provided that most of the time they use keys that map to different subcaches.

### DIFF
--- a/frameworks/C/h2o/src/cache.c
+++ b/frameworks/C/h2o/src/cache.c
@@ -1,0 +1,97 @@
+/*
+ Copyright (c) 2018 Anton Valentinov Kirilov
+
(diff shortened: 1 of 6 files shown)


volyrique:
Let's see how it performs first. Besides, these changes have been done from the perspective of a user of `libh2o` (i.e. trying to reuse as much as possible), so they may not be optimal if one has the freedom to modify the underlying implementation completely.

PR #3905, opened 2018-07-03
Add Language VB

benaadams:
VB.NET using aspnetcore


michaelhixson:
It's not too often we get a new language in here anymore.  This is great.  Thanks, @benaadams.

PR #3907, opened 2018-07-05
smart-socket 

smthing:

### DIFF
--- a/frameworks/Java/smart-socket/README.md
+++ b/frameworks/Java/smart-socket/README.md
@@ -1,0 +1,23 @@
+# smart-socket Benchmarking Test
+
+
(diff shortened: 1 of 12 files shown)


msmith-techempower:
What is the status of this pull request? It is marked `WIP` - is it still a work in progress?

smthing:
pls merge this pr

msmith-techempower:
@smthing I merged this in, but then was notified that the implementation was not in [compliance](https://github.com/TechEmpower/FrameworkBenchmarks/pull/3907/files#diff-c9e286386d4af63732244034856bfa2aR35) with the rules, as the `json` test must actually do serialization and cannot simply be sending a string (otherwise it does not differ from plaintext in implementation).

Feel free to open another pull request with that fix, and we'll get it merged.

PR #3909, opened 2018-07-05
Csharp span json

dv00d00:

### DIFF
--- a/frameworks/CSharp/aspnetcore/benchmark_config.json
+++ b/frameworks/CSharp/aspnetcore/benchmark_config.json
@@ -97,6 +97,24 @@
       "notes": "",
       "versus": "aspcore"
     },
+    "mw-spanjson": {
+      "json_url": "/spanjson",
+      "port": 8080,
+      "approach": "Realistic",
+      "classification": "Micro",
+      "database": "None",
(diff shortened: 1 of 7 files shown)


dv00d00:
@Tornhoof could you take a look?

Tornhoof:
I think it looks ok, but honestly that message size is kinda the worst case scenario, string into utf8 and a very short one at that, I imagine it will perform worse than e.g. utf8json.

frameworks/CSharp/aspnetcore/benchmark_config.json
```diff
@@ -97,6 +97,24 @@
...
+      "database": "None",
+      "framework": "ASP.NET Core",
+      "language": "C#",
+      "orm": "Raw",
+      "platform": ".NET",
+      "flavor": "CoreCLR",
+      "webserver": "Kestrel",
+      "os": "Linux",
+      "database_os": "Linux",
+      "display_name": "ASP.NET Core, Middleware, Utf8Json",
```
benaadams: Utf8Json -> SpanJson

dv00d00: thx

PR #3918, opened 2018-07-10
Uvicorn app improvements

tomchristie:
* Use prepared statements where possible.
* Use single `executemany` in update test type.
* Store `ujson.dumps` as `ujson_dumps`.
* Pre-initialize JSON/PlainText/HTML response dictionaries.
* Call `randint(1, 10000)` instead of using `random_int = partial(randint, 1, 10000)`


frameworks/Python/uvicorn/app.py
```diff
@@ -84,62 +100,53 @@ def __init__(self, scope):
         pass
 
     async def __call__(self, receive, send):
+        row_id = randint(1, 10000)
```
michaelhixson: Same comment here as in your other PR.  Pretty sure this should be `randint(1, 10001)`.

tomchristie: Same again - randint is inclusive of bounds.

PR #3919, opened 2018-07-10
Add Starlette

tomchristie:
Add [Starlette](https://github.com/encode/starlette) to the Python frameworks.

### DIFF
--- a/frameworks/Python/starlette/app.py
+++ b/frameworks/Python/starlette/app.py
@@ -1,0 +1,131 @@
+import asyncio
+import asyncpg
+import os
+import jinja2
+from starlette import (
+    asgi_application, HTMLResponse, JSONResponse, Path, PlainTextResponse, Router
+)
+from random import randint
+from operator import itemgetter
+from urllib.parse import parse_qs
+from ujson import dumps as ujson_dumps
+
+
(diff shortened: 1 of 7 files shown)


frameworks/Python/starlette/app.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+@asgi_application
+def json_serialization(request):
+    return UJSONResponse({'message': 'Hello, world!'})
+
+
+@asgi_application
+async def single_database_query(request):
+    row_id = randint(1, 10000)
```
michaelhixson: Not that it matters a whole lot, but the highest number this generates is 9999, correct?  If so, I think this should be `randint(1, 10001)`.  There is a record with id=10000 in the database.

tomchristie: Nope, it’s 1-10000 inclusive.

Python’s docs for randint are: ‘Return a random integer N such that a <= N <= b’

(Unlike Pythons list slices)

michaelhixson: Oh my bad.  I was looking at the docs for numpy.random.randint.

tomchristie: Thanks for the review @michaelhixson - Have a fab weekend! ✨

PR #3923, opened 2018-07-11
ULib: enable db pipeline plus some optimization

stefanocasazza:
I hope this time is ok...

### DIFF
--- a/frameworks/C++/ulib/src/db.usp
+++ b/frameworks/C++/ulib/src/db.usp
@@ -9,30 +9,31 @@ static void usp_fork_db() { World::handlerForkSql(); }
 -->
 <!--#header
 -->
+<!--#vcode
+if (UServer_Base::startParallelization()) return;
+-->
 <!--#code
 World::initOneResult();
-
-World::pworld_query->id = World::rnumber[0];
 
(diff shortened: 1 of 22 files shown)


michaelhixson:
Sorry @stefanocasazza, but I can't merge this.  Even though you added a Sync between each query, you're still using batch queries.

https://github.com/stefanocasazza/FrameworkBenchmarks/blob/c5affc8e658d4f762ae3d82c04af8859711415dd/frameworks/C%2B%2B/ulib/src/query.usp#L35-L44

```
   while (true)
      {
      World::sendQueryPrepared(i);

      if (++i == num_queries) break;

      (void) U_SYSCALL(PQbatchPutSyncOnQueue, "%p", World::conn); // NB: make the difference between pipeline vs batch...
      }

   (void) U_SYSCALL(PQbatchSendQueue, "%p", World::conn);
```

This approach is exactly what we mean by batch queries.  It's what I was describing in my email to you:

> The thing we want to avoid in the multi-query test is implementations
that are written like this:
>
> * For each incoming HTTP request, construct a batch of queries.
> * In one operation, send the entire batch as a unit to the database.

PR #3925, opened 2018-07-11
Bump Finagle benchmarks to 18.7

dschobel:
* Finagle to 18.7
* Finatra to 18.7


michaelhixson:
The Finatra build is failing.  It's complaining that the `jcl-over-slf4j` and `commons-logging` libraries contain conflicting classes in the `org.apache.commons.logging` package.  Any ideas how to fix it?

dschobel:
thanks for taking a look @michaelhixson, should be all good now.

PR #3937, opened 2018-07-18
OfficeFloor thread-per-request thread injection

sagenschneider:
OfficeFloor provides thread injection with various means to manage threading of the applications.  Providing this implementation as the "bar" to beat in the various thread implementations.

Also, actually configures Afterburner for XML rendering


sagenschneider:
Seems to be some issue in starting...  will fix locally and then re raise PR

PR #3950, opened 2018-07-23
Fixes for OfficeFloor

sagenschneider:
Changes include:
  - actually registering Afterburner
  - thread per request configuration as baseline for other threading tests configured (thread injection allows various threading, and seems keeping database connections down in the database tests gets better through put - therefore adding thread per request as baseline to beat with other threading models)
  - correct JDBC connection string to connect to PostgreSql


sagenschneider:
OfficeFloor seems to be passing test.   But build failing due to other frameworks 😕

michaelhixson:
I recommend rebasing against our master branch or recreating your fork at some point.  That would make these PRs contain fewer commits and would make Travis test fewer frameworks.

PR #3955, opened 2018-07-26
Fastify - NodeJs Framework

Talento90:
I would like to add [Fastify framework](https://github.com/fastify/fastify) created by Matteo Collina and Tomas Della Vedova and it's sponsored by NearForm. At this moment I have to admit that I am impressed with the results it seems to be the best alternative to develop API's using NodeJs.

### DIFF
--- a/frameworks/JavaScript/fastify/package.json
+++ b/frameworks/JavaScript/fastify/package.json
@@ -5,13 +5,13 @@
   "main": "app.js",
   "private": true,
   "dependencies": {
-    "fastify": "^1.9.0",
-    "handlebars": "~4.0.11",
-    "mongoose": "~5.2.5",
-    "mysql2": "~1.5.3",
-    "pg": "~7.4.1",
-    "pg-hstore": "~2.3.2",
-    "point-of-view": "^1.1.0",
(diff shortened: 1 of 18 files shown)


frameworks/JavaScript/fastify/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "private": true,
+  "dependencies": {
+    "fastify": "^1.9.0",
+    "handlebars": "~4.0.11",
+    "mongoose": "~5.2.5",
+    "mysql2": "~1.5.3",
+    "pg": "~7.4.1",
+    "pg-hstore": "~2.3.2",
+    "point-of-view": "^1.1.0",
+    "sequelize": "~4.37.3"
```
msmith-techempower: Not sure of the syntax here, but could you lock these to exact versions? A problem we frequently encounter is dependencies causing trouble because someone marks "at least version 1.1.0" and on a much later benchmarking run, the newest version is 1.2.0 and it breaks the implementation.

Talento90:
I completely agree with you! I have locked the versions to avoid those kind of problems 👍

PR #3956, opened 2018-07-27
Submission for the java ratpack framework

jringuette:
Submission for the java ratpack framework


frameworks/Java/ratpack/ratpack.dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM gradle:4.7.0-jdk8 as gradle
+USER root
+WORKDIR /ratpack
+COPY build.gradle build.gradle
+COPY src src
+RUN gradle shadowJar
```
msmith-techempower: `gradle-wrapper.jar` either needs to be downloaded here when building the image, or in a dependency management script.

frameworks/Java/ratpack/ratpack.dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM gradle:4.7.0-jdk8 as gradle
```
msmith-techempower: Presumably, this puts `gradlew` on the path, and therefore the inclusion of that file in this codebase is unnecessary.

frameworks/Java/ratpack/ratpack.dockerfile
jringuette: Done, I've removed it. It is not required for building/running via docker. It was just a convenience to build via intellij without having to install a system wide gradle.

frameworks/Java/ratpack/ratpack.dockerfile
jringuette: See above

PR #3958, opened 2018-07-27
Actix: use async postgresql driver

fafhrd91:
* actix-web 0.7.2
* use async postgresql driver


gjaegle:
it was worth the wait. congrats

PR #3983, opened 2018-08-09
Updated and aligned common python frameworks' components

gi0baro:
The aim is to have the same conditions on all the frameworks for common requirements.

Alignments:
- using same docker images (updated to 2.7.15-stretch and 3.6.6-stretch)
- using same platform requirements

The only exception is for *web2py* framework, which refuses to start with gunicorn 19.9.0.

The update of docker images is also fixing the issues with nginx-uwsgi tests.

### DIFF
--- a/frameworks/Python/weppy/weppy-nginx-uwsgi.dockerfile
+++ b/frameworks/Python/weppy/weppy-nginx-uwsgi.dockerfile
@@ -1,8 +1,8 @@
-FROM python:2.7.14
+FROM python:3.6.6-stretch
 
 RUN curl -s http://nginx.org/keys/nginx_signing.key | apt-key add -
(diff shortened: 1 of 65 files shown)


gi0baro:
Ready for merge

frameworks/Python/weppy/weppy-nginx-uwsgi.dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM python:2.7.14
+FROM python:3.6.6-stretch
```
nbrady-techempower: Did you mean to bump major version here? If so would you mind updating the notes section in https://github.com/TechEmpower/FrameworkBenchmarks/blob/084e0f47bf0704d8bc3e82c1bab5bb9268daeab9/frameworks/Python/weppy/benchmark_config.json#L90

gi0baro:
@nbrady-techempower yup, just pushed and update as you required.

PR #3984, opened 2018-08-09
Fixes Giraffe, and some vulns reported by GH

msmith-techempower:
There's a lot of formatting noise, so I recommend viewing the diff while ignoring whitespace changes.

This PR fixes:

1. The toolset suggesting that Giraffe did not specify a default test
2. hapi's postgres driver having a security vulnerability
3. nodejs's postgres driver having a security vulnerability
4. sailsjs's postgres driver having a security vulnerability


msmith-techempower:
The FSharp and JavaScript tests passed, so I'm merging this in.

PR #3987, opened 2018-08-11
Actix: optimizae async pg connection

fafhrd91:
* increase async pg connections per thread
* optimize pg connection usage
* update to latest actix-web version

### DIFF
--- a/frameworks/Rust/actix/src/db_pg_direct.rs
+++ b/frameworks/Rust/actix/src/db_pg_direct.rs
@@ -1,0 +1,154 @@
+use std::io;
+
+use actix::prelude::*;
+use futures::{stream, Future, Stream};
+use rand::{thread_rng, Rng};
+use tokio_postgres::{connect, Client, Statement, TlsMode};
+
+use models::{Fortune, World};
+
+/// Postgres interface
+pub struct PgConnection {
+    cl: Client,
+    fortune: Statement,
+    world: Statement,
+    update: Statement,
+}
+
+impl PgConnection {
+    pub fn connect(db_url: &str) -> impl Future<Item=PgConnection, Error=()> {
+        let hs = connect(db_url.parse().unwrap(), TlsMode::None);
+
+        hs.map_err(|_| panic!("can not connect to postgresql"))
+            .and_then(|(cl, conn)| {
+                Arbiter::spawn(conn.map_err(|e| panic!("{}", e)));
+                cl.prepare("SELECT id, message FROM fortune")
+                    .map_err(|_| ())
+                    .and_then(move |fortune| {
+                        cl.prepare("SELECT id, randomnumber FROM world WHERE id=$1")
+                            .map_err(|_| ())
+                            .and_then(move |world| {
+                                let st = (fortune, world);
+                                cl.prepare("SELECT id FROM world WHERE id=$1")
+                                    .map_err(|_| ())
+                                    .and_then(|update| {
+
+                                        Ok(PgConnection {
+                                            cl,
+                                            fortune: st.0,
+                                            world: st.1,
+                                            update
+                                        })
+                                    })
+                            })
+                    })
+            })
+    }
+}
+
+impl PgConnection {
+
+    pub fn get_world(&self) -> impl Future<Item=World, Error=io::Error> {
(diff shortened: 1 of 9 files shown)


frameworks/Rust/actix/src/db_pg_direct.rs
```diff
@@ -0,0 +1,154 @@
...
+            .and_then(|(cl, conn)| {
+                Arbiter::spawn(conn.map_err(|e| panic!("{}", e)));
+                cl.prepare("SELECT id, message FROM fortune")
+                    .map_err(|_| ())
+                    .and_then(move |fortune| {
+                        cl.prepare("SELECT id, randomnumber FROM world WHERE id=$1")
+                            .map_err(|_| ())
+                            .and_then(move |world| {
+                                let st = (fortune, world);
+                                cl.prepare("SELECT id FROM world WHERE id=$1")
```
michaelhixson: Out of curiosity, what is this query used for?

frameworks/Rust/actix/src/db_pg_direct.rs
```diff
@@ -0,0 +1,154 @@
...
+            .collect()
+            .and_then(move |mut worlds| {
+                let mut update = String::with_capacity(120 + 6 * num as usize);
+                update
+                    .push_str("UPDATE world SET randomnumber = temp.randomnumber FROM (VALUES ");
+
+                for w in &worlds {
+                    update.push_str(&format!("({}, {}),", w.id, w.randomnumber));
+                }
+                worlds.sort_by_key(|w| w.id);
```
michaelhixson: I think you can remove this sort operation.  The `ORDER BY 1` in your update statement should take care of the batch update deadlock issue, since that sorts the `VALUES` by id.  (My theory about this code, which may be wrong, is that `worlds.sort_by_key` was originally added to this code to address deadlocks, but now it has drifted into the wrong place in the code.  It would have to be before the `for w in &worlds` loop to have its intended effect, but it's redundant because of the `ORDER BY 1` anyway.)

frameworks/Rust/actix/src/db_pg_direct.rs
```diff
@@ -0,0 +1,154 @@
...
+                    .push_str(" ORDER BY 1) AS temp(id, randomnumber) WHERE temp.id = world.id");
+
+                cl.batch_execute(&update)
+                  .map_err(|e| io::Error::new(io::ErrorKind::Other, e))
+                  .and_then(|_| Ok(worlds))
+            })
+    }
+
+    pub fn tell_fortune(&self) -> impl Future<Item=Vec<Fortune>, Error=io::Error> {
+        let mut items = Vec::with_capacity(16);
```
michaelhixson: The `16` here looks suspicious.  As a reminder, this is one of our fortunes test requirements:

Could you change this so that you do not specify an initial capacity?  Would `Vec::new()` work?

michaelhixson:
I noticed that `actix-raw` is doing very well in the database tests in the latest continuous benchmarks.  What's the reason for this boost in performance, in your opinion?  Is it using pipelining a.k.a. multiplexing - that is, it sends more queries out over each connection while previous queries are still in flight on that same connection?

By the way, out of my comments here, only the fortunes list pre-sizing issue needs to be addressed before we merge the PR.  The rest of the comments and questions are optional.

frameworks/Rust/actix/src/db_pg_direct.rs
fafhrd91: it is for db updates, line 96

frameworks/Rust/actix/src/db_pg_direct.rs
fafhrd91: i am not sure about that, i remember there was some problems with query without this sort. i will test again.

frameworks/Rust/actix/src/db_pg_direct.rs
fafhrd91: done. Vec::with_capacity() is just pre-allocation in heap.

fafhrd91:
i switched to new async pg driver, it allows to use multiplexing. `actix-pg` should perform better to but it hit bug in base actix package, it should work better after this pr.

PR #4002, opened 2018-08-17
OfficeFloor: using initial PostgreSql ADBA work

sagenschneider:


sagenschneider:
ADBA needs little more work.   Will recreate once in better state.

PR #4013, opened 2018-08-23
Add Dart Angel Framework

thosakwe:
Hello!

This PR adds benchmarks for [Angel](https://github.com/angel-dart/angel), a Dart server framework.

I've included code for running Angel with both MongoDB and PostgreSQL.

Thanks in advance.


frameworks/Dart/angel/.gitignore
```diff
@@ -0,0 +1,4 @@
+!bin/
+.dart_tool
+.packages
+*.g.part
```
michaelhixson: Did you mean for this to be `*.g.dart` instead?

If the `.g.dart` files are being included intentionally, I'm wondering if they can instead be generated as one of the build steps in docker and if we can omit them from source control.

thosakwe: No, I specifically meant `.g.part`. Updates to Dart's `source_gen` now include a `CombiningBuilder`, which generates `.g.part` files.

`.g.dart` is included intentionally.

The current version of `source_gen` *should* delete the `.part` files, but it currently does not; not sure why.

thosakwe: A JSON serialization generator. Should I do this in the Dockerfile instead?

michaelhixson: Yes, I would prefer the `.g.dart` files be generated as a build step in the Dockerfile.

This is coming from my general aversion to generated code in source control, not from any knowledge of this framework or of Dart.  If everyone who deals with these files would in fact put them in source control, I'd love to hear why.  Even if the reason is bad, the mere fact that everyone else does it might be reason enough for us to do it too.  Maybe.  But knowing none of that, these files look like they shouldn't be here.  (It's noteworthy to me that GitHub auto-minimizes them by default in the "Files changed" tab, as if they aren't worth reviewing.)

thosakwe:
Sounds good, I’ll have it done by tomorrow morning.

thosakwe:
Actually, I've just fixed it now.

thosakwe:
Thanks for the merge!

PR #4015, opened 2018-08-25
Improve aspcore DB benchmarks

bgrainger:
* Updates MySqlConnector and npgsql to latest published versions
* Improves MySqlConnector connection string
* Uses explicitly prepared commands (MySqlConnector only)

### DIFF
--- a/frameworks/CSharp/aspnetcore/Benchmarks/Data/RawDb.cs
+++ b/frameworks/CSharp/aspnetcore/Benchmarks/Data/RawDb.cs
@@ -64,6 +64,8 @@ namespace Benchmarks.Data
             id.Value = _random.Next(1, 10001);
             cmd.Parameters.Add(id);
 
+            (cmd as MySql.Data.MySqlClient.MySqlCommand)?.Prepare();
+
             return cmd;
         }
 
@@ -146,6 +148,8 @@ namespace Benchmarks.Data
 
                 db.ConnectionString = _connectionString;
                 await db.OpenAsync();
+
+                (cmd as MySql.Data.MySqlClient.MySqlCommand)?.Prepare();
 
                 using (var rdr = await cmd.ExecuteReaderAsync(CommandBehavior.CloseConnection))
                 {
--- a/frameworks/CSharp/aspnetcore-mono/Benchmarks/Benchmarks.csproj
+++ b/frameworks/CSharp/aspnetcore-mono/Benchmarks/Benchmarks.csproj
@@ -16,20 +16,19 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="2.1.0-rc1-final" />
-    <PackageReference Include="Microsoft.AspNetCore.Mvc" Version="2.1.0-rc1-final" />
-    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel" Version="2.1.0-rc1-final" />
-    <PackageReference Include="Microsoft.AspNetCore.StaticFiles" Version="2.1.0-rc1-final" />
-    <PackageReference Include="Microsoft.Extensions.Configuration" Version="2.1.0-rc1-final" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.CommandLine" Version="2.1.0-rc1-final" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="2.1.0-rc1-final" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="2.1.0-rc1-final" />
-    <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="2.1.0-rc1-final" />
-    <!-- Fix https://github.com/aspnet/KestrelHttpServer/pull/2562  -->
-    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets" Version="2.1.0-rtm-30793" />
+    <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="2.1.1" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc" Version="2.1.2" />
+    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel" Version="2.1.3" />
+    <PackageReference Include="Microsoft.AspNetCore.StaticFiles" Version="2.1.1" />
+    <PackageReference Include="Microsoft.Extensions.Configuration" Version="2.1.1" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.CommandLine" Version="2.1.1" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="2.1.1" />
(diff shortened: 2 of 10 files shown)


frameworks/CSharp/aspnetcore/Benchmarks/Data/RawDb.cs
```diff
@@ -64,6 +64,8 @@ DbCommand CreateReadCommand(DbConnection connection)
             id.Value = _random.Next(1, 10001);
             cmd.Parameters.Add(id);
 
+            (cmd as MySql.Data.MySqlClient.MySqlCommand)?.Prepare();
```
bgrainger: @sebastienros Based on 14ad208cd437b09c0cbe6bbf9a58fed76e738cdf I decided to only opt MySqlConnector into this. It doesn't have npgsql's auto-prepare capability, so this improves performance by 10-15% (in my local Docker testing).

I tested changing this to just `cmd.Prepare()` with `aspcore-ado-pg` and got these results:

Test | Before | After | Change
--- | --- | --- | ---
Fortunes | 43492.60 | 43695.85 | -0.47%
DB | 46917.18 | 47240.47 | +0.69%

I'm not sure if this is significant, or just within the experimental margin-of-error.

Perhaps @roji has an opinion on whether explicit `Prepare()` should be used for both npgsql and MySqlConnector or left like this.

bgrainger: Also we should probably update npgsql from 4.0.0-rc1 to 4.0.2, but I didn't address that in this PR.

roji: In principle an explicit `Prepare()` should be best for Npgsql. But IIRC there was some unexplained weirdness at some point where using auto-prepare was actually faster than explicit `Prepare()`. I didn't have any time to look into it (and don't have any right now either...), but maybe @sebastienros can re-confirm this...

sebastienros:
I confirm autoprepare increased the perf significantly over Prepare().

bgrainger:
@sebastienros Once this is merged, can you change the four instances of 

     (cmd as MySql.Data.MySqlClient.MySqlCommand)?.Prepare();

to

    cmd.Prepare();

and check to see if you can still repro a performance decrease in `aspcore-ado-pg` and `aspcore-mw-ado-pg` when `cmd.Prepare()` is used?

bgrainger:
That's fine if you want to measure before and after this change; based on my local testing I'm pretty confident in the net effect being about +20%.

However, I really don't know if `cmd.Prepare()` is bad for npgsql so that's why I'm wondering if you could test that hypothesis without my having to push a potentially-negative change to TFB.

sebastienros:
That's my concern, regression npgsql because I tested it by the past. Maybe there used to be an issue with how `Prepare()` works. @roji when `Prepare()` is called, are you checking for existing prepared statements on the pooled connection before issuing the command to the server?

I will test this branch today for both dbs and update here.

sebastienros:
My local results:


Test | Before | After | Change
-- | -- | -- | --
Npgsql - Update Libraries | 104,905 | 106,004 | 1.05%
Npgsql - Prepare() | 106,004 | 104,039 | -1.85%
MySql | 76,880 | 80,213 | 4.34%

Note I wasn't able to test it of the same infra as TE (for reasons I have to investigate).

Conclusion is that 
- Updating to the latest Npgsql version improves the results
- Call `Prepare()` explicitly in Npgsql is confirmed to be slower than `AutoPrepare`
- You changes do have some positive impact (though I am not measuring 20%)

I'd say we go with this PR for now, and you can update the Npgsql version or we do it in a separate one.

roji:
Thanks for measuring the Npgsql side, this is really odd. In a couple months I will start getting more free and will try to investigate why explicit `Prepare()` is slowing things down.

sebastienros:
@roji I just want to insist on the fact that this table has `AutoPrepare` on in both cases. And I prefer this to the explicit call as this also benefits the EF/Dapper scenarios.

roji:
Oh, ok... Is it possible to do a run with explicit `Prepare()` but without `AutoPrepare`? Just to get the full picture.

Yeah, `AutoPrepare` is indeed you're only option when using Dapper/EF, but in theory we still want the raw ADO.NET perf to be as good as possible.

bgrainger:
@sebastienros Looks like your benchmark suite is much more accurate than my local Docker results. Here's what I got from comparing the [before](https://www.techempower.com/benchmarks/#section=test&runid=ede0f29b-899b-4b74-939c-c3267a2129eb) and [after](https://www.techempower.com/benchmarks/#section=test&runid=6feb4a73-1fca-4d1c-9cdc-cdd46df81a82) runs:

Benchmark | Framework | Before | After | Delta
--- | --- | --- | --- | ---
Fortunes | aspcore-ado-my | 181,534 | 190,574 | +5.0%
Fortunes | aspcore-vb-mw-ado-my| 170,391 | 170,549 | +0.1%
Fortunes | aspcore-mw-ado-my | 163,827 | 170,023 | +3.8%
Fortunes | aspcore-mw-dap-my	| 155,365 | 162,011 | +4.3%
Fortunes | aspcore-mvc-ado-my | 94,900 | 96,662 | +1.9%
| | | |
DB | aspcore-ado-my | 217,032 | 231,045	| +6.5%
DB | aspcore-mw-ado-my | 191,171 | 201,157 | +5.2%
DB | aspcore-vb-mw-ado-my | 191,207 | 197,575 | +3.3%
DB | aspcore-mvc-ado-my | 133,124 | 138,630 | +4.1%
| | | |
Query | aspcore-mw-ado-my | 10,213 | 18,839 | +84.5%
Query | aspcore-mvc-ado-my | 11,059 | 18,580 | +68.0%
Query | aspcore-vb-mw-ado-my | 12,258 | 17,670 | +44.2%
| | | |
Update | aspcore-mvc-ado-my | 2,747 | 2,795 | +1.7%
Update | aspcore-mw-ado-my | 2,341 | 2,550 | +8.9%
Update | aspcore-vb-mw-ado-my | 2,736 | 2,440 | -10.8%
Update | aspcore-ado-my | 2,538 | 	2,288 | -9.9%

One interesting thing is comparing `aspcore-vb-mw-ado-my` to `aspcore-mw-ado-my`. As far as I know, the code is generally supposed to be the same (although I haven't verified that the same IL is generated), the MySqlConnector version and connection string are (now) the same; the only difference is that I added the `cmd.Prepare()` call to the C# version. Thus we can see that for "Fortunes", preparing apparently makes no difference (i.e., C# and VB are tried), but in "DB" and "Query", there may be some benefit to preparing the query for MySQL.

I don't understand the "Update" benchmarks at all (but we've discussed that separately).

patricka2:
.NET Core enable "LogLevel" default to "Error" by default even by clear "LogLevel" in appsettings.json (Like now) in both production and development mode.
Is not it better to set "LogLevel" default to "Error" instant "Warring" to disable logging like this:
problem

patricka2:
.NET Core enable "LogLevel" default to "Error" by default even by clear "LogLevel" in appsettings.json (Like now) in both production and development mode.
Is not it better to set "LogLevel" default to "Error" instant "Warring" to disable logging like this:
eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee

PR #4020, opened 2018-08-28
Update Crystal and frameworks

vlazar:
Updates Crystal and all Crystal frameworks to the latest versions.

Successor of #3884

### DIFF
--- a/frameworks/Crystal/amber/amber.dockerfile
+++ b/frameworks/Crystal/amber/amber.dockerfile
@@ -1,4 +1,4 @@
-FROM crystallang/crystal:0.24.1
+FROM crystallang/crystal:0.26.1
 
 WORKDIR /amber
 COPY config config
--- a/frameworks/Crystal/amber/config/initializers/database.cr
+++ b/frameworks/Crystal/amber/config/initializers/database.cr
@@ -1,3 +1,4 @@
-require "granite_orm/adapter/pg"
+require "granite/adapter/pg"
 
-Granite::ORM.settings.logger = Logger.new(nil)
+Granite.settings.logger = Logger.new(nil)
+Granite::Adapters << Granite::Adapter::Pg.new({name: "pg", url: ENV["DATABASE_URL"]})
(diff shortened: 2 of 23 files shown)


vlazar:
I'm aware of https://tfb-status.techempower.com and I'm pretty sure I've seen a diff view of benchmark results between different git revisions somewhere, but couldn't find it.

Is there a link somewhere on UI where so I could see the diff for Crystal results for this PR has once it gets merged?

cjnething:
Hi @vlazar, it seems like you already found this but I'll go through the steps just in case. If you go to [the status site](https://tfb-status.techempower.com/) and click on "visualize" for each run ID, that will take you to a view [like this](https://www.techempower.com/benchmarks/#section=test&runid=9d5522a6-2917-467a-9d7a-8c0f6a8ed790) and each run is tied to a particular commit, which you can see if you click on "details" for each run ID, [like here](https://tfb-status.techempower.com/results/9d5522a6-2917-467a-9d7a-8c0f6a8ed790). The benchmarks run continuously so you would just have to monitor the status site for the run that starts after you PR gets merged. If that's not what you're looking for let me know!

vlazar:
@cjnething Sorry I wasn't clear. What I want to see is the difference my PR made on some test result, eg for some tab like Fortunes there will be a result for particular Crystal implementation like `crystal-radix` which is now about `20.4%`.

So I would like to see the number after my PR merged in diffed with an old result. SOmething like:

```diff
crystal-radix    +12.5%
```

vlazar:
@cjnething I guess I've seen something like this https://github.com/TechEmpower/FrameworkBenchmarks/pull/3229#issuecomment-362018875 - is this on UI somewhere?

cjnething:
Hi @vlazar, I asked the team and it seems that we do have a tool for that but it's internal. Sorry about that!

vlazar:
@cjnething Ok, thanks. Not a big deal but can be useful with some UI externally to compare runs or test rounds. I'll use some json diff tool instead meanwhile.

bhauer:
I've created a new issue (#4022) concerning adapting that internal tool for external use.

PR #4031, opened 2018-09-07
Allow text/html as a Content-Type in headers

davidlee-techempower:
Modified verifications.py to allow html content type of "text/html", "text/html; chartset=UTF-8", or "text/html; chartset=utf-8". Brought up in issue [77](https://github.com/TechEmpower/TFB-Documentation/issues/77) from docs repo.


michaelhixson:
We should fix the documentation rather than change this requirement.  The fortunes response body includes UTF-8 content, so the response headers should specify UTF-8 as the charset.  I'd expect web browsers to render the response incorrectly otherwise, and indeed I see that incorrect rendering locally when I try this.

davidlee-techempower:
That makes sense, utf-8 is the standard now and websites really shouldn't be using the old ISO format. I will make an update to the docs to reflect this.

PR #4038, opened 2018-09-12
Adds an Nginx configuration for Hapi and Postgres

jcoleman-techempower:


frameworks/JavaScript/hapi/hapi-nginx.dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM node:10.10.0
+
+RUN apt update
+RUN apt install nginx -y
+
+WORKDIR /
+
+COPY ./ ./
+RUN chmod +x start-servers.sh
+RUN mkdir -p /tmp/nginx/cache
```
msmith-techempower: This is unused, right?

frameworks/JavaScript/hapi/start-servers.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+PORT=3000 node create-server.js &
+PORT=3001 node create-server.js &
+PORT=3002 node create-server.js &
+PORT=3003 node create-server.js &
+PORT=3004 node create-server.js &
+PORT=3005 node create-server.js &
+PORT=3006 node create-server.js &
+PORT=3007 node create-server.js &
```
nbrady-techempower: Is the intention here to spin up 8 cores because that's what you were testing on? To scale correctly and test in other environments, could you do this dynamically? https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Dart/stream/start-servers.sh as an example. You can build the nginx conf the same way.

jcoleman-techempower: Changing now. Will run a verify

frameworks/JavaScript/hapi/hapi-nginx.dockerfile
jcoleman-techempower: Totally unused. Removed

frameworks/JavaScript/hapi/start-servers.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+CPU_COUNT=$(nproc)
+P=3000
+END=$(($P+$CPU_COUNT))
+
+while [ $P -lt $END ]; do
+  PORT=$P node create-server.js &
```
nbrady-techempower: In this loop, you could use `sed` to insert your `server` lines under `upstream workers {` in your `nginx.conf` or use a separate script. Either way, that has to be created dynamically as well.

jcoleman-techempower: ah, right. Thank you for pointing that out

PR #4040, opened 2018-09-13
completed graphql-express-mysql implementation and passed tests

jenriquez-techempower:


jcoleman-techempower:
Wouldn't this just belong with Express?

jenriquez-techempower:
> Wouldn't this just belong with Express?

Possibly. It could be moved.

.travis.yml
```diff
@@ -81,6 +81,7 @@ env:
      - "TESTDIR=Java/wildfly-ee7"
      - "TESTDIR=Java/wizzardo-http"
      - "TESTDIR=Scala/cask"
+     - "TESTDIR=JavaScript/graphql-express-mysql"
```
msmith-techempower: Remove this please, the line below will ensure it gets tested with the other JavaScript impls.

frameworks/JavaScript/graphql-express-mysql/src/server.js
```diff
@@ -0,0 +1,316 @@
...
+    res.status(200);
+    res.contentType('application/json', 'charset=UTF-8');
+    res.setHeader('date', dateFormat(now, "yyyy-mm-dd hh:MM:ss o"));
+    res.setHeader('Server', 'GraphQL-Express-MySQL');
+    res.setHeader('Content-Length', jsonLength);
+};
+
+app.listen(port, () => {
+    console.log(`Listening on localhost:${port}`);
+});
```
msmith-techempower: This code looks fine to me, but I think that this **should** be another permutation in the `express` folder (since this is express most of the way, and we're just treating GraphQL as a complicated ORM).

jenriquez-techempower:
Changes made

frameworks/JavaScript/express/graphql-mysql-app.js
```diff
@@ -0,0 +1,295 @@
...
+
+        res.real_end(htmlToRet);
+    }
+};
+
+const spoofHTML = arr => {
+
+    return new Promise((resolve, reject) => {
+
+        var count = 0;
```
nbrady-techempower: Super JS nit for me: we have `let`, `const`, and `var` all being used here.. can we replace all the `var` declarations with `let`

jenriquez-techempower:
Made the changes Nate.

PR #4041, opened 2018-09-14
Adding the Armeria framework to the tests

davidlee-techempower:
Adding Armeria framework to the suite of tests.

This framework tests for all 6 tests and uses Postgres for the database.

### DIFF
--- a/frameworks/Java/armeria/pom.xml
+++ b/frameworks/Java/armeria/pom.xml
@@ -35,6 +35,12 @@
       <artifactId>compiler</artifactId>
       <version>0.9.5</version>
     </dependency>
+    <dependency>
+      <groupId>com.zaxxer</groupId>
+      <artifactId>HikariCP</artifactId>
+      <version>2.7.8</version>
+      <scope>compile</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/frameworks/Java/armeria/pom.xml
+++ b/frameworks/Java/armeria/pom.xml
@@ -12,9 +12,8 @@
 
   <properties>
     <!-- Compiler options -->
-    <maven.compiler.compilerVersion>1.8</maven.compiler.compilerVersion>
-    <maven.compiler.source>1.8</maven.compiler.source>
-    <maven.compiler.target>1.8</maven.compiler.target>
+    <maven.compiler.source>10</maven.compiler.source>
+    <maven.compiler.target>10</maven.compiler.target>
 
     <!-- Dependency versions -->
     <armeria.version>0.71.1</armeria.version>
--- a/frameworks/Java/armeria/pom.xml
+++ b/frameworks/Java/armeria/pom.xml
@@ -1,0 +1,72 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
+                             http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>hello</groupId>
+  <artifactId>hello</artifactId>
+  <version>1.0-SNAPSHOT</version>
+
+  <properties>
+    <!-- Compiler options -->
+    <maven.compiler.compilerVersion>1.8</maven.compiler.compilerVersion>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
+
+    <!-- Dependency versions -->
+    <armeria.version>0.71.1</armeria.version>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.linecorp.armeria</groupId>
+      <artifactId>armeria</artifactId>
+      <version>${armeria.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.postgresql</groupId>
+      <artifactId>postgresql</artifactId>
+      <version>42.1.4</version>
+    </dependency>
+    <dependency>
+      <groupId>com.github.spullara.mustache.java</groupId>
+      <artifactId>compiler</artifactId>
+      <version>0.9.5</version>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+
+      <!-- Build an executable JAR with dependencies -->
+      <plugin>
+        <artifactId>maven-assembly-plugin</artifactId>
+        <version>3.1.0</version>
+        <configuration>
+          <archive>
+            <manifest>
+              <mainClass>hello.App</mainClass>
+            </manifest>
+          </archive>
+          <descriptorRefs>
+            <descriptorRef>jar-with-dependencies</descriptorRef>
+          </descriptorRefs>
+        </configuration>
+        <executions>
+          <execution>
+            <id>make-assembly</id>
+            <phase>package</phase>
+            <goals>
+              <goal>single</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+
+    </plugins>
+  </build>
+
+</project>
--- a/frameworks/Java/armeria/src/main/java/hello/App.java
+++ b/frameworks/Java/armeria/src/main/java/hello/App.java
@@ -1,0 +1,25 @@
+package hello;
+
+import hello.services.HelloService;
+import hello.services.PostgresDbService;
+import hello.services.PostgresFortunesService;
+
+import com.linecorp.armeria.common.SessionProtocol;
+import com.linecorp.armeria.server.Server;
+import com.linecorp.armeria.server.ServerBuilder;
+
+public final class App
+{
+  public static void main(String[] args)
+  {
+    ServerBuilder sb = new ServerBuilder();
+
+    sb.port(8080, SessionProtocol.HTTP).annotatedService("/",
+        new HelloService()).annotatedService("/",
+        new PostgresDbService()).annotatedService("/",
+        new PostgresFortunesService());
+
+    Server server = sb.build();
+    server.start().join();
+  }
+}
--- a/frameworks/Java/armeria/src/main/java/hello/App.java
+++ b/frameworks/Java/armeria/src/main/java/hello/App.java
@@ -8,16 +8,14 @@ import com.linecorp.armeria.common.SessionProtocol;
 import com.linecorp.armeria.server.Server;
 import com.linecorp.armeria.server.ServerBuilder;
 
-public final class App
-{
-  public static void main(String[] args)
-  {
+public final class App {
+  public static void main(String[] args) {
     ServerBuilder sb = new ServerBuilder();
 
-    sb.port(8080, SessionProtocol.HTTP).annotatedService("/",
-        new HelloService()).annotatedService("/",
-        new PostgresDbService()).annotatedService("/",
-        new PostgresFortunesService());
+    sb.port(8080, SessionProtocol.HTTP)
+      .annotatedService("/", new HelloService())
+      .annotatedService("/", new PostgresDbService())
+      .annotatedService("/", new PostgresFortunesService());
 
     Server server = sb.build();
     server.start().join();
--- a/frameworks/Java/armeria/src/main/java/hello/helpers/HttpHeadersHelper.java
+++ b/frameworks/Java/armeria/src/main/java/hello/helpers/HttpHeadersHelper.java
@@ -10,7 +10,7 @@ import com.linecorp.armeria.common.HttpStatus;
 import com.linecorp.armeria.common.MediaType;
 
 public class HttpHeadersHelper {
-  public HttpHeaders getHttpHeader(MediaType mediaType) {
+  public static HttpHeaders getHttpHeader(MediaType mediaType) {
     return HttpHeaders
         .of(HttpStatus.OK)
         .add(HttpHeaderNames.SERVER, "armeria")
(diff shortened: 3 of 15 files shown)


frameworks/Java/armeria/src/main/java/hello/services/PostgresFortunesService.java
```diff
@@ -0,0 +1,112 @@
...
+    {
+    }
+    return fortunes;
+  }
+
+  private String buildMustacheTemplate(List<Fortune> fortunes)
+      throws IOException
+  {
+    // for some reason cannot access mustache template with file name.
+    // manually getting the file and converting to a stream
```
michaelhixson: It looks like this is reading the Mustache template file and recompiling it on each request.  You want to use the `compile(String)` method of `DefaultMustacheFactory` because that will cache the template by name after compiling it once.

I take it from your comment that you tried `mustacheFactory.compile("fortunes.mustache")` at first but that wasn't working.  What error were you getting?

frameworks/Java/armeria/src/main/java/hello/services/PostgresDbService.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+    catch (JsonProcessingException e)
+    {
+      return HttpResponse.of(HttpStatus.INTERNAL_SERVER_ERROR);
+    }
+  }
+
+  private int getRandomNumber(int min, int max)
+  {
+    Random r = new Random();
```
michaelhixson: You generally want to reuse instances of `Random` rather than creating a new instance for each number that you generate.  And if the `Random` is going to be used by many threads, which I think this is, then that's what `ThreadLocalRandom` is designed for.  That class also conveniently has a `nextInt(minInclusive, maxExclusive)` function, unlike `Random`.  AFAIK most of the other Java frameworks in TFB are using `ThreadLocalRandom` for this.

Here's one example:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/cffabe9d29f4bf4b3a9ad845d70726a8ac87fd11/frameworks/Java/undertow/src/main/java/hello/HelloWebServer.java#L254-L256

frameworks/Java/armeria/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+
+  <groupId>hello</groupId>
+  <artifactId>hello</artifactId>
+  <version>1.0-SNAPSHOT</version>
+
+  <properties>
+    <!-- Compiler options -->
+    <maven.compiler.compilerVersion>1.8</maven.compiler.compilerVersion>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
```
michaelhixson: Since this is running on Java 10 (according to the dockerfile), I think this should be source=10 and target=10.

In newer versions of Java there's a `release` option that's supposed to be used instead of `source` and `target`, but I forget whether that's supported in Maven or not.

frameworks/Java/armeria/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>hello</groupId>
+  <artifactId>hello</artifactId>
+  <version>1.0-SNAPSHOT</version>
+
+  <properties>
+    <!-- Compiler options -->
+    <maven.compiler.compilerVersion>1.8</maven.compiler.compilerVersion>
```
michaelhixson: I'm not familiar with the `maven.compiler.compilerVersion` property...  Did you check if it works without this?

frameworks/Java/armeria/src/main/java/hello/App.java
```diff
@@ -0,0 +1,25 @@
...
+import com.linecorp.armeria.server.ServerBuilder;
+
+public final class App
+{
+  public static void main(String[] args)
+  {
+    ServerBuilder sb = new ServerBuilder();
+
+    sb.port(8080, SessionProtocol.HTTP).annotatedService("/",
+        new HelloService()).annotatedService("/",
```
michaelhixson: (Usually we don't comment on highly-subjective matters of code style like this in TFB.  The following is more of a TechEmpower-internal comment than a TFB comment.)

When doing chained method calls like this, prefer to insert line breaks before each `.` rather than breaking between method parameters.  So in this case, something more like this:

```java
sb.port(8080, SessionProtocol.HTTP)
    .annotatedService("/", new HelloService())
    .annotatedService("/", new PostgresDbService())
    ...
```

Here's one explanation from Google's Java style guide, which pretty closely matches TechEmpower's preferred Java style. 
 See "4.5.1 Where to break":  https://google.github.io/styleguide/javaguide.html#s4.5.1-line-wrapping-where-to-break
> The prime directive of line-wrapping is: prefer to break at a **higher syntactic level**.

In this case, the `.annotatedService(...)` method calls occur at a higher syntactic level than the lists of input parameters.

I also really like the "rectangle rule" explanation from an older version of Google's JavaScript style guide.  They removed this section from more recent versions of the guide, but I still think it's good advice.  https://github.com/google/styleguide/blob/d012c726248bee61bdc20bd68a40fd67778739ae/jsguide.html#L650-L684
> When a source file is formatted, each subtree gets its own bounding rectangle, containing all of that subtree's text and none of any other subtree's.

In this case, consider the subtree `.annotatedService("/", new HelloService())`.  Under your current formatting, it's not possible to draw a rectangle around `.annotatedService("/", new HelloService())` that excludes all other code.

frameworks/Java/armeria/src/main/java/hello/services/HelloService.java
```diff
@@ -0,0 +1,57 @@
...
+{
+  private static final byte[]       PLAINTEXT      = "Hello, World!".getBytes(
+      StandardCharsets.UTF_8);
+  private static final ObjectMapper MAPPER         = new ObjectMapper();
+  private static final DateFormat   DATE_FORMATTER = new SimpleDateFormat(
+      "E, dd MMM yyyy HH:mm:ss z");
+
+  @Get("/plaintext") public HttpResponse plaintext()
+  {
+    HttpHeaders headers = HttpHeaders.of(HttpStatus.OK).add(
```
michaelhixson: See my previous comment about line breaks near the `annotatedService(...)` method calls.  The same advice applies here.

frameworks/Java/armeria/src/main/java/hello/services/HelloService.java
```diff
@@ -0,0 +1,57 @@
...
+import com.linecorp.armeria.common.MediaType;
+import com.linecorp.armeria.server.annotation.Get;
+import com.linecorp.armeria.server.annotation.ProducesJson;
+
+public class HelloService
+{
+  private static final byte[]       PLAINTEXT      = "Hello, World!".getBytes(
+      StandardCharsets.UTF_8);
+  private static final ObjectMapper MAPPER         = new ObjectMapper();
+  private static final DateFormat   DATE_FORMATTER = new SimpleDateFormat(
```
michaelhixson: This looks unsafe to me.  A single `SimpleDateFormat` instance should not be used concurrently from multiple threads.  Unless this framework somehow guarantees that only one thread will be executing any one of your request handler methods at a given time (which would have to be disastrous for performance, wouldn't it?), this code is incorrect.

In cases where you have to use `SimpleDateFormat`, I suggest constructing a new instance each time you have to format a date.  In other words, replace uses of `DATE_FORMATTER` with `new SimpleDateFormat(...)`.  (You may be able to optimize that by using a `ThreadLocal<SimpleDateFormat>`, but it would make the code slightly more complicated.  In most circumstances I don't bother with `ThreadLocal` for stuff like this.)

However, both `SimpleDateFormat` and `java.util.Date` should be avoided in modern Java code.  Those are considered part of the legacy date-time libraries in Java.  Java 8 introduced the `java.time` packages as a replacement.  In particular, `java.time.format.DateTimeFormatter` is the thread-safe alternative to `SimpleDateFormat`.  It has a built-in constant for the standard HTTP date format.

Try out the code from this StackOverflow answer: https://stackoverflow.com/a/26367834

```java
DateTimeFormatter.RFC_1123_DATE_TIME.format(ZonedDateTime.now(ZoneOffset.UTC))
```

frameworks/Java/armeria/src/main/java/hello/services/PostgresFortunesService.java
```diff
@@ -0,0 +1,112 @@
...
+  @Get("/fortunes") public HttpResponse fortunes()
+  {
+    List<Fortune> fortunes = getFortunes();
+    fortunes.add(new Fortune(0, "Additional fortune added at request time."));
+    fortunes.sort(Comparator.comparing(Fortune::getMessage));
+
+    try
+    {
+      HttpHeaders headers = HttpHeaders.of(HttpStatus.OK).add(
+          HttpHeaderNames.SERVER, "armeria").add(HttpHeaderNames.DATE,
```
michaelhixson: Another subjective nit-picky style comment:

You repeat this logic enough times (6) that it should be extracted to a helper method.  By "this logic" I'm referring to the logic that adds the `Server` and `Date` headers specifically; those will be the same everywhere in your application whereas the status code and `Content-Type` are likely to vary between call sites.

I suggest something like this:

```java
static HttpHeaders addCommonHeaders(HttpHeaders headers) {
    return headers.add(HttpHeaderNames.SERVER, "armeria")
                  .add(HttpHeaderNames.DATE, ...);
}
```

or this:

```java
static HttpHeaders responseHeaders(HttpStatus status, MediaType contentType) {
    return HttpHeaders.of(status)
        .add(HttpHeaderNames.SERVER, "armeria")
        .add(HttpHeaderNames.DATE, ...);
        .contentType(contentType);
}
```

davidlee-techempower: Thank you for all the great feedback. I will make the updates. As for the formatting, I used IntelliJ auto-formatting after importing the settings for the wiki. Not sure why I am getting bad formatting, I might have imported the settings incorrectly up. Either way, I will correct this.

michaelhixson: It could also be that the settings on our wiki are wrong.  I'm not familiar with them.  I'll send you the code style settings that I use.

frameworks/Java/armeria/src/main/java/hello/helpers/HttpHeadersHelper.java
```diff
@@ -0,0 +1,22 @@
...
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+
+import com.linecorp.armeria.common.HttpHeaderNames;
+import com.linecorp.armeria.common.HttpHeaders;
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.MediaType;
+
+public class HttpHeadersHelper {
+  public HttpHeaders getHttpHeader(MediaType mediaType) {
```
michaelhixson: This method doesn't depend on any internal state so it can be `static`:

```java
public static HttpHeaders getHttpHeader(MediaType mediaType) {
```

That way you don't need to create a new instance of `HttpHeadersHelper` in order to use the method.  Where you call it, it'll look like:
```java
HttpHeadersHelper.getHttpHeader(...)
```
rather than:
```java
new HttpHeadersHelper.getHttpHeader(...)
```

frameworks/Java/armeria/src/main/java/hello/services/PostgresDbService.java
```diff
@@ -0,0 +1,161 @@
...
+        return 500;
+      }
+      return intCount;
+    } catch (NumberFormatException e) {
+      return 1;
+    }
+  }
+
+  private World getWorld(int number) throws SQLException {
+    Connection connection = DriverManager.getConnection(
```
michaelhixson: Oh I didn't notice this code before.  There are a couple of issues here that should be addressed.

1. **Close all of the JDBC objects you create when running a query.**

Specifically I mean the `Connection`, `PreparedStatement`, and `ResultSet`.  Use the "try with resources" language feature for this.  A block of code using JDBC tends to look like this:

```java
try (Connection connection = ...;
     PreparedStatement statement = connection.prepareStatement(...)) {
  statement.setInt(1, ...);
  try (ResultSet resultSet = statement.executeQuery()) {
    // resultSet.next() or while (resultSet.next()) ...
  }
}
```

(If you don't need to set any parameter values on the `PreparedStatement`, then the `ResultSet` declaration can be lumped into the first `try` statement.  This should be true for the fortunes query.)

Otherwise you probably have a resource leak.  I'd guess that when you benchmark this code (not in verify mode), performance deteriorates very quickly as it piles up an unbounded number of connections to the database.

2. **Use connection pooling.**

I don't know what `DriverManager.getConnection(url, username, password)` does because I've never used it, but I'm assuming it creates a *new* connection to the database.  That's an expensive thing to be doing on each request.  You want to reuse connections.  This holds true for pretty much any application that talks to a database; it's not a micro-optimization that we only make in benchmark scenarios.

I take it that the Ameria web framework doesn't come with its own JDBC connection pool.  In that case, try [HikariCP](https://brettwooldridge.github.io/HikariCP/).  Setting up the connection pool is easy:

```java
HikariConfig config = new HikariConfig();
config.setJdbcUrl("jdbc:postgresql://tfb-database:5432/hello_world");
config.setUsername("benchmarkdbuser");
config.setPassword("benchmarkdbpass");
config.setMaximumPoolSize(48); // This number is questionable but good enough
DataSource dataSource = new HikariDataSource(config);
```

A `DataSource` has a parameterless `getConnection()` method that retrieves a connection from the pool.  When you close that connection, it returns the `Connection` object to the pool without closing the underlying physical connection.

If you make this `DataSource` a non-static member of the `PostgresDbService` initialized in the constructor, then wherever you need a connection you'd get it like this:

```java
try (Connection connection = dataSource.getConnection();
     PreparedStatement statement = connection.prepareStatement(...)) {
  ...
}
```

davidlee-techempower: Thanks for the info, I should have another commit up soon.

michaelhixson:
Oops, it's not actually being tested by Travis yet.  Please add this framework's directory to `.travis.yml`

https://github.com/TechEmpower/FrameworkBenchmarks/blob/8d22d1a01b938413505323f8fa46e0c022303970/.travis.yml#L39-L41

PR #4046, opened 2018-09-18
SolarLint Code Review

MatheusRV:
The code was submitted to the SolarLint code reviewer, in which 16 were updates that required better study a good part to review the complexity of the code. Two conditions could be aligned because they generated the same output. Some commented code snippets have been removed.


nbrady-techempower:
Thanks for doing this. We're not going to merge in those TODOs but appreciate the look.

PR #4047, opened 2018-09-18
Bump finagle/finatra: 18.8 -> 18.9

nepthar:
* finagle to 18.9
* finatra to 18.9


nepthar:
Oops. Looks like some logging dependencies need to change. I'll update shortly.

nepthar:
There we are. Please take a look!

PR #4048, opened 2018-09-19
Swiftperfect

jenriquez-techempower:


frameworks/Swift/perfect/Sources/Perfect-MySQL/main.swift
```diff
@@ -0,0 +1,336 @@
...
+        self.m = Int(pow(2.0, 31.0)) //2^31 or 2147483648
+        self.shift = 16
+    }
+ 
+    init(a: Int, c: Int, m: Int, shift: Int) {
+        self.a = a
+        self.c = c
+        self.m = m //2^31 or 2147483648
+        self.shift = shift
+    }
```
nbrady-techempower: This initializer is unused and can be removed.

frameworks/Swift/perfect/Sources/Perfect-MySQL/main.swift
```diff
@@ -0,0 +1,336 @@
...
+ 
+    func seed(seed: Int) -> Void {
+        state = seed;
+    }
```
nbrady-techempower: This can be removed as well.

frameworks/Swift/perfect/benchmark_config.json
```diff
@@ -0,0 +1,39 @@
...
+      "platform": "Perfect",
+      "webserver": "Perfect",
+      "os": "Linux",
+      "database_os": "Linux",
+      "display_name": "Perfect",
+      "notes": ""
+    },
+    "mysql": {
+      "update_url": "/updates?queries=",
+      "query_url": "/queries?queries=",
```
nbrady-techempower: You can add these two urls to the default test and have the default test do plaintext, json, update, and query

frameworks/Swift/perfect/perfect.dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM swift:4.1
+
+ADD . /perfect
+WORKDIR /perfect
+RUN apt update -yqq && apt install -yqq libpq-dev && apt install -y xsltproc docbook-xsl uuid-dev
+RUN apt-get -y install libmysqlclient-dev
```
nbrady-techempower: You can probably turn this into:

`RUN apt update -yqq && apt install -yqq libpq-dev xsltproc docbook-xsl uuid-dev libmysqlclient-dev`

I'd also make it a habit to add your apt packages before adding your framework files (`ADD . /perfect`). This way as you're troubleshooting, and changing the files in `./perfect`, your apt packages remain cached by docker.

frameworks/Swift/perfect/Sources/Perfect/main.swift
```diff
@@ -0,0 +1,78 @@
...
+var routes = Routes()
+routes.add(method: .get, uri: "/json", handler: jsonHandler)
+routes.add(method: .get, uri: "/plaintext", handler: plaintextHandler)
+routes.add(method: .get, uri: "/**",
+		   handler: StaticFileHandler(documentRoot: "./webroot", allowResponseFilters: true).handleRequest)
+try HTTPServer.launch(name: "localhost",
+					  port: 8080,
+					  routes: routes,
+					  responseFilters: [
+						(PerfectHTTPServer.HTTPFilter.contentCompression(data: [:]), HTTPFilterPriority.high)])
```
nbrady-techempower: Indentation nit

frameworks/Swift/perfect/Sources/Perfect-MySQL/main.swift
```diff
@@ -0,0 +1,336 @@
...
+    return returnObj
+}
+
+func updateOneFromWorld() -> [String: Any] {
+
+    var returnObj = [String: Any]()
+    var errorObject = [String: Any]()
+
+    let rand = numGenerator.random() % 10000
+    let rand2 = numGenerator.random() % 10000
```
nbrady-techempower: This is actually 0 - 9999 inclusive. Will need to change all of these to:

`numGenerator.random() % 10000 + 1`

frameworks/Swift/perfect/benchmark_config.json
jenriquez-techempower: I probably am not understanding how the benchmark-config works, but I was under the assumption that the default would hit perfect while each other would hit perfect-DB. If I added queries to perfect (which doesn't have a db) wouldn't it just fail?

frameworks/Swift/perfect/Sources/Perfect/main.swift
jenriquez-techempower: Made the changes in all of the main.swift files

frameworks/Swift/perfect/Sources/Perfect-MySQL/main.swift
jenriquez-techempower: Done. Thanks!

frameworks/Swift/perfect/benchmark_config.json
nbrady-techempower: You'd add the DB to that config as well and have all routes registered in `/Perfect/main.swift`. The database connection can still be created when the server is started, but if unused in the `/plaintext`, `/json` routes it shouldn't have any effect on performance. A lot of tests will combine the plaintext and json results with the first db variant and then do a second db variant in another config.. but when you only have one it should look something like this: https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Lua/lapis/benchmark_config.json

jenriquez-techempower: Thanks. I have included 3 dbs now though. I wrote it to look like the express benchmark_config. Should I still make changes to it now that there are 3 db variants?

nbrady-techempower: Oh cool, I hadn't noticed. This is fine then as it matches what we do with gemini. Thanks!

jenriquez-techempower: Awesome! Thanks!

frameworks/Swift/perfect/Sources/Perfect-MySQL/main.swift
```diff
@@ -0,0 +1,322 @@
...
+
+    response.appendBody(string: htmlToRet)
+    
+    setHeaders(response: response, contentType: "text/html")
+    response.setHeader(.custom(name: "CustomLength"), value: String(describing: htmlToRet.count + 32))
+
+    response.completed()
+}
+
+func updatesHandler(request: HTTPRequest, response: HTTPResponse) {
```
bgrainger: I don't believe this meets the requirements for "Test type 5: Database updates":

> 4. The request handler must retrieve a set of World objects, equal in count to the `queries` parameter, from the World database table.

I don't see this happening anywhere in this request handler. It calls `updateOneFromWorld` _n_ times, but that function simply executes one `UPDATE` statement. There is no `SELECT` statement that I can see, particularly one that occurs before the DB is updated.

> 6. At least the `randomNumber` field must be read from the database result set.

The code on lines 138-143 fakes a `returnObj` _without_ reading it from the database. The presence of that code strongly suggests that `randomNumber` is not being read from the database result set. No such code should be necessary in a conformant implementation.

> 7. Each World object must have its `randomNumber` field updated to a new random integer between 1 and 10000.
> 8. Each World object must be persisted to the database with its new `randomNumber` value.

There does not appear to be any in-memory update of a "World" domain object that happens before the DB is updated. The update is a direct update of a DB row (on line 113) with an interpolated string, not a dynamically-generated query based on ORM-style persistence.

PR #4061, opened 2018-09-26
Build techempower/tfb Dockerfile for tests if changed

nbrady-techempower:
From the work done on #3807 @michaelhixson pointed out that the main `Dockerfile` tagged as `techempower/tfb` isn't being built during the travis run if it has changed. This is because we're pulling the image directly from dockerhub.

This PR checks to see if the `Dockerfile` is among the files changed and builds it before each test.

Since that Dockerfile still needs to be published to dockerhub if changed, I've added a new tag to identify that need when a PR is merged.

To test this PR, I've made a couple of minor changes to the base `Dockerfile`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,7 @@ env:
      - "TESTDIR=Java/jlhttp"
      - "TESTDIR=Java/jooby"
      - "TESTDIR=Java/light-java"
-     - "TESTDIR=Java/micronaut"     
+     - "TESTDIR=Java/micronaut"
      - "TESTDIR=Java/minijax"
      - "TESTDIR=Java/nanohttpd"
      - "TESTDIR=Java/netty"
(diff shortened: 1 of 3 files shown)


nbrady-techempower:
@msmith-techempower @michaelhixson Thinking this through.. do we need to publish the base `Dockerfile` to dockerhub? We went back on this with `wrk` for ease of use and maybe we should do the same here. Instead, we could just do `docker build -t techempower/tfb - < Dockerfile` [here](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/tfb#L103). 

Some pros and cons:

  * It would add ~ 1 min to each travis test group
  * It would build after the first use of `tfb` and then be cached for additional uses.
  * It would allow us to work on it locally and the toolset would rebuild it for us
  * We would no longer have to manually sync changes to dockerhub. (This would actually eliminate the need for dockerhub altogether)

msmith-techempower:
@nbrady-techempower Agreed; make it so.

michaelhixson:
I also agree.  Let's build the tfb image as part of the script.

PR #4066, opened 2018-09-27
finagle/finatra: Update to 18.9.1 point release.

ryanoneill:
This commit updates the Finagle and Finatra versions used for benchmarks from 18.9.0 to 18.9.1, which is now the latest version.


ryanoneill:
Thank you.

PR #4074, opened 2018-09-29
Fix failing Haskell builds

nbrady-techempower:


nbrady-techempower:
This fixes yesod and gets snap past the same point of failure, but snap is still breaking for other reasons.

PR #4075, opened 2018-09-29
[h2o.cr] add /json end point

S-YOU:
Added json end point to h2o.cr


S-YOU:
@michaelhixson, could you review this please?

PR #4079, opened 2018-10-01
Align all stable Rust versions on 1.29.1

whitfin:
This PR will align all stable Rust images on `1.29.1` - previously there were a couple of different versions used, which doesn't really provide a fair benchmark. I would suggest in future that versions are not changed without changing all others.

### DIFF
--- a/frameworks/Rust/actix/actix-diesel.dockerfile
+++ b/frameworks/Rust/actix/actix-diesel.dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.29.0
+FROM rust:1.29.1
 
 ADD ./ /actix
 WORKDIR /actix
--- a/frameworks/Rust/actix/actix-pg.dockerfile
+++ b/frameworks/Rust/actix/actix-pg.dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.29.0
+FROM rust:1.29.1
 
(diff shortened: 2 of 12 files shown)


nbrady-techempower:
@whitfin Thanks! would you mind merging master into this.. I just merged in a change to the `travis.yml` file because Rust frameworks were exceeding the timeout limit.

whitfin:
@nbrady-techempower you got it - should be good to go now :)

fafhrd91:
I wouldn't enforce "same version" policy for all frameworks. Personally I am interested in outdated frameworks like iron, and in sync frameworks in general.

whitfin:
@fafhrd91 part of benchmarking is ensuring that you're running in the same environments - changing the language version (and using potential optimizations included in newer versions) definitely violate that rule.

fafhrd91:
I have no any interest to fix old and outdated frameworks. TFB provides value beyond just benchmarking of different frameworks

whitfin:
@fafhrd91 sure, but updating stable Rust versions doesn't require fixing old frameworks - Rust promises to be backwards compatible in that regard. Anyway! Not for this issue :)

fafhrd91:
maintaining same version of compiler is fine. enforcing same version of compile is BS

PR #4080, opened 2018-10-02
ULib: remove debug comment

stefanocasazza:
issue noticed by @michaelhixson


michaelhixson:
Thanks!

FTR, this fixes the issue where ulib's TFB implementation code -- the world- and fortunes-related logic, and so on -- was coming from an external repository rather than our own.

PR #4098, opened 2018-10-08
Initial commit of NestJS tfb configuration - included tests - json, p…

jenriquez-techempower:
…laintext, updates, multiple, single


frameworks/TypeScript/nest/src/bench/bench.service.ts
```diff
@@ -0,0 +1,63 @@
...
+            const worldArr = [];
+            const total = this.sanitizeQueries(totalQueries);
+            for (let i = 0; i < total; i++) {
+                let worldToUpdate = await this.getOne();
+                worldToUpdate.randomnumber = Math.floor(Math.random() * 10000) + 1;
+                worldToUpdate = await this.worldRepository.save(worldToUpdate);
+                worldArr.push(worldToUpdate);
+            }
+
+            if (worldArr.length === total) {
```
nbrady-techempower: Should you remove this condition? What happens if the total isn't reached? Does the server respond? Does it error with a rejected promise? If you remove the condition and the total is wrong, the toolset will complain about the number of results returned versus the number of results expected.

frameworks/TypeScript/nest/tslint.json
```diff
@@ -0,0 +1,55 @@
+{
```
nbrady-techempower: Linting stuff can also be removed, but kept locally if you want it.

frameworks/TypeScript/nest/src/bench/bench.service.ts
jenriquez-techempower: Removed. It was unnecessary.

frameworks/TypeScript/nest/src/app.service.ts
```diff
@@ -0,0 +1,8 @@
+import { Injectable } from '@nestjs/common';
```
jcoleman-techempower: We may be able to do away with all the boilerplate code and just leave the `BenchController`

frameworks/TypeScript/nest/src/bench/bench.controller.ts
```diff
@@ -0,0 +1,45 @@
...
+    async getMultipleRandomWorlds(@Query('queries') query){
+
+        return await this.benchService.getMultiple(query);
+    }
+
+    @Get('/updates')
+    @Header('Server', 'NestJS')
+    async updateRandomWorlds(@Query('queries') query){
+
+        return await this.benchService.updateMultiple(query);
```
jcoleman-techempower: We don't need to await this line. Only the line where this function is called as `async updateRandomWorlds` will already return the Promise.

jcoleman-techempower: You can apply the same logic to the other methods in this file.

frameworks/TypeScript/nest/src/bench/bench.service.ts
```diff
@@ -0,0 +1,57 @@
...
+
+@Injectable()
+export class BenchService {
+    constructor(
+        @InjectRepository(WorldEntity)
+        private worldRepository: Repository<WorldEntity>){}
+
+    async getOne(){
+
+        return await this.worldRepository.findOne({ where: {id: Math.floor(Math.random() * 10000) + 1}});
```
jcoleman-techempower: No need to `await` here, either.

PR #4103, opened 2018-10-09
Docker file tuning

bitstorm:

### DIFF
--- a/frameworks/Java/wicket/wicket.dockerfile
+++ b/frameworks/Java/wicket/wicket.dockerfile
@@ -1,8 +1,10 @@
-FROM maven:3.5.3-jdk-8-slim as maven
+FROM maven:3.5.3-jdk-10-slim as maven
 WORKDIR /wicket
 COPY src src
 COPY pom.xml pom.xml
 RUN mvn compile war:war -q
+
+FROM openjdk:10-jdk
 WORKDIR /resin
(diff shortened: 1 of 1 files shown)


bitstorm:
Hi @zloster,

which line are you referring to?

frameworks/Java/wicket/wicket.dockerfile
```diff
@@ -3,11 +3,9 @@ WORKDIR /wicket
 COPY src src
 COPY pom.xml pom.xml
 RUN mvn compile war:war -q
-
-FROM openjdk:8-jdk
```
zloster: This one.

michaelhixson: I prefer to keep this as well.  @bitstorm Could you re-add `FROM openjdk:8-jdk`?

In general, I like having a second `FROM` after the build/compile steps because it better illustrates what is required for a minimal working production environment.  Having Maven or your original source code in production is probably unnecessary, for example.  And I think the consumers of these benchmarks benefit (if only a little bit) from seeing what a minimal working production environment looks like.

I don't feel especially strongly about this, and we don't take this approach consistently throughout TFB, but at the same time I don't think we should be removing these `FROM`s without good reason.

bitstorm: @michaelhixson  Thank you for your explanation! I've re-added the second FROM clause and  I've updated both the images to Java 10 as @zloster  suggested.

PR #4104, opened 2018-10-09
Play Framework version updated to 2.6.20

KoenDG:
Docker image updated to latest in its series.

As a result:
Scala version used by the projects has been updated to version 2.12.7
SBT version to use has been set to version 1.2.3

A single test run with ftb was done on play2-java-ebean-hikaricp, confirmed that build used correct versions, build succeeded and passed.

PR #4105, opened 2018-10-09
For round17, Play framework update

KoenDG:
Backport of #4104 for round 17, if that's allowed.

I'm not seeing results of round17 on the site yet, but the branch exists.

When will they start? 2 iterations ago was February, I think? So is this in preparation for coming February?

Anyway, I made a PR for master and making one for round17 as well, in case it can still be accepted before the tests begin.

Where can I find info on this?


nbrady-techempower:
Sorry @KoenDG Round 17 is closed. We only accept PRs for master.

KoenDG:
Okay, thanks.

PR #4110, opened 2018-10-11
initial commit of javalin framework with two tests - json and plaintext

jenriquez-techempower:


frameworks/Java/javalin/src/main/java/com/lostsys/test/javalin/Bench.java
```diff
@@ -0,0 +1,30 @@
...
+public class Bench {
+
+    public static void main(String[] args) {
+
+        Javalin app = Javalin.create().start(8080);
+        app.get("/plaintext", ctx -> ctx.result("Hello, World!"));
+
+        app.get("/json", ctx -> {
+            Map<String, String> hello = new HashMap<>();
+            hello.put("message", "Hello, World!");
```
michaelhixson: Use `Collections.singletonMap("message", "Hello, World!")` instead (from `java.util.Collections`).  `singletonMap` uses a dedicated `Map` implementation for a single key and value that is much more efficient than `HashMap`.  The impact of this choice on the JSON test results is measurable in some of the faster Java frameworks, IIRC.

If this were a more recent version of Java (after 8), I would say to use `Map.of("message", "Hello, World!")`.

frameworks/Java/javalin/src/main/java/com/lostsys/test/javalin/Bench.java
```diff
@@ -0,0 +1,30 @@
...
+            ctx.json(hello);
+        });
+
+    }
+
+    private static Map<String, String> sayHello() {
```
michaelhixson: Remove this or use it.  IMO it's simple enough to be inlined; no need for a helper method.

frameworks/Java/javalin/javalin.dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM gradle:4.7.0-jdk8
+USER root
+WORKDIR /javalin
+COPY build.gradle build.gradle
+COPY ./ ./
```
michaelhixson: In general, please avoid copying the entire working directory (`COPY ./ ./`) in favor of copying specific files or directories, as you do with `build.gradle` and `src`.  Do this for TFB at least, but I recommend it when using Docker in any context.

Advantages include:
* The dockerfile is more clear about what it requires to build correctly.
* You use Docker's build cache more effectively.
* You're less likely to accidentally include files in the Docker image that you didn't mean to include.

There may be other dockerfiles in this project that do `COPY ./ ./`, but I would like to fix those eventually and want to avoid adding more.

frameworks/Java/javalin/build.gradle
```diff
@@ -0,0 +1,44 @@
...
+}
+
+/* ****************** */
+/*    Application     */
+/* ****************** */
+run {
+    if(project.hasProperty('args')){
+        args project.args.split(',')
+    }
+    jvmArgs = ['-XX:+UseParallelGC','-XX:+AggressiveOpts']//['-XX:+UseNUMA','-XX:+UseParallelGC','-XX:+AggressiveOpts']
```
michaelhixson: Did you mean to leave `-XX:+UseNUMA` commented out?  I haven't measured its performance impact for any framework in the Citrine environment, but if you're going for consistency with other frameworks, I suggest leaving it enabled.

frameworks/Java/javalin/build.gradle
```diff
@@ -0,0 +1,44 @@
+apply plugin: 'java'
+apply plugin: 'eclipse'
```
michaelhixson: The eclipse plugin isn't required for running the benchmark, is it?  If so, can you please remove this line and the Eclipse section at the bottom of this file?  We try to keep the code in TFB the minimum required to run the benchmark, so stuff like functional tests and IDE configuration aren't included here even if they are useful in real-world projects.

frameworks/Java/javalin/javalin.dockerfile
```diff
@@ -0,0 +1,7 @@
...
+COPY src src
+CMD ["gradle", "--no-daemon", "--refresh-dependencies", "run", "-Pargs=8080,production"]
```
michaelhixson: Can you look into whether it's possible to separate this into a `gradle build` step and a `java -jar ...` step?  See some of the other Java frameworks that use Gradle as an example.

We'd make better use of Docker's caching this way.  As you have it now, it does not cache the result of compiling the code so it'll do that on every run even if there was no change to the implementation.  Also, I don't know exactly what `--refresh-dependencies` does, but that definitely sounds like something we don't want to happen in a `CMD` (every time we benchmark the framework) and perhaps not at all (if running this code now pulls version X of a dependency, and running it later will pull version X+1 of that same dependency, that introduces unwanted instability to our builds).

frameworks/Java/javalin/src/main/java/com/lostsys/test/javalin/Bench.java
```diff
@@ -0,0 +1,30 @@
...
+import io.javalin.Javalin;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Bench {
+
+    public static void main(String[] args) {
+
+        Javalin app = Javalin.create().start(8080);
```
michaelhixson: You have `8080` hardcoded here but you also pass `8080` as an argument in the dockerfile's `CMD`.  I assume that one of these is unnecessary.

michaelhixson:
You'll need to add a line to `.travis.yml` for this framework for it to be tested.

frameworks/Java/javalin/build.gradle
jenriquez-techempower: Added them. This is my first time working with Java. I don't understand much of what's going on with the arguments. I just frankenstein'd from implementations of other Java frameworks until it ran.

frameworks/Java/javalin/build.gradle
jenriquez-techempower: removed.

frameworks/Java/javalin/javalin.dockerfile
jenriquez-techempower: I mostly took after the jawn implementation. Made the changes.

frameworks/Java/javalin/src/main/java/com/lostsys/test/javalin/Bench.java
jenriquez-techempower: Done. Thanks for the tip!

frameworks/Java/javalin/build.gradle
michaelhixson: These JVM arguments aren't necessary, but other people maintaining the Java implementations in this project have added them in an attempt to improve the performance results.  Here's what I know about each one:

* `+UseParallelGC`: Use Parallel GC instead of the default garbage collector.  Parallel GC performs better on throughput (which is the primary performance metric we use in TFB) than other collectors, but worse in other metrics such as worst-case latency.  We generally don't use Parallel GC in real-world projects, but it probably has a measurable benefit to requests/s in TFB.
* `+UseNUMA`: Enable optimizations that only apply for [NUMA](https://en.wikipedia.org/wiki/Non-uniform_memory_access) architecture, which the Citrine environment has.  As I understand it, this flag only does anything when Parallel GC is enabled.
* `+AggressiveOpts`: Enable a grab-bag of various optimizations with questionable benefits.  I believe this flag was deprecated in later versions of Java in favor of using more specific flags.

At some point I would like to test the impact of each of these flags in the Citrine environment on a few different representative Java frameworks, halting the continuous benchmarks while we run these tests manually.  Then we can understand once and for all how they affect the results.  Until then, using the same JVM arguments as other frameworks seems fine.  Omitting all of these flags would also be fine.

PR #4119, opened 2018-10-13
Update iter to Python3

MatheusRV:
PEP 469 -- Migration of dict iteration code to Python 3

https://www.python.org/dev/peps/pep-0469/


nbrady-techempower:
Don't see the need for this, so going to close it for now. But if there's a good reason to get this in I'll reopen.

PR #4125, opened 2018-10-16
Update docker-py to 3.5.0

### LINKED ISSUE #4118: Update Docker-Py
I'm trying to implement a simpler way to do tasks inside Dockers, but because Docker-Py is in version 0.3.5 I'm having some bizarre errors, for example, it's not possible to find /bin/bash when running the [docker_helper.py #L177](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/utils/docker_helper.py#L177) It returns me that the file does not exist even though it is there. I already ran several tests and the error persists as https://github.com/docker/docker-py/issues/2149 . How to proceed to update Docker-Py, it usually breaks not closing Docker after RUN, so the next execution will return that it is still running and can not access it.

Last version is 3.5.0 as seen: https://github.com/docker/docker-py/releases/latest

nbrady-techempower:
Resolves #4118

https://github.com/docker/docker-py/releases gives us mostly bug fixes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ RUN apt -qqy install -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--
   libmysqlclient-dev            `# Needed for MySQL-python` \
   libpq-dev                     `# Needed for psycopg2`
 
-RUN pip install colorama==0.3.1 requests MySQL-python psycopg2-binary pymongo docker==3.1.0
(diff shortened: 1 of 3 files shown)


msmith-techempower:
Does this get built every time or does it need to be pushed to Dockerhub?

nbrady-techempower:
@msmith-techempower  We no longer need to push to Dockerhub

MatheusRV:
Awesome, thanks alot. This fix a strange bug when calling a command inside container.run.

PR #4131, opened 2018-10-17
Fix vertx reported vulnerability

nbrady-techempower:
https://nvd.nist.gov/vuln/detail/CVE-2018-12544


zloster:
@nbrady-techempower IMO `verx-web` should be updated also:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/8c1f79fef242e31d5a191741ba2acda79f16c2f7/frameworks/Java/vertx-web/pom.xml#L10-L24

PR #4150, opened 2018-10-23
finagle/finatra: update to 18.10.0

luciferous:
Bumps finagle and finatra benchmarks to the latest release.

### DIFF
--- a/frameworks/Scala/finagle/build.sbt
+++ b/frameworks/Scala/finagle/build.sbt
@@ -1,4 +1,4 @@
-lazy val finagleVersion = "18.9.1"
+lazy val finagleVersion = "18.10.0"
 
 name := "finagle-benchmark"
 scalaVersion := "2.12.5"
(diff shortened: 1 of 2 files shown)


luciferous:
Tested both finatra and finagle with `./tfb`.

```
$ ./tfb --mode verify --test finagle
[...snip...]
Verification Summary
| finagle
|       plaintext     : PASS
|       json          : PASS
```

frameworks/Scala/finagle/build.sbt
```diff
@@ -1,4 +1,4 @@
-lazy val finagleVersion = "18.9.1"
+lazy val finagleVersion = "18.10.0"
 
 name := "finagle-benchmark"
 scalaVersion := "2.12.5"
```
cacoco: should we update scala versions? we're on [2.12.7](https://github.com/twitter/finagle/blob/develop/build.sbt#L58)

PR #4152, opened 2018-10-24
Updating to Hamlet 3.2

vasily-kartashov:


frameworks/PHP/hamlet/Benchmark/Resources/HelloResource.php
```diff
@@ -0,0 +1,30 @@
...
+    private $json;
+
+    public function __construct(bool $json)
+    {
+        $this->json = $json;
+    }
+
+    public function getResponse(Request $request): Response
+    {
+        if ($this->json) {
```
nbrady-techempower: Out of curiosity, might you see better results if you create two different resources that implement `WebResource`, like `PlaintextResource` and `JsonResource`? Having to hit this conditional on every request seems like an unnecessary expense.

vasily-kartashov: Generally speaking the resources in Hamlet should be organised not around formats, but around data they send back.

`Resource` -> checks method types, negotiates content type, produces data, defines appropriate response code, wraps data into appropriate entity

`Entity` -> knows how to format data and send appropriate entity-specific headers.

`Response` -> pretty much a collection of status code + headers + entity.

nbrady-techempower: I see what you're saying. Just pointing out that this `HelloResource` class could be broken up into two different classes instead of instantiating the same class with a json boolean and performing that extra conditional on every request. For benchmarking purposes, hundreds of thousands of requests checking on that boolean is some non-zero expense. I was thinking:

```
class PlaintextResource implements WebResource
{
     public function getResponse(Request $request): Response
    {
       $entity = new PlainTextEntity('Hello, World!');
        return new SimpleOKResponse($entity);
    }
}
```

And another for the json route. In the same way you have the db, update, and fortunes resources. Just a thought to save some ticks.

PR #4154, opened 2018-10-25
Fixed startup host IP selection and added single DB tests.

nathantippy:


nathantippy:
Unable to get travis ci to do clean build, will try again.

PR #4174, opened 2018-11-05
OfficeFloor 3.3.0

sagenschneider:
[ci fw-only Java/officefloor]


frameworks/Java/officefloor/src/woof_benchmark/src/main/java/net/officefloor/benchmark/Logic.java
```diff
@@ -36,76 +32,29 @@ public void db(EntityManager entityManager, ObjectResponse<World> response) {
...
-			int index = i;
-			GetEntry entry = new GetEntry(random.nextInt(1, 10000));
-			flows.getEntry(entry, (escalation) -> {
-				worlds[index] = entry.world;
-				loaded[0]++;
-				if (loaded[0] >= count) {
-					response.send(worlds);
-				}
-			});
+			worlds[i] = entityManager.find(World.class, random.nextInt(1, 10000));
```
nbrady-techempower: Sorry we missed this before, but ThreadLocalRandom's `nextInt()` upper bound is exclusive. Could you change all these to 10001

sagenschneider: No worries... will update PR :+1:

PR #4178, opened 2018-11-06
create dotnet core FastHttpApi http server

IKende:
dotnet core FastHttpApi


frameworks/CSharp/FastHttpApi/beetlex-fasthttpapi.dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM microsoft/dotnet:2.1-sdk-stretch AS build
+WORKDIR /app
+COPY Benchmarks .
+RUN dotnet publish -c Release -o out
+
+WORKDIR /app
+COPY --from=build /app/out ./
```
nbrady-techempower: This doesn't work for me locally because you're not using multi-stage builds. Can you remove the `COPY` line and change to `WORKDIR /app/out`

nbrady-techempower:
After making the change to the dockerfile I suggested above, I can get the server running but it's refusing connections

IKende:
> After making the change to the dockerfile I suggested above, I can get the server running but it's refusing connections
This code can run on Linux Ubuntu\ Windows and pass the wrk4.1.0 test.
This code is not using ASP.net core, is the use of HTTP TCP base, do I need additional configuration?

misiek08:
I'm not sure - trying to check it myself, but it will take me some time.

Check if it's not a problem with IPv6. I tried your image alone and it works, but requests are served in IPv6 and maybe that's the problem. Force the server to listen on 0.0.0.0.

misiek08:
Container with your server is dying right after starting. Console.Read (and Console.ReadLine) won't work as daemon mode. When container is run as background job it don't get any input or even tty, so it can't wait for reading anything and exits immediately.

Similar issues:
https://stackoverflow.com/questions/38549006/docker-container-exits-immediately-even-with-console-readline-in-a-net-core-c
https://stackoverflow.com/questions/40506122/control-lifetime-of-net-core-console-application-hosted-in-docker

IKende:
> Container with your server is dying right after starting. Console.Read (and Console.ReadLine) won't work as daemon mode. When container is run as background job it don't get any input or even tty, so it can't wait for reading anything and exits immediately.
> 
> Similar issues:
> https://stackoverflow.com/questions/38549006/docker-container-exits-immediately-even-with-console-readline-in-a-net-core-c
> https://stackoverflow.com/questions/40506122/control-lifetime-of-net-core-console-application-hosted-in-docker

Thank you very much！

nbrady-techempower:
@IKende can you change your directory name to `beetlex-fasthttpapi` (all lowercase) or `beetlex` or `fasthttpapi` not sure which is your framework name

IKende:
> @IKende can you change your directory name to `beetlex-fasthttpapi` (all lowercase) or `beetlex` or `fasthttpapi` not sure which is your framework name

Okay I'll change it.

frameworks/CSharp/FastHttpApi/Benchmarks/Program.cs
```diff
@@ -0,0 +1,76 @@
...
+        public class JsonMessage
+        {
+            public string message { get; set; }
+        }
+
+
+    }
+
+
+    public class BeetleXHttpServer : BackgroundService
```
davidfowl: You can just use IHostedService instead of BackgroundService. That way you don't have to spin on the token.

IKende: Thank you. I'm going to modify it.

frameworks/CSharp/FastHttpApi/Benchmarks/Program.cs
```diff
@@ -0,0 +1,76 @@
...
+            mDate = DateTime.Now.ToUniversalTime().ToString("r");
+            var builder = new HostBuilder()
+                .ConfigureServices((hostContext, services) =>
+                {
+                    services.AddHostedService<BeetleXHttpServer>();
+
+                });
+
+            var result = builder.RunConsoleAsync();
+            result.Wait();
```
davidfowl: Don’t do run console async just use Run

frameworks/CSharp/FastHttpApi/PlatformBenchmarks/Program.cs
```diff
@@ -0,0 +1,137 @@
...
+            mServer.Open();
+            Console.WriteLine("BeetleX base server");
+            Console.WriteLine($"ServerGC:{System.Runtime.GCSettings.IsServerGC}");
+            Console.Write(mServer);
+            return Task.CompletedTask;
+        }
+
+        public async Task StopAsync(CancellationToken cancellationToken)
+        {
+            return;
```
davidfowl: You probably want to stop the server here

IKende: Can I do this?
```
            var builder = new HostBuilder()
                .ConfigureServices((hostContext, services) =>
                {
                    services.AddHostedService<BeetleXHttpServer>();
                });
            var result = builder.Start();
```
```
        public virtual async Task StopAsync(CancellationToken cancellationToken)
        {
            mApiServer.BaseServer.Dispose();
            return;
        }
```

IKende:
@nbrady-techempower 
What do I need to do merge this pr?

frameworks/CSharp/FastHttpApi/PlatformBenchmarks/HttpParse.cs
```diff
@@ -0,0 +1,102 @@
...
+            //PlaintextBytes
+            string helloword = "Hello, World!";
+            sb.AppendLine("HTTP/1.1 200 OK");
+            sb.AppendLine("Content-Type: text/plain");
+            sb.AppendLine("Content-Length: " + helloword.Length);
+            sb.AppendLine("Server: BeetleX");
+            sb.AppendLine("Date: " + DateTime.Now.ToUniversalTime().ToString("r"));
+            sb.AppendLine("");
+            sb.Append(helloword);
+            PlaintextBytes = Encoding.UTF8.GetBytes(sb.ToString());
```
davidfowl: I don’t think this is legal. You need to create the response every single request, you can’t cache it.

frameworks/CSharp/FastHttpApi/PlatformBenchmarks/Program.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        private void OnJson(PipeStream stream)
+        {
+            HttpParse.Json(stream);
+        }
+
+        public virtual void SessionReceive(IServer server, SessionReceiveEventArgs e)
+        {
+            PipeStream stream = e.Stream.ToPipeStream();
+            string line = null;
```
davidfowl: I haven’t looked but does this assume you’ll get the entire buffer in a single call?

frameworks/CSharp/FastHttpApi/PlatformBenchmarks/HttpParse.cs
nbrady-techempower: @IKende please look at the test requirements. For the json test, you must instantiate and serialize an object on every request. It's also not acceptable to cache the headers: https://frameworkbenchmarks.readthedocs.io/en/latest/Project-Information/Framework-Tests/#test-types

frameworks/CSharp/FastHttpApi/Benchmarks/Benchmarks.csproj.user
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
IKende: okay

frameworks/CSharp/FastHttpApi/PlatformBenchmarks/Program.cs
IKende: Yes, I think about it. There is a bug here, in the multi-receive with request, I need to fix it！

frameworks/CSharp/FastHttpApi/PlatformBenchmarks/HttpParse.cs
IKende: I think  can use this approach and see that other platform languages are also using this approach 
 https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/netty/src/main/java/hello/HelloServerHandler.java
i removed this test.

frameworks/CSharp/FastHttpApi/Benchmarks/Program.cs
```diff
@@ -0,0 +1,72 @@
...
+namespace Benchmarks
+{
+
+    [BeetleX.FastHttpApi.Controller]
+    class Program
+    {
+        private static string mDate;
+        public static void Main(string[] args)
+        {
+            mDate = DateTime.Now.ToUniversalTime().ToString("r");
```
davidfowl: You can’t cache the date header either. It needs to update (we update it every second)

frameworks/CSharp/FastHttpApi/Benchmarks/Program.cs
```diff
@@ -0,0 +1,72 @@
...
+namespace Benchmarks
+{
+
+    [BeetleX.FastHttpApi.Controller]
+    class Program
+    {
+        private static string mDate;
+        public static void Main(string[] args)
+        {
+
```
davidfowl: nit: Remove extra space.

IKende: thank

nbrady-techempower:
This looks good now. One last thing for me, could you rename the `FastHttpApi/` directory to `beetlex` and we can get this merged. Thanks!

PR #4186, opened 2018-11-10
Change max children processes

joanhey:
There are almost no differences in req/s and the latency is greater.
So we revert the change to 1024 again.
Perhaps we will try with less children later.

And also introduced errors when using postgresql.


joanhey:
Hi, again
I was really busy the last 6 months, but I will continue optimizing the php stack.

joanhey:
The job exceeded the maximum time limit for jobs, and has been terminated.

Possibly it will be necessary to separate the php fws in 2 jobs.

nbrady-techempower:
Welcome back @joanhey! If you'd like to take a crack at separating PHP out into two or three jobs in the `.travis.yml` in this PR that'd be great! Keeping in mind, it'd be good to keep closely related frameworks together to utilize the docker cache in each job. Otherwise I can probably get to it by Tuesday.

joanhey:
Separated PHP frameworks in two jobs for travis.

@vasily-kartashov Could you check hamlet-amp and hamlet-roadrunner?

vasily-kartashov:
@joanhey both will be removed very soon. Will submit a PR in a couple of days

joanhey:
@javiereguiluz Could anybody check the symfony test fail ?
https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/461202934#L5350

Thank you

ostrolucky:
symfony test works for me locally. try restarting the job

joanhey:
It's working for me too locally.
For that I send an small commit `Changed php fortune with <?=` for rerun the job and failed again.

Like I will add another job only for the php (for add more servers with php raw), we will check if happen again.

joanhey:
The funny thing is that symfony also fails in the continuous bechmark.
Exactly with the update test too.
https://tfb-status.techempower.com/results/663adbe9-d01e-47d6-aedd-26b7415800f5

joanhey:
Hamlet-amp and hamlet-roadrunner also fail without this PR.

Deleted changes in zen1 and kohana to avoid merge conflicts.

This PR is ready to be merged.

PR #4193, opened 2018-11-12
Not using default JVM memory limits setting Xms and Xmx to expected max range.

nathantippy:


nathantippy:
Last dashboard showed that greenlighting did not startup since it could not allocate all the needed memory due to the default limit of Java.  Please merge this change before the next test cycle starts.

PR #4212, opened 2018-11-24
Add Json benchmark support to aspnetcore-corert

MichalStrehovsky:
This is pretty much just a copypaste of the Json parts of the aspnetcore benchmark.

The notable difference is the Utf8Json.Generated.cs file - this file was generated using the Utf8Json.UniversalCodeGenerator tool provided by the Utf8Json author. Utf8Json relies on Reflection.Emit by default (and the presence of a JIT). The Reflection.Emit part wouldn't work on CoreRT.


frameworks/CSharp/aspnetcore-corert/PlatformBenchmarks/Utf8Json.Generated.cs
```diff
@@ -0,0 +1,163 @@
+﻿#pragma warning disable 618
```
jkotas: Add a comment here that this file was generated by the tool?

MichalStrehovsky: Sure, done. But for the record, I'm not a huge fan of hand-editing generated files. I made sure the file name clearly states it was generated and that the commit description has the details of how.

PR #4215, opened 2018-11-26
Refine threading and reduce kernel calls (Trying to get docker container to run same everwhere)

nathantippy:


nathantippy:
Please merge quickly so it gets into the next run. Thanks

PR #4224, opened 2018-11-29
update to 1.0.10.1128-beta

smthing:


frameworks/Java/smart-socket/src/main/java/org/smartboot/http/Bootstrap.java
```diff
@@ -35,9 +35,10 @@ public void doHandle(HttpRequest request, HttpResponse response) throws IOExcept
             }
         });
         processor.route("/json", new HttpHandle() {
+            byte[] b = JSON.toJson(new Message("Hello, World!"));
```
nbrady-techempower: Is this getting created on every request? If `processor.route` just sets up the route and `public void doHandle()` is what gets called on each request to the the route, this would have to be moved back there to comply with the test rules.

smthing: this problem has been solved.

PR #4238, opened 2018-12-07
D/Hunt: The old project Kiss has been renamed to Hunt

Heromyth:
We have improved this project. A simple HTTP parser has been integrated into Hunt.  It is hoped greatly that Hunt will get a chance to be shown on the results website.


nbrady-techempower:
@Heromyth The ldc docker base doesn't have git installed. You'll have to do that in the dockerfiles

Heromyth:
Did you have any plan to upgrade the LDC compiler?

nbrady-techempower:
@Heromyth We generally rely on contributors and test maintainers to do that, unless we spot it and have the time on our own. Thanks for taking care of this :)

PR #4240, opened 2018-12-07
PHP with h2o server

joanhey:
In local the performance is lower than nginx.


joanhey:
Not found this 2 libs when run `make`
```
Could NOT find LIBUV (missing:  LIBUV_LIBRARIES LIBUV_INCLUDE_DIR)
Could NOT find WSLAY (missing:  WSLAY_LIBRARIES WSLAY_INCLUDE_DIR)
```
Perhaps will affect the performance. But also happen in the h2o built https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/C/h2o

@volyrique could you check it ?

proyb6:
You can compile without those libs as you are not compile as H2O libraries, they have no impact on the performance. More info is on H2O page.

joanhey:
Thank you @proyb6

PR #4242, opened 2018-12-07
Remove php zend1 end of life in 2016

joanhey:
https://framework.zend.com/blog/2016-06-28-zf1-eol.html


joanhey:
The next month we can remove php5, EOL 1 jan 2019
http://php.net/supported-versions.php

joanhey:
Is there a date to update the ubuntu 16.04 docker?

msmith-techempower:
> Any date to update the ubuntu 16.04 docker?

Is there a pressing need to use 18?

joanhey:
No
But actually using docker is really easy to update individually or globally.

And now there are dockers using ubuntu:18.04 and 18.10  (31)
https://github.com/TechEmpower/FrameworkBenchmarks/search?q=ubuntu%3A18&unscoped_q=ubuntu%3A18

Could we use it in our frameworks ?

msmith-techempower:
> Could we use it in our frameworks ?

Oh, of course. **Your** `Dockerfile`s are your to handle as you wish.

I thought you were referring to upgrading [the toolset](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/Dockerfile#L1) to use Ubuntu 18, and I do not see a need for that.

PR #4247, opened 2018-12-09
Add tests for Ktor using an async database layer

Whoops:
This adds a couple additional tests with Ktor, but using non-blocking database clients. In theory this should reduce the overhead of using blocking JDBC calls with a work pool.

I actually went ahead and implemented two different clients since I'm curious how they perform. [Jasync-sql](https://github.com/jasync-sql/jasync-sql) is interesting in that it has Kotlin coroutines integration, which makes the code actually more natural than the JDBC based tests. [Reactive-PG-Client](https://reactiverse.io/reactive-pg-client/) seems more feature full, and is the same client used in the Vert.x tests.


nbrady-techempower:
Since this is a permutation for ktor, we'd prefer this to be in the `Kotlin/ktor` directory. You can add these tests to that `benchmark_config.json`

Whoops:
Done, I've moved it and the existing tests into subdirectories under the existing `ktor` directory. I'm hoping that approach is acceptable, and I'm not sure how I'd integrate them into a single build.

nbrady-techempower:
Yeah, that approach is great. No need to make them a single build.

PR #4250, opened 2018-12-11
Spring webflux r2dbc

davidkiss:
Add new benchmarks for spring-webflux using r2dbc and rxjava2


davidkiss:
@jringuette I finally created this PR that uses webflux.fn with added R2DBC and RxJava benchmarks. Let me know your thoughts

michaelhixson:
@davidkiss Do you mind cleaning up this PR so that it only contains your changes to spring-webflux? Currently, the "Files changed" tab in GitHub shows a whole lot of stuff that I don't think you meant to change as part of this PR.  Perhaps a merge went wrong somewhere.

davidkiss:
Sure, I'll be submitting a new cleaned up PR then

PR #4257, opened 2018-12-13
Update php to Ubuntu18.10

joanhey:
First modern kernel.

But more important is that **ubuntu 18.10 use nginx 1.15.5, with 16.04 was nginx 1.10.3**.

Only h2o fail the build, so use again ubuntu 16.04.

And phalcon only to ubuntu 18.04 now.


joanhey:
Always fail hamlet-amp and hamlet-roadrunner.

Perhaps will be better to delete both.

nbrady-techempower:
ping @vasily-kartashov for Hamlet fixes

PR #4259, opened 2018-12-14
Ngx php embeded php in nginx

joanhey:
Still in development but promising.

This will help to develop in the right direction.


joanhey:
Only fail hamlet.

joanhey:
It's ready to merge.

Fails only hamlet, like always.

PR #4268, opened 2018-12-18
Nginx simplified and faster dockerfile

joanhey:


nbrady-techempower:
@joanhey We prefer to stay away from things like `latest` and `rolling` because it can break tests without warning and it's harder to back test.

joanhey:
Normally I don't like too.
But in this case always come a new version with nginx, and there is nothing more than 1 line.

But I will chage it now.

PR #4271, opened 2018-12-18
Update php to v7.3

joanhey:
Only Phalcon, amp and php_ngx are not ready for php 7.3.

Added faster composer install without the dev dependencies.


joanhey:
Build failed, looks like a toolset problem:

`APIError: 400 Client Error: Bad Request ("Conflict. The container name "/tfb-database" is already in use by container "94a5f3d3ec2d49a25439218f8e28503058352a06e0d6a31756942e52160eb501". You have to remove (or rename) that container to be able to reuse that name.")`

https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/469631264

ping @nbrady-techempower

joanhey:
In my travis job, was working OK. But with less frameworks, perhaps is one of that missing frameworks that create the problem.

Only exits a problem with symfony for don't send correctly the ENV, perhaps for that symfony 4 was slower than symfony 3.
Now sending a commit for that problem.

joanhey:
Still fail 
https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/469658504

I can't find winch framework use the  container (0a646c33a743276c051753f0eeaf7e9b2dc14ea4d28e94e36a4a21bba69af6c2) that fail.

But it is strange that the tool set open a new container before close the previous completely.

nbrady-techempower:
@joanhey Thanks, we have an open issue for that. It happens every so often that the database container takes longer than it should to die after it gets a kill signal. Looking into it. I've restarted the jobs.

joanhey:
I tested locally the laravel framework, as was the last test working before fail.
`tfb --test laravel --mode verify`

After that I can't test any other framework. Give me the same error than travis.
Don't close mysql, nginx and php-fpm atter finish the test.

I send an htop picture after the test finished, that show that are still working.
htop

I open a lot the htop in another vagrant ssh, for check cpu utilization, cpu distribution,...

Perhaps help to solve the issue.

nbrady-techempower:
@joanhey thanks. I found the issue. There's an error bubbling up from the toolset. I'll have a fix up in a moment.

nbrady-techempower:
@joanhey Please merge in the current master; should solve the issue. Locally, you may have to do a `docker ps` and `docker kill` of the database container.

joanhey:
Perfect @nbrady-techempower 

This PR is ready to be merged.

Thinking about comment the 2 hamlet that fail in the `benchmark_config.json` until is solved.

joanhey:
For me is ready to be merged.

The update problem is an intermittent problem in travis.
In one run are OK and the next NOT.
Not sure where is the problem, in travis or toolset.
But I think that is in travis that sometime give spontaneous fails.

PR #4279, opened 2018-12-20
Composer faster autoload

joanhey:


joanhey:
A good demonstration of the spontaneous travis fails. :)

joanhey:
Ready to be merged false fail
@nbrady-techempower

PR #4282, opened 2018-12-21
Add benchmark tests for FastAPI

tiangolo:
This PR adds benchmark tests for [FastAPI](https://github.com/tiangolo/fastapi).

I didn't modify `.travis.yml` as all the Python tests seem to be covered by the `TESTLANG=Python` env var.

And the tests for Starlette (in which FastAPI is based) don't have any additions to `.travis.yml`. The same with Responder, which is also based on Starlette.


tiangolo:
Great! Thanks! :tada:

PR #4290, opened 2018-12-23
Hexagon

jaguililla:
Update framework version and benchmark code.


jaguililla:
The build error was not triggered by Hexagon framework

PR #4301, opened 2018-12-27
rename config parameters

cyberz-eu:
I changed some names of the configurations, if some build breaks, this will fix it


nbrady-techempower:
@cyberz-eu Looks like this didn’t work

nbrady-techempower:
Not sure what your intentions here were, but feel free to re-open once you've got this working the way you'd like.

PR #4303, opened 2018-12-27
aspnetcore: add platform multiple queries and tweak postgresql connectionstring

n-stefan:
Added Platform level Multiple Queries for ASP.NET Core.

Additionally tweaked the PostgreSQL connectionstring; will undo if it proves detrimental to performance.


n-stefan:
Do I need to ask for a review? Thanks.

PR #4305, opened 2018-12-28
Optimize FastAPI server config and JSON route responses

tiangolo:
This PR adds performance optimizations for FastAPI, defined after several local tests.


tiangolo:
Thanks! :tada:

PR #4306, opened 2018-12-28
Rails: refresh the entire project

spizm:
It looks like the rails project here is a bit out of date. It's time for a refresh!
- upgrade ruby to 2.6.0
- upgrade rails to 5.2.2
- upgrade all the gems
- use more idiomatic ruby syles
- use the default rails 5.2 configurations
- don't load all the rails modules


nbrady-techempower:
Fantastic! Thanks @spizm!

PR #4307, opened 2018-12-28
Adds hug framework with uwsgi configuration to tfb project

jcoleman-techempower:

### DIFF
--- a/frameworks/Python/hug/benchmark_config.json
+++ b/frameworks/Python/hug/benchmark_config.json
@@ -19,24 +19,6 @@
         "display_name": "Hug",
         "notes": "CPython 2",
         "versus": "wsgi"
-      },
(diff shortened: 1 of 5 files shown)


frameworks/Python/hug/benchmark_config.json
```diff
@@ -0,0 +1,43 @@
...
+        "orm": "Raw",
+        "platform": "Falcon",
+        "webserver": "None",
+        "os": "Linux",
+        "database_os": "Linux",
+        "display_name": "Hug",
+        "notes": "CPython 2",
+        "versus": "wsgi"
+      },
+      "gunicorn": {
```
nbrady-techempower: the `hug-gunicorn.dockerfile` is missing so the whole thing will fail

PR #4320, opened 2019-01-04
Hexagon data update test improvements

jaguililla:


jaguililla:
Travis failures are produced by other frameworks' tests

PR #4325, opened 2019-01-07
Update to next GL version

nathantippy:


nathantippy:
In the RAW JSON there are no results yet there is also no record of a failure for this current round. I can wait for the log file however the JSON generated last run had multiple fields and reported 0 requests. The current run only shows that start/stop time and nothing else appears in the JSON.

PR #4330, opened 2019-01-07
Elixir Cowboy 2

l1x:
Updating to Cowboy 2, Elixir 1.7 and moving to use a proper release


l1x:
I think the app should listen on another IP other than 127.0.0.1

nbrady-techempower:
@l1x Try binding to `0.0.0.0` in the config.

nbrady-techempower:
@l1x can you remove the files in the `rel` plugin folder. I don't think they're needed for this test.

l1x:
> @l1x can you remove the files in the `rel` plugin folder. I don't think they're needed for this test.

Depends what you are trying to achieve here. 
I have moved the project to use a proper release and setup.sh to use the released build. In this case rel/vm.args can be used to tune the VM behaviour for the optimal performance. This way it is easier to modify the setup for the next person who wants to change this because we follow the standards. In this sense I do think rel folder is necessary. Let me know if you disagree. Also I need to fine tune a bit the vm.args. Can I see the results of the test run somewhere? I have seen the CI saves the results but I am not sure where can I see that file. Thanks in advance.

nbrady-techempower:
Ok, that sounds fine. With this merged, you can see the results on https://tfb-status.techempower.com/ during the next run. For local results, data is saved to a results folder in the top level of our project

PR #4331, opened 2019-01-08
optimize spider-gazelle benchmark

stakach:


frameworks/Crystal/spider-gazelle/src/controllers/benchmark.cr
```diff
@@ -24,7 +22,7 @@ class Benchmark < Application
   # Postgres Test 2: Single database query
   get "/db", :db do
     results = {} of Symbol => Int32
-    if world = World.find rand(1..ID_MAXIMUM)
+    if world = World.find(Random.rand(ID_MAXIMUM).succ)
```
nbrady-techempower: According to Crystal docs `Random.rand` Generates a random integer which is greater than or equal to 0 and less than max. The range was fine before, but this change would never get ID 10,000 and sometimes fail with ID 0.

stakach: The `succ` adds 1 to the result to prevent 0.
Good catch on the max. Thanks!

nbrady-techempower: Oh, sorry. I think you were fine then. It adds 1 always? So if `ID_MAXIMUM` is `9999` you'll get 1 to 10000 inclusive? Changing `ID_MAXIMUM` to `10000` might get you 10001 sometimes?

stakach: No you were right about the maximum.
‘.rand(2)’ will result in 0 or 1 so adding 1 will move the range to 1 or 2 so the max should be 10000

nbrady-techempower: Ah, right. Thanks!

PR #4332, opened 2019-01-08
Luminus updates

l1x:
Just minor version updates


l1x:
Sorry wrong branch.

l1x:
The git commits are still there but the change is no limited to a single file. I am not sure if I can remove the unnecessary commits with git somehow. Anyways it is good to go.

PR #4333, opened 2019-01-09
Tweak aspnet pg connection strings

benaadams:
Pull the main pool down to 256 to reduce the scan time and the Update pool down to 64

/cc @sebastienros @n-stefan


sebastienros:
Oh! didn't realize it was merged

sebastienros:
"Pending" ... :/

image

sebastienros:
@n-stefan

Did you actually run some tests that proved it was faster? With how many client connections?
Techempower will run database scenario with up to 512 client connections, so the pool should be at least that big.

n-stefan:
Yes, using the TechEmpower toolset, here is Platform Updates with pool size at 1024:

[Plt_Updates_1024.zip](https://github.com/TechEmpower/FrameworkBenchmarks/files/2738953/Plt_Updates_1024.zip)

And at 64:
[Plt_Updates_64.zip](https://github.com/TechEmpower/FrameworkBenchmarks/files/2738957/Plt_Updates_64.zip)

nbrady-techempower:
The latest run on https://tfb-status.techempower.com/ has this change in it. Do you guys want to wait to see how that turns out?

benaadams:
Would seem wise :)

n-stefan:
Yes, please :)

benaadams:
Looks like it drops with the higher connection counts (256, 512) for Fortune
However does look like its significantly higher for Updates

Might want to change the connection pool size based on scenario?

n-stefan:
In the current unfinished run the best Updates result has gone from 4,900 to 13,718.
So maybe 512 for Updates and 1024 for the rest, but is it legal to use different connection strings based on scenario?

benaadams:
I imagine 512 for updates wouldn't change much over 1024.

64 might be reducing write contention, hence the boost; but reducing the read concurrency hence the drop; at a guess...

/cc @roji

roji:
Interesting...

The connection pool is currently implemented as a simple fixed-size array that's scanned sequentially until an idle connector is found. As a result, if the array is big the scans may start to take long if an idle connection happens to be near the end. On the other hand, if the pool is too small, then there may not be any idle connections and the attempt must wait until a connection is released, which should be pretty bad for perf. In other words, the sweet spot is the lowest number where nobody needs to wait... I guess this number is going to vary by scenario - that makes sense.

I do have some ideas about improving the pool design for the next version, but nothing concrete yet.

benaadams:
>  and the attempt must wait until a connection is released

Thinking this might be where the improvements to Updates is as it reduced the write contention on the db.

>The connection pool is currently implemented as a simple fixed-size array that's scanned sequentially until an idle connector is found
> ...
> I do have some ideas about improving the pool design for the next version, but nothing concrete yet.

Try `Channel.CreateBounded<Connection>` https://github.com/dotnet/corefx/blob/a10890f4ffe0fadf090c922578ba0e606ebdd16c/src/System.Threading.Channels/tests/BoundedChannelTests.cs#L148-L166

Point me at the repo ;)

roji:
Here you go: https://github.com/npgsql/npgsql/blob/dev/src/Npgsql/ConnectorPool.cs. I've been wanting to take a look at channels for a while now...

Uh oh, @benaadams is looking at my repo :)

roji:
Without diving too deep into channels, here are some first probable issues:

* The pool should actually behave like a stack, where the connections return last are returned first. This would allow allow idle connections to be pruned after they've been hanging around in the pool for a long time (this actually currently doesn't happens: https://github.com/npgsql/npgsql/issues/2224). I'm not 100% sure, but channels seem to implement FIFO semantics?
* Npgsql currently tries to provide first-class support for sync I/O in parallel to async I/O, but channels don't seem to provide APIs for blocking synchronously. This is also a problem for adopting Pipelines, but maybe at some point I'll bite the bullet and see about dropping sync support.

There may some other issues, no sure yet...

bhauer:
For whatever it's worth, in our in-house framework, we use a similar approach for the connection pool: a fixed-size array.  My thinking at the time was that maintenance of a more complex data structure (such as a queue of available connections) would in most real-world cases cause greater overhead than scanning an array for a free connection.  I figure that is true for smaller connection pools and eventually becomes untrue at some connection pool size.

A couple small optimizations we have:

* Each thread has an affinity for a specific connection and attempts to acquire a lock on that connection first and, only if it is unavailable, asks the pool to scan for an available connection.  In many cases, our apps are deployed with a number of request worker threads equal to the number of connections in the pool, so each thread is effectively hard-wired to a single database connection.

* The pool maintains a shared atomic index "cursor" so that we're not continuously scanning the first N connections looking for something available, leaving a bulk of later connections idle.  If we last found an available connection at position 10, the next scan begins at position 11 (and not back at 1).

roji:
@bhauer that was exactly the kind of improvements I was thinking for vNext. Regarding thread affinity and connections, how did you end up implementing this efficiently?

Note my comment above about the desirability of stack semantics, so that we can occasionally scan and prune idle connections which have been dormant for a while - this makes things more tricky...

bhauer:
@roji The thread affinity is just a ThreadLocal that retains an index or reference to a connection.  When the worker thread asks the pool for a connection, the pool checks if the connection referenced by the ThreadLocal is available.  If it has been locked by another thread (unlikely if the number of request worker threads is equal to the number connections, but still checked anyway), the pool will do a scan to find an available connection.

We don't do dynamic resizing of the pool data structure.  But we do close idle connections and then re-open them if the slot gets acquired/locked and then used.  In other words, the data structure doesn't get resized but the number of open and active connections will vary over time between some minimum and the total size of the pool.

But some thoughts on supporting dynamic resizing:

Pruning: Scan the pool and attempt to acquire a lock on each connection.  Assuming the connection is available, check its idle time and close/purge as needed.  Leave a placeholder "permanently-locked" pseudo-connection in the pool at that position.  Periodically copy the array to a resized clone and update the array reference, and in so doing, remove the permanently-locked placeholders.

Growing: (Re)adding connections can be done to these placeholder slots if they exist (that is, before a periodic array resize has happened) until they're all re-occupied, at which time the array will need to be resized and copied again to make room.

The cursor index will need to be maintained on resizes (maybe a mod or reset to 0).  Could be a small race condition to deal with when resizing and maintaining the cursor.

roji:
> @roji The thread affinity is just a ThreadLocal that retains an index or reference to a connection. When the worker thread asks the pool for a connection, the pool checks if the connection referenced by the ThreadLocal is available. If it has been locked by another thread (unlikely if the number of request worker threads is equal to the number connections, but still checked anyway), the pool will do a scan to find an available connection.

Interesting, I wasn't sure whether using thread-local storage wouldn't actually slow things down more than any advantage it would bring in avoiding the array scan. It seems that this isn't true - although what kind of pool size do you typically have? The default in my case is 100 (which is also the PostgreSQL/server-side default).

> We don't do dynamic resizing of the pool data structure. But we do close idle connections and then re-open them if the slot gets acquired/locked and then used. In other words, the data structure doesn't get resized but the number of open and active connections will vary over time between some minimum and the total size of the pool.

We don't do resizing either - we're fortunate that way. When a connection to a database is made, the user specifies the maximum pool size (100 by default). This value is part of the key to the pool, so different pools exist for different values of MaxPoolSize (or any other parameter). So for each given pool, the array size is totally constant.

> Pruning: Scan the pool and attempt to acquire a lock on each connection. Assuming the connection is available, check its idle time and close/purge as needed. Leave a placeholder "permanently-locked" pseudo-connection in the pool at that position. Periodically copy the array to a resized clone and update the array reference, and in so doing, remove the permanently-locked placeholders.

That's somewhat similar - we simply have null in "locked" slots. However, if you make no attempt to enforce LIFO semantics, doesn't that mean that connections never stay idle in the pool for long? Here's what I mean. Let's assume there's a load spike and 100 physical connections are made and used. Load now drops, and so now we have around 100 idle connections in the pool, waiting to be used. If the pool implements FIFO semantics, then even with very low load all the connections will be cycled through, and so will not have a very long idle time, and won't be pruned. If it implements LIFO semantics, then only the first X connections are ever out, leaving the rest of the connections idle and therefore prunable.

benaadams:
Updated PR.

Makes sense to pull the main pool down to 256 (looking at past results) to reduce the scan time and the Update pool down to 64 for now as per @n-stefan's change; pending a better number?

n-stefan:
Beat me to it, heheh :)

bhauer:
> Interesting, I wasn't sure whether using thread-local storage wouldn't actually slow things down more than any advantage it would bring in avoiding the array scan. It seems that this isn't true - although what kind of pool size do you typically have? The default in my case is 100 (which is also the PostgreSQL/server-side default).

I recommend microbenchmarking to validate ThreadLocal performance.  We're on the JVM and the ThreadLocal overhead there is small; my *guess* is that it's small on .NET as well.  We often configure the pool to 50, 100, 200, or more connections.  I don't think the size of the pool affects the performance of the ThreadLocal lookup, though; that should be a fixed cost regardless of pool size.

> That's somewhat similar - we simply have null in "locked" slots. However, if you make no attempt to enforce LIFO semantics, doesn't that mean that connections never stay idle in the pool for long? Here's what I mean. Let's assume there's a load spike and 100 physical connections are made and used. Load now drops, and so now we have around 100 idle connections in the pool, waiting to be used. If the pool implements FIFO semantics, then even with very low load all the connections will be cycled through, and so will not have a very long idle time, and won't be pruned. If it implements LIFO semantics, then only the first X connections are ever out, leaving the rest of the connections idle and therefore prunable.

With the thread affinity for connections, we are effectively pushing that concern up to the thread pool that manages request worker threads.  After a traffic burst, as traffic returns to a lower baseline, if the request worker pool doesn't use a worker thread, the connection to which that thread has an affinity is very likely to go idle and get closed.

When the pool is scanned because a thread's affinity-mapped connection is locked, the shared cursor causes the "hot spot" in the array data structure to shift.  In a fashion, this also contributes to allowing truly idle connections to be left cold and ready to purge.

sebastienros:
@benaadams changes look good, just extra stuff to maintain.

I wish there was a way to define some configuration for each type of run (physical/cloud, per scenario) so we could uses these values in the scripts. Like thread counts, connections strings, ... Ideally a bash script for each environment, so we can define custom ENV settings to be used in docker images or in the applications themselves.

@nbrady-techempower maybe it's already here and I am just blind ;)

nbrady-techempower:
@sebastienros It's not, but we could set some environment variable on the hosts that you can check in the docker image. Let me draft a quick issue.

roji:
@bhauer 

>> Interesting, I wasn't sure whether using thread-local storage wouldn't actually slow things down more than any advantage it would bring in avoiding the array scan. It seems that this isn't true - although what kind of pool size do you typically have? The default in my case is 100 (which is also the PostgreSQL/server-side default).
>
> I recommend microbenchmarking to validate ThreadLocal performance. We're on the JVM and the ThreadLocal overhead there is small; my guess is that it's small on .NET as well. We often configure the pool to 50, 100, 200, or more connections. I don't think the size of the pool affects the performance of the ThreadLocal lookup, though; that should be a fixed cost regardless of pool size.

OK, thanks for the info! The size of the pool shouldn't affect the performance of the ThreadLocal lookup, but the bigger it is, the more time it takes to traverse it, and so the more the (constant) price of the ThreadLocal lookup is worth it... Put otherwise, if the pool really is very small then there isn't anything to be gained by the ThreadLocal lookup (indeed it could be slower).

I'm not sure I understood how the shared cursor leaves idle connections cold - do you have some code I could look at?

benaadams:
@nbrady-techempower happy for this to be merged before next run ;)

benaadams:
@n-stefan very interesting discovery :)

image

n-stefan:
Indeed :)
I'm wondering several things now:
maybe 64 isn't even the optimal pool size
maybe mysql can also benefit from this
some of the frameworks above asp.net core are using postgresql pipelining, aren't they?

n-stefan:
Yes, pool size of 64 instead of 1024 :)

divega:
Everyone, this is a great improvement! And a very interesting path to explore. Thank you! 

Just in case it went unnoticed, in the [same run](https://www.techempower.com/benchmarks/#section=test&runid=ee5b26dc-6606-4925-8b30-584584cb5931) in which I see the improvement to the Data Update results, I also see a ~10% regression in Fortunes.

Is it possible that the regression was caused by something in this PR?

n-stefan:
It was caused by a pool size of 64, however this PR fixes it (pool size of 64 only for Data Updates, 256 for everything else).
In the current unfinished run Platform Fortunes is back up at 299,797 for aspcore-ado-pg.

divega:
@n-stefan got it. Thanks!

bhauer:
Awesome work guys! It's your accomplishment, but we really appreciate that it's happening in this context. It makes us feel great about the utility of the benchmarking project. Thanks as always for being part of this!

PR #4334, opened 2019-01-09
Rework of http4k entry, bump to J11 etc..

daviddenton:
- Bump http4k, Kotlin & dependencies
- Bump Gradle/Java Dockerfile version to use Gradle 5 + Java 11
- add new Ktor-CIO backend
- add configuration for using reactive Postgres Driver for Apache Server
- general refactor of DB code to support dual PG/Reactive PG drivers and to tune possible inefficient connection logic in updates code
- removed configuration for SunHttp (this is a development server configuration only)


daviddenton:
After a bit of a battle with the new driver implementation, this is now ready to merge. If we could get it into the continuous benchmarking soon then that would be awesome.

Keep up the great work! :)

daviddenton:
thanks! :)

Issue #4336, opened 2019-01-10
ENV variable to identify host environment


nbrady-techempower:
Let's talk about setting some kind of environment variable on the host machines that tests can use to identify where the benchmarks are being run like:

`BENCHMARK_ENV=citrine`

This would allow tests to tune to specific environments without necessarily creating multiple test variants. 
 
Per discussion in #4333

sebastienros:
Would it allow conditional statements in Docker images though? Like call `COPY` or `ENV` conditionally based on these settings?

If not then maybe calling a specific bash script before invoking the Docker image would be useful.

nbrady-techempower:
No, but you could check the environment in a `RUN` command and run whichever bash script you wanted

msmith-techempower:
I am not entirely against the notion of using environment variables, but we already have to tweak the run script (our continuous scripts for starting/stopping a run) on each environment manually; so, I think probably adding a simple optional flag to `run-tests.py` like `--env citrine` does the job simply.

nbrady-techempower:
Then to @sebastienros point you could have a dockerfile that runs some setup script like:

`RUN ./${BENCHMARK_ENV}.sh || default.sh`  which would run `./citrine.sh` if you had it or `default.sh` in some other environment if you didnt have a `vagrant.sh` or `travis.sh` or whatever

volyrique:
FYI `h2o` currently detects the environment it is running on, though in a fragile manner - details [here](https://github.com/TechEmpower/FrameworkBenchmarks/blob/3f5ae4d173f330cd9a047b9439aa9b3cea84e16b/frameworks/C/h2o/start-servers.sh#L13).

mkurz:
Besides an environment variable that let's you identify on which platform the tests run (cloud vs. citrine) I suggest two more environment variables:

- One for "raw" CPU count - which is the CPU count of the host machine _but without hyper threading threads_.
- One for threads count (that is CPU count _including_ hyper threading threads)

If these two would be available, there wouldn't be a need to switch or customize scripts per environment when the config file syntax of a framework or the app itself can read environment variables (Play2 for example).
E.g. in a Play config file you could just define something like:
```
db.poolsize = ${CPU_COUNT}
// or maybe even (not sure)
db.poolsize = ${CPU_COUNT * 2}
// or
app.workerthreads = ${THREAD_COUNT}
```
The `${...}` variables would be resolved via env variables obviously.

What do you think?

[...]

PR #4337, opened 2019-01-10
Add a --benchmark-env option to the toolset

nbrady-techempower:
This allows us to pass an environment variable to help framework tests identify which environment the benchmarks are being run in.

Usage:
`tfb --mode benchmark --benchmark-env citrine`

In test dockerfiles, to access:

`ARG BENCHMARK_ENV`
`RUN echo $BENCMARK_ENV`

result => `citrine`

Per #4336

### DIFF
--- a/toolset/run-tests.py
+++ b/toolset/run-tests.py
@@ -182,6 +182,10 @@ def main(argv=None):
         nargs='+',
         default=[1, 10, 20, 50, 100],
         help='List of cached query levels to benchmark')
+    parser.add_argument(
+        '--benchmark-env',
+        default='none',
+        help='Hostname/IP for database server')
 
     # Network options
     parser.add_argument(
--- a/.travis.yml
+++ b/.travis.yml
@@ -127,4 +127,4 @@ script:
   # run-ci.py runs the diffing to see if travis needs to test this framework. Ideally/eventually,
   # we'd like to try and do the diffing before travis_clean & setup.
   # This will run the tests exactly as you would in your own vm:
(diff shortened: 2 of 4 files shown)


toolset/run-tests.py
```diff
@@ -182,6 +182,10 @@ def main(argv=None):
         nargs='+',
         default=[1, 10, 20, 50, 100],
         help='List of cached query levels to benchmark')
+    parser.add_argument(
+        '--benchmark-env',
+        default='none',
```
msmith-techempower: Might need a better default here.

nbrady-techempower: Couldn't think of one and not sure it matters? But any suggestions?

sebastienros:
Could this also include the scenario it's currently running if it's called for each of them?
We might need to tweak values for each of them too. Otherwise it require a distinct docker file.

nbrady-techempower:
@sebastienros The scenario it's currently running would require rebuilding the dockerfile as the new arg would bust the cache. I don't think that's feasible. I'm not sure I have a solution for that.

nbrady-techempower:
Yeah, strike that. My solution was to pass the current test type in at run time, but that wouldn't work. We currently start the server up (run the dockerfile) and then test all scenarios. To do that, we'd have to stop the container, then restart for each scenario. The best bet right now @sebastienros if you think you need different configurations for different test types is to separate them out into different dockerfiles. At least if the dockerfiles are the same until some final config, the cache won't make it expensive.

nbrady-techempower:
Gonna get this in with "none" as the default for now. Think that's better than an empty string.

PR #4339, opened 2019-01-11
Use multi_accept off in nginx

joanhey:
So scale better nginx #4312 for the rest of php frameworks.


joanhey:
We could merge, later I will check these errors.

Laravel update also fail in the last runs.

And now fat-free fail less, that the last runs. 
[fat-free] failed: db query update
[fat-free-raw] failed: fortune query update
Now only fail the fat-free-raw update.

@nbrady-techempower

PR #4340, opened 2019-01-11
Optimize Nginx conf

joanhey:
Sendfile and tcp_nopush are for static files.
And will be placed in the location part for static files.
If you are not using a CDN.


joanhey:
If the results are good, will be changed to all the nginx config.

PR #4350, opened 2019-01-15
Update Finch to 0.27

vkostyukov:


nbrady-techempower:
@vkostyukov looks like this update isn't passing travis

vkostyukov:
Should be good now! @nbrady-techempower

PR #4351, opened 2019-01-15
set new pageSize

smthing:


nbrady-techempower:
@smthing this is getting stuck at `mvn compile assembly:single -q`

PR #4352, opened 2019-01-16
Optimize keepalive

joanhey:
Most keepalive libs use only the timeout and unlimited requests.


joanhey:
Results:

Test | Before | After | Difference + | % + |
----- |  ------: | -------: | --: | --: | 
Plaintext | 2,410,783 | 3,789,272 | 1,378,489 | 57.89

But scale worst and with failed connections with the max 16,384 concurrent connections.

Waiting for check the errors log, to fine tune the number or observe a memory leak.

PR #4366, opened 2019-01-21
update facil.io version

boazsegev:
facil.io version 0.7.0.beta7 is probably the last beta version before an 0.7.0 version release.

It includes changes to the hash map data structure, added security features, and numerous fixes, such as unaligned memory access fix for SipHash.

These changes could effect performance (some for the better and some might negatively impact performance in favor of security or memory consumption).

Thanks!
   Bo.


boazsegev:
Thanks, @nbrady-techempower !

PR #4367, opened 2019-01-21
Update H2o conf

joanhey:


joanhey:
It's the same problem with the update and php 7.3.
But now with this change appear too with h2o, so it is configured more like nginx.
Before h2o simply didn't send enough food (requests) to saturate the php-fpm.

Ready to merge.

PR #4368, opened 2019-01-21
Less children

joanhey:
Try to solve the update bench with php 7.3 and check performance with less children.

If the results are OK, we will change the rest of php frameworks.


joanhey:
OK to be merged!

PR #4341, opened 2019-01-11
Faster Fastcgi in nginx

joanhey:
# Normal config for applications with php files
It is very easy, using try_files, but with poor performance. The one that you see everywhere.
And also used in almost all php frameworks.

```
location / {
            try_files $uri $uri/ /index.php?_url=$uri&$args;
            # 1 stat() $uri
            # 2 stat() $uri/
            # 3 redirect  index.php....
            # 4 check locations
            # 5 check regexp locations
        }

location ~ \.php$ {
             
            fastcgi_pass   fastcgi_backend;
            fastcgi_keep_conn on;
            fastcgi_index  index.php;
            fastcgi_param  SCRIPT_FILENAME  $document_root$fastcgi_script_name;
            include        /etc/nginx/fastcgi_params;
}

```
--------------------------------------

# Most php frameworks use the front controller pattern
## Faster and more segure
No regexp only execute index.php (front controller) no other files.
No fastcgi_index after /.

```
location / {
            try_files $uri $uri/ /index.php?_url=$uri&$args;
            # 1 stat() $uri
            # 2 stat() $uri/
            # 3 redirect  index.php....
            # 4 check locations
        }

location = /index.php {
             
            fastcgi_pass   fastcgi_backend;
            fastcgi_keep_conn on;
            #fastcgi_index  index.php;
            fastcgi_param  SCRIPT_FILENAME  $document_root$fastcgi_script_name;
            include        /etc/nginx/fastcgi_params;
}

```
## Better alternative
Only check for $uri (file) and using named location @php.
In this way the front controller file can be anywhere, outside of the root.
Advert: if you leaf the index.php in the root, will be dowloaded not executed.
```
location / {
            try_files $uri @php;
            # 1 stat() $uri
            # 2 redirect  @php
            
        }

location @php {
             
            fastcgi_pass   fastcgi_backend;
            fastcgi_keep_conn on;
            #fastcgi_index  index.php;
            fastcgi_param  SCRIPT_FILENAME  /path/to/index.php;
            fastcgi_param  PATH_INFO        $uri;
            fastcgi_param  QUERY_STRING     _url=$uri&$query_string;
            include        /etc/nginx/fastcgi_params;
}
```
-------------------------

## Fastest
The one that I use normally, and I have not seen elsewhere.
Normally a lot of apps don't use static files, so it is optional.
Also most websites use cdn for static files.

No stat() that are very expensive and use filesystem, no redirect, no regexp
No index.php

```
location / {
[...]

PR #4369, opened 2019-01-21
Faster FCGI conf

joanhey:
Phpixie #4341 

Phalcon and Slim reverted for now


joanhey:
I updated vagrand and virtualbox and they aren't working for test in local.
Till I solve my local tests I will not send more commits to this PR.

But we could merge for Phpixie

PR #4384, opened 2019-01-25
Lower keepalive requests

joanhey:
And faster FCGI config.


joanhey:
Ready to be merged.

Still investigating the update problem.

PR #4386, opened 2019-01-27
Faster FCGI config

joanhey:
Fat-free, slim, cygnite and codeigniter fail with the fastest one.
So using another not so fast #4341, still using one stat() in each request.

Anyone that know more about these frameworks, can change it to the fastest.


joanhey:
Ready.

About the update problem happen in travis, citrine and vagrant.
Only with php7.3, with php7.2, php5 or hhvm it is all OK.
So I will create a bug report in php.net.

PR #4399, opened 2019-01-31
 Revert to 1024 children

joanhey:
Better results with 1024.

Tried for the update problem, but it is a bug in php 7.3.
It's solved, in 2-3 weeks arrive in PHP 7.3.2.


joanhey:
Ready to be merged.

PR #4404, opened 2019-02-02
Remove wheezyweb as last update in 2015

hal8k:
Last commit was in 2015 https://bitbucket.org/akorn/wheezy.web/commits/all


akornatskyy:
It has been moved here: https://github.com/akornatskyy/wheezy.web

PR #4417, opened 2019-02-07
Update php-fpm.conf

joanhey:
rlimit_files, rlimit_core, epoll,...

Still searching where is the bottleneck.
In the fortune test php-raw7 have only an avg latency of 0.9ms and max of 25.2ms. Less than other frameworks with a lot more req/s.
And at 128 concurrent it's the peek, and later degrade.


joanhey:
Ready

PR #4418, opened 2019-02-07
aspnetcore: add platform tests using libuv

n-stefan:
Added ASP.NET Core platform tests using the Libuv transport for Kestrel.


sebastienros:
Need to talk with the team about this one. Also the perf will vary with the custom thread count, the default value won't make it justice on citrine.

n-stefan:
Got it, thanks!

sebastienros:
We discussed within the ASP.NET Core team and we feel that it’s best to leave out the libuv measurements from the TechEmpower numbers because it’s no longer a default scenario.

n-stefan:
I see.
Thank you!

PR #4423, opened 2019-02-09
Php pools

joanhey:
Now will use 4 php-fpm pools

**Don't merge.**

Only for check that the update test problem it's solved. And other connections problems.

Now it's working perfect. But will be necesary to clean the docker cache in citrine.
PHP don't have for select minor versions, like this one 7.3.2.
Only php 7.3, and the docker cache will never update to that version.

And this error create a lot of 52Mb logs.


joanhey:
Please clean the docker cache before the next run.
After that we can close this PR.

nbrady-techempower:
@joanhey we don't do it manually. it will reset apprx every 2 runs.

nbrady-techempower:
I missed your *do not merge* comment and got trigger happy. Just re-open when you want it merged please.

PR #4430, opened 2019-02-11
Suggestion: Improve ENDTIME accuracy

benaadams:
Use `time` rather than `date` which gives the number of seconds with decimal places rather than whole numbers (with indeterminate rounding) 

The parser might need changes to deal with the decimals?

### DIFF
--- a/toolset/utils/results.py
+++ b/toolset/utils/results.py
@@ -155,11 +155,11 @@ class Results:
                             if m != None:
                                 rawData['5xx'] = int(m.group(1))
                         if "STARTTIME" in line:
-                            m = re.search("[0-9]+", line)
-                            rawData["startTime"] = int(m.group(0))
+                            m = re.search("[0-9]+\.*[0-9]*", line)
+                            rawData["startTime"] = float(m.group(0))
                         if "ENDTIME" in line:
-                            m = re.search("[0-9]+", line)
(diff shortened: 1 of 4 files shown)


benaadams:
So I was wondering about this... and whether trusting `wrk` would be better? i.e. using the value passed to `$duration`?

Though still not using `date` for the second measurement e.g. `$duration + $STARTTIME` (will need some kind of parsing of `$duration`)

msmith-techempower:
`wrk` does not have the granularity that `time` does, and often when we specify "15 seconds" for `wrk`, it reports that it was over `15.10s`, [for example](http://tfb-logs.techempower.com/round-17/final/citrine-results/20180903024112/gemini/plaintext/raw.txt).

I think this implementation is fine, overall; however, I think we might want to consider moving away from timing things ourselves and instead simply rely on `wrk`'s reporting. We fairly often get questions along the lines of "in round X, framework Y shows Z R/S on the website, but the raw file is different - why?" and that could be completely avoided by simply using `wrk`'s reporting.

benaadams:
>  and that could be completely avoided by simply using wrk's reporting.

Ah, from the summary output. More complex to parse the elapsed time from, but good idea

talawahtech:
One approach might be to just grab the req/s value directly from the summary and avoid doing a separate calculation altogether.

PR #4436, opened 2019-02-12
updated Go/iris

chunariov:
Fully updated Go/iris, which passes all tests.


nbrady-techempower:
@chunariov Thanks for your contribution.

We don't want to add 800 vendor files in this repo. Those should all be pulled from github as one of the docker commands using `go get` like https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Go/martini/martini.dockerfile#L10

chunariov:
Vendor files are deleted

frameworks/Go/iris/iris.dockerfile
```diff
@@ -1,14 +1,11 @@
-FROM golang:1.10.1
+FROM golang:latest
```
nbrady-techempower: Please keep this at a specific version number

chunariov: Done

PR #4438, opened 2019-02-13
Add benchmark tests for the Drogon framework

an-tao:
This PR adds benchmark tests for the [Drogon](https://github.com/an-tao/drogon) framework.

### DIFF
--- a/frameworks/C++/drogon/drogon.dockerfile
+++ b/frameworks/C++/drogon/drogon.dockerfile
@@ -1,0 +1,47 @@
+FROM ubuntu:18.04
+
+COPY ./ ./
+
+RUN  apt update -yqq && \
+     apt-get install -yqq software-properties-common && \
+	 apt install -yqq sudo curl wget cmake locales git \
+     openssl libssl-dev \
+     libjsoncpp-dev \
+     uuid-dev \
+     zlib1g-dev \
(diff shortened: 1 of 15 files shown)


frameworks/C++/drogon/drogon.dockerfile
```diff
@@ -0,0 +1,47 @@
...
+ENV AR=gcc-ar-8
+ENV RANLIB=gcc-ranlib-8
+
+ENV IROOT=/install
+ENV DROGON_ROOT=$IROOT/drogon
+ENV TEST_PATH=/drogon_benchmark/build
+WORKDIR $IROOT
+
+RUN git clone https://github.com/an-tao/drogon
+
```
nbrady-techempower: Can you checkout a specific version of drogon, or a commit hash? We want to avoid always pulling latest from master to keep the tests as reproducible as possible.

an-tao: Ok, I will make a new PR again. Thanks!

nbrady-techempower: No need to make a new PR! Just need a line here `RUN git checkout <hash>`

an-tao: I've added the line to the docker file.

PR #4448, opened 2019-02-14
Changed go ver to 1.11.5 in fasthttp-postgresql; added iris-prefork, iris-quicktemplate; fixed go-std; quicktemplate to go-std

chunariov:


chunariov:
Why TravisCI didn't check last commit?

PR #4457, opened 2019-02-17
Fix FastAPI implementation

tiangolo:
This fixes the implementation of FastAPI, updating the version (which fixes an incompatibility in a dependency). And updates the CPUs used.


tiangolo:
Thanks!

PR #4460, opened 2019-02-19
Adds Onyx Framework for Crystal

CaDs:
Hello.

Added a new framework for crystal, [Onyx](https://onyxframework.org)


frameworks/Crystal/onyx/onyx.dockerfile
```diff
@@ -0,0 +1,16 @@
...
+COPY run.sh run.sh
+COPY src src
+COPY shard.lock shard.lock
+COPY shard.yml shard.yml
+
+ENV DATABASE_URL postgres://benchmarkdbuser:benchmarkdbpass@tfb-database:5432/hello_world?initial_pool_size=56&max_pool_size=56&max_idle_pool_size=56
+ENV TEST_HOST tfb-server
+
+RUN shards install
+RUN crystal build --no-debug src/server_postgres.cr
```
vladfaust: Um, I think that the server should be build with `--release` tag

vladfaust:
```suggestion
RUN crystal build --no-debug src/server_postgres.cr
RUN crystal build --release --no-debug src/server_postgres.cr
```

nbrady-techempower:
Hi @CaDs - Looks like this is failing because you're trying to copy a `shards.lock` file that's not in this PR.

CaDs:
@nbrady-techempower thanks a lot. 
I fixed the dockerfile.

CaDs:
@vladfaust I saw the changes you shared with me on your gist and I used some of your code.
Still, when running the benchmarker test`./tfb --mode verify --test onyx` db and update actions are failing.

The reason is that the benchmarker is sending sometimes "bad" parameters such as 
```
for http://tfb-server:8080/updates?queries=foo
```
And with your implementation, queries with wrong kind of parameteres seems to be just skipped, while the benchmarker is expecting to fallback to a default of 1
I looked into [how to handle this errors](https://github.com/onyxframework/rest#errors) but I couldn't make it work here.

Other thing I have notice is that if we set 
```crystal
Onyx.render(:json)
```
the content-type for plaintext action is being set to `application/json` even if we attemtpt to overwrite it here

```cr
get "/plaintext" do |env|
  env.response.content_type = "text/plain"
  env.response << "Hello, World!"
end
```
↓
```
onyx: DEBUG [02:04:42.354 #8] 1.407ms
{'X-Response-Time': '147', 'X-Request-ID': 'e00d0984-f40c-4ede-89f7-d42bf068e36f', 'Content-Length': '13', 'Server': 'Onyx', 'Connection': 'keep-alive', 'Date': 'Thu, 21 Feb 2019 02:04:42 GMT', 'Access-Control-Allow-Origin': '*', 'Content-Type': 'application/json; charset=utf-8'}
Hello, World!
```

Also, probably it would be better to leave the sql logger disabled in order to gain some speed.

vladfaust:
@CaDs,

> ```
> for http://tfb-server:8080/updates?queries=foo
> ```

You can specify params like this:

```crystal
struct MyAction
  include Onyx::REST::Action
  
  params do
    query do
      type queries : String | Int32 = 1
    end
  end

  def call
    pp typeof(params.query.queries) # String | Int32
  end
end
```

You get the idea.

---

> the content-type for plaintext action is being set to `application/json` even if we attemtpt to overwrite it

True. This is a bug, I think. Strange, the valid code was once on my machine :thinking: I'll fix it ASAP.

vladfaust:
> Also, probably it would be better to leave the sql logger disabled in order to gain some speed.

As seen at https://github.com/onyxframework/onyx/blob/master/src/onyx/sql.cr#L10, once the app built in production-mode (with `CRYSTAL_ENV` equals to `production`), the logging stops.

CaDs:
@vladfaust 
Done, I have updated the PR with your code and added some small fixes.
Other than that, is using all the components you wanted, so it should be a pretty accurate benchmark for the framework.

I did pass the CRYSTAL_ENV in production mode in a previous commit, but still, I'm seeing INFO logs

```
onyx:  INFO [01:32:45.058 #10] [2e490d70]    GET /updates 200 2.041ms
```

Other benchmarks such as Raze or Kemal ones are removing the logs completely in order to gain some speed.
But this can be done later take a look and let me know if you are ok with the current version.

PR #4465, opened 2019-02-20
 update to 4.0.10

helyho:

### DIFF
--- a/frameworks/Java/voovan/src/main/java/org/voovan/VoovanTFB.java
+++ b/frameworks/Java/voovan/src/main/java/org/voovan/VoovanTFB.java
@@ -11,40 +11,46 @@ import org.voovan.tools.TObject;
 import org.voovan.tools.json.JSON;
 import org.voovan.tools.log.Logger;
 
+import java.util.Map;
+
 
 public class VoovanTFB {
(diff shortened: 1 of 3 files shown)


frameworks/Java/voovan/src/main/java/org/voovan/VoovanTFB.java
```diff
@@ -11,40 +11,46 @@
...
-            public void process(HttpRequest req, HttpResponse resp) throws Exception {
-                resp.header().put("Content-Type", "application/json");
-                resp.write(JSON.toJSON(TObject.asMap("message", "Hello, World!")));
-            }
-        });
-
-        webServer.serve();
-    }
+	private static final byte[] HELLO_WORLD = "Hello, World!".getBytes();
+	private static final Map MAP = TObject.asMap("message", "Hello, World!");
```
zloster: @nbrady-techempower @helyho 
This change is not OK. The JSON object must be instantiated for each request. This is per test rules. More details for example here: #2888

nbrady-techempower: Whoops! Thanks @zloster!

helyho:
OK  got it

PR #4470, opened 2019-02-21
Jooby 2.0.0.M1

jknack:
This is Jooby 2.0.0.M1 https://jooby.io. For now I would like to keep [Jooby 1.x]() tests due there is no stable release yet for 2.0.0

Also, Jooby supports multi-server deploy want to see Jooby performance on netty, undertow and jetty and not sure if what I did is enough for running tests on all them. I followed dropwizard setup and create a dockerfile for each of the three servers.

Thanks.


jknack:
Jooby 2 tests are good, but there are errors in other frameworks

PR #4471, opened 2019-02-22
Many updates in go-std, iris

chunariov:
1. drivers

- pq(go-std, iris)
- pgx(go-std, iris)
- mgo(go-std)
- mysql(go-std)

2. quicktemplate(go-std, iris)

3. easyjson(go-std)


nbrady-techempower:
@chunariov There's a number of conflicts here that are preventing travis to run

PR #4474, opened 2019-02-23
Fixed iris and go-std

chunariov:


nbrady-techempower:
@chunariov There's an overwhelming number of test permutations here. 20 for `go-std` is just too many. Can we cut this back some?

PR #4490, opened 2019-02-28
Add Quart

K900:
This adds support for Quart, a Python ASGI framework with a Flask-compatible API. Everything but cached queries is supported, because Quart doesn't provide a caching layer.


K900:
There seems to be a problem with the database occasionally deadlocking under heavy load - I'll investigate it during the weekend probably.

nbrady-techempower:
Thanks @K900

Issue #4493, opened 2019-03-01
Limit the amount of memory the framework container can use


michaelhixson:
Limit the amount of memory that any framework's docker container can consume.  Ideally, the framework will be able to use *almost* all available memory, but the toolset will ensure that just enough memory remains on the system for it to be functional.  A framework with a memory leak should be unable to crash the toolset.

We don't want to hard-code system-specific memory limits anywhere.  This feature should set a dynamic limit for the container based on the resources available on the host.  For example, the limit might be `available_memory * 0.95` or `available_memory - 500mb`.

I don't have an opinion on what exactly should happen when the framework reaches the limit as long as the toolset doesn't crash or exit (when in benchmark mode).

Currently, php/codeigniter is an example of a framework that appears to consume all available memory and then possibly crashes the toolset.  Citrine manages to plow through this (though SSH session become unresponsive), but Azure cannot recover.  Therefore, you could test this feature against the current version of our php/codeigniter implementation.  The current commit is  https://github.com/TechEmpower/FrameworkBenchmarks/commit/813d3ab452a6528dc058c885f1e1a214b184a0d1

This feature could impact performance of some frameworks so if we implement it, we should call this out in the blog post for the next round.

michaelhixson:
I guess this is fixed: https://github.com/TechEmpower/FrameworkBenchmarks/pull/4494

PR #4494, opened 2019-03-01
Impose Memory Limit on Test Containers

nbrady-techempower:
Still testing this locally but figured I'll let travis have a crack at it too

Solves #4493

### DIFF
--- a/toolset/utils/docker_helper.py
+++ b/toolset/utils/docker_helper.py
@@ -14,7 +14,7 @@ from toolset.utils.database_helper import test_database
 from psutil import virtual_memory
 
 # total memory limit allocated for the test container
-mem_limit = virtual_memory().total * .95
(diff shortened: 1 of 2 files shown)


toolset/utils/docker_helper.py
```diff
@@ -11,6 +11,10 @@
 from toolset.utils.output_helper import log
 from toolset.utils.database_helper import test_database
 
+from psutil import virtual_memory
+
+# total memory limit allocated for the test container
+mem_limit = virtual_memory().total * .95
```
msmith-techempower: Running inside a docker container doesn't have trouble with this?

nbrady-techempower: No, currently the docker containers have all the virtual memory of the host

Issue #4291, opened 2018-12-24
Responses including cached Date header


an-tao:
I did some testing in the benchmark project and I found that in the tests 'Json' and 'Plaintext', the response 'Date' header of some frameworks is always the same as the first response. It hints that the response entire text string including headers and the body is cached in memory for every request. According to the benchmark document, this seems to be illegal. But the framework that uses caching occupies a high position in the rankings. Am I wrong with the test rules? Does the test client check the Date header to ensure that each response is generated separately?

an-tao:
I think the ulib framework is caching the response, maybe I was wrong, but when I followed it's dockerfile to run the test, I got responses by the curl command as follows:
<img width="821" alt="2018-12-25 9 40 41" src="https://user-images.githubusercontent.com/21002501/50408868-2b05ea00-0829-11e9-956f-b5db5d974f17.png">
As you can see, all the Date headers are the same. Do you think that what I suspect is reasonable? If I am wrong, please correct me. Thank you!

dividedbynil:
@stefanocasazza

stefanocasazza:
I simple follow as other frameworks to not update the Date header for performance reason. So far it is not a requirement to have a Date header timed aligned

fafhrd91:
Rust top performing frameworks do update date header.

nbrady-techempower:
Per the test rules, the response must be composed on the spot. This should include the proper date, though we currently do not check that the date header is different from request to request, just that it's present. I'll look to add this as part of the verification step before the next round, marking tests that don't fix this in the future as failed.

@bhauer @msmith-techempower @michaelhixson Anything to add here?

benaadams:
Date header should be changing every second otherwise it wouldn't meet the "production-grade" requirement?

zloster:
@nbrady-techempower Some test implementation are using time-expiration caching for the date header. They refresh the date header once a second or so. Out my head I can recall:
- https://github.com/TechEmpower/FrameworkBenchmarks/blob/d157dfc58f2cae4c0f57ba6eddfca0b26fd73525/frameworks/Java/netty/src/main/java/hello/HelloServerHandler.java#L78-L87
- https://github.com/TechEmpower/FrameworkBenchmarks/blob/d157dfc58f2cae4c0f57ba6eddfca0b26fd73525/frameworks/Java/spring-webflux/src/main/java/benchmark/DateHandler.java#L11-L19

[...]

PR #4508, opened 2019-03-05
Add Oj to Ruby On Rails

jonathantribouharet:
Hi,
I made a little change to the configuration of the Ruby On Rails framework.
You are using the default serializer of Ruby which is known to not be performant.
The Oj gem is not provided out of the box with Rails but it's a must have that every Rails developers have installed.
It's not an opinatied choice, there are only 2 major known / used serializers in Ruby, the default of Ruby and Oj.

Regards,
John


nbrady-techempower:
@jonathantribouharet We've recently enforced this requirement in #4291 

It looks like rails is one of those frameworks that is caching the date header. I doubt the gem you added is responsible for this, but it won't pass until that's taken care of.

jonathantribouharet:
The problem is `config.action_dispatch.default_headers.merge!('Date' => Time.now.httpdate, 'Server' => 'WebServer')` in `config/application.rb`, this line set a header, so the header 'Date' will always be the same value.
I have two questions:
- what is the usage / goal of this header?
- do you want me to fix it in this PR or another?

nbrady-techempower:
@jonathantribouharet In this PR is fine, as we don't want to merge anything new until it passes the travis tests. The headers are just required as part of the tests:

 - The response headers must include either Content-Length or Transfer-Encoding.
 - The response headers must include Server and Date.

We generally don't do the test implementations; we leave that up to the community. So feel free to adjust as you see fit. Thanks!

nbrady-techempower:
Nice work! Thanks for the updates!

PR #4514, opened 2019-03-06
Fix #4492

### LINKED ISSUE #4492: Codeigniter framework issues
CodeIgniter is causing some issues in our Azure and Citrine environments, most recently leaving 1026 php-fpm processes open and using 30gb of memory. Though Citrine can handle it, it's causing docker to shutdown on Azure.  The same is not happening for the `codeigniter-hhvm` tests, though those tests are getting several hundred 5xx errors.

@joanhey as my go to for PHP related issues, do you have any time to take a look at this?

Edit: CakePHP also opens 1026 php-fpm processes but only uses 2-3gb of memory

joanhey:
Fixes #4492
Disable db persistent connections (memory leak).
Also update codeigniter core and nginx config.

Now I'll create a issue in codeigniter repo, as it is the only framework with this problem.


joanhey:
No way to use the BENCMARK_ENV.
Using: tfb -m benchmark --test codeigniter --benchmark-env citrine

Help needed

joanhey:
Delete codeigniter hhvm as v4 is not compatible with PHP
https://hhvm.com/blog/2019/02/11/hhvm-4.0.0.html

Will be necessary delete the rest of hhvm in php frameworks.

PR #4515, opened 2019-03-06
Delete hhvm v4 PHP is unsupported

joanhey:
https://hhvm.com/blog/2019/02/11/hhvm-4.0.0.html

and

https://hhvm.com/blog/2019/02/28/hhvm-4.0.3.html


joanhey:
Only deleted from frameworks.
Later I will try that work with vanilla PHP.

PR #4516, opened 2019-03-06
[ci skip] Update continuous benchmarking scripts

michaelhixson:
* Bumped up the expected required free disk space to complete a run from
  20G to 50G based on observation.
* Build the techempower/tfb docker image before each run.  Previously,
  we were pulling the image from docker hub, but we would not always
  publish a new image to docker hub every time we changed our Dockerfile.
* Move looping behavior from the bash script into the service.  That way,
  we can update the script without restarting the service (which would
  end the current run), and the changes will affect the next run.
* Added more logging.
* Added --benchmark-env flag to tfb command.

### DIFF
--- a/toolset/continuous/tfb-shutdown.sh
+++ b/toolset/continuous/tfb-shutdown.sh
@@ -3,17 +3,31 @@
 set -e
 
 docker_clean() {
+  echo "restarting docker"
   sudo service docker restart
+
+  echo "running 'docker stop'"
   docker ps --all --quiet | xargs --no-run-if-empty docker stop
+
+  echo "running 'docker rm'"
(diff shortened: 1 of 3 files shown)


michaelhixson:
@nbrady-techempower This made me realize we have `--results-environment` and `--benchmark-env` arguments to the `tfb` command, which seem redundant.  Also, the help text for `--benchmark-env` is wrong.

https://github.com/TechEmpower/FrameworkBenchmarks/blob/f9eef703f67ad13c1fac5aa82773589f765c96f2/toolset/run-tests.py#L97-L100

https://github.com/TechEmpower/FrameworkBenchmarks/blob/f9eef703f67ad13c1fac5aa82773589f765c96f2/toolset/run-tests.py#L185-L188

PR #4529, opened 2019-03-07
Separate Go tests in travis, easyjson in Go/fasthttp

chunariov:


nbrady-techempower:
@chunariov We want to group some of these tests if we can. Can you group tests that have similar dockerfiles like so:

`'TESTDIR="Go/aah Go/beego Go/chi"'`
`'TESTDIR="Go/echo Go/evio Go/falcore"'`

.travis.yml
```diff
@@ -13,111 +13,108 @@ services:
...
+    - "TESTLANG=Clojure"
+    - "TESTLANG=Crystal"
+    - "TESTLANG=D"
+    - "TESTLANG=Dart"
+    - "TESTLANG=Elixir"
+    - "TESTLANG=Erlang"
+    - "TESTLANG=FSharp"
+    - 'TESTDIR="GO/aah GO/chi GO/falcore GO/gin GO/goji GO/kami GO/martini GO/revel GO/webgo"'
+    - 'TESTDIR="GO/beego GO/echo"'
+    - 'TESTDIR="GO/evio GO/fasthttp GO/go-std GO/iris"'
```
nbrady-techempower: This is case sensitive in our differ, so needs to be `Go` for all the tests

PR #4537, opened 2019-03-11
Fix Angel Framework tests

### LINKED ISSUE #4432: Dart Angel framework failing
@thosakwe Looks like the framework is no longer building correctly possibly because some of the dependencies in the pubspec were updated. Could you update this if you have the time? If not we can just remove it for now and you can add it back when you're ready. Thanks!

thosakwe:
Resolves #4432. 😄


nbrady-techempower:
Woo! Thanks @thosakwe

PR #4543, opened 2019-03-11
Add ability to override dockerfile and CMD in benchmark_config

nbrady-techempower:
This PR adds 2 new optional properties to the `benchmark_config.json`

`"dockerfile"`: Allows you to override the dockerfile your test will use. Currently all tests look for the `test-name.dockerfile` by default. This allows you to use the same dockerfile across multiple tests.

`"docker_cmd"`: This allows you to override the `CMD` line of the dockerfile the test is using. This lets you reuse the same dockerfile for building if you're only changing the `CMD`.

For a test run, I found that `Go/aah` uses the same setup for all three tests while only changing the `CMD`.

This PR also adds two new environment variables for tests to consume: `TFB_TEST_NAME` and `TFB_TEST_DATABASE`. This allows you to dynamically alter your run environment based on your test name or the database it's using. For an example, nodejs tests were all using the same docker build steps while changing an environment variable to determine which handler to use. I've updated this test to use a single dockerfile while consuming the `TFB_TEST_NAME` arg to determine which handler to use.

Overall this QOL improvement should make it easier for maintainers to build out implementations without copying/pasting dockerfiles, or duplicating server code. This will also help ensure that the docker cache is being utilized properly. Before, if a maintainer had the same build steps across multiple dockerfiles, but they were out of order, it would bust the docker cache.


toolset/utils/docker_helper.py
```diff
@@ -131,8 +134,13 @@ def build(self, test, build_log_dir=os.devnull):
                 log_prefix=log_prefix,
                 path=test.directory,
                 dockerfile=test_docker_file,
-                tag="techempower/tfb.test.%s" % test_docker_file.replace(
-                    ".dockerfile", ""))
```
msmith-techempower: lol, what terrible code I wrote. "Make `test_docker_file` by combining `test.name` and ".dockerfile", then remove ".dockerfile" to find the name of the test." 👍

toolset/utils/docker_helper.py
```diff
@@ -182,9 +190,14 @@ def watch_container(docker_container, docker_file):
...
             }]
 
+            docker_cmd = ''
+            if hasattr(test, 'docker_cmd'):
+                docker_cmd = test.docker_cmd
+
             container = self.server.containers.run(
                 "techempower/tfb.test.%s" % test.name,
                 name=name,
+                command=docker_cmd,
```
msmith-techempower: I suspect it does, but this will override the `CMD` present in their dockerfile, right?

nbrady-techempower: Right. And if it’s an empty string it will keep the CMD in the dockerfile.

nbrady-techempower:
I’m not sure where ENTRYPOINT would fit in here but I suspect it would make it harder to run the test dockerfile independent of the toolset which is something I didn’t want to break with this PR.

nbrady-techempower:
Documentation has been updated

PR #4556, opened 2019-03-18
Play2: Change throughput and switch to thread-pool-executor (for db benchmarks)

mkurz:
Given that the db benchmarks always hit the database, let's try the "Highly synchronous" configuration described here:
https://www.playframework.com/documentation/2.7.x/ThreadPools#Highly-synchronous

For all non-database benchmarks I increased the throughput to 64.

Let's see how this effects performance.
If it's bad, we can just rollback later.


mkurz:
[The build](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/builds/507913543) is green :heavy_check_mark:

PR #4557, opened 2019-03-18
chore: Upgrade phoenix to 1.4 and elixir 1.8

yuhama:
Attempt at upgrading the Phoenix implementation to Phoenix 1.4.x and Elixir 1.8.x. There's no effort towards improving performance.

Not sure if it makes sense to reorganize the code to use the 1.4 layout too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,7 @@ build/
 
 # haskell
 .stack-work
+
+# Elixir
+_build/
+deps/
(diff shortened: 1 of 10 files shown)


nbrady-techempower:
Hi @yuhama - Thanks for doing this! If you have the time to make these implementations look more like a current real world example, please do! I'll merge these updates in now

PR #4559, opened 2019-03-19
[ci fw-only Java/servlet] Use autocommit=true for the SQL DB interractions

zloster:


zloster:
Hmm, all the Java project builds are OK in Travis.

PR #4562, opened 2019-03-19
ffead-cpp version 3.0

sumeetchhetri:
Bumped up ffead-cpp version to v3.0

Apache web server and nginx spawn multiple child processes in order to handle requests, in such a scenario, some children may be initialized before other child nodes. Cache priming generally should happen only once. The web servers first start listening on the ports and in the background initialize the child process and wait for them to complete init. To delay/fix this problem of child processes being initialized without waiting for the cache to be primed, we need to introduce an attribute "wait_before_sending_requests" in the test configuration, this setting will add additional delay before the framework in general is ready to accept requests.

Also the ffead-cpp embedded server first listens on the port and only then proceeds with any database/cache connection pool setup or any pre-initialization routines before handling any requests, in this case as well, the benchmarker will falsely signal that the server is ready to accept requests, whereas actually it is still priming the cache, which needs to happen before there is any actual request for a cached resource.


nbrady-techempower:
@sumeetchhetri Looks like the redis ones are failing and the cached apache test

sumeetchhetri:
> @sumeetchhetri Looks like the redis ones are failing and the cached apache test

Resolved.

PR #4580, opened 2019-03-26
New(JavaScript): Add Polkadot candidate

lukeed:
I did my best to model the entry after existing frameworks. I did not see a place within `.travis.yml` for me to insert the new record (as per PR template note).

I've added all database drivers using the "raw" client libraries:

* MySQL
* PostgreSQL
* MongoDB

Similarly, each of these have their own dockerfile. There's also a plain "default" dockerfile without any DB drivers attached. (I saw this in some of the other directories)

Thanks~

### DIFF
--- a/frameworks/JavaScript/polkadot/.npmrc
+++ b/frameworks/JavaScript/polkadot/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false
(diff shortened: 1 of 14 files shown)


nbrady-techempower:
Hey @lukeed -

This looks great. Popped over to the repo and it looks like fun. I'm going to spin up a server with it later and mess around.

Merging this in now, so you should see results on http://tfb-status.techempower.com as soon as the next run kicks off.

lukeed:
Awesome  🙌 thank you~!

PR #4585, opened 2019-03-27
Stop docker containers on run_test failure/error

nbrady-techempower:
Make sure all server and database containers are stopped when a test fails; whether it's a normal failure or an error that bubbled up from the toolset.


michaelhixson:
I still suggest using a `finally` block instead.  Then it's really easy to see that the containers always get stopped.  The way it is in this PR now, we are stopping the containers one way if success and another if failure, and you have to think about whether `__exit_test` is always called and with what `success` value.  It seems more brittle than it could be.

msmith-techempower:
A net loss in lines of code: always good for a project.

PR #4586, opened 2019-03-28
* [ci fw-only Java/baseio] update to 3.2.9.beta10

kyle2008:
bug fix


kyle2008:
@nbrady-techempower  Sorry, I add it for a test, but forget to push to other branch ..... ,will handle it few soon.

nbrady-techempower:
Ah, ok. Thanks!

PR #4571, opened 2019-03-24
Add framework Warp

naushadh:
- Resurrect WAI that was removed in https://github.com/TechEmpower/FrameworkBenchmarks/pull/4522: now with full test coverage, using 2 different database libraries, latest GHC+deps, _quickest_ possible deps
- Shared docker image for both tests so we can exploit docker caching and finish in under 30 minutes
- Travis run: https://travis-ci.org/naushadh/FrameworkBenchmarks/jobs/510469874
- Performance in comparison with `servant`

.|JSON Serialization|Single Query|Mutiple queries|Fortunes|Updates|Plaintext
---|---|---|---|---|---|---
servant (ffec2c1)|18,120|4,057|376|3,476|175|21,478
servant-mysql-haskell (3f9f25c)|20,116|5,610|362|5,161|94|20,396
warp (65526da)|30,677|3,884|702|4,112|164|54,811
warp-hasql (65526da)|30,164|3,131|410|4,327|176|52,160
warp-mysql-haskell (65526da)|29,063|5,470|346|5,993|71|52,780

PR #4595, opened 2019-04-01
 Restructure Haskell/Warp for db modularity.

naushadh:
Was about to make some changes to Warp but quickly realized how cumbersome it was to re-apply patches by hand to each db. So decided to do some upfront one off refactoring to deduplicate db logic.

The database libs are now generic enough it can be trivially re-used with any server/framework. But that will require making tooling changes to make Docker run from parent dir so we can have say `servant` and `warp` share the reusable database libs. Will create a separate issue/PR for that.

Baseline stats from last PR (https://github.com/TechEmpower/FrameworkBenchmarks/pull/4571):

.|JSON Serialization|Single Query|Mutiple queries|Fortunes|Updates|Plaintext
---|---|---|---|---|---|---
warp (65526da)|30,677|3,884|702|4,112|164|54,811
warp-hasql (65526da)|30,164|3,131|410|4,327|176|52,160
warp-mysql-haskell (65526da)|29,063|5,470|346|5,993|71|52,780

Current stats:

.|JSON Serialization|Single Query|Mutiple queries|Fortunes|Updates|Plaintext
---|---|---|---|---|---|---
warp|24,600|3,919|713|4,255|164|54,588
warp-hasql|25,787|4,865|410|4,623|173|53,855
warp-mysql-haskell|28,740|5,193|356|6,133|69|50,784

Looks mostly the same or better except for some odd regressions for `JSON Serialization` -- even that seems to get better on the later runs (`warp-mysql-haskell` vs `warp`) in the same test using the same code. I'm not too worried myself.


nbrady-techempower:
Thanks @naushadh 

As far as the context that the docker container pulls in, I'll see if I can add another benchmark_config option to pull in context from a directory above the framework test. I don't want do this automatically for all frameworks because I don't want to pull in the entire context of the language directory as it will make each test container larger. As it is, all these test containers take up a ton of disk space.

naushadh:
See comment for performance results: https://github.com/TechEmpower/FrameworkBenchmarks/pull/4571#issuecomment-480446366

PR #4602, opened 2019-04-02
[ci fw-only CSharp/aspcore] update URLs

n-stefan:
Update a couple of URLs in benchmark_config.json.


sebastienros:
@benaadams remember when we were benchmarking this and the changes would not show any improvement?

sebastienros:
I assume this issue is due to the PR where we used a different connection string for updates?

n-stefan:
Not sure, but currently the Linux Transport (rhtx) result isn't showing up in Updates, that's what the PR hopes to fix.

PR #4603, opened 2019-04-02
Update Grizzly

zloster:
Current status Grizzly:

- [x] Based the code implementation on 2.4.x branch;

- [x] Added basic error handling - HTTP 404 for requests;

- [x] Added README.md;

- [x] Updated libraries - Jackson to 2.9.8 and Grizzly to 2.4.4.

Current status Grizzly-Jersey or Jersey-Grizzly:

- [x] Based the code structure to be similar to Grizzly implementation;

- [x] Added basic exception handling;

- [x] Added implementation for DB updates test;

- [x] Added separate endpoint for the multiple queries test;

- [x] Added README.md;

- [x] Updated libraries - Jackson to 2.9.8, Grizzly to 2.4.4, Mustache.java to 0.9.6.

@nbrady-techempower Some thoughts about #4534
It seems that `Grizzly-Jersey` is more like `Jersey-Grizzly`. Most of the code comes from the `Jersey` repositories.


nbrady-techempower:
> Some thoughts about #4534 It seems that Grizzly-Jersey is more like Jersey-Grizzly. Most of the code comes from the Jersey repositories.

Ok, I don't mind that. If you want to separate that out into a `Java/jersey` directory, I think that's fine. It was the `grizzly` `grizzly-jersey` thing that bothered me and made me think (without much looking into it) that they could be grouped under `grizzly`.

I'll merge this as is though.

PR #4607, opened 2019-04-03
jooby 2.0.0.M2

jknack:
- upgrade to latest jooby

### DIFF
--- a/frameworks/Java/jooby2/src/main/java/com/techempower/App.java
+++ b/frameworks/Java/jooby2/src/main/java/com/techempower/App.java
@@ -1,50 +1,55 @@
 package com.techempower;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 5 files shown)


frameworks/Java/jooby2/src/main/java/com/techempower/App.java
```diff
@@ -1,131 +1,160 @@
...
+
+          try (PreparedStatement statement = connection.prepareStatement(updateSql.toString())) {
+            int i = 0;
+            for (World world : result) {
+              world.randomNumber = nextRandom(rnd);
+              statement.setInt(++i, world.id);
+              statement.setInt(++i, world.randomNumber);
+            }
+            statement.executeUpdate();
+          }
```
msmith-techempower: This implementation is batching all the updates into a single query, which is against the rules.

PR #4617, opened 2019-04-07
Use the CoreRT compiler from the new blob feed

MichalStrehovsky:
* Use CoreRT compiler from the new blob feed. See dotnet/corert#7268
* Opt out of compiler defaults. These options became the defaults in the past months, but the benchmark doesn't need them. This removes almost 10 MB of cruft from the produced executable (which likely helps working set and code locality).


MichalStrehovsky:
@benaadams Could you have a look please?

PR #4622, opened 2019-04-08
Fail JSON tests that return object where array is required

### LINKED ISSUE #4612: [php-ngx] Multiple queries test regression
The #4596 is separating `single query` and `multiple queries` tests in the PHP frameworks. However there is a missed configuration about the `php-ngx`:
https://github.com/TechEmpower/FrameworkBenchmarks/pull/4596/files#diff-d064e3038c12f323a8ab7749aa0d07e3R136

I guess that @joanhey will fix this easily.

IMO there is _also a problem with the verification code_. Check the build output in Travis:
https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/489982007#L1256-L1285

It only issues `warnings` but doesn't fail the verification. _It should fail because the code is always returning a single result_. https://github.com/TechEmpower/FrameworkBenchmarks/blob/ba86c0d71263d5ddd5253007a1d1ffa24736b37e/frameworks/PHP/php/dbraw.php#L10-L15

The other related configurations:

- https://github.com/TechEmpower/FrameworkBenchmarks/blob/ba86c0d71263d5ddd5253007a1d1ffa24736b37e/frameworks/PHP/php/deploy/nginx_php.conf#L54-L60
 - https://github.com/TechEmpower/FrameworkBenchmarks/blob/ba86c0d71263d5ddd5253007a1d1ffa24736b37e/frameworks/PHP/php/php-ngx.dockerfile#L29

nbrady-techempower:
We should be failing tests that return a single object when an array is required. It's almost never the right number of elements and it's less bytes than we require.

Resolve #4612

### DIFF
--- a/toolset/benchmark/test_types/verifications.py
+++ b/toolset/benchmark/test_types/verifications.py
@@ -222,10 +222,10 @@ def verify_randomnumber_list(expected_len,
         return problems
 
     # This path will be hit when the framework returns a single JSON object
-    # rather than a list containing one element. We allow this with a warn,
-    # then verify the supplied object
+    # rather than a list containing one element.
     if type(response) is not list:
-        problems.append(('warn', 'Top-level JSON is an object, not an array',
+        problems.append((max_infraction,
+                         'Top-level JSON is an object, not an array',
                          url))
         problems += verify_randomnumber_object(response, url, max_infraction)
         return problems


joanhey:
I think that is correct now.
But some frameworks fail now.

I could change the php framework in a moment.
The rest of frameworks that fail, will need to change the code too.

zloster:
There seems to be a false positive case for the new checks:
https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/517302629#L6039-L6064
`ringojs` manages to return 501 objects in the right case.

nbrady-techempower:
@zloster I think this is ok? We haven't changed warning on 501+ results returned and after debugging it locally, it looks like it is indeed returning an array of 501 properly formed objects. So that warning is fine.

zloster:
My point was that maybe the failing rule is too strict. Not all frameworks have different implementations for `single` and `multiple` queries. With the current checks they will fail the verification.

nbrady-techempower:
Ah, I see what you're saying. I'm not sure I'm worried about that. Looking at travis, there certainly are a few, but I would rather those be fixed. I think the closer we get to parity/strictness of expected return results, the better. But if others don't feel that way, I'm fine to hold off on this.

msmith-techempower:
I agree with Nate; let's get a bit more draconian.

joanhey:
Sorry but you don't need 2 different implementations.

I change the php activerecords with 1.
https://github.com/TechEmpower/FrameworkBenchmarks/pull/4606/files

joanhey:
Without this change, you can see things like that.
https://www.techempower.com/benchmarks/#section=test&runid=8ab47a8b-0b0d-43f7-ad12-1e96b72ff42d&hw=ph&test=query

Check the php-ngx req/s.

zloster:
@nbrady-techempower @msmith-techempower It's your call.
Maybe before going draconian to consider another option. Reordering the checks:
If `?queries=501` returns 500+ array of objects than issue warning. And don't force failing for the cases requiring result in the form of one element array. Because the check will catch if the framework is not returning enough data.

@joanhey Yes, it is not required to have two implementations. IMO changing the current code for given framework or implementing two separate handlers will be almost equal as effort.

joanhey:
I think that will file in both situations.
And
```
WARN for http://tfb-server:8080/db?queries=501
     JSON array length of 501 != expected length of 500
```
And this one will be a fail too, not a warning.

The tests help to create better code.
If you are sending 100 o 502 when need to be 500.
Only help the programmer that create the code.

nbrady-techempower:
After some review, I think we're going to merge this in. It also doesn't require a rule change. The rules for multiple queries state:

> Each World object must be added to a list or array.

I think for the purpose of benchmarking and adhering to the rules,  this change makes sense. 

I also agree with @joanhey that more than 500 elements should probably also be a failure. We put these rules in place to create some branching logic (though not spectacular in this case) to test the performance under heavy load. I'm not sure that returning 501 elements is a benefit vs doing a logic check and returning 500, but it's another one of those things where ensuring parity means we can focus more on benchmarking and less on who's skirting this rule & that rule. With this one though, it's more likely that we're pointing out a problem for which the tester meant to correct.

zloster:
IMO I haven't explained it well enough:
I don't mind having stricter verification of the rules. I had considerations about the execution of the change. In my mind:

1. Change verification for the multiple queries to fail if there is a single result for the request "?queries=501" - this was the original problem in the linked issue. Merge this code;
1. Announce (by pinning the issue and/or mailing the list) the planned stricter checks;
1. Wait 1 or 2 months and than implement the change. This gives an opportunity for some of the participants to fix the implementations. 

The result should be "everyone is happy and there are stricter checks" but IMO the transition is smoother.

nbrady-techempower:
@zloster I see. I think I misunderstood your original issue. Well, the ship on this has sailed. There aren't many tests that fail because of this. @joahney grabbed all the php ones. I can tackle a couple this week and that's pretty much it. At least the rules always said *must* so we're not getting any stricter with the rules, we're getting stricter with the enforcement. I'll make a pinned issue this week describing some of the changes that we've made, like #4291

zloster:
@nbrady-techempower So, everything is clear.
I'm preparing a change for the Java/restexpress.

PR #4624, opened 2019-04-08
Fix debug mode

### LINKED ISSUE #4284: tfb --mode debug not working
After the change to docker the debug mode don't work.
Perhaps the docker is not exposing the ports to vagrant.

The process stop correctly
`Entering debug mode. Server has started. CTRL-c to stop.`

But noway to connect to localhost:28080
`http: error: ConnectionError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer')) while doing GET request to URL: http://localhost:28080/fortune`

Will be good if the message add the 'localhost:28080' link.
`Entering debug mode. Server http://localhost:28080 has started. CTRL-c to stop.`

nbrady-techempower:
This problem was introduced when we switched to docker.  This exposes the test port to vagrant or whatever host is running docker but only in `--mode debug`

Sorry this took so long to get to @joanhey !

resolves #4284

### DIFF
--- a/toolset/utils/docker_helper.py
+++ b/toolset/utils/docker_helper.py
@@ -202,12 +202,18 @@ class DockerHelper:
             if hasattr(test, 'docker_cmd'):
                 docker_cmd = test.docker_cmd
 
+            # Expose ports in debugging mode
+            ports = {}
+            if self.benchmarker.config.mode == "debug":
+                ports = {test.port: test.port}
+
             container = self.server.containers.run(
                 "techempower/tfb.test.%s" % test.name,
                 name=name,
                 command=docker_cmd,
                 network=self.benchmarker.config.network,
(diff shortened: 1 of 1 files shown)


toolset/utils/docker_helper.py
```diff
@@ -202,12 +202,18 @@ def watch_container(docker_container, docker_file):
...
+            if self.benchmarker.config.mode == "debug":
+                ports = {test.port: test.port}
+
             container = self.server.containers.run(
                 "techempower/tfb.test.%s" % test.name,
                 name=name,
                 command=docker_cmd,
                 network=self.benchmarker.config.network,
                 network_mode=self.benchmarker.config.network_mode,
+                ports=ports,
```
msmith-techempower: Ah yes, without this you would need to run a docker container on the `tfb` docker network in order to curl. Good catch.

PR #4626, opened 2019-04-08
Update Fuel query to pass the test

joanhey:
Fix for the change in tests #4622


joanhey:
The travis-ci passed, but didn't update here.

PR #4638, opened 2019-04-10
Add docker_context benchmark_config property

nbrady-techempower:
This adds a new property to the `benchmark_config.json` called `docker_context`. This allows you to define a path relative to your test to be available during the build step. The primary use case for this, as asked for by @naushadh, is to share modules, config, etc across multiple frameworks.

In the included nodejs example, I've added:

```
"docker_context": "../",
```

to the default test's benchmark config. This places the root context of our dockerfile in the `frameworks/JavaScript` directory, meaning I can now add any files from any of the other JS frameworks to my dockerfile. This also means that in order to add all the files for my framework, I had to change `COPY ./ ./` to `COPY ./nodejs ./`

Concerns:

 - I'm still not :100: that I want this. For most tests, you can run the Dockerfile without the toolset to spin up a server. This makes it a little less obvious that you'd have to provide a custom context if you were to do that.
 - I did check that this doesn't increase the image size. The extra context is only available at build time.
 - I like the idea of reducing copying/pasting for people that submit and maintain multiple frameworks and use a lot of the same files. However, I don't like that this would tightly couple implementations across framework folders. I'd have to do some serious magic in the travis differ to trigger a framework test because it's using the context of a file that was edited.

Before merge:

- [ ] conversation and approvals
- [ ] remove nodejs example
- [ ] add docs to wiki

### DIFF
--- a/frameworks/JavaScript/nodejs/benchmark_config.json
+++ b/frameworks/JavaScript/nodejs/benchmark_config.json
@@ -2,6 +2,7 @@
   "framework": "nodejs",
   "tests": [{
     "default": {
+      "docker_context": "../",
       "json_url": "/json",
       "plaintext_url": "/plaintext",
       "db_url": "/db",
--- a/frameworks/JavaScript/nodejs/nodejs.dockerfile
+++ b/frameworks/JavaScript/nodejs/nodejs.dockerfile
@@ -2,7 +2,7 @@ FROM node:10.12.0
 
 ARG TFB_TEST_NAME
 
-COPY ./ ./
+COPY ./nodejs ./
 
 RUN npm install
 
--- a/toolset/utils/docker_helper.py
+++ b/toolset/utils/docker_helper.py
@@ -1,4 +1,6 @@
 import os
+from io import BytesIO
+import tarfile
 import socket
 import json
 import docker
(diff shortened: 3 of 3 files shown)


msmith-techempower:
I don't understand the use-case from your code changes or description. Does this make a `nodejs` "image" available to other JavaScript frameworks by way of the `custom_context`?

Can you show me a concrete example of where this is useful and what it gains us?

nbrady-techempower:
@msmith-techempower #4614 is an example. Say one framework like `Haskell/servant` has a shared directory. So it has a path of `Haskell/servant/shared` and you want to use that in `Haskell/warp`. By adding `"docker_context": "../"` in `Haskell/warp`, that dockerfile now has access to everything in `Haskell/*` so you can do this in `warp.dockerfile`:

```
FROM haskell:8.6.3

# copy that shared folder we want from another framework since our
# root context is Haskell/* and not Haskell/warp/*
COPY ./servant/shared ./shared

# Now add our framework files
COPY ./warp ./
```

My description probably over-complicated. It simply changes the root of what the dockerfile has access to.

msmith-techempower:
Is there some limitation of docker where the following, without `custom_context`, would not work:

```Docker
FROM haskell:8.6.3

# copy that shared folder we want from another framework since our
# root context is Haskell/* and not Haskell/warp/*
COPY ../servant/shared ./shared
#    ^ relevant difference

# Now add our framework files
COPY ./warp ./
```

nbrady-techempower:
Correct, you cannot do that. Docker wants you to explicitly declare the context your file has access to.

msmith-techempower:
I dislike how opaque the context is to the implementation of the dockerfile. My concern is that some contributors will wonder why Haskell frameworks can reference one another but their implementation cannot, not knowing that the `benchmark_config.json` holds a magic key to allow it.

Maybe add to the scaffolding to include the `custom_context` by default and make it `.`?

Also, if we merge this, [this page](https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Codebase-Framework-Files) will need documentation for it.

nbrady-techempower:
> Maybe add to the scaffolding to include the custom_context by default and make it `.`?

Not a bad idea

> will need documentation for it.

Yes, is part of my checklist above

michaelhixson:
> I'm still not 💯 that I want this. For most tests, you can run the Dockerfile without the toolset to spin up a server. This makes it a little less obvious that you'd have to provide a custom context if you were to do that.

I don't like this aspect of it either.

We're basically adding support for shared libraries now?  The layout in your example doesn't look desirable to me.  Warp shouldn't be pulling in something from Servant.  Next, someone will wonder why that shared code isn't in `Frameworks/Haskell/Shared` or something like that.  I don't think it's good for us to go down this road.

The idea of the docker context being configurable also seems scary from a security standpoint, although maybe that doesn't apply in this case.  Since the toolset itself is running inside docker, even if a framework sets its context to be `/`, it can't see any sensitive files on the host user's machine, can it?

nbrady-techempower:
@michaelhixson right, the `/` would refer to the toolset container's root. But I agree with your overall point. I think I'm going to scratch this idea. If a stronger case can be made for why it's needed, and maybe a better way to do it, I'll consider it again.

naushadh:
> I'm still not 💯 that I want this. For most tests, you can run the Dockerfile without the toolset to spin up a server. This makes it a little less obvious that you'd have to provide a custom context if you were to do that.

@nbrady-techempower The server could still be spun up without the toolset. The build tool being used (stack) has the ability to reference library modules from arbitrary/shared directories in the filesystem. My only limiting factor was having a sane way to get docker to source in a sibling scope dir.

@michaelhixson: Warp was never intended to pull code from Servant or vice-versa. The intent was indeed to have shared code stored in `Frameworks/Haskell/Shared` before being re-used between Servant, Warp and later on Snap.

naushadh:
> ... going to scratch this idea. If a stronger case can be made for why it's needed, and maybe a better way to do it, I'll consider it again.

I'm surprised none else ever has wanted to share code between frameworks. I suppose my options are 
- keep copy-pasting code between frameworks
- or host the shared libs in a separate git repo before importing it in as a dependency

nbrady-techempower:
@naushadh I think, generally, the people submitting tests here are more single-framework-focused. If it’s all the same to you, I think it’s better if you copy/paste across frameworks that way all the significant lines of code to spin up a server are visible here and if you end up abandoning them for whatever reason, someone else can edit them easily.

zloster:
@naushadh Hi and welcome.
I'm one of the few contributors here that are not framework-only focused. I have contributed some test implementations but I'm also trying to help with the maintenance.

Now on the question: IMO it is better to copy-paste the shared code. Why? Take a look at #2554.

The code is not that much - at least in the Java part you have:

- 2 (two) classes for the objects representation;
- couple of methods for the random number generation;
- couple of methods for the JSON serialisation. Major part of the frameworks are using Jackson for this.

PR #4546, opened 2019-03-13
Dropwizard JDBI update

FredDeschenes:
Updated Dropwizard to latest version (1.3.9) and switched JDBI from v2 to v3 (last v2 release was over 2 years ago).
I've also modified the "updates" test (test type 5) to batch the updates as it's now allowed as per point 9 in the requirements.

PR #4639, opened 2019-04-10
Attempt to fix deadlock in JDBI test

FredDeschenes:
As mentioned here : https://github.com/TechEmpower/FrameworkBenchmarks/pull/4546#issuecomment-481678409

I can also just go back to a non-batched update if this doesn't fix the issue. Also, as mentioned in #2684, I cannot repro this on the 2 machines I tried it on.

### DIFF
--- a/frameworks/Java/dropwizard/src/main/java/com/example/helloworld/db/jdbi/WorldRepository.java
+++ b/frameworks/Java/dropwizard/src/main/java/com/example/helloworld/db/jdbi/WorldRepository.java
@@ -6,6 +6,10 @@ import com.example.helloworld.db.WorldDAO;
 import com.example.helloworld.db.model.World;
 import com.example.helloworld.resources.Helper;
 
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+
 public class WorldRepository implements WorldDAO {
 	private final Jdbi jdbi;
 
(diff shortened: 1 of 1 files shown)


zloster:
@FredDeschenes I've just had an idea for the reproduction:

 The test is updating a table with 10 000 rows. To trigger the problem there must be high enough TPS hitting the database. Lowering the random number upper limit to (200, 500 or 1000) should provoke the problem to happen at lower TPS.
How much lower will depend on the processing power/IO capacity of the given computer/machine.

FredDeschenes:
@zloster : Sorry I'm not too familiar with actually running the tests, are you talking about changing the query being sent to the server to send a lower "queries" value or just hardcoding it to a lower value (let's say 200) directly in the code?

zloster:
I'm not sure. Here is a little longer explanation:
Here is the update code:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/c1c5d1ce032d967841bce9be6329046c6aa18c06/frameworks/Java/dropwizard/src/main/java/com/example/helloworld/db/jdbi/WorldRepository.java#L36-L38
Select random world record from the DB. Update the value. Update the record in the DB.

The random number is produced here:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/c1c5d1ce032d967841bce9be6329046c6aa18c06/frameworks/Java/dropwizard/src/main/java/com/example/helloworld/resources/Helper.java#L31-L33
If the value returned from `randomWorld()` is lower (say 1000), than the code will work with the first 1000 records only (instead of all 10000). So the chance of deadlock should be greater for given RPS rate.

FredDeschenes:
Oh right that makes more sense. I'll have some time to look at this either tomorrow or this weekend, I'll keep you updated.

PR #4646, opened 2019-04-11
[ci skip] Easier debug

joanhey:
Added to `--mode debug`  http://localhost:28080 to open with Ctrl+click.
It will not be necessary to look at the documentation.

### DIFF
--- a/toolset/benchmark/benchmarker.py
+++ b/toolset/benchmark/benchmarker.py
@@ -170,7 +170,7 @@ class Benchmarker:
 
             # Debug mode blocks execution here until ctrl+c
             if self.config.mode == "debug":
-                log("Entering debug mode. Server has started. CTRL-c to stop.",
+                log("Entering debug mode. Server http://localhost:28080 has started. CTRL-c to stop.",
                     prefix=log_prefix,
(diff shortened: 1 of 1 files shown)


nbrady-techempower:
@joanhey This only works for people that use port 8080 and then are using vagrant which is mapping 28080 to 8080. If you're testing through some other means, or using curl on another screen in vagrant, it will be the actual test port.

joanhey:
From vagrant is more than 90% of cases.
Perhaps we can add to the test: from vagrant  http://localhost:28080

nbrady-techempower:
It's not just that it's vagrant, it's also that the test has to be using port 8080 which we don't enforce. I use vagrant too, but I never hit the debug endpoint from outside of vagrant; I'm usually using screen or tmux. Let me think on this a moment. We can probably test that the user is using vagrant and then add a message like:
```
http://localhost:8080 or http://localhost:28080 outside vagrant
```

or if they're not using vagrant and the port isn't 8080 just
```
http://localhost:<test.port>
```

nbrady-techempower:
Addressed this in #5135

PR #4648, opened 2019-04-11
Update Eloquent ORM to return objects

joanhey:
By requirements:

> The row should be converted to an object using an object-relational mapping (ORM) tool. Tests that do not use an ORM will be classified as "raw" meaning they use the platform's raw database connectivity.

### PHP-Eloquent
Eloquent was using arrays, changed to use objects.
In the fortunes tests it is noticeable, using only arrays and sorting directly with `asort`.

### PHP-lavravel-query-builder
It's only a query builder not an ORM, but it returns stdObject.
So I changed to `orm="Micro"` for now. 
There is no clear definition between full and micro.
But for me a query builder it is NOT an ORM.

### Hamlet
It's also using arrays not objects.
So for now changed to `orm="Raw"`
I don't know enough Hamlet to change it.
@vasily-kartashov Please try to fix it.

Perhaps that happens in other languages, as the tests can't check it. Only with code review you can identify it.


joanhey:
Passed the travis-ci for these changes.

joanhey:
After more checking, there are more ORM using array.
The first time I only searched for `asort()`

nbrady-techempower:
@joanhey Did you want to do anything else here or is this ready?

joanhey:
It's ready

PR #4660, opened 2019-04-16
Updating Hamlet framework

vasily-kartashov:
using entities instead of raw ORM, removing roadrunner, react and amp integration.


nbrady-techempower:
@vasily-kartashov don't know if you saw, there's some broken stuff here

PR #4662, opened 2019-04-17
aspnetcore: pg single update statement

n-stefan:
Use a single UPDATE statement for postgres (mysql doesn't seem to support UPDATE ... FROM).
Locally it is faster than multiple UPDATEs; would be interesting to see how it behaves on Citrine.
If it wasn't any good, I would revert the changes.


nbrady-techempower:
@n-stefan Should any of this be run by the aspnetcore team before submitting?

n-stefan:
I'm not sure, have to ask.
/cc @sebastienros

nbrady-techempower:
I'll get this in now since we're about to bring all the servers back up and revert if needed.

PR #4669, opened 2019-04-18
Syntax error

MatheusRV:
A piece of code could not be parsed due to syntax errors.

Syntax errors prevent code from executing correctly. If a piece of code contains syntax errors, this most likely indicates that it is never run and thus is dead code that should be removed.


nbrady-techempower:
This code works as intended. Without that line, we're not selecting the proper database in mongo.

PR #4671, opened 2019-04-18
Play2 enhancements

mkurz:
- Upgrades to Play 2.7.1
- Reverts a commit from #4556 because of bad performance (with akka-http backend, not with netty)
- Let's see if `parallelism-max = 64` has an impact


mkurz:
This is ready to merge.

PR #4675, opened 2019-04-19
Update firenio to 1.1.0, try nodelay

kyle2008:


frameworks/Java/firenio/src/main/java/hello/TestHttpLoadServer.java
```diff
@@ -133,6 +134,12 @@ public void channelOpened(Channel ch) throws Exception {
                 }
             });
         }
+        ByteTree cachedUrls = null;
+        if (cachedurl){
```
nbrady-techempower: Could you explain a little bit what `cachedUrls` is?

kyle2008: It used to cut down String instance. For example when decode the first line of http protocol, the url may always in the server configured path, so no more “new String(url bytes)” is needed.

PR #4691, opened 2019-04-24
Reitit + plaintest & db

ikitommi:
Added plaintext & db-test for reitit, the clojure web library.

### DIFF
--- a/frameworks/Clojure/reitit/benchmark_config.json
+++ b/frameworks/Clojure/reitit/benchmark_config.json
@@ -3,10 +3,12 @@
   "tests": [{
     "default": {
       "json_url": "/json",
+      "plaintext_url": "/plaintext",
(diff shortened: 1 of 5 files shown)


frameworks/Clojure/reitit/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '2'
```
zloster: Hi,
you don't need to embed PostgreSQL in the framework's `Docker` file. The TFB toolset is taking care of starting the appropriate database container for the tests. The configurations are [accessible here](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/toolset/databases).

ikitommi: Thanks! Removed that part and polished the implementation.

PR #4701, opened 2019-04-27
Ubiquity framework cleaning

jcheron:
Slight optimization:
- Remove unused files
- Use PHPs spaceship operator instead of `strcmp` in `Fortunes`
- Update `nginx` config


jcheron:
Thanks,
no rest for the brave!

PR #4714, opened 2019-05-01
Upgrade CakePHP to version 3.7

oozone:
Upgrade CakePHP to version 3.7


frameworks/PHP/cakephp/src/Controller/FortunesController.php
```diff
@@ -0,0 +1,42 @@
...
+        $this->loadModel('Fortune');
+	$this->set('title_for_layout', 'Fortunes');
+
+        $fortunesTable = TableRegistry::getTableLocator()->get('Fortune');
+
+        $query = $fortunesTable->find('all');
+
+        // Calling all() will execute the query
+        // and return the result set.
+        $results = $query->all()->toArray();
```
joanhey: Please, use objects to have Full ORM.

> Using an ORM, all Fortune objects must be fetched from the Fortune table, and placed into a list data structure.

joanhey:
It's not necessary to close the PR every time.

Please try to delete unnecessary files.

frameworks/PHP/cakephp/src/Controller/FortunesController.php
oozone: Will do!

joanhey: Example for sorting the objects:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/PHP/yii2/app/controllers/SiteController.php#L52-L62

joanhey:
Ready to be merged :+1:

nbrady-techempower:
Thanks for reviewing @joanhey!

PR #4719, opened 2019-05-04
Add support for specifying multiple test types

amichair:
This allows one to specify multiple types on the command line, e.g. '--type db query fortune'. Convenient for comparison, analysis, development, etc. Backwards compatible with default, explicit 'all' and single argument.


amichair:
The build failure looks unrelated to this change...

nbrady-techempower:
tested locally; thanks!

PR #4740, opened 2019-05-07
Change docker containers to shut down gracefully

amichair:
This gives containers up to a couple of seconds to shut down gracefully before killing it as it did before.

This allows collecting and logging statistics at the end of the tests, closing logs and other files that need to be persisted, maybe do cleanup, and in theory might prevent some other issues caused by killing the containers violently. Although it adds up to 2 seconds per framework to the test, I think this is negligible overall compared to the test times or to adding a new framework.

I chose the 2 second timeout since it's enough for the frameworks I tested and doesn't add much, though you might want to tweak it in the future if necessary (the default for this parameter is 10 seconds which I thought might be too much in the worst case).

### DIFF
--- a/toolset/utils/docker_helper.py
+++ b/toolset/utils/docker_helper.py
@@ -250,7 +250,7 @@ class DockerHelper:
     @staticmethod
     def __stop_container(container):
         try:
-            container.kill()
+            container.stop(timeout=2)
             time.sleep(2)
         except:
             # container has already been killed


amichair:
Sorry about the typo, fixed.

nbrady-techempower:
@amichair Unfortunately, if you look through the history of this file, you'll see that we've tried this in the past. A lot of containers simply don't stop this way. Waiting 2 seconds is just a guess, and will lead to a number of `start/stop` problems that we've fixed with this.

An alternative solution is to maybe stop, sleep(5) ish, and then kill. But they all need to be killed to be sure.

amichair:
That's exactly what container.stop does - it sends a SIGTERM and if the container is not terminated by the timeout (grace period) then it also sends a SIGKILL. The container.kill method just send the SIGKILL directly. I did leave the 2 extra seconds of sleep that were there after the kill (which presumably was added to work around some issue), so it looks like eventually it does the same as before, except it gives the container a chance to do things right for 2 seconds first.

I see the history, but don't think this has been tried before - there was a stop and a kill without a sleep, which I guess had issues, but once the sleep was added to solve things, it was not tried with a proper stop again.

Although if you know this has been tried before and doesn't work for some reason, I can't argue with that :-)

(The travis build failure above look unrelated.)

nbrady-techempower:
Yeah that's what container.stop() is supposed to do, but if I remember the issue correctly, dockerpy was returning early from it with false positives and SIGKILLs were not sent. Since the current run is almost complete, I'll throw this in and see what happens. We can compare problems with containers starting/stopping. If it's the same or better, we'll leave this in.

amichair:
Fair enough - thanks! Also as you suggest, if the kill never happens for some reason, we can add an explicit one after the stop.

PR #4748, opened 2019-05-08
Unify stack config

joanhey:
## Unify configs for all 

Use **nginx** `keepalive 40`

Use **php-fpm** `pm.max_children = 1024`

So will be easier to change all configs easily.
And now all have the same stak config, necessary for a good benchmark.

### Commented nginx open_file_cache
The bench don't use static files.


joanhey:
Fail:
hhvm that we know
hamlet-peachpie than this PR don't touch, and fail in the last continous run also.

Ready to be merged

PR #4753, opened 2019-05-09
Add Ubiquity-react

jcheron:
Add benchs for Ubiquity with internal server based on ReactPHP.
All tests are green this time


frameworks/PHP/ubiquity/benchmark_config.json
```diff
@@ -18,10 +41,10 @@
...
         "orm": "Full",
         "platform": "None",
-        "webserver": "nginx",
+        "webserver": "react",
         "os": "Linux",
         "database_os": "Linux",
-        "display_name": "ubiquity",
+        "display_name": "ubiquity-react",
         "notes": "",
         "versus": "php"
```
joanhey: Ideally will be versus `"react"`, but there isn't a vanilla react test.

So perhaps you can compare vs `"ubiquity"` and you will know the diference.
Will be more informative than vs "php".

What do you thing is the best @nbrady-techempower

jcheron: Hi @joanhey ,
Indeed, I asked myself the question.
versus php is not really satisfactory

joanhey:
Hi @jcheron 
Like you know react, Could you make a vanilla react test in a new PR ?

Most people try react or road-runner with his framework, but nobody add a vanilla test for them.

jcheron:
@joanhey 
It's okay for the vanilla React test

frameworks/PHP/ubiquity/benchmark_config.json
nbrady-techempower: Like you said @joanhey, vs a vanilla react test would be idea, but totally up to you guys.

jcheron: To look at it more closely:
- hamlet swoole is vs php
- Laravel-swoole is vs swoole
Which is not very consistent.

Waiting for the vanilla React test,
what should I use: Ubiquity or php ?

jcheron: If I use `php` it allows me to see the performance of the `Ubiquity-React` couple compared to `PHP`.
If I use `Ubiquity`, it is to highlight React's performance
It seems to me that the comparison with PHP is more relevant for the moment

nbrady-techempower: I think that's fine, after all, you can look at the charts and compare it against anything you want.

jcheron: Thank you,
vanilla React test to come....

PR #4758, opened 2019-05-10
Update Resin to 4.0.61 and OpenJDK 11

zloster:
Resin and Java 11, the dependencies are also updated (maybe there are omissions):
- [x] Servlet
- [x] Gemini
- [x] wicket
- [x] revenj-jvm
- [x] activeweb
- [x] curacao
- [x] Kotlin/hexagon
- [x] Clojure/compojure

Only `Resin` updated. Stays on `Java 8`. `Grails` code is very old and the update will require effort.
- [x] Groovy/grails - too much effort will be required. Very old version currently.


nbrady-techempower:
This is fantastic @zloster. Really appreciate it!

PR #4761, opened 2019-05-12
Update to Java 11 and Hibernate

zloster:
- Dropwizard
- Officefloor
- Spark
- undertow-jersey

**Please note**: one of the `Officefloor` variants (Spring Boot) is currently failing. I haven't investigated the problem - the dependencies are the latest (2.1.4.Release). I don't know what is a better approach - to leave it on the old versions of Java or leave it failing but configured the same as all the others.


nbrady-techempower:
@zloster If officefloor is passing on the older version I'd rather leave it that way or remove it altogether.

zloster:
@nbrady-techempower Only [1 from 6 variants of `Officefloor` is failing](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/531516953#L2498).
I'll downgrade only the failing to Java 8.

zloster:
`Officefloor` variant `SpringBoot` is not updated. The framework version is `3.4.0`, the OpenJDK is 10, Maven is updated to OpenJDK 11-based image. It seems the problem is somewhere in the actual code for the benchmark - when I downgraded the JDK from 11 to 10, the error was the same (as before the downgrade).
@nbrady-techempower This should be OK for merge now.

PR #4762, opened 2019-05-13
Update reitit JVM-settings

ikitommi:
Update reitit JVM & Pool settings.


ikitommi:
Just peeked at the [upcoming results for this](https://tfb-status.techempower.com/results/bb73698b-f41d-40e3-98ea-fb71d521d3cd) and it seems that the db got 0 responses in all the tests :(

* Is there any way to get the logs to see what errors there are? (pool too big, unclosed connections, invalid response etc.)
* Is it possible to run locally with the prod postgresql settings to test how different things like pool size effect the perf

ikitommi:
Fix https://github.com/TechEmpower/FrameworkBenchmarks/pull/4795

PR #4766, opened 2019-05-13
Format document using shfmt

MatheusRV:

### DIFF
--- a/tfb
+++ b/tfb
@@ -55,8 +55,8 @@ _prepend_dir_context_if_necessary() {
 
 _prepend_path_if_relative() {
     case "$2" in
-        /* ) printf '%s\n' "$2" ;;
-         * ) printf '%s\n' "$1/$2" ;;
+    /*) printf '%s\n' "$2" ;;
+    *) printf '%s\n' "$1/$2" ;;
     esac
 }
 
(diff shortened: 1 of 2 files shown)


tfb
```diff
@@ -98,9 +98,9 @@ SCRIPT_PATH="$(realpath "$0")"
 SCRIPT_ROOT="$(dirname "$SCRIPT_PATH")"
 
 if ! docker network inspect tfb >/dev/null 2>&1; then
-  docker network create tfb >/dev/null
+    docker network create tfb >/dev/null
 fi
 
 test -t 1 && USE_TTY="-t"
-docker build -t techempower/tfb - < ${SCRIPT_ROOT}/Dockerfile
+docker build -t techempower/tfb - <${SCRIPT_ROOT}/Dockerfile
```
nbrady-techempower: This will break the docker build?

MatheusRV: No, this isnt a broken PR. Works perfectly.

nbrady-techempower: Sorry to be a nit but I prefer leaving the space between `<` and the `Dockerfile`. It's both easier to read and matches the docker documentation.

https://docs.docker.com/engine/reference/commandline/build/

image

nbrady-techempower:
I'm going to close this for now. I'm ok with the double quoting but I'd prefer a space before and after redirection operators, like ` > somefile.txt`

PR #4767, opened 2019-05-13
Package 'python-software-properties' has no installation candidate

MatheusRV:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN apt -qqy install -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--
   cloc dstat                    `# Collect resource usage statistics` \
   python-dev \
   python-pip \
-  python-software-properties \
+  software-properties-common \
   libmysqlclient-dev            `# Needed for MySQL-python` \
   libpq-dev                     `# Needed for psycopg2`
 


Dockerfile
```diff
@@ -10,7 +10,7 @@ RUN apt -qqy install -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--
   cloc dstat                    `# Collect resource usage statistics` \
   python-dev \
   python-pip \
-  python-software-properties \
+  software-properties-common \
```
nbrady-techempower: Hey @MatheusRV -- Are you saying that currently this is failing to install this package? And if so, that would seem to me like we don't need this package?

MatheusRV:
```
Get:1 http://security.ubuntu.com/ubuntu cosmic-security InRelease [79.6 kB]         
Hit:2 http://us.archive.ubuntu.com/ubuntu cosmic InRelease         
Get:3 http://us.archive.ubuntu.com/ubuntu cosmic-updates InRelease [83.2 kB]            
Hit:4 http://ppa.launchpad.net/webupd8team/java/ubuntu cosmic InRelease              
Hit:5 http://us.archive.ubuntu.com/ubuntu cosmic-backports InRelease                       
Fetched 163 kB in 2s (82.8 kB/s)               
Reading package lists... Done             
Updated successfully.               
Reading package lists... Done                
Building dependency tree             
Reading state information... Done
Package python-software-properties is not available, but is referred to by another package.            
This may mean that the package is missing, has been obsoleted, or        
is only available from another source         
However the following packages replace it:         
  software-properties-common
```

PR #4774, opened 2019-05-14
update smart-socket to 1.4.1

smthing:


frameworks/Java/smart-socket/src/main/java/org/smartboot/http/JSON.java
```diff
@@ -2,12 +2,22 @@
...
+import java.util.Map;
+
 /**
  * @author 三刀
  * @version V1.0 , 2018/8/12
  */
 public class JSON {
-    public static byte[] toJson(Object obj) {
-        return JSONObject.toJSONBytes(obj);
+    private static Map<String, byte[]> cacheMap = new HashMap<>();
```
nbrady-techempower: Hi @smthing -

We're not going to allow caching of JSON serialization. We'll make it a point to be more specific about this in the rules.

Issue #4771, opened 2019-05-14
actix-pg doesn't read randomNumber


sebastienros:
As per https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#database-updates on point .vi `At least the randomNumber field must be read from the database result set.`

It looks like `actix-pg` is not doing it.

https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Rust/actix/src/db_pg.rs#L169

The other rust implementation (`actix-core`) is using __diesel__ which should read all values by itself.

nbrady-techempower:
Resolved by #4777

PR #4777, opened 2019-05-14
Actix: update deps and plaintext/json impl

fafhrd91:
* Read rundomNumber #4771 
* Update dependencies
* Add custom plaintext and json impl

### DIFF
--- a/frameworks/Rust/actix/src/db_pg_direct.rs
+++ b/frameworks/Rust/actix/src/db_pg_direct.rs
@@ -1,3 +1,4 @@
+use std::fmt::Write;
 use std::io;
 
 use actix_http::Error;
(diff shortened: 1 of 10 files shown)


frameworks/Rust/actix/src/db_pg_direct.rs
```diff
@@ -132,7 +133,8 @@ impl PgConnection {
                 );
 
                 for w in &worlds {
-                    update.push_str(&format!("({}, {}),", w.id, w.randomnumber));
+                    let _ =
+                        write!(&mut update, "({}, {}),", w.id, rng.gen_range(1, 10_001));
```
sebastienros: > Each World object must have its randomNumber field updated to a new random integer between 1 and 10000.

You need to update the `World` object, then you can get the value from it. It's also necessary as the new value needs to be returned in json.

fafhrd91: right. fixed.

fafhrd91:
@nbrady-techempower could this be merged before next test run

PR #4778, opened 2019-05-15
Update grizzly-jersey and undertow-jersey 

zloster:
Changes:
- introducing JPA EntityManager instead of Hibernate SessionFactory;
- the code uses `unwrap()` to get access to the Hibernate session;
- domain objects getters/setters and `private`;
- `update` test sorting before saving (transaction deadlocks);
- `undertow-jersey` gets `update` test;

Now the two projects are almost the same. They could be moved in new directory with three build options.


michaelhixson:
Thanks for doing this.

> Now the two projects are almost the same. They could be moved in new directory with three build options.

To clarify, you're thinking that a new `frameworks/Java/jersey` directory would be created, and it would have three variants: `grizzly`, `undertow`, and `undertow-hikaricp`?  And `frameworks/Java/undertow-jersey` would be deleted, and `frameworks/Java/grizzly` would contain the raw Grizzly implementation only, no Jersey code?

That would make much more sense to me.  I think we require one of the variants to be the default, and I suppose that should be Grizzly, since the Grizzly-Jersey connector is something the Jersey developers made ([Jersey connecctors](https://github.com/eclipse-ee4j/jersey/tree/master/connectors)), whereas the Undertow-Jersey connector is something inlined in this repository (and is perhaps not suitable for production; it doesn't implement async response support for example).

zloster:
Yes, new `frameworks/Java/jersey` with `grizzly`, `undertow` and `undertow-hikaricp`.
My comment was referring to this (at the end of my first comment): https://github.com/TechEmpower/FrameworkBenchmarks/pull/4603

zloster:
I've forgot one important point about the implementation:
It's using Jersey HK2 DI framework for injecton of the EntityManagerFactory with `javax.annotations.@Inject`. There are examples with [javax.persistence.@PersistenceContext](https://github.com/eclipse-ee4j/jersey/blob/master/examples/bookmark-em/src/main/java/org/glassfish/jersey/examples/bookmark_em/resource/UsersResource.java#L41-L43) or [javax.persistence.@PersistenceUnit](https://github.com/eclipse-ee4j/jersey/blob/master/examples/bookmark/src/main/java/org/glassfish/jersey/examples/bookmark/resource/UsersResource.java#L40-L41). But they are using embedded `glassfish` (check the `pom`s), not `grizzly`. There are various sample integration projects for `grizzly` and `JPA` but I couldn't find a sample using `HK2`.
It is also possible to have managed and injected instances for `EntityManager` if you want to use request scope processing.

PR #4786, opened 2019-05-17
Update php to Ubuntu 19.04 and nginx 1.15.9

joanhey:
Ubuntu 19.04 use nginx 1.15.9


joanhey:
Only fail hhvm that we know.

Ready to be merged.

PR #4827, opened 2019-06-01
Add async-db test, update deps

ikitommi:
* Add reactive SQL test for Clojure/Reitit.
* Update dependencies
* Fix reflection warning
* Set worker thread pool to 256

### DIFF
--- a/frameworks/Clojure/reitit/README.md
+++ b/frameworks/Clojure/reitit/README.md
@@ -11,11 +11,11 @@ This is the [Reitit](https://github.com/metosin/reitit) portion of a [benchmarki
 The dependencies are documented in [project.clj](hello/project.clj),
 but the main ones are:
 
-* [Clojure 1.10.0](http://clojure.org/)
-* [ikitommi/immutant-web "3.0.0-alpha4"](https://github.com/ikitommi/immutant) - a performant fork of [Immutant]((http://immutant.org/))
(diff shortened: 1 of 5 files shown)


ikitommi:
Hi. Any change of this getting into round18 results? on my tests, ~doubles the db-perf, making Clojure one of the fastest entries there. Would like to see that in the official results.

cheers.

nbrady-techempower:
@ikitommi Unfortunately, probably not. We've captured citrine results for round 18 and we're running Azure (we only have a limited number of credits) off of that same commit.

ikitommi:
Credits? When the citrine results for b8e862a759 (with this in) arrive, if the results are what I expect, is it possible to add more credits to run azure too with it? How much would be needed? If the price is known, could ask our company to sponsor the needed credits, if you take test credit donations. Many ifs :)

ikitommi:
It seems that the changes are actually in ae69f390a8ec33f6159bbfc5c2f7947af06d5626, which only just started. Is it possible to sponsor an extra round for azure? Anyway, thanks for the benchmarks.

nbrady-techempower:
@ikitommi I'll pass this along but I think there's too many moving pieces to get that all approved from our end, too. We've been pretty busy with our paying clients so we can keep the lights on. :wink: We are probably evaluating the two runs we captured for the round this week to make sure there are no glaring issues so we can get the round published. If that changes, I'll let you know. Our "credits" reset every month, sponsored by the MS team. We usually do some basic tests and then one full run on Azure.

PR #4831, opened 2019-06-04
Adding Quarkus (https://quarkus.io) framework to the benchmark suite.

jringuette:


frameworks/Java/quarkus/benchmark_config.json
```diff
@@ -0,0 +1,30 @@
...
+        "fortune_url": "/fortunes",
+        "port": 8080,
+        "approach": "Realistic",
+        "classification": "fullstack",
+        "database": "Postgres",
+        "framework": "Quarkus",
+        "language": "Java",
+        "flavor": "None",
+        "orm": "full",
+        "platform": "Wildfly",
```
zloster: IMO this should be `JAX-RS`. See for [example the `grizzly-jersey`](https://github.com/TechEmpower/FrameworkBenchmarks/blob/4a05d28ae2846ce0f6d17ed3168c88f026e497c8/frameworks/Java/grizzly/benchmark_config.json#L38) framework.

jringuette: JAX-RS is only an API specification and not and implementation. It would make more sense to put the actual implementation here, which is RESTEasy. What do you think?

zloster: I was thinking about the following case: I'm looking at the result dashboard and I [want to know how `JAX-RS` implementations compare](https://www.techempower.com/benchmarks/#section=test&runid=52587667-bee1-4b9b-b34d-929b616d5e4c&hw=ph&test=fortune&p=zik0zj-zik0zj-zhxjwf-4fti4f).
Currently most of the projects bare the implementation name in the `variant name`. `minijax` is an exception - the project provides its own implementation. But yes, the current scheme is not very suitable for `quarkus`.
Is my explanation making more sense now?

jringuette: It makes sense. I will update it

jringuette: Done

PR #4843, opened 2019-06-10
update beetlex

IKende:


frameworks/CSharp/beetlex/Benchmarks/Benchmarks.csproj
```diff
@@ -7,9 +7,18 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="BeetleX.FastHttpApi" Version="1.4.9.8" />
```
nbrady-techempower: @IKende Same with BeetleX, please. We don't want a copy of the framework in our repo. It should be installed the way a real world user would install it.

IKende: thank，all library has changed

PR #4846, opened 2019-06-11
Upgrade to Hunt 1.2.x

Heromyth:


Heromyth:
@nbrady-techempower This fix is done. Can you merge it?

frameworks/D/hunt/mmap/DemoProcessor.d
```diff
@@ -0,0 +1,267 @@
...
+
+            default:
+                respondWith404();
+                break;
+            }
+        }
+
+
+        private void respondSingleQuery() {
+            int id = uniform(1, 10000);
```
nbrady-techempower: It looks like the upper bound for `uniform()` isn't inclusive: https://dlang.org/phobos/std_random.html#uniform

Could you update all of those to `uniform(1, 10001)`

PR #4860, opened 2019-06-16
Update ubiquity-* Fix React minimum-stability

jcheron:
### Updated
- Fix composer ``minimum-stability`` problem with Ubiquity-react
- Move Db connection for swoole (Waiting for a non-blocking Db connection)


joanhey:
Please change the `benchmark_config.json`.
The swoole version will be vs `'swoole'` not to `'php'`.

frameworks/PHP/ubiquity/benchmark_config.json
```diff
@@ -5,10 +5,10 @@
...
         "plaintext_url": "/Plaintext",
-        "db_url": "/Db_",
-        "update_url": "/Db_/update/",
-        "query_url": "/Db_/query/",
-        "fortune_url": "/Fortunes_",
+        "db_url": "/Db",
+        "update_url": "/Db/update/",
+        "query_url": "/Db/query/",
+        "fortune_url": "/Fortunes",
         "port": 8080,
```
joanhey: Please change the benchmark_config.json.
The swoole version will be vs 'swoole' not to 'php'.

jcheron: It's not really interesting to know how much slower it is than ``swoole``, which uses a pool of non-blocking connections to the database.
This version uses Mysql PDO, and Swoole is just used as a Server.
In this context, It is more interesting to see the gain compared to ``php``.

In the same spirit, ``laravel-swoole`` is vs ``php`` and not vs ``swoole``.

jcheron: It's not really interesting to know how much slower it is than ``swoole``, which uses a pool of non-blocking connections to the database.
This version uses Mysql PDO, and Swoole is just used as a Server.
In this context, It is more interesting to see the gain compared to ``php``.

In the same spirit, ``laravel-swoole`` is vs ``php`` and not vs ``swoole``.

jcheron: It's okay: fixed

joanhey: Thank you @jcheron 

In #4865 changed swoole to platform. And examples about others languages.
Swoole is not only a webserver. Nobody will use it only like a web server.
The tests are not only about db connections, also check the swoole code in this bench for non-blocking connections.

jcheron: @joanhey I agree with that.
As for non-blocking db connections, it's a little more complicated than reproducing the swoole bench code, the framework I use has to implement them itself.

PR #4863, opened 2019-06-18
Added Vert.x Web Scala.

NikolajLeischner:
This adds a Scala version of the Vert.x Web framework. The test is closely derived from the Java version, as to show if there are any differences in performance between the Java version and the Scala wrapper.


NikolajLeischner:
Thanks for reviewing; I removed the file.

PR #4864, opened 2019-06-18
Config now show the webserver h2o

joanhey:
Useful when comparing results
Also when used like a lib.

### DIFF
--- a/frameworks/C/h2o/benchmark_config.json
+++ b/frameworks/C/h2o/benchmark_config.json
@@ -18,7 +18,7 @@
       "flavor": "None",
       "orm": "Raw",
       "platform": "None",
-      "webserver": "None",
+      "webserver": "h2o",
       "os": "Linux",
       "database_os": "Linux",
       "display_name": "H2O",
--- a/frameworks/Crystal/h2o.cr/benchmark_config.json
+++ b/frameworks/Crystal/h2o.cr/benchmark_config.json
@@ -14,7 +14,7 @@
         "flavor": "None",
         "orm": "Raw",
         "platform": "None",
-        "webserver": "None",
+        "webserver": "h2o",
         "os": "Linux",
         "database_os": "Linux",
         "display_name": "crystal-h2o",


volyrique:
FYI I unset the field in #3670 because in the case of `C/h2o` it is a little deceiving - the latter is not using the standalone H2O Web server. I guess the question is how exactly is that field supposed to be interpreted.

joanhey:
The more information the better.

Are you using `h2o server` standalone or like a lib ? Yes or Not.

That's what you see now, when you select h2o like webserver:
https://www.techempower.com/benchmarks/#section=test&runid=14aa155b-efb0-46c2-8354-c2b8debd36bb&hw=ph&test=plaintext&w=zik0zj-z8kflr-e7

That's is deceiving, because have 2 really fast implementations.

joanhey:
The h2o config for h2o_mruby is not optimized.
I will create a PR.

volyrique:
> The more information the better.

Nobody is arguing the opposite, but more information is better as long as you have a good understanding of it. Read on:

> That is deceiving, because h2o have 2 more really fast implementations.

Using the standalone server or the library directly can have very different performance characteristics. For example, `h2o` (the C test implementation which is based on `libh2o`, not the standalone server) uses the `SO_REUSEPORT` socket option (and I believe `h2o.cr` as well), while the standalone server doesn't. So, it can be argued equally well that the two different groups (using the standalone server or not) are not directly comparable. This is the reason why I unset the field in #3670 (initially when I pushed the code it was set in the same way as in your patch). Also, the documentation of the field states:

> webserver (metadata): name of the web-server (also referred to as the "front-end server")

which can be interpreted as implying that the test implementation uses a front-end server (that communicates with a back-end via FastCGI or something similar), which `h2o` (and probably `h2o.cr`) doesn't.

Anyway, I don't have strong feelings on the subject, so I am not opposed to your changes - I just wanted to explain some of the subtleties involved.

PR #4865, opened 2019-06-18
[ci skip] Config add swoole platform to php frameworks using it

joanhey:
It's not only a webserver but a completely different platform.

Also useful when comparing results in visualize mode.

And will be with `Front End Server / Web Server = none`.


joanhey:
Examples:

Fw/Platform | Language | Platform | WebServer |
----- |  ------ | ------- | -- | 
Nodejs | js | njs | none
Peachpie | php | .Net | kestrel
Spring | java | Tomcat | none
vertx-web | java | vertx | none

PR #4881, opened 2019-06-25
add r2dbc-pool to avoid lack of connections

davecramer:


zloster:
@davecramer Hi,
if you have time can you check this comment: https://github.com/TechEmpower/FrameworkBenchmarks/issues/4885#issuecomment-506957575
I've got some strange behaviour from `spring-webflux` (before this change).

PR #4904, opened 2019-07-04
improvements for pyramid framework (python)

slav0nic:
improved pyramid benchmarks:

- Baked Queries
- filter().one() -> get()
- use more generators
- remove unused code
- bump packages

in my local bench got ~+20-40%


nbrady-techempower:
This is just a note for myself about sqlalchemy baked queries:

> This extension presents a very specific kind of optimization that is not ordinarily needed. As noted above, it does not cache queries, only the string formulation of the SQL itself.

PR #4916, opened 2019-07-11
Update Drogon to the latest version and make some optimizations

an-tao:
In this PR:
* Bump Drogon to the latest version.
* Increase the number of DB connections.


nbrady-techempower:
@an-tao Thanks for the updates! Could you please remove that clang-format file? Keep that in your local repo; we only want files that are necessary for the test to run in this repo

an-tao:
@nbrady-techempower Thank you for your reply. I've removed that clang-format file.

PR #4922, opened 2019-07-12
Update ubiquity react (128 workers)

jcheron:
### Updated
- Increase `php-pm` workers to 128
- Update to Ubiquity 2.2.0


jcheron:
force composer updating

jcheron:
I don't think there's a problem, it's a whim of Travis-ci.

PR #4930, opened 2019-07-15
Update Rust edition in Gotham tests

whitfin:
This sets the Rust edition in the Gotham manifest, which appears to be causing errors in the build logs. Based on the errors reported in [this log](https://tfb-status.techempower.com/unzip/results.2019-07-12-04-35-24-534.zip/results/20190708104112/gotham/build/gotham.log).


whitfin:
Actually, hold off on merging this for now... other Rust projects are showing the same errors. I will update this PR with fixes for them as well - give me a few minutes!

nbrady-techempower:
@whitfin Take your time; let me know when it's ready.

whitfin:
@nbrady-techempower ok, should be good now!

The actual fix for the builds here is simply adding the `edition` flag to each `Cargo.toml` and updating the Docker images to use `latest` (to avoid having to keep each one updated periodically).

The other changes are a result of running `cargo fmt` inside each project, which just formats using the Rust standard. I also removed a `Cargo.lock` file which was pinning dependencies more strictly than was required.

whitfin:
@nbrady-techempower okay, Travis finally said it's ready for merging :D so should be ready whenever you get time to take a look! Hopefully the next build on http://tfb-status.techempower.com/ should include Gotham/Hyper/Iron successfully.

frameworks/Rust/actix/actix-core.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:1.35
+FROM rust:latest
```
msmith-techempower: Can you lock all these down to a particular version?

It is important that each run be reproducible based on the git commit hash, and if you use `rust:latest`, we will not be able to guarantee the version used on any particular run.

whitfin: @msmith-techempower you got it; done in latest commit!

whitfin:
@msmith-techempower pretty much just the Rust version and the `edition` flags. There were some minor changes in line with these updates (fixing compiler warnings, stuff like that) but nothing that should change how anything behaves beyond that.

The rest is just formatting, yes.

Issue #4788, opened 2019-05-18
libreactor json benchmark


fafhrd91:
[libreactor](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/C/libreactor/src/main.c#L65) uses [libclo](https://github.com/fredrikwidlund/libclo/) library for json serialization 
readme states `It does this by constructing the JSON structures as compound literals, which does not require any runtime memory allocation or system calls.`
that sounds like violation of rules

zloster:
@fafhrd91 
The rules define only that before serialisation there should be a memory allocation for the data structure that will be serialised. They do not imply restrictions of how the JSON serialisation should be achieved.

`libreactor` is calling [this macro](https://github.com/fredrikwidlund/libclo/blob/555dd56fd537effd4e8d7713b7cf66ab284d7329/src/clo.h#L9) which seems to be allocating a `C struct`. This `struct` is sent for serialisation immediately after creation. That's my understanding of the code.

Could you add a little bit more explanation why you think there is a violation of the rules.

fafhrd91:
It is about ‘literal’ in lib description. It may avoid actual json serialization, but I don’t know c to be sure

zloster:
Here is my interpretation of the code. IMO it seems OK - there is escaping for the special symbols and handling of the UTF-8 encoding.

The framework is calling this method:
https://github.com/fredrikwidlund/libclo/blob/555dd56fd537effd4e8d7713b7cf66ab284d7329/src/clo.c#L184

Later the code will move through this portion:
https://github.com/fredrikwidlund/libclo/blob/555dd56fd537effd4e8d7713b7cf66ab284d7329/src/clo.c#L130-L147
The case is `CLO_OBJECT` (check the defined macro).

- First start with writing the "{" in a buffer;
- Mark that this is the first field (`first=1`);
- Cycle over key-value pairs (`clo_pair`)
- (inside the `for` loop) - some logic about the `first` variable;
- Call for `key` string encoding. The code seems to the necessary activities - some escaping and such:
https://github.com/fredrikwidlund/libclo/blob/555dd56fd537effd4e8d7713b7cf66ab284d7329/src/clo.c#L91-L122
- Write `:` to the buffer;
- Recursion call with the `value` data;
- Write the closing `}` in the buffer.

Again, this is my interpretation. Let's see what will say someone with better knowledge of `C`.

volyrique:
[...]

PR #4931, opened 2019-07-16
H2O: Various changes again

volyrique:
* Switch most of the calls to memory allocation routines to h2o_mem_alloc() - leads to cleaner code.
* Remove the multi-process functionality from the h2o.sh script - it is no longer necessary.
* Make sure that calls to abort() are preceded by printing an error message


zloster:
@volyrique Sorry to bother you. I want to ask you to take a look at #4788.
The issue is open for quite some time and it needs someone with good C experience to confirm or to reject the stated problem. I've already written my understanding of the code. My guess is that it shouldn't take more than an hour to review the comments and the code.

PR #4932, opened 2019-07-17
adding restana and 0http frameworks

jkyberneees:
In this PR we submit the following JavaScript frameworks:
- 0http
- restana
> Both frameworks are listed as TOP node.js frameworks at: https://github.com/the-benchmarker/web-frameworks

Only the following tests are supported:
- json
- plaintext


nbrady-techempower:
looks good @jkyberneees, thanks for the additions!

PR #4945, opened 2019-07-22
updating 0http framework version

jkyberneees:
Updating `0http` to version `1.2.0` in order to use the newly created `low` http server: https://github.com/jkyberneees/0http#low-server

### DIFF
--- a/frameworks/JavaScript/0http/app.js
+++ b/frameworks/JavaScript/0http/app.js
@@ -1,8 +1,9 @@
-const cluster = require('cluster')
-const cpus = require('os').cpus()
-
 const cero = require('0http')
-const { router, server } = cero()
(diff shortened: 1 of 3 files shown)


frameworks/JavaScript/0http/app.js
```diff
@@ -1,25 +1,26 @@
...
-
 const cero = require('0http')
-const { router, server } = cero()
+const low = require('0http/lib/server/low')
+
+const { router, server } = cero({
+  server: low()
+})
+
+const date = new Date().toUTCString()
```
nbrady-techempower: You either have to have some mechanism that updates this on the server every second or so or generates this date in the handlers for each route.

jkyberneees:
Will do ASAP, thanks for supporting.

Regards

PR #4955, opened 2019-07-26
Upgrading to dotnet core 3.0 preview7

sebastienros:
- Updating docker image versions
- Removing SpanJson and Utf8Json scenarios
- Temporarilly removing RHTX scenarios until it's compatible with 3.0


frameworks/CSharp/aspnetcore/aspcore-ado-my.dockerfile
```diff
@@ -1,11 +1,10 @@
-FROM mcr.microsoft.com/dotnet/core/sdk:2.2 AS build
+FROM mcr.microsoft.com/dotnet/core/sdk:3.0.100-preview7 AS build
```
benaadams: Better to just use 
```
FROM mcr.microsoft.com/dotnet/core/sdk:3.0 AS build 
```
Then it will automatically update without needing to make more PRs?

sebastienros: I intentionally decided not to. I don't want the benchmarks to be unstable, either because there is a breaking change, or because the perf is volatile.

frameworks/CSharp/aspnetcore/Benchmarks/Data/EfDb.cs
```diff
@@ -42,7 +45,10 @@ public async Task<World[]> LoadMultipleQueriesRows(int count)
             {
                 var id = _random.Next(1, 10001);
 
-                result[i] = await _firstWorldQuery(_dbContext, id);
+                // compiled queries are not supported in EF 3.0-preview7
```
davidfowl: Make these TODO: so we can find them easily.

frameworks/CSharp/aspnetcore/Benchmarks/Middleware/JsonMiddleware.cs
```diff
@@ -26,23 +26,20 @@ public JsonMiddleware(RequestDelegate next)
             _next = next;
         }
 
-        public Task Invoke(HttpContext httpContext)
+        public async Task Invoke(HttpContext httpContext)
```
davidfowl: Why is this async? Just return the Task directly.

murbanowicz:
Microsoft claims it is stable and there will be no breaking changes so it could be merged already as the API for 3.0 is stable already and many people are eager to see the results with 3.0 as it introduced so many performance improvements

frameworks/CSharp/aspnetcore/aspcore-ado-my.dockerfile
sebastienros: In the end we decide to use 3.0. Hope you like it

sebastienros:
@nbrady-techempower you can merge now if you are ok with the changes

jorgeesquer:
This latest run includes this commit, right? https://tfb-status.techempower.com/results/26a79c95-5eec-4572-8c94-dd710df659d7

Aren't we supposed to see some notable improvements with Core 3.0?

slang25:
Yes that run has the changes in, I was expecting to see the same 🤔

sebastienros:
I compiled some numbers to have a cleaner idea. I focused on platform and middleware scenarios.

```
fortunes        
        aspcore-ado-pg          243,637         242,567
        aspcore-mw-ado-pg       209,169         210,625

updates
        aspcore-ado-pg-up       14,867          15,712
        aspcore-mw-ado-pg-up    14,461          15,900

multi-query     
        aspcore-ado-pp          19,747          21,020
        aspcore-mw-ado-pg       18,591          20,942

single query    
        aspcore-ado-pg          296,089         287,514
        aspcore-mw-ado-pg       227,154         238,676

json            
        aspcore                 774,470         756,883
        aspcore-mw-json         582,155         653,790

plaintext       
        aspcore                 6,110,963       6,768,327
        aspcore-mw              2,914,070       3,735,976
```

My conclusions from these numbers:
- We are still missing the RHTX numbers which were our fastest versions. The PR to migrate to 3.0 is still pending but it shouldn't take that long to have it merged.
- Plaintext has improved a lot.
- All middleware based scenarios are faster.
- Fortunes like most db scenarios haven't improved much, mostly due to them being bound to the db driver which hasn't changed. We are still impacted by not having pipelining support.
- Json results are much different than our own benchmarks, where we get 900K for platform and 800K for middleware, using the same hardware. Usually we see the opposite, not being able to be as fast as TE. I hope it's a bad run and next ones will improve.

benaadams:
> Aren't we supposed to see some notable improvements with Core 3.0?

There are other factors, such as the Linux TCP SACK PANIC CVE patch 
https://twitter.com/BrianHauerTSO/status/1156702173322858498 earlier this month

Which for example knocked 9% off aspnet core's plaintext score (from 6,988,971 -> 6,159,250) so a change to 6,768,327 is a gain, however doesn't completely mitigate the effects of the security patch 😢

PR #4956, opened 2019-07-27
Add Nginx Unit platform to PHP

joanhey:
[NGINX Unti](https://unit.nginx.org) is a dynamic web and application server, designed to run applications in multiple languages. It's an alternative to php-fpm.

The information about processes is not too clear, so we try with 64 till receive more information from the maintainers.

The nginx/unit docker only have php 7.0, later I will try to create a docker to use php 7.3.


joanhey:
The results are very good.

| Test | FPM/FastCGI | Unit Nginx
------- | ------------------: | -----------:
Plaintext | 173,500  | 198,636 
Json | 172,707 | 202,307
Single query | 97,203 | 105,750
Fortunes | 73,471 | 84,208

And after checking the data table, scale better Unit.

Unit is using php 7.0 that is about 15-20% slower than php 7.3.

### TODO
Update the dockerfile to use php7.3
Adjust better the processes number

The php.ini for now I see it correct.

PR #4965, opened 2019-07-31
Add Ubiquity raw + update to 2.3.0

jcheron:
### Added
- `Ubiquity-raw` (without ORM and without Template engine)

### Changed
- Update Ubiquity to `2.3.0`

### Removed
- `Ubiquity-react`
I hope it's temporary and that the Php-pm team will find a solution to the problem (see https://github.com/php-pm/php-pm/issues/456)... 
The current results are very unsatisfactory, and I prefer not to make any bad publicity (neither at Php-pm, nor at Ubiquity).
I tried to increase the number of workers to increase performance, but Travis-ci tests do not pass with 512, nor even with 256 workers.


frameworks/PHP/ubiquity/app/controllers/RawFortunes.php
```diff
@@ -0,0 +1,22 @@
...
+namespace controllers;
+
+class RawFortunes extends \Ubiquity\controllers\Controller {
+	
+	public function index() {
+		$fortunes = \Ubiquity\db\Database::start()->fetchAll('SELECT id,message FROM Fortune',\PDO::FETCH_KEY_PAIR);
+		$fortunes[0] = 'Additional fortune added at request time.';
+		\asort($fortunes);
+		?>
+		<!DOCTYPE html><html><head><title>Fortunes</title></head><body><table><tr><th>id</th><th>message</th></tr>
```
joanhey: I don't think that this is a realistic approach.
I understand that use it the vanilla php test. 
But not in a full stack framework.

Also the requeriments say: 
 - The sorted list must be provided to a server-side template and rendered to simple HTML (see below for minimum template). The resulting HTML table displays each Fortune's id number and message text.
 - The HTML generated by the template must be sent as a response.
 - The implementation is encouraged to use best practices for templates such as layout inheritence, separate header and footer files, and so on. However, this is not required. We request that implementations do not manage assets (JavaScript, CSS, images). We are deferring asset management until we can craft a more suitable test.

joanhey: Close and open php in the middle of a class is terrible.
Minimum use will be do an include. But it's also not realistic in a full stack framework. Will be an striped test.

jcheron: I agree with you @joanhey . Consider this as just a test to evaluate the cost of the template engine on performance.
As soon as I have the results, I will submit a code that is more in line with good practices (and benchmark specifications).

joanhey: Wait a little.
In the last run ubiquity fail to start in all platforms.

jcheron: It was a problem with `composer.json` files, I had the bad idea to prepare the release of Ubiquity version `2.3.0` at the wrong time

joanhey: I want to see the ubiquity results with swoole.
As the hamlet with swoole is failing in a lot of tests: db, query and update.

jcheron: `Ubiquity-swoole` is unlikely to have the same problem as `Hamlet-swoole`.
`Hamlet-swoole` implements its own pool of connections to Mysql using `Swoole\Coroutine\MySQL`.

The error displayed in the [logs](https://tfb-status.techempower.com/unzip/results.2019-08-01-06-50-52-747.zip/results/20190728133632/hamlet-swoole/run/hamlet-swoole.log) is related to one of this classes:

>Fatal error: Uncaught Error: Can use "yield from" only with arrays and Traversables in /php/vendor/hamlet-framework/db-mysql-swoole/src/MySQLSwooleProcedure.php:57

While `Ubiquity` uses the same PDO-based connection for all tests (Swoole or not)

jcheron: I do a PR for the template engine (it is ready).
I will do another one if Swoole has a problem in the next results.
What do you think of that?

joanhey: If ubiquity use PDO connections will be slower than plain php, even using swoole, in all db tests.

jcheron: I know that PDO is blocking, and that Swoole's gain is very limited accordingly.
But on closer examination of the results, I felt that for the time being, it was not appropriate to develop a specific Database class for Swoole.

>If ubiquity use PDO connections will be slower than plain php, even using swoole, in all db tests.

The Ubiquity-swoole results are not that bad:
- On `multiple-queries`, it does well with 1 and 5 requests (it's not as good for the others),
- and it's in front on the `data-update` (better than Swoole).

So not in all db tests ;-)

It should not be forgotten that in addition to having the PDO blocking, there is always the weight of the framework, and the operations performed by the ORM...

joanhey: The next week will be kumbia-unit first.

joanhey: And in october-november arrive php 7.4.
And will boost all php frameworks by 30-60% with the new preload.

jcheron: > The next week will be kumbia-unit first.
what is kumbia-unit?
you didn't talk about that on https://github.com/KumbiaPHP/KumbiaPHP
or it was in Spanish....

joanhey: It's using Nginx Unit. Now I added to plain php #4956 . The next week to kumbiaphp.
In local the results in plaintext and json were almost doubled. And using php7.0.
First I will update to php7.3 and optimize the php.ini, but the results are positive.

jcheron: That's cool,
I'll wait and see how it goes in the next results,
 and eventually start working on it.

PR #4972, opened 2019-08-02
Ubiquity-raw fortunes : the return of the template

jcheron:
### Updated
- Restore the template engine in `Fortunes` for the raw tests (For compliance with good practices and test specifications) see https://github.com/TechEmpower/FrameworkBenchmarks/pull/4965#discussion_r310093302
- Delete unused files (react)


frameworks/PHP/ubiquity/benchmark_config.json
```diff
@@ -26,26 +26,28 @@
...
-            "platform": "php-fpm",
-            "webserver": "nginx",
-            "os": "Linux",
-            "database_os": "Linux",
-            "display_name": "ubiquity-raw",
-            "notes": "",
-            "versus": "php"
-          },
+        "json_url": "/Json",
+        "plaintext_url": "/Plaintext",
```
joanhey: Please remove the json and plaintext.
Are the same that with ORM.

jcheron: The framework configuration is slightly different, including for `Plaintext` and `Json`:
The configuration does not need to start the ORM cache in this raw bench.

jcheron: >For me that is an arquitectural problem.

I'd rather say it's code factorization.

In all tests except `Raw`, the ORM cache is started in the file `services.php`, used in all requests.

If I start the ORM cache only in the Db and Fortunes controllers (it is possible), I am forced to duplicate all the code for Swoole, for which ORM starts only once, and not at each request...

By adding the `Plaintext` and `Json` URLs to the `Raw` bench,
I can now evaluate these 2 tests without unnecessary cache, and without the duplication due to Swoole :-)

jcheron: so do I leave it unchanged or do I correct, by duplicating the `Db` and `Fortunes` controllers for `Swoole`?

joanhey: Why do you create new controllers for the raw version and not for swoole ?

I will show you some disadvantages:

1. Every run in the continuous bench, need to run 2 extra tests . And add extra noise to the visualize.
2. Almost nobody will understand the json and plaintext with and without ORM.
3. You are showing a bad example of how to use your framework.

frameworks/PHP/ubiquity/benchmark_config.json
```diff
@@ -26,26 +26,28 @@
...
+        "update_url": "/Raw/update/",
+        "port": 8080,
+        "approach": "Realistic",
+        "classification": "Fullstack",
+        "database": "MySQL",
+        "framework": "Ubiquity",
+        "language": "PHP",
+        "flavor": "PHP7",
+        "orm": "Raw",
+        "platform": "php-fpm",
```
joanhey: Platform changed to `FPM/FastCGI` to avoid confusions in visualize mode.

frameworks/PHP/ubiquity/benchmark_config.json
jcheron: It's okay, @joanhey .
I close this PR and will open another one to correct.

frameworks/PHP/ubiquity/benchmark_config.json
jcheron: Okay, I saw

PR #4974, opened 2019-08-04
[ci skip] Better names for the platforms

joanhey:
to prevent confusions in the visualize mode.

Changed `php-fpm` to `FPM/FastCGI` as the php is redundant and only show the php in the visualize. Now is easier to check it.
DeepinScreenshot_Seleccionar área_20190802201755

Also changed` nginx-unit` to `Unit Nginx` for the same. Avoid confusion with nginx and Unit Nginx.


joanhey:
The changes are OK. In reality these changes don't need the travis tests.

### One travis job exceeded the maximum 50 min

So first moved the ubiquity framework to the other job that only need 19 min.

Later laravel and lumen, that are also using swoole to distribute better the time between jobs.
Every docker using swoole need 2-4 min to create the docker. I think that each one is implementing swoole different and they can't take advantage of the docker cache.

Now the jobs are more distributed and need less time to finish when we do a global php change.

### Still there is a problem with the PostgreSQL docker 
A GPG signing error :
https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/567772185#L1440
That I leave to @nbrady-techempower

PR #4979, opened 2019-08-05
symfony project update to 4.3 + flex 1.4

kaznovac:
- remove surplus symfony configuration parameters
- docker file update, add .dockerignore, reorder dockerfile for better use of docker image cache
- delete unused apache virtual host config
- add readme
- separate controller into benchmark respective controllers
- use simple controller class (without AbstractController)
- dont call controller actions actions (as per best practices)


frameworks/PHP/symfony/src/Controller/BenchController.php
```diff
@@ -2,108 +2,134 @@
...
-      $worlds = array();
+      $worlds = [];
       $conn = $this->get('database_connection');
 
       for ($i = 0; $i < $queries; ++$i) {
-          $id = mt_rand(1, 10000);
-          $randomNumber = mt_rand(1, 10000);
-          $conn->executeUpdate('UPDATE world SET randomNumber=? WHERE id=?', array($randomNumber, $id));
-          $worlds[] = array('id' => $id, 'randomNumber' => $randomNumber);
+          $id = random_int(1, 10000);
```
Tobion: as this is a performance test, using random_int as cryptographically secure function will make this code slower for sure.

frameworks/PHP/symfony/templates/bench/fortunes.html.twig
```diff
@@ -2,4 +2,11 @@
 
 {% block title "Fortunes" %}
 
-{% block content %}<table><tr><th>id</th><th>message</th></tr>{% for fortune in fortunes %}<tr><td>{{ fortune.id }}</td><td>{{ fortune.message }}</td></tr>{% endfor %}</table>{% endblock %}
```
Tobion: wasn't this done on purpose to save some bytes and make it faster?

frameworks/PHP/symfony/src/Entity/World.php
```diff
@@ -10,59 +10,37 @@
...
     }
 
-    /**
-     * Set randomNumber
-     *
-     * @param integer $randomNumber
-     * @return World
-     */
-    public function setRandomNumber($randomNumber)
+    public function setRandomNumber(int $randomNumber): World
```
Tobion: does not look like we need to make this method chainable.

frameworks/PHP/symfony/src/Controller/BenchController.php
```diff
@@ -2,108 +2,134 @@
...
-class BenchController extends Controller
+class BenchController extends AbstractController
 {
-    use ControllerTrait;
-
-    public function plaintextAction()
+    /**
+     * @Route("/plaintext")
+     */
+    public function plaintextAction(): Response
```
Tobion: best practice is to not use the `Action` suffix anymore

frameworks/PHP/symfony/src/Controller/BenchController.php
```diff
@@ -2,108 +2,134 @@
...
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
-use App\Entity\Fortune;
-use App\Entity\World;
+use Symfony\Component\Routing\Annotation\Route;
 
-class BenchController extends Controller
+class BenchController extends AbstractController
```
Tobion: it would probably be faster to not use AbstractController and only inject explicit dependencies as that lowers the number of dependencies. on the other hand, it makes dependencies like doctrine more lazy. so maybe needs measurement.

frameworks/PHP/symfony/src/Controller/BenchController.php
kaznovac: true

most of the other PHP frameworks are using the the same random so using any other would be unfair to use faster function (e.g. https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/PHP/laravel/app/Http/Controllers/Controller.php#L24)

frameworks/PHP/symfony/src/Controller/BenchController.php
kaznovac: thanks, i've missed that, will fix

frameworks/PHP/symfony/templates/bench/fortunes.html.twig
kaznovac: yes definitely, a micro optimization that one would never do in a production code...

on this dataset count that ~15 '\n' characters are 'overhead'

frameworks/PHP/symfony/src/Entity/World.php
kaznovac: i agree - not a fan of 'fluid' methods neither.
i can change this.

frameworks/PHP/symfony/src/Controller/BenchController.php
kaznovac: definitely worth a test, i'll try and post results

frameworks/PHP/symfony/src/Controller/BenchController.php
kaznovac: missed this, will fix

frameworks/PHP/symfony/src/Controller/BenchController.php
Tobion: it's not only about speed but semantics. we don't need cryptographically secure integers here. so no point in using it. you could also argue it's not fair for php when compared to other languages.

kaznovac: i agree, changed it back to `mt_rand`

frameworks/PHP/symfony/src/Controller/BenchController.php
kaznovac: no noticeable difference in performance injecting over `AbstractController`, performance loss even occurs when injecting services in actions

but i've managed to get significant boost ~7% when i split mono-controller into several dedicated simple controllers

Tobion: 👍 makes sense

frameworks/PHP/symfony/config/services.yaml
```diff
@@ -7,15 +10,12 @@ services:
...
 
     # makes classes in src/ available to be used as services
     # this creates a service per class whose id is the fully-qualified class name
     App\:
         resource: '../src/*'
-        exclude: '../src/{Entity,Migrations,Tests}'
+        exclude: '../src/{DependencyInjection,Entity,Migrations,Tests,Kernel.php}'
 
     # controllers are imported separately to make sure services can be injected
     # as action arguments even if you don't extend any base controller class
```
Tobion: I'd remove the `tags: ['controller.service_arguments']` so it can't be used at all. You just need to set `public: true` instead.

frameworks/PHP/symfony/src/Controller/FortunesController.php
```diff
@@ -0,0 +1,54 @@
...
+
+    public function __construct(Environment $twig, FortuneRepository $fortuneRepository)
+    {
+        $this->twig = $twig;
+        $this->fortuneRepository = $fortuneRepository;
+    }
+
+    /**
+     * @Route("/fortunes")
+     * @throws
```
Tobion: those seem irrelevant

kaznovac: i was reluctant to add them - phpstorm marks half of the controller code as unchecked function calls :|

i should probably remove them

frameworks/PHP/symfony/src/Controller/BenchController.php
kaznovac: findings: while `random_int` is 30-40 times slower than `mt_rand` it has no significant effect on test performance (it can be theoretically called max 500 times, in practice this number is 20)

frameworks/PHP/symfony/config/services.yaml
kaznovac: will try,
this is compile pass so i don't expect any boosts in performance

kaznovac:
@Tobion if you have time i have few questions (and potential tasks for symfony):

- application does not use localization but there is `Symfony\Component\HttpKernel\EventListener\LocaleListener::setDefaultLocale` Event listener registered - i can't find a config to unregister it (it was probably added when route localizations was introduced). can we unregister it?

- twig generates form template cache, there is no symfony/form component installed, templating is disabled, form templates are somehow in FrameworkBundle?! can it be disabled?

- is there a way to unregister cli commands (per command or altogether)? for this app, and most of the microservices there is no need for commands other than `cache:clear` and `rm -rf var/cache` is sufficient

joanhey:
I recommend to read https://github.com/symfony/symfony/issues/27624 .
And ask for review there too.

joanhey:
### Missing extensions
About the missing extensions, check the `phpinfo()`. OPCache is running for all php frameworks.

All use the same extensions, and only add an extra `RUN` to the dockerfile when a framework need an extra one.
So we can take advantage from the docker cache.

### Changes in php.ini for opcache
I explained here https://github.com/symfony/symfony/issues/27624#issuecomment-450574677
I recommend that all use the same config for equality. And for future global changes to all php frameworks with an easy `search and replace`.

kaznovac:
@joanhey just reverted the changes to deploy config (will look into docker later but i'd prefer those docker changed to stay [i'll see to update dockerfiles for the rest of the frameworks in another PR])

please note that symfony (and ubiquity) has changed `;opcache.save_comments=1` in order for annotations to work other frameworks have `opcache.save_comments=0`

joanhey:
Yes, the frameworks that don't need to save the comments are disabled, to use less memory.
The annotations are a feature of the framework not from the php language.
I think it is the only option that change in the php.ini.

Use the phpinfo() for check the files used and memory in symfony. And for sure that use less than 2.000 files and 64Mb.

OPCache its installed by default when you install php.

frameworks/PHP/symfony/templates/bench/fortunes.html.twig
kaznovac: now there are 9 '\n' in an output of 1204 chars, i think this is acceptable

kaznovac:
@joanhey 
> OPCache its installed by default when you install php, in the latest versions.

this depends of php build configuration and distribution packaging - in most cases when you use package manager this is not true (some distributions provide meta-packages/groups for easier lamp stack installation)

with ubuntu (the image we are building container from) most php extensions are packed separately - i.e. opcache is provided by `php7.3-opcache` package, curl is in `php7.3-curl`, etc.

actually all of the frameworks running on ubuntu based container (few are based on other images [php, nginx/unit, .net core && aspnet core??] - i haven't checked opcache there) are **not** using opcache as it simply isn't installed.

joanhey:
Info about the package:
https://packages.ubuntu.com/disco/php7.2-fpm
depends from opcache, ...

With the odrej PPA for php 7.3
```
apt show php7.3-fpm
Package: php7.3-fpm
Version: 7.3.7-2+0~20190725.42+debian9~1.gbp848ca5
Priority: optional
Section: php
Source: php7.3
Maintainer: Debian PHP Maintainers <team+pkg-php@tracker.debian.org>
Installed-Size: 4.797 kB
Provides: php-fpm, phpapi-20180731
Depends: libmagic1, mime-support, php7.3-cli, php7.3-common (= 7.3.7-2+0~20190725.42+debian9~1.gbp848ca5), php7.3-json, php7.3-opcache, tzdata, ucf, libapparmor1 (>= 2.7.0~beta1+bzr1772), libargon2-1 (>= 0~20171227), libc6 (>= 2.15), libpcre2-8-0 (>= 10.33), libsodium23 (>= 1.0.14), libssl1.1 (>= 1.1.0), libsystemd0, libxml2 (>= 2.8.0), zlib1g (>= 1:1.1.4)
Suggests: php-pear
Homepage: http://www.php.net/
Download-Size: 1.384 kB
APT-Manual-Installed: yes
APT-Sources: https://packages.sury.org/php stretch/main amd64 Packages
Description: server-side, HTML-embedded scripting language (FPM-CGI binary)
 This package provides the Fast Process Manager interpreter that runs
 as a daemon and receives Fast/CGI requests.  Note that MOST Apache users
 probably want the libapache2-mod-php7.3 package.
 .
 The following extensions are built in: Core date filter hash libxml openssl
 pcre Reflection session sodium SPL standard zlib.
 .
 PHP (recursive acronym for PHP: Hypertext Preprocessor) is a widely-used
 open source general-purpose scripting language that is especially suited
 for web development and can be embedded into HTML.

```
Also depends from php7.3-opcache.
All the dockerfiles for php install opcache.

kaznovac:
@joanhey you are right, sorry my bad - opcache runs in containers :|

this is from symfony container:

```
root@3705563b0057:/symfony# apt show php7.3-fpm
Package: php7.3-fpm
Version: 7.3.7-2+ubuntu19.04.1+deb.sury.org+1
Priority: optional
Section: php
Source: php7.3
Maintainer: Debian PHP Maintainers <team+pkg-php@tracker.debian.org>
Installed-Size: 4868 kB
Provides: php-fpm, phpapi-20180731
Depends: libmagic1, mime-support, php7.3-cli, php7.3-common (= 7.3.7-2+ubuntu19.04.1+deb.sury.org+1), php7.3-json, php7.3-opcache, tzdata, ucf, libapparmor1 (>= 2.7.0~beta1+bzr1772), libargon2-1 (>= 0~20171227), libc6 (>= 2.29), libpcre2-8-0 (>= 10.33), libsodium23 (>= 1.0.14), libssl1.1 (>= 1.1.0), libsystemd0, libxml2 (>= 2.8.0), zlib1g (>= 1:1.1.4)
Suggests: php-pear
Download-Size: 1424 kB
APT-Manual-Installed: yes
APT-Sources: http://ppa.launchpad.net/ondrej/php/ubuntu disco/main amd64 Packages
Description: server-side, HTML-embedded scripting language (FPM-CGI binary)
```

i'm reluctant to remove requested-installed packages if we explicitly require them (there is docker cache on the other side)
if they are marked as automatically installed as a dependency - they can also automatically get uninstalled if they are no longer dependency - it feels safer this way and more transparent what of the environment we expect (if dev sees opcache it runs opcache, otherwise one has to remember/look into dependencies of installed packages)

php-common can be safely removed (it is shared between different php sapis
php-cli i'd leave as we occasionally run php commands from the command line (event if it is a dependency of php-fpm [and it doesn't make much sense to me why it is stated as dependency])

but i can remove it

joanhey:
Remember `phpinfo()` is your best friend. :+1:

kaznovac:
thanks joanhey

nbrady-techempower:
Let me know when this is ready for review, please

frameworks/PHP/symfony/symfony.dockerfile
```diff
@@ -5,30 +5,20 @@ ARG DEBIAN_FRONTEND=noninteractive
 RUN apt-get update -yqq && apt-get install -yqq software-properties-common > /dev/null
 RUN LC_ALL=C.UTF-8 add-apt-repository ppa:ondrej/php
 RUN apt-get update -yqq > /dev/null && \
-    apt-get install -yqq nginx git unzip php7.3 php7.3-common php7.3-cli php7.3-fpm php7.3-mysql  > /dev/null
-RUN apt-get install php7.3-xml  > /dev/null
+    apt-get install -yqq nginx git unzip php7.3 php7.3-common php7.3-cli php7.3-fpm php7.3-mysql > /dev/null
```
joanhey: If you delete only 1 space, like in this line.
The docker cache will not work. :(

frameworks/PHP/symfony/symfony.dockerfile
```diff
@@ -5,30 +5,20 @@ ARG DEBIAN_FRONTEND=noninteractive
...
+RUN apt-get install -yqq php7.3-curl php7.3-intl php7.3-mbstring php7.3-opcache php7.3-xml > /dev/null
```
joanhey: Some extensions are not necessary in this bench.
Using exactly :
`RUN apt-get install -yqq php7.3-mbstring php7.3-xml  > /dev/null`

It will also use the docker cache, as this line is used in others frameworks.

frameworks/PHP/symfony/symfony.dockerfile
joanhey: Only in the php directory, exist 47 dockerfiles.
Using cache each container is build in 2-6 seconds.
Without cache 2-4 minutes.

If only 10 dockerfiles break the docker cache, it's about 40 minutes to build the container vs 40 seconds.

frameworks/PHP/symfony/config/packages/doctrine.yaml
```diff
@@ -1,22 +1,17 @@
-parameters:
-    # Adds a fallback DATABASE_URL if the env var is not set.
-    # This allows you to run cache:warmup even if your
-    # environment variables are not available yet.
-    # You should not need to change this value.
-    env(DATABASE_URL): ''
-
 doctrine:
```
Tobion: The production ready config for doctrine seems to be missing so that the metadata is cached: https://github.com/symfony/recipes/blob/master/doctrine/doctrine-bundle/1.6/config/packages/prod/doctrine.yaml

frameworks/PHP/symfony/symfony.dockerfile
kaznovac: i was planning to create separate pr for the rest of the docker files, but that pr can include this one...

frameworks/PHP/symfony/config/packages/doctrine.yaml
joanhey: It is allowed the metadata cache.

frameworks/PHP/symfony/symfony.dockerfile
kaznovac: also curl made quite a speedup when installing composer packages on my machine, maybe keep it in a separate line of `ext-intl`?

frameworks/PHP/symfony/symfony.dockerfile
kaznovac: here are the other variations of `ext-mbstring` installation so there are also candidates for fixing for the docker cache

```
        cakephp.dockerfile  (1 usage found)
            8     apt-get install -yqq nginx git unzip php7.3 php7.3-common php7.3-cli php7.3-fpm php7.3-mysql php7.3-xml php7.3-mbstring php7.0-mcrypt php7.3-intl > /dev/null
        fuel.dockerfile  (1 usage found)
            8     apt-get install -yqq nginx git unzip php7.3 php7.3-common php7.3-cli php7.3-fpm php7.3-mysql php7.3-xml php7.3-mbstring php7.0-mcrypt  > /dev/null
        hhvm.dockerfile  (1 usage found)
            10     apt-get install -yqq hhvm nginx git unzip php7.3 php7.3-common php7.3-cli php7.3-fpm php7.3-mysql php7.3-xml php7.3-mbstring php7.3-mongodb  > /dev/null
        laravel.dockerfile  (1 usage found)
            9 RUN apt-get install -yqq php7.3-mbstring php7.3-xml  > /dev/null
        lumen.dockerfile  (1 usage found)
            9 RUN apt-get install -yqq php7.3-mbstring php7.3-xml  > /dev/null
        php-eloquent.dockerfile  (1 usage found)
            8     apt-get install -yqq nginx git unzip php7.3 php7.3-common php7.3-cli php7.3-fpm php7.3-mysql  php7.3-mbstring > /dev/null
        php-laravel-query-builder.dockerfile  (1 usage found)
            8     apt-get install -yqq nginx git unzip php7.3 php7.3-common php7.3-cli php7.3-fpm php7.3-mysql  php7.3-mbstring > /dev/null
        symfony-raw.dockerfile  (1 usage found)
            9 RUN apt-get install -yqq php7.3-mbstring php7.3-xml  > /dev/null
        symfony.dockerfile  (1 usage found)
            9 RUN apt-get install -yqq php7.3-mbstring php7.3-xml  > /dev/null
        yii2-raw.dockerfile  (1 usage found)
            8     apt-get install -yqq nginx git unzip php7.3 php7.3-common php7.3-cli php7.3-fpm php7.3-mysql php7.3-mbstring  > /dev/null
        yii2.dockerfile  (1 usage found)
            8     apt-get install -yqq nginx git unzip php7.3 php7.3-common php7.3-cli php7.3-fpm php7.3-mysql php7.3-mbstring  > /dev/null
        zend.dockerfile  (1 usage found)
            9 RUN apt-get install -yqq php7.3-xml php7.3-mbstring  > /dev/null
```

joanhey: I have a PR ready to send to optimize docker cache.
But I wait for your changes or I will break your PR.

kaznovac: can we create a list of desired php extensions and fix the rest of the dockerfiles (in another pr)?

frameworks/PHP/symfony/symfony.dockerfile
joanhey: Do you think that is it necessary the php-intl ?

This extension is only used for localization.
This bench is only in English and without any localization.

If the framework install the polifill without being used, for me that is an architectural fail.

kaznovac: also i've found interesting problem
i was tripped by docker cache as i couldn't just add 
`RUN apt-get install -yqq php7.3-intl > /dev/null` as the third line - package lists/remote repo has been changed (php was updated to 7.3.8 and this command tried to install intl 7.3.7 which was no longer available)

i've could add `apt-get update` before install but that just looks like a hack.

any suggestions?

frameworks/PHP/symfony/config/packages/doctrine.yaml
kaznovac: on this container it's request time about the same and sometimes even slower what this production ready config is enabled?!

frameworks/PHP/symfony/symfony.dockerfile
joanhey: I have a PR ready for these changes.
But if I send it, it will break your PR.

joanhey: The solution is to join the lines when it is posible.

kaznovac: i'll drop `intl` (it's used for localization of routes and domains)
for your pr please consider adding `curl` it really speeds up composer

don't worry about this pr, i'll gladly merge any changes.

frameworks/PHP/symfony/composer.json
```diff
@@ -1,49 +1,68 @@
...
+        "psr-4": {
+            "App\\Tests\\": "tests/"
+        }
+    },
+    "replace": {
+        "doctrine/doctrine-migrations-bundle": "*",
+        "paragonie/random_compat": "2.*",
+        "symfony/polyfill-ctype": "*",
+        "symfony/polyfill-iconv": "*",
+        "symfony/polyfill-mbstring": "*",
```
jderusse: "symfony/polyfill-intl-idn"

jderusse:
> application does not use localization but there is Symfony\Component\HttpKernel\EventListener\LocaleListener::setDefaultLocale Event listener registered - i can't find a config to unregister it (it was probably added when route localizations was introduced). can we unregister it?

You can create a compiler passe that remove services. See https://stackoverflow.com/a/32426218

> twig generates form template cache, there is no symfony/form component installed, templating is disabled, form templates are somehow in FrameworkBundle?! can it be disabled?

does it matter? those are just cache which are not used at runtime.

> is there a way to unregister cli commands (per command or altogether)? for this app, and most of the microservices there is no need for commands other than cache:clear and rm -rf var/cache is sufficient

same trick than EventListener. every thing is a service and can be removed. Beware some services (like the post install script in composer relly on commanded like cache clear).
But does it matter ? commands are not loaded when running a http request.

An other optimization is to switch to a `.0` version (see http://fabien.potencier.org/symfony4-performance.html why) like `4.0` or `5.0`

kaznovac:
thanks tobion, joanhey, jderusse

@nbrady-techempower this is ready for merge

frameworks/PHP/symfony/composer.json
```diff
@@ -1,49 +1,69 @@
...
+    "require": {
+        "php": "^7.3.7",
+        "ext-ctype": "*",
+        "ext-iconv": "*",
+        "ext-mbstring": "*",
+        "symfony/console": "4.3.*",
+        "symfony/dotenv": "4.3.*",
+        "symfony/flex": "^1.3.1",
+        "symfony/framework-bundle": "4.3.*",
+        "symfony/orm-pack": "^1.0",
```
Tobion: you don't need to require the orm-pack. you should require the real packages, so you don't need to replace the doctrine-migrations-bundle at all.

frameworks/PHP/symfony/config/packages/doctrine.yaml
Tobion: That would be strange and a sign that something is not correctly configured.

PR #4984, opened 2019-08-07
Toolset and Database Updates

nbrady-techempower:
Since we just completed Round 18, I thought this was a good time to get some updates in.

- All toolset dockerfiles (tfb, wrk, and databases) now use the same source: `buildpack-deps:bionic`. This will speed up time to first test for most tests as well as free up some space. Also cleaned up these files a bit so they only do one `apt-get update` saving more time.
- Updated MySQL from 5.7 to 8.0
- Updated MongoDB from 3.6 to 4.0
- Updated docker-py from 3.5 to 4.0
- Allows for increase in vagrant disksize

### Vagrant

If you're running multiple tests and need more disk space without wanting/needing to remove docker images, you can increase the size of `/dev/sda1`.

First install the necessary vagrant plugin `vagrant plugin install vagrant-disksize`
Then set the environment variable `TFB_DISKSIZE` to something like `20GB` before
running `vagrant up`


nbrady-techempower:
And I gave this the @msmith-techempower special: `+52 −114`

joanhey:
PHP5 can't connect to mysql 8.0.
**It's time to remove php5 as it has reached the end of life**. Everybody know that php7 is faster now.

Still there is a framework using it _Lithium_, I tried before to update to php7 without luck.
And checking the lithium github, it is not update for more than 2 years.
So perhaps it's better to delete it. 
No framework using php5 will have a fair benchmark.

joanhey:
Well php5 can use mysql 8.0, but it's better to remove it from the bench.
But also had the problems earlier versions of php7

> MySQL 8
> 
> When running a PHP version before 7.1.16, or PHP 7.2 before 7.2.4, set MySQL 8 Server's default password plugin to mysql_native_password or else you will see errors similar to The server requested authentication method unknown to the client [caching_sha2_password] even when caching_sha2_password is not used.

Unit use php7.0 for now (I will update to php7.3) and have this problem too.

And easy fix:
`alter user 'username'@'localhost' identified with mysql_native_password by 'password';`

But I will create a unit dockerfile using php7.3 the next week.

nbrady-techempower:
@joanhey I did add this line in the inital sql file:

`CREATE USER 'benchmarkdbuser'@'localhost' IDENTIFIED WITH mysql_native_password BY 'benchmarkdbpass';`

joanhey:
https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/569430975#L2762

I will try to investigate why exists the problem.
But it is faster update unit to php7.3

PR #4985, opened 2019-08-07
Bump Finagle and Finatra versions

kevinoliver:
Monthly release cycle:
* Finagle to 19.8.0
* Finatra to 19.8.0


kevinoliver:
Thanks @nbrady-techempower for the quick merge

Issue #4147, opened 2018-10-22
Lost history


ghost:
I'm looking for Lwan results in Round 10, which according to the blog was the top performer in many areas for that round.

I cannot find lwan anywhere in any of the charts of round 10 or any other rounds. You seem to have lost recorded history.

Also, Lwan is missing in new rounds. It was removed due to the code living outside the repo, right? Will it come back?

Currently Actix and Fasthttp top plaintext charts, which is kind of misleading given that in my own tests they score *way* below Lwan and other C servers.

nbrady-techempower:
Hey @alexhultman -

Thanks for opening this issue. lwan isn't in the current rounds because there isn't an implementation in our repo. As far as it being removed from previous rounds, I can only speak to not being able to review those implementations.

As far as it being misleading that there are tests that top the plaintext charts in your own testing, please feel free to create implementations for those tests. We can only test the implementations we have. I'd love to see the C servers that score way above! Also, we're totally open to lwan coming back with a test. I think we've pinged @lpereira a few times on that and last I heard he was working on the first major release. Would love to see it come back!

Also, there are fully transparent live results happening here: https://tfb-status.techempower.com/ so if you get some tests in, you'll be able to see how they stack up right away.

alexhultman:
What I mean with the lost history is that your own blog: https://www.techempower.com/blog/2015/04/21/framework-benchmarks-round-10/

Say

[...]

For Round 10, Lwan has taken the crown.

[...]

Yet it's not there in the round 10 charts. Why did the *old* charts change?

lpereira:
Yes, you're remembering it right, @nbrady-techempower.  It's entirely my fault that the implementation never made its way to the TWFB repository.

As far as reviewing the implementation: it's been always possible to review not only the Lwan source code, but the benchmark harness too.  The scripts (in this repo) that pulled in Lwan fetched a specific Git revision; due to the "tamper-resistant" nature of Git, the implementation was always reviewable and reproducible... maybe not as convenient as having it in the TWFB repository, but considering that the API is wasn't set in stone and no releases were made, it was the best I could do at the time.  However, as they say, "when in Rome...", so, yeah, maybe someday I'll publish the benchmark harness in this repo.  I got interested in Lwan again, at least. :)

(And, yes, I've been puzzled by the same thing, @alexhultman.  Learned that old results were removed when showing the page to a prospective employer... kind of unsettling, as I was very proud of that result.  No hard feelings, though.)

[...]

PR #4986, opened 2019-08-08
Delete php5 as reached EOL

joanhey:
Also can't connect to mysql 8.0

Deleted php5 and sllim php5.

**Hoping not to lose history if they solve #4147**


joanhey:
I don't know why fail unit, the changes don't affect it. 
Also I created another PR to update the processes number in unit without problems.

PR #4998, opened 2019-08-14
Marking core-rt as stripped

sebastienros:
Removes CoreRT from the results as this is an unsupported framework.


RUSshy:
Why censorship ? again, that's the proof someone at MS want to kill CoreRT

PR #5000, opened 2019-08-15
Do not censor the result

RUSshy:
I don't understand why some people want to kill CoreRT

Benchmark proved it is superior to Mono, don't dare tell me mono asp.net is supported by Microsoft, because that would be super sad based on the poor performance

image

https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/CSharp/aspnetcore-mono/benchmark_config.json#L8


msmith-techempower:
I am closing this pull request.

@sebastienros works for Microsoft and the core team, so I trust his judgment on what is and what is not a stripped implementation of the benchmarks.

Viewers can always show the stripped frameworks on the results website, but we do not show them by default.

PR #5005, opened 2019-08-17
GL, update version cleanup epoll

nathantippy:


nbrady-techempower:
@nathantippy There are no changes here?

PR #5006, opened 2019-08-18
beetlex-core update .net core to v3.0

IKende:
beetlex-core update .net core to v3.0


frameworks/CSharp/beetlex/beetlex-core.dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM mcr.microsoft.com/dotnet/core/sdk:2.2 AS build
+FROM mcr.microsoft.com/dotnet/core/sdk:3.0 AS build
 WORKDIR /app
 COPY PlatformBenchmarks .
 RUN dotnet publish -c Release -o out
 
-FROM mcr.microsoft.com/dotnet/core/aspnet:2.2 AS runtime
+FROM mcr.microsoft.com/dotnet/core/aspnet:3.0 AS runtime
 ENV COMPlus_ReadyToRun 0
```
sebastienros: This ENV can be removed. In 3.0 Tiered Compilation is on by default and this flag becomes unnecessary (as in it has no visible impact on perf)

PR #5020, opened 2019-08-25
add postgresql to rails

gregors:
Having done quite a few Rails related contracts over the last 5 years I've seen a lot more Postgres than Mysql. I also suggest that most new projects are using Postgres. 

I've added Postgres to the test suite. With the recent release of Rails 6 I'd  like to update a few things in here but this seems to be the most glaring omission. I followed the paradigm that I saw in the Sinatra Mysql vs Postgres project.

### DIFF
--- a/frameworks/Ruby/rails/benchmark_config.json
+++ b/frameworks/Ruby/rails/benchmark_config.json
@@ -23,6 +23,28 @@
       "notes": "",
       "versus": "rack-puma-mri"
     },
+    "postgresql": {
+      "json_url": "/hello_world/json",
+      "db_url": "/hello_world/db",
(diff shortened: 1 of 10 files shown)


frameworks/Ruby/rails/benchmark_config.json
```diff
@@ -23,6 +23,28 @@
       "notes": "",
       "versus": "rack-puma-mri"
     },
+    "postgresql": {
+      "json_url": "/hello_world/json",
```
nbrady-techempower: You can remove the json and plaintext properties here as they're already tested

gregors: done

PR #5022, opened 2019-08-26
Update ngx_php version

joanhey:
Ngx_php v0.0.19
Nginx v1.17.3

Update fortunes test.


joanhey:
After checking `phpinfo()` looks like ngx_php don't take the `ENV`s for now.
I will close this PR till @rryqszq4 find a solution, via `ENV` or with `getHostByName()`

Later I will send another PR only for update the version.

PR #5030, opened 2019-08-31
Update drogon

an-tao:
In this PR:

* Update drogon to the latest version;

* Optimize the performance of some tests;

* Modify the templates for the fortune test;

### DIFF
--- a/frameworks/C++/drogon/drogon_benchmark/plugins/SyncPlugin.cc
+++ b/frameworks/C++/drogon/drogon_benchmark/plugins/SyncPlugin.cc
@@ -1,0 +1,51 @@
+/**
+ *
+ *  SyncPlugin.cc
+ *
+ */
+
+#include "SyncPlugin.h"
+#include <drogon/drogon.h>
+
+using namespace drogon;
+
+void SyncPlugin::initAndStart(const Json::Value &config)
+{
+    /// Initialize and start the plugin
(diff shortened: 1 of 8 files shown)


frameworks/C++/drogon/drogon_benchmark/plugins/SyncPlugin.cc
```diff
@@ -0,0 +1,51 @@
...
+    drogon::app().registerSyncAdvice(
+        [](const HttpRequestPtr &req) -> HttpResponsePtr {
+            if (req->method() != Get)
+            {
+                return HttpResponsePtr{};
+            }
+            switch (req->path().length())
+            {
+                case 5:
+                    if (req->path() == "/json")
```
nbrady-techempower: Curious why you're checking the path string if you're already switching over the length? Why not one or the other?

an-tao: Consider the `/db` path of the single query test. When processing a request with this path, if you use a switch statement like this, there is no comparison of the string, otherwise there will be two string comparisons. I think this has a little benefit on performance.
If in the actual environment, many requests have the same prefix or there are more valid path strings to compare, I believe this will bring more benefits.
And if an invalid request with a length of 5 is processed here, it is necessary to perform a string check after switching over the length.

PR #5045, opened 2019-09-09
Symfony - persistent connections

jderusse:
Current TestSuite is failing because of number of connection to the database. 
The solution is to use "persistent connection", but this is not currently supported by Doctrine (see https://github.com/doctrine/dbal/issues/2315) because setting ATTR_PERSISTENT conflict with ATTR_STATEMENT_CLASS.

nb: This issue has been fixed in Doctrine 3 (not yet released nor installable)

This PR patch doctrine to support persistent connection in PHP/symfony

note: Drawback is: caching results won't works anymore. Which not an issue in our case because caching result is not permitted in the benchmark.

This PR also replaces calls to aliases by a call to the original method. Less function calls === more performances. Tell me if you prefer a dedicated PR

ping @raziel057 @kaznovac

### DIFF
--- a/frameworks/PHP/symfony/config/packages/doctrine.yaml
+++ b/frameworks/PHP/symfony/config/packages/doctrine.yaml
@@ -2,12 +2,13 @@ doctrine:
     dbal:
         # configure these for your database server
         driver: 'pdo_mysql'
-        server_version: '5.7'
+        server_version: '8.0'
         charset: utf8mb4
         default_table_options:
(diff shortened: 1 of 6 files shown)


raziel057:
Sounds good to me.

nbrady-techempower:
@jderusse Remove the `WIP` from title when this is ready to merge

jderusse:
Ready to merge @nbrady-techempower

PR #5079, opened 2019-09-21
Add apidaora python framework

dutradda:
Add [apidaora](https://dutradda.github.io/apidaora/) tests to python frameworks


frameworks/Python/apidaora/app.py
```diff
@@ -0,0 +1,202 @@
...
+        status_code=HTTPStatus.OK,
+        body=content
+    )
+
+
+async def database_updates(
+    req: MultipleDatabaseRequest
+) -> MultipleDatabaseResponse:
+    num_queries = get_num_queries(req.query['queries'])
+    updates = [(randint(1, 10000), randint(1, 10000)) for _ in range(num_queries)]
```
nbrady-techempower: The updates test requires that you query `num_queries` world objects from the database first, then update that world object's random number field. If I'm reading this correctly, you're just creating random world objects without querying them here.

dutradda: @nbrady-techempower I will update the test. Thanks for replying!

I copied the tests from fastapi, so I think the fastapi tests are wrong too.

dutradda: the code queries the results first

```
await stmt.fetchval(id_)
```

nbrady-techempower: Hmm, ok. I'm going to get back to you on this shortly and look at some of the other tests. There may have been a discussion about this already but I don't remember. I think other python tests will need to be fixed. The data should be read from the database, mapped to an object,  and then have its `randomNumber` field updated. In this implementation, `statement.fetchval` could do literally anything because the result is never read.

dutradda: OK! Now I understand your point @nbrady-techempower 

I will update my test to set the retrived values on objects, update its values with random numbers, then update the database values.

when my fixed tests is OK, I can updated others python tests too in other PR

nbrady-techempower: Thanks! That'd be great!

dutradda: @nbrady-techempower
I updated the tests, now I think it does what you describe.

nbrady-techempower:
Thanks @dutradda!

PR #5085, opened 2019-09-24
Update CoreRT to .NET Core 3.0 RTM

MichalStrehovsky:
Since we still use the mcr.microsoft.com/dotnet/core/sdk:2.2 container to build the thing, this has to be updated manually. The vanilla CoreCLR bits use 3.0 SDK and get the update through that.


MichalStrehovsky:
@msmith-techempower Seems like the CI finished - could you please merge?

PR #5088, opened 2019-09-24
Workerman async MySQL

joanhey:

### DIFF
--- a/frameworks/PHP/workerman/benchmark_config.json
+++ b/frameworks/PHP/workerman/benchmark_config.json
@@ -41,7 +41,7 @@
       "webserver": "None",
       "os": "Linux",
       "database_os": "Linux",
-      "display_name": "workerman postgres",
+      "display_name": "workerman-postgres",
       "notes": "",
       "versus": "php"
     },
@@ -60,7 +60,7 @@
       "webserver": "None",
       "os": "Linux",
       "database_os": "Linux",
-      "display_name": "workerman async db",
+      "display_name": "workerman-async-db",
       "notes": "",
       "versus": "php"
     }
--- a/frameworks/PHP/workerman/benchmark_config.json
+++ b/frameworks/PHP/workerman/benchmark_config.json
@@ -44,6 +44,25 @@
       "display_name": "workerman postgres",
       "notes": "",
       "versus": "php"
+    },
+    "async": {
+      "db_url": "/db",
+      "fortune_url": "/fortune",
+      "port": 8080,
+      "approach": "Realistic",
+      "classification": "Platform",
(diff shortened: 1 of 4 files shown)


frameworks/PHP/workerman/benchmark_config.json
```diff
@@ -44,6 +44,25 @@
       "display_name": "workerman postgres",
       "notes": "",
       "versus": "php"
+    },
+    "async": {
+      "db_url": "/db",
+      "fortune_url": "/fortune",
+      "port": 8080,
+      "approach": "Realistic",
+      "classification": "Platform",
```
msmith-techempower: This does not look correct to me - `platform` classification is for barebones infrastructure for servicing HTTP requests, but does not include a framework at all. Given what the source code looks like, I would say the classification of `workerman` is either `micro` or `fullstack` depending on what functionality is provided.

I apologize, it appears that our definitions for these have been removed from the website and not yet added to the wiki - I will get them in there.

frameworks/PHP/workerman/benchmark_config.json
```diff
@@ -44,6 +44,25 @@
...
+      "database": "MySQL",
+      "framework": "None",
+      "language": "PHP",
+      "flavor": "PHP7",
+      "orm": "Raw",
+      "platform": "workerman",
+      "webserver": "None",
+      "os": "Linux",
+      "database_os": "Linux",
+      "display_name": "workerman async db",
```
msmith-techempower: Every display name so far seems to conform to hyphens rather than spaces - `workerman-async-db`

frameworks/PHP/workerman/benchmark_config.json
joanhey: I think that is a platform. The code only is for servicing HTTP request.
But using php in cli mode, whitout any webserver, fastcgi, ...
It's like nodejs but in php.

msmith-techempower:
Whoops, merged thinking you were done \>_<

joanhey:
I only changed the config json, and don't need to pass the travis check.

frameworks/PHP/workerman/benchmark_config.json
joanhey: ready

frameworks/PHP/workerman/benchmark_config.json
joanhey: And the definitions are in : https://www.techempower.com/benchmarks/#section=motivation

msmith-techempower: Basically, yeah.

PR #5093, opened 2019-09-28
Double workers in workerman

joanhey:


joanhey:
Really good results:

With MySql

|Test | 1xcore | 2xcore | 
----- |  ------: | -------: |
Single query | 171,821 | 250,567 | 
Multiple queries | 11,527 | 15,482 | 
Fortunes | 150,597 | 216,584 |
Update | 2,509 | 2,733 | 281 |

And much better with PostreSQL.

PR #5095, opened 2019-09-29
Ubiquity update Swoole & Workerman

jcheron:
### Updated
- composer refs 
#### Swoole
- pool initialization in `onWorkerStart` event
- update Swoole to 4.4.7

#### Workerman
- Fix `Wrk` read errors
- Start Database on startup


frameworks/PHP/ubiquity/README.md
```diff
@@ -20,22 +21,24 @@ The tests are separated into 7 controllers:
...
 * [Ubiquity 2.3.*](https://ubiquity.kobject.net/)
 * [PHP Version 7.3.*](http://www.php.net/) with FPM and APC
 * [nginx 1.14](http://nginx.org/)
-* [Swoole](https://www.swoole.com/)
-* [Ubiquity-swoole](https://github.com/phpMv/ubiquity-swoole)
-* [Workerman](https://github.com/walkor/Workerman)
-* [Ubiquity-workerman](https://github.com/phpMv/ubiquity-workerman)
+* [Swoole](https://www.swoole.com/), [Ubiquity-swoole](https://github.com/phpMv/ubiquity-swoole)
+* [Workerman](https://github.com/walkor/Workerman), [Ubiquity-workerman](https://github.com/phpMv/ubiquity-workerman)
 * [MySQL 5.7](https://dev.mysql.com/)
```
joanhey: Now this benchmark use Mysql 8.0

frameworks/PHP/ubiquity/ubiquity-swoole.dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM php:7.3
 
-ENV SWOOLE_VERSION=4.4.5
+ENV SWOOLE_VERSION=4.4.7
```
joanhey: Why not use the pecl install, will be faster and updated. 
Like do hamlet:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/PHP/hamlet/hamlet-swoole.dockerfile#L3-L5

joanhey: Also you can install a specific version:
`pecl install swoole-4.4.7`

jcheron: Well done!
ok

frameworks/PHP/ubiquity/ubiquity-swoole.dockerfile
```diff
@@ -23,7 +19,7 @@ RUN deploy/run/install-composer.sh
 RUN apt-get update -yqq > /dev/null && \
     apt-get install -yqq git unzip > /dev/null
 
-RUN php composer.phar require phpmv/ubiquity-devtools:dev-master phpmv/ubiquity-swoole:dev-master --quiet
+RUN php composer.phar require phpmv/ubiquity-devtools:dev-techempower-benchmarks phpmv/ubiquity-swoole:dev-techempower-benchmarks --quiet
```
joanhey: I don't think, that have a branch only for the benchmark, is considered a production grade.
I understand if it is to do a test.

frameworks/PHP/ubiquity/ubiquity-workerman.dockerfile
```diff
@@ -27,7 +27,7 @@ RUN deploy/run/install-composer.sh
 RUN apt-get update -yqq > /dev/null && \
     apt-get install -yqq git unzip > /dev/null
 
-RUN php composer.phar require phpmv/ubiquity-devtools:dev-master phpmv/ubiquity-workerman:dev-master phpmv/ubiquity-mysqli:dev-master --quiet
+RUN php composer.phar require phpmv/ubiquity-devtools:dev-techempower-benchmarks phpmv/ubiquity-workerman:dev-techempower-benchmarks phpmv/ubiquity-mysqli:dev-techempower-benchmarks --quiet
```
joanhey: All test implementations should be production-grade for the benchmark.

frameworks/PHP/ubiquity/ubiquity-swoole.dockerfile
jcheron: It's just like a security airlock.
It allows me to prepare the next bench (on the `dev-techempower` branch) without taking any risk on `dev-master`.
There are times when the opposite is also likely: I make modifications on `master` that make Techempower tests no longer pass...
Once the test is passed, if it is a success, I make a merge between master and techempower branchs.
The Techempower branch is not intended to be very far from the master branch...

Once the test is passed, if it is a success, I make a merge between master and techempower

joanhey:
Hi @jcheron 
You have another problem, that is easy to fix:
https://tfb-status.techempower.com/unzip/results.2019-09-27-18-21-14-704.zip/results/20190923214828/ubiquity-workerman/run/ubiquity-workerman.log

You need to change the php.ini to use more memory.

jcheron:
Thnak you @joanhey 
> You need to change the php.ini to use more memory.
```Allowed memory size of 134217728 bytes exhausted (tried to allocate 69632 bytes)```
These values are already huge.
I think I have a memory leak problem in my Mysqli classes.

joanhey:
Yes, it is a lot of memory for 28 workers.

But I think that is a problem with the PHP bulit-in Mysqli class.
I needed to stop Mysqli persistent connections in CodeIgniter(use by default Mysqli) because have a memory leak, that breaks the continuos benchmark server. 

We need to investigate more the problem with Mysqli and create a bug in php.net.

jcheron:
I can try to go to 512 in the meantime.
`memory_limit = 512M`

joanhey:
A lot of errors in last run with swoole.
https://tfb-status.techempower.com/unzip/results.2019-10-05-23-14-37-606.zip/results/20191002030153/ubiquity-swoole/run/ubiquity-swoole.log

joanhey:
Also try to return errors, by default Swoole don't return it.
See #5121.
So you could check the errors in the bench or app.

PR #5096, opened 2019-09-29
Update Swoole to 4.4.7

jcheron:
- Update Swoole to 4.4.7
I'm not doing it for Laravel and Lumen, but they're still in 4.3

### DIFF
--- a/frameworks/PHP/swoole/swoole.dockerfile
+++ b/frameworks/PHP/swoole/swoole.dockerfile
@@ -1,6 +1,6 @@
 FROM php:7.3
 
-ENV SWOOLE_VERSION=4.4.5
+ENV SWOOLE_VERSION=4.4.7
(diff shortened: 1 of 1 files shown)


frameworks/PHP/swoole/swoole.dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM php:7.3
 
-ENV SWOOLE_VERSION=4.4.5
+ENV SWOOLE_VERSION=4.4.7
```
joanhey: Why not use the pecl install, will be faster and updated. 
Like do hamlet:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/PHP/hamlet/hamlet-swoole.dockerfile#L3-L5

PR #5104, opened 2019-10-01
Ngx_php move to own framework dir

joanhey:
Ngx_php was limited to the plain php tests.

So could grow alone. Also the travis tests for it, will be faster.
Now use only the `app.php` file.

I have kept the name, hoping not to lose the history. Bug #4147

@rryqszq4 I created the php-async.php, so you can add the Mysql async pool variant.


rryqszq4:
@joanhey HI，
The message was received and it worked very well. I will spend some time filling it.

PR #5105, opened 2019-10-01
Ubiquity-workerman: Add memory_limit to php.ini

jcheron:
see https://github.com/TechEmpower/FrameworkBenchmarks/pull/5095#issuecomment-536919785

### DIFF
--- a/frameworks/PHP/ubiquity/deploy/conf/php-async.ini
+++ b/frameworks/PHP/ubiquity/deploy/conf/php-async.ini
@@ -6,3 +6,5 @@ opcache.enable_file_override=1
 opcache.huge_code_pages=1
 
 mysqlnd.collect_statistics = Off
+
+memory_limit = 512M


joanhey:
The results are impresive, if the test is correct.
We need to update the plain workerman to reflect it.

joanhey:
I think the code have anything wrong.
This result is impossible;
https://www.techempower.com/benchmarks/#section=test&runid=42f65a64-69b2-400d-b24e-20ecec9848bc&hw=ph&test=query

jcheron:
I have the same doubts as you do...
The tests never check the content of the request or the content-length when there is concurrent access?

joanhey:
The travis tests check all that.
But can't check any cache inside the code.
Perhaps you left any cache in the code.

joanhey:
Anyway It will be good to investigate from where come that gain.
Even if it is from a cache.

jcheron:
I'll check for the cache, but I don't think that's it. This should be explicitly stated in the configuration.

joanhey:
Try to change to use PDO or pgsql instead of mysqli.
And we will check what happens.

joanhey:
I will try in the plain workerman to use mysqli.

jcheron:
Yes, and I would like workerman-mysqli to be much faster than the current Ubiquity-workerman

PR #5108, opened 2019-10-02
Use postgres connection pooling for lisp frameworks

gwbrown:
The Lisp frameworks (woo, Ningle, and ninglex) do not currently use postmodern's built-in connection pooling, which leads to benchmark performance well below what can be achieved by the frameworks themselves.  This commit changes the test applications to use connection pooling.

### DIFF
--- a/frameworks/Lisp/ningle/ningle.ros
+++ b/frameworks/Lisp/ningle/ningle.ros
@@ -166,7 +166,7 @@ exec ros -Q -- $0 "$@"
     (apply #'woo:run
       (lambda (env)
         ;; preprocessing
-        (let ((res (postmodern:with-connection '("hello_world" "benchmarkdbuser" "benchmarkdbpass" "tfb-database")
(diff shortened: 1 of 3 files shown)


gwbrown:
Some notes on just how much this improves performance: I ran the Woo benchmark on a 12-core server I have locally with and without the connection pooling.  Here are the (unscientific, load and application on same box, so take with a grain of salt) results from the `fortune` test using the highest level of load concurrency: 

measurement | without pooling | with pooling
-------------- | --------------- | -----------
latencyAvg      | 292.05ms         | 27.19ms
latencyMax     | 623.36ms         | 241.10ms
totalRequests | 25,741               | 274,447
5xx                  | 19,279              | 0

Moral of the story: Connection pooling can be really important.

---

It's also worth noting that the Racket test application also does not use connection pooling, but in my local tests, using Racket's connection pooling gives a slight performance boost at lower levels of load concurrency, but makes performance *significantly worse* at the higher end - possibly because Racket's servlet framework is single-threaded.

PR #5109, opened 2019-10-02
Update armeria to latest version, use more idiomatic armeria code, an…

anuraaga:
…d ensure DB calls are not on the event loop.

- For server level headers, it is appropriate to use a server decorator to add the headers to all responses instead of doing the logic at every call site

- For JSON responses, armeria's annotated services will automatically serialize if the `@ProducesJson` annotation is applied which is the idiomatic way of returning JSON from an annotated service

- Services are run on a non-blocking event loop. Because database calls are blocking, such calls must be moved to the `blockingExecutor` or throughput will be hurt significantly by blocking the event loop

- Slightly optimizes returning the mustache template

@trustin @minwoox @ikhoon can you review too?


frameworks/Java/armeria/src/main/java/hello/App.java
```diff
@@ -1,20 +1,32 @@
...
     sb.http(8080)
       .annotatedService("/", new HelloService())
       .annotatedService("/", new PostgresDbService())
-      .annotatedService("/", new PostgresFortunesService());
+      .annotatedService("/", new PostgresFortunesService())
+      .decorator((delegate, ctx, req) -> {
+        ctx.addAdditionalResponseHeader(HttpHeaderNames.SERVER, "armeria");
+        ctx.addAdditionalResponseHeader(HttpHeaderNames.DATE,
+                                        DateTimeFormatter.RFC_1123_DATE_TIME.format(
+                                                ZonedDateTime.now(ZoneOffset.UTC)));
```
trustin: I wonder we should improve Armeria to add these headers by default.

frameworks/Java/armeria/src/main/java/hello/services/PostgresDbService.java
```diff
@@ -88,71 +74,87 @@ private int getSanitizedCount(String count) {
...
+            statement.setInt(1, number);
+
+            try (final ResultSet resultSet = statement.executeQuery()) {
+              resultSet.next();
+              return new World(resultSet.getInt(1), resultSet.getInt(2));
+            }
+          } catch (SQLException e) {
+            throw new IllegalStateException("Database error", e);
+          }
+        }, ServiceRequestContext.current().blockingTaskExecutor());
```
trustin: Could make `ServiceRequestContext` injected by Armeria?

frameworks/Java/armeria/src/main/java/hello/App.java
anuraaga: Yup that sounds good to me.

frameworks/Java/armeria/src/main/java/hello/services/PostgresFortunesService.java
```diff
@@ -44,43 +48,51 @@ public PostgresFortunesService() {
...
+          }
+          return fortunes;
+        }, ServiceRequestContext.current().blockingTaskExecutor()
+    );
   }
 
-  private String buildMustacheTemplate(List<Fortune> fortunes)
-      throws IOException {
+  private byte[] buildMustacheTemplate(List<Fortune> fortunes) {
     Mustache mustache = mustacheFactory.compile("fortunes.mustache");
```
trustin: Shouldn't we compile the template only once?

anuraaga: Mustache apparently automatically caches so this should be ok - it was recommended in the initial submission. https://github.com/TechEmpower/FrameworkBenchmarks/pull/4041#discussion_r217892498

trustin: Ah OK! Thanks for education.

trustin:
I wonder if making use of the new fluent route API makes any difference as well.

anuraaga:
@trustin Do you mean implementing `Service` instead of annotated service and mounting with the route API? I guess it could be a little faster, but for these sort of methods I guess it's most idiomatic to use annotated services and think we should stick with it - I think annotated service generates the routes in the background anyways.

frameworks/Java/armeria/src/main/java/hello/App.java
trustin: https://github.com/line/armeria/issues/2137

frameworks/Java/armeria/src/main/java/hello/services/PostgresDbService.java
```diff
@@ -36,37 +32,26 @@ public PostgresDbService() {
...
-    	HttpHeadersHelper.getHttpHeader(MediaType.JSON),
-        HttpData.of(MAPPER.writeValueAsBytes(getWorld(getRandomNumber())))
-        );
+  @ProducesJson
+  public CompletableFuture<World> db(ServiceRequestContext ctx) throws Exception {
+    return getWorld(getRandomNumber(), ctx);
   }
 
   // need to use regex as /queries/{count} doesn't work when count is null
   @Get("regex:^/queries/(?<count>.*)$")
```
trustin: How about splitting this into `@Get("/queries/")` and `@Get("/queries/{count}")` ?

ikhoon:
Great job! 👍🙏

trustin:
No idea why the build has failed. The only difference I can see is the `content-type` header which was previously `application/json`, which is now `application/json; charset=utf-8`.

msmith-techempower:
The only thing I see in the Travis-CI logs are the following:
```
armeria: SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
armeria: SLF4J: Defaulting to no-operation (NOP) logger implementation
armeria: SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
armeria: ERROR: Framework is not accepting requests from client machine
```
I restarted the Travis-CI job (just your test) to see if it was just a fluke.

msmith-techempower:
Second run attempt errored out the same way.

anuraaga:
Was able to repro the failure locally - seem to have fixed it by switching to spring-boot-maven-plugin from assembly plugin for creating the fatjar, I guess it handles runtime dependencies more reliably.

Issue #5107, opened 2019-10-01
Can't build the Actix docker files


sebastienros:
Since 9/26 I am getting these errors when building the actix-raw docker.
And I can see that the current build started after that date is displaying errors too.

```
...
[11:33:02.305]    Compiling yarte_config v0.3.3
[11:33:06.594]    Compiling state_machine_future v0.2.0
[11:33:06.682]    Compiling tokio-postgres v0.4.0-rc.2 (https://github.com/fafhrd91/rust-postgres.git#eb54cf28)
[11:33:12.669]    Compiling actix-web v1.0.8
[11:33:12.669]    Compiling actix v0.8.3
[11:33:13.510]    Compiling askama_derive v0.8.0
[11:33:13.886]    Compiling onig_sys v69.2.0
[11:33:15.638] error: failed to run custom build command for `onig_sys v69.2.0`
[11:33:15.638]
[11:33:15.638] Caused by:
[11:33:15.638]   process didn't exit successfully: `/actix/target/release/build/onig_sys-d412d643b50ef4eb/build-script-build` (exit code: 101)
[11:33:15.638] --- stdout
[11:33:15.638] cargo:warning=couldn't execute `llvm-config --prefix` (error: No such file or directory (os error 2))
[11:33:15.638] cargo:warning=set the LLVM_CONFIG_PATH environment variable to a valid `llvm-config` executable
[11:33:15.638]
[11:33:15.638] --- stderr
[11:33:15.638] thread 'main' panicked at 'Unable to find libclang: "couldn\'t find any valid shared libraries matching: [\'libclang.so\', \'libclang-*.so\', \'libclang.so.*\'], set the `LIBCLANG_PATH` environment variable to a path where one of these files can be found (invalid: [])"', src/libcore/result.rs:999:5
[11:33:15.638] note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
[11:33:15.638]
[11:33:15.638] warning: build failed, waiting for other jobs to finish...
[11:33:37.880] error: build failed
[11:33:38.196] The command '/bin/sh -c RUSTFLAGS="-C target-cpu=native" cargo build --release' returned a non-zero code: 101
[11:33:38.197]
[11:33:38.197]
```

msmith-techempower:
I am looking into this and will report back when I have more.

msmith-techempower:
[...]

PR #5110, opened 2019-10-02
Fix Actix bindgen issue

msmith-techempower:
Addresses #5107


msmith-techempower:
Well, \^ is a good sign.

PR #5116, opened 2019-10-05
Add swoole without async db pool

joanhey:
The actual swoole test is using async db pool connection.

So we can compare without it. Using the normal PDO from PHP.

The rest of frameworks can compare and learn.
Also we will know the envent loop performance.

A benchmark it is NOT a competition. It is a tool that help to create better and faster code.

### DIFF
--- a/frameworks/PHP/swoole/swoole-server-noasync.php
+++ b/frameworks/PHP/swoole/swoole-server-noasync.php
@@ -1,0 +1,155 @@
+<?php
+
+use Swoole\Http\Request;
+use Swoole\Http\Response;
+
+$server = new swoole_http_server('0.0.0.0', 8080, SWOOLE_BASE);
+$server->set([
+    'worker_num' => NUMCORES
+]);
+
+$pdo = new PDO("mysql:host=tfb-database;dbname=hello_world", "benchmarkdbuser", "benchmarkdbpass", [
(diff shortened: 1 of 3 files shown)


frameworks/PHP/swoole/swoole-server-noasync.php
```diff
@@ -0,0 +1,155 @@
...
+
+use Swoole\Http\Request;
+use Swoole\Http\Response;
+
+$server = new swoole_http_server('0.0.0.0', 8080, SWOOLE_BASE);
+$server->set([
+    'worker_num' => NUMCORES
+]);
+
+$pdo = new PDO("mysql:host=tfb-database;dbname=hello_world", "benchmarkdbuser", "benchmarkdbpass", [
```
jcheron: It is necessary to create a pdo instance for each Worker, otherwise you only have one PDO instance, which will not support concurrent access.
```php
$server->on('workerStart',function() {
	// create pdo connection
});
```

joanhey: I know, but I can create the PDO instance there directly. So first I tried so and found 2 problems.

I tried to maintain the original anonymous functions, but it's no posible if we create the $db in the workerstart.
So I will use normal functions.

Issue #5118, opened 2019-10-06
Ubiquity-workerman (incredible) results in last run


jcheron:
I wish that were true,
The `Ubiquity-workerman` couple would be the fastest on the Multiple queries test (see [42f65a64-69b2-400d-b24e-20ecec9848bc](https://www.techempower.com/benchmarks/#section=test&runid=42f65a64-69b2-400d-b24e-20ecec9848bc&hw=ph&test=query)).
Ubiquity is fast, and so is Workerman. But it seems more than unlikely that the workerman-raw version will be slower than the one associated with a Framework (even if Ubiquity uses Mysqli, and workerman PDO), and this is not the only point of comparison that can raise suspicion...

I didn't see anything suspicious in the [query raw file](https://tfb-status.techempower.com/unzip/results.2019-10-05-23-14-37-606.zip/results/20191002030153/ubiquity-workerman/query/raw.txt).
The [logs](https://tfb-status.techempower.com/unzip/results.2019-10-05-23-14-37-606.zip/results/20191002030153/ubiquity-workerman/run/ubiquity-workerman.log) just indicate an error that occurred on 1 worker

I'll try to investigate this on my side...

joanhey:
Well Kumbia-Workerman will be faster :) jeje

joanhey:
We will check #5122 results in the next run in the plain Workerma with Mysqli.
If the results are good, will benefit any framework using Mysqli.

jcheron:
So there was a problem in the **Ubiquity-workerman** tests, not revealed by the initial test phase, nor by the logs in the last run.
A good part of the Db queries were not executed, as soon as the concurrency level increased.

To confirm that, I made a script checking that the number of Db queries executed, and that the number of rows read on the server side corresponds to the executed code, at a given concurrency level.
In principle:
- I store the initial values on the DB side.
- I clear the cache of requests.
- I run a test (ab or wrk) with `xx` requests and a certain level of concurrent access.
- I measure the values again and compare the difference with an expected `yy` number.

For Mysql, my script looks like this:
```php
class QueriesCheck {
	/**
	 * @var Database
	 */
	private $db;
	public function __construct($dbOffset='default'){
		$this->db=Database::start($dbOffset);
	}
	public function run($url,$concurrency=512,$count=15000,$expectedQueries=15000,$expectedRows=15000):void{
		$queries=$this->getQueries();
		$rows=$this->getRows();
		$this->resetCache();
		
[...]

PR #5122, opened 2019-10-07
Add Workerman with Mysqli

joanhey:
To check if Mysqli is really much faster than PDO.

For now use a worker per core.
And raise memory limit to 512Mb as Mysqli looks like use more memory.


joanhey:
I will revert this PR.
Mysqli it isn't faster, see https://github.com/TechEmpower/FrameworkBenchmarks/issues/5118#issuecomment-539438748

PR #5126, opened 2019-10-07
Fix Ubiquity-workerman pb

jcheron:
### Fixed
- `Ubiquity-workerman` db connection pb with `Mysqli` see https://github.com/TechEmpower/FrameworkBenchmarks/issues/5118 , moving the connection in onWorkerStart event
- Doubles the number of workers


joanhey:
Well Kumbia-Workerman will be faster  :) jeje

joanhey:
We will check #5122 results in the next run in the plain Workerma with Mysqli.

joanhey:
Will be good to check Mysqli with double number of workers.
But I think that Mysqli is using asynchronous connections and it will benefit less from more workers.

We will need to wait to the next run.

PR #5128, opened 2019-10-08
Update APIDaora Python Framework tests

dutradda:
Create a new test suite called `apidaora-core` and update the other test `apidaora`.


nbrady-techempower:
@dutradda We don't want to create new directories for different flavors of the same framework. This should remain under the `Python/apidaora` and added as a `core` test in that benchmark config.

dutradda:
@nateberkopec 
I updated the tests

PR #5146, opened 2019-10-11
Use compiled queries

murbanowicz:
EF Core 3.0 supports compiled queries.

https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.ef.compileasyncquery?view=efcore-3.0

### DIFF
--- a/frameworks/CSharp/aspnetcore/Benchmarks/Data/EfDb.cs
+++ b/frameworks/CSharp/aspnetcore/Benchmarks/Data/EfDb.cs
@@ -30,11 +30,8 @@ namespace Benchmarks.Data
         public Task<World> LoadSingleQueryRow()
         {
             var id = _random.Next(1, 10001);
-
-            // TODO: compiled queries are not supported in EF 3.0-preview7
-            // return _firstWorldQuery(_dbContext, id);
             
-            return _dbContext.World.FirstAsync(w => w.Id == id);
+            return _firstWorldQuery(_dbContext, id);
         }
 
         public async Task<World[]> LoadMultipleQueriesRows(int count)
@@ -44,11 +41,8 @@ namespace Benchmarks.Data
             for (var i = 0; i < count; i++)
             {
                 var id = _random.Next(1, 10001);
-
-                // TODO: compiled queries are not supported in EF 3.0-preview7
-                // result[i] = await _firstWorldQuery(_dbContext, id);
-
-                result[i] = await _dbContext.World.FirstAsync(w => w.Id == id);
+                
+                result[i] = await _firstWorldQuery(_dbContext, id);
             }
 
             return result;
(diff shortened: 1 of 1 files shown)


frameworks/CSharp/aspnetcore/Benchmarks/Data/EfDb.cs
```diff
@@ -86,11 +77,9 @@ public async Task<World[]> LoadMultipleUpdatesRows(int count)
         public async Task<List<Fortune>> LoadFortunesRows()
         {
             var result = await _dbContext.Fortune.ToListAsync();
-
-            // TODO: compiled queries are not supported in EF 3.0-preview7
-            // await foreach (var element in _fortunesQuery(_dbContext))
-
-            result.Add(new Fortune { Message = "Additional fortune added at request time." });
+            
+            await foreach (var element in _fortunesQuery(_dbContext))
```
sebastienros: This block is wrong. I think the comment I left mislead you.

murbanowicz: can you elaborate

sebastienros:
This code is adding a new fortune for each existing one. We need only one added. So the whole foreach is unnecessary. Can you check how the code was before I removye the compiled queries?

frameworks/CSharp/aspnetcore/Benchmarks/Data/EfDb.cs
sebastienros: This code is adding a new fortune for each existing one. We need only one added. So the whole foreach is unnecessary. Can you check how the code was before I removye the compiled queries?

sebastienros: The solution is this:

```cs
public async Task<List<Fortune>> LoadFortunesRows()
{
  var result = await _fortunesQuery(_dbContext).ToListAsync();
  result.Add(new Fortune { Message = "Additional fortune added at request time." });
  result.Sort();
  
  return result;
}
```

murbanowicz: It has to be something like:

```
            var result = new List<Fortune>();
            var fortunes = _fortunesQuery(_dbContext).GetAsyncEnumerator();
            while (fortunes.Current != null)
            {
                result.Add(fortunes.Current);
                await fortunes.MoveNextAsync();
            }
            result.Add(new Fortune { Message = "Additional fortune added at request time." });
            result.Sort();
  
            return result;
```

because it can only give AsyncEnumerator.
Am I right?

nbrady-techempower:
closing this for now as it looks like it's awaiting feedback and has some conflicts.

PR #5153, opened 2019-10-14
Phoenix performance

tomciopp:
@michalmuskala @josevalim Tested these changes on my local machine and saw a small bump in performance. 

Updated elixir to v1.9.1 which is self explanatory. Changed the SQL queries so that they can be run in parallel to show off the concurrency features of elixir.

We reduce the pool_size, since our storage is an SSD and we are running on 14 cores. More threads only perform better when there is blocking which is common with HDDs due to seeking with the spindle/platter, this doesn't exist with an SSD. Since we aren't blocking on I/O, we can't get more work done unless we add more cores to the processor. Therefore any open connections above the number of cores only adds to overhead and diminished performance. So we tune it to 14 which is what the TechEmpower people are using with their hardware.

### DIFF
--- a/frameworks/Elixir/phoenix/web/controllers/page_controller.ex
+++ b/frameworks/Elixir/phoenix/web/controllers/page_controller.ex
@@ -1,41 +1,40 @@
 defmodule Hello.PageController do
+  alias Hello.{Fortune, World}
+
   use Hello.Web, :controller
-  alias Hello.World
-  alias Hello.Fortune
+
+  @json "application/json"
+  @plain "text/plain"
 
   def index(conn, _params) do
     conn
-    |> put_resp_content_type("application/json", nil)
+    |> put_resp_content_type(@json, nil)
     |> send_resp(200, Jason.encode_to_iodata!(%{"TE Benchmarks\n" => "Started"}))
   end
 
   # avoid namespace collision
   def _json(conn, _params) do
     conn
-    |> put_resp_content_type("application/json", nil)
+    |> put_resp_content_type(@json, nil)
     |> send_resp(200, Jason.encode_to_iodata!(%{"message" => "Hello, world!"}))
   end
 
   def db(conn, _params) do
     conn
-    |> put_resp_content_type("application/json", nil)
+    |> put_resp_content_type(@json, nil)
     |> send_resp(200, Jason.encode_to_iodata!(Repo.get(World, :rand.uniform(10000))))
   end
 
   def queries(conn, params) do
-    q = try do
-      case String.to_integer(params["queries"]) do
-        x when x < 1    -> 1
-        x when x > 500  -> 500
-        x               -> x
-      end
-    rescue
-      ArgumentError -> 1
-    end
+    json =
+      params["queries"]
+      |> query_range()
+      |> parallel(fn _ -> Repo.get(World, :rand.uniform(10000)) end)
+      |> Jason.encode_to_iodata!()
 
     conn
-    |> put_resp_content_type("application/json", nil)
-    |> send_resp(200, Jason.encode_to_iodata!(for _ <- 1..q, do: Repo.get(World, :rand.uniform(10000))))
+    |> put_resp_content_type(@json, nil)
+    |> send_resp(200, json)
   end
 
   def fortunes(conn, _params) do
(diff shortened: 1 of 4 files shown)


frameworks/Elixir/phoenix/web/controllers/page_controller.ex
```diff
@@ -50,32 +49,41 @@ defmodule Hello.PageController do
...
-      num = :rand.uniform(10000)
-      w = Repo.get(World, id)
-      changeset = Ecto.Changeset.change(w, randomnumber: num)
-      Repo.update!(changeset)
-    end
+    json =
+      params["queries"]
+      |> query_range()
+      |> parallel(fn _ -> Repo.get(World, :rand.uniform(10000)) end)
+      |> parallel(fn w ->
```
josevalim: We should probably merge those two parallel into a single one You can also wrap both queries into a Repo.checkout in Ecto 3.0+.

tomciopp: I was honestly considering an upsert, but I didn't know if that was in the spirit of the test. I guess since it is a feature of Ecto it _should_ be ok.  Will refactor it in a little bit.

nbrady-techempower: FYI, per the test requirements for updates, you must query the world from the database, update the random number on that object, and then update the database.

tomciopp: Ok, won't change.

jtwebman:
@tomciopp Are you sure the core and connection pool per core makes sense? Or is this because the db only has 14 cores so it could be doing more than 14 at a time? Seems like 2 or 3 per core would be better. Did you guys benchmark both and find this was better?

tomciopp:
@jtwebman The logic is sound, I will point you to the following links for more details.

https://github.com/brettwooldridge/HikariCP/wiki/About-Pool-Sizing
https://www.youtube.com/watch?v=xNDnVOCdvQ0&feature=emb_title

I didn't do extensive testing to optimize the exact pool_size, but I did do testing on my own hardware and saw some decent improvements. A pool_size of 15 or 16 might be optimal since there is a slight I/O wait time with an SSD but it is unlikely to make a huge difference in performance.

jtwebman:
@tomciopp thank you for sharing!

josevalim:
I really don’t think you want to limit the pool size to 14. If you limit it to 14, then you receive 14 requests, one per core, and then they all wait in the DB, the next will block on the pool. This would be extremely common.

On the other hand, 256 is way too many, and it will lead to poor memory usage, por cache hits, etc. The first link suggests to have at least twice of connections per the number of cores, so I would go with 30. Unfortunately I did not see the pool_soze change when I reviewed the PR, otherwise I would have dropped a comment before.

tomciopp:
@josevalim Reread the last sentence in that paragraph.

```
A formula which has held up pretty well across a lot of benchmarks for years is
that for optimal throughput the number of active connections should be somewhere
near ((core_count * 2) + effective_spindle_count). Core count should not include
HT threads, even if hyperthreading is enabled. Effective spindle count is zero if
the active data set is fully cached, and approaches the actual number of spindles
as the cache hit rate falls. ...

There hasn't been any analysis so far regarding how well the formula works with SSDs.
```

We only have the opportunity to do work when blocking creates opportunities for executing. Since we are running on an SSD we have a negligible amount of time where we are blocking on I/O compared to a HDD. 14 may be too small of a number but I'm willing to bet that the optimal size is not far from there. 
 
Rather than guessing what an optimal `pool_size` is, can we actually get some data?

tehprofessor:
So... I think it's pretty safe to safe this has empirically slowed down the Phoenix results as you can see with this commit's visualization: https://www.techempower.com/benchmarks/#section=test&runid=19b49b58-2cc2-4bb0-a224-90e242cb31da&hw=ph&test=query (and others after it which have less errors are still slower despite the other changes like a bunch in elixir version).

In general, the network overhead on both sides, is going to cause issues if the pool size is too small or too big. Right now it's too small, and needs to be **at least doubled** as the connection to postgres _locally_ adds 60% of the request time (there are a lot of reason why this benchmark and the pathological cases talked about in the linked material are likely different, the size and complexity of the dataset being by far the most significant).

An easy way to test this, with the given data (I grabbed the sql files in the other part of this repository):

Using the `/fortune` endpoint, I ran a benchmark with _one_ connection `ab -n 5000 -c 1 http://localhost:8080/fortune` hitting psql as it is. I then ran the same `ab` settings against the endpoint using a module attribute (which is a copy pasta of the return value from `Repo.all`).

**With the database calls**, I would average `~1450 reqs/s`.
**Without the database**, I would average `~2300 reqs/s`.

As, there is damn near zero chance Ecto and Postgrex are to blame for that much overhead (which is several magnitudes more when done over a network), it's safe to say that Phoenix is _waiting_ at least 60% of the time on `psql` for this data-set in the best case scenario (a local interface). _And_ postgres is likely waiting on Phoenix, but I didn't bother measuring that... 

I think @josevalim is probably pretty damn spot on with his suggestion of `30`, depending on the network conditions and perhaps the other queries, I'd think it _might_ be able to go as high as `60` but that's just my own, likely wrong, two-cents...

Also, using `Task.stream_async` would cleanup the call `parallel`:

```
defp parallel(collection, func)
  Task.async_stream(collection, func)
  |> Enum.map(fn {:ok, value} -> value end)
end
```

and make it easier to tune the amount of asynchronous work being done, because right now if I understand it correctly it (Task) will default to the number of schedulers available, even if that's not optimal for the given task because of other bottlenecks (like I/O).

josevalim:
> We only have the opportunity to do work when blocking creates opportunities for executing. Since we are running on an SSD we have a negligible amount of time where we are blocking on I/O compared to a HDD. 

I think you are underestimating the time involved in writing this information on the socket and the time waiting spent on Postgres to process it, encode results, write it back on the socket and for us to receive it. All of this is included in the query time - it is not only the time spent on SSD.

Also, because you are doing Repo.checkout, it means that time spent inside Elixir is now used to block the connection too.

Taking into consideration the results @tehprofessor mentioned above, I would suggest bumping to 30-40, and continue watching out the results.

tomciopp:
@josevalim @tehprofessor I don't think it's quite clear that we've been slowed down.

Let's take another look at another run where there were no errors reported, and compare it to round 18 which is the latest report published on the site.

https://www.techempower.com/benchmarks/#section=test&runid=7f65c127-fad2-4a88-a6cb-5333c68362ef&hw=ph&test=update

I'm going to use actix-core for a baseline comp.

```
                      |   Fortunes  | Multi  | Single  | Update  |
actix-core (7f65c127) |     634,754 | 35,473 | 678,074 | 24,533  |
phoenix    (7f65c127) |      48,024 |  3,414 |  53,441 |  1,531  |
actix-core (R18)      |     702,165 | 46,968 | 886,499 | 25,513  | 
phoenix    (R18)      |      53,383 |  4,013 |  63,467 |  2,032  |

Fortune (new / old)   7.56% / 7.60%   
Multi   (new / old)   9.62% / 8.54%
Single  (new / old)   7.88% / 7.16%
Update  (new / old)   6.24% / 7.96%
```

Over both runs actix and phoenix are slower. What we care about performance wise is how many operations are we doing compared to our fastest competitors. (actix-core is pretty fast and available over both runs) 

As you can see we stayed about the same in fortunes, gained a percent in multi and three quarters of a percent in single while slowing down in updates. 

What does concern me is the error rate, that seems to have gone up a lot when glancing over the last few runs. This does seems to indicate that we have a problem. Again I don't think it will drastically alter performance too much if we increase the pool size to ~20-40.

josevalim:
@tomciopp in later tests they had to increase the queue_target and queue_interval to avoid getting timeout errors. This shows we are spending more time waiting for a connection then before, which is a direct consequence of a low pool size.

PR #5157, opened 2019-10-16
Use PDO exeption in errors

joanhey:
To avoid the anormal results in swoole-no-async db tests.
And disabled persistent db connection.
Hoping that now will show more errors, and lower the results numbers.

PD: possibly will not change anything.

### DIFF
--- a/frameworks/PHP/swoole/swoole-server-noasync.php
+++ b/frameworks/PHP/swoole/swoole-server-noasync.php
@@ -9,7 +9,7 @@ $server->set([
 ]);
 
 $pdo = new PDO("mysql:host=tfb-database;dbname=hello_world", "benchmarkdbuser", "benchmarkdbpass", [
-    PDO::ATTR_PERSISTENT => true
+        PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION
 ]);
 
 /**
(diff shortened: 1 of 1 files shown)


frameworks/PHP/swoole/swoole-server-noasync.php
```diff
@@ -9,7 +9,7 @@
 ]);
 
 $pdo = new PDO("mysql:host=tfb-database;dbname=hello_world", "benchmarkdbuser", "benchmarkdbpass", [
-    PDO::ATTR_PERSISTENT => true
+        PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION
```
jcheron: I don't think you'll get any additional error messages.
I displayed the Mysql error stats on the second last run travis-ci, and for `Swoole-no-async`, you had nothing (see https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/598272528#L3623)

joanhey: By default PDO use PDO::ERRMODE_SILENT.
https://www.php.net/manual/en/pdo.error-handling.php

And perhaps the error is for reusing the same db connection in the pool.

joanhey: And some errors:
https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/598272528#L3607-L3608

jcheron: > And some errors:
> https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/598272528#L3607-L3608

Good point, you have your answer:
too many simultaneous requests on the same PDO connection

joanhey: But I need to add the PDO Exception, or fail silently.

jcheron: You can do both:
- Display errors
- Limit the chances of having one by creating the pdo connection for each worker

joanhey: I know, I used so in Workerman and nginx. But it's not posible using the original anonymous functions.

joanhey:
A new revision will be send.

PR #5170, opened 2019-10-21
Update amp

joanhey:
Update: 
* PHP v7.3.
* Amp LIbs
* Libuv instead of libev
* the code

Ping @kelunik to review


frameworks/PHP/amp/server.php
```diff
@@ -88,15 +95,15 @@ private function doHandleRequest($request) {
 
             $items = [];
 
-            $statement = yield $this->mysql->prepare("SELECT * FROM World WHERE id = ?");
+            $statement = yield $this->mysql->prepare('SELECT * FROM World WHERE id = ?');
 
-            for ($i = 0; $i < $queries; $i++) {
```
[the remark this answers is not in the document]
joanhey: We don't use the `$i` variable.
The `while($queries--)` is faster.

frameworks/PHP/amp/server.php
```diff
@@ -12,42 +12,46 @@
...
-use Psr\Log\NullLogger;
+use Monolog\Logger;
+
+define('DB_HOST', gethostbyname('tfb-database'));
 
 Amp\Loop::run(function () {
     $sockets = yield [
-        Cluster::listen("0.0.0.0:8080"),
-        Cluster::listen("[::]:8080"),
+        Cluster::listen('0.0.0.0:8080')
```
kelunik: This should probably pass a customized `BindContext` and increase the default backlog size.

https://github.com/amphp/socket/blob/abc18f1786363ceee2555ffd7fe87e4eead2ce24/src/BindContext.php#L12

joanhey: Yes, 128 is very low for a server.
But I don't know enough Amp.
Please could you help with that?

kelunik: Sure. All you need to do is create an instance of `(new Amp\Socket\BindContext)->withBacklog($size)` as second  parameter.

joanhey: Added socket bind context with `backlog(102400)` as the stream server use 2 sockets.
And with `reuseport` and `tcpnodelay`.

But I miss and option to use `keepalive` in the `BindContest class`.

kelunik: @joanhey I stumbled upon that yesterday, too, so I just opened https://github.com/amphp/socket/issues/67.

joanhey: Without `keepalive`, never will be fast, as each request need to open a new tcp connection.

kelunik: Oh, we're talking about different keep alives. HTTP/1.1 keep alive / connection reuse is automatically enabled.

The thing I just opened is about Tcp keep alive and only affects idle connections.

PR #5174, opened 2019-10-22
Update Rocket for more performance

bhansconnect:
Increased the max number of database connections. Increased the
number of workers because rocket is synchronous and may have
many blocked threads. Changed the templating engine to yarte.
Update the versions of dependencies. Add extra compile time
parameters.

For some reason this is failing my personal Travis CI. Cargo claims to
not be able to find Cargo.toml when building the dockerfile. It builds
locally in vagrant and in some vms I have setup. Cargo.toml is 
definitely available, so the error seems to be with Travis CI.


bhansconnect:
Any ideas on why this is failing on Travis would be greatly appreciated. I attempted to add an call to `ls` in one version of the dockerfile to see the results in travis. The Cargo.toml file is definitely there. Am I missing something?

You can see the results of the version with `ls` here: https://travis-ci.org/bhansconnect/FrameworkBenchmarks/jobs/601462731

bhansconnect:
Did some more testing. It looks like the old version that is currently on master also fails in Travis CI. I decided to test a build where all I did was bump the version number of the final package from `0.1.0` to `0.1.1`, and it fails for the exact same reason: 
`error: could not find `Cargo.toml` in `/rocket` or any parent directory`

https://travis-ci.org/bhansconnect/FrameworkBenchmarks/jobs/601566086

nbrady-techempower:
Hi @bhansconnect,

I'll try and take a look into this. The first thing we want to do though is remove the `FROM rustlang/rust:nightly` in `rocket.dockerfile` and change it to a static version like  `FROM rustlang/rust:1.38`. We want to make sure the tests can be reproducable as much as possible.

bhansconnect:
The rocket framework currently only compiles on nightly. So it can not be built with a stable rust compiler. That being said, I did some more testing yesterday and at least when you specify a stable version of rust, it recognizes the Cargo.toml file. That makes me believe it is an issue with the rust nightly compiler/container. The one thing I still don't understand is why it works locally but not in Travis.

nbrady-techempower:
Try this out. This worked locally for me. I don't know what it is about travis and nightly that borked, but maybe both of the toolchains will help:

```
FROM rust:1.38

ADD ./ /rocket
WORKDIR /rocket

RUN rustup default nightly
RUN apt-get update && apt-get install -yqq clang-6.0

ENV DATABASE_URL=postgres://benchmarkdbuser:benchmarkdbpass@tfb-database/hello_world
RUN cargo clean
RUN RUSTFLAGS="-C target-cpu=native" cargo build --release

CMD ./target/release/rocket
```

msmith-techempower:
How is that `RUN rustup default nightly` not going to also install the nightly version?

nbrady-techempower:
@msmith-techempower It *is* going to *also* install nightly as well as the other toolchain which may include some missing piece for cargo.

msmith-techempower:
Doesn't that also go against our principle of repeatability? The toolchain may literally change tomorrow night.

nbrady-techempower:
I agree with you, but we allowed it already. And, it does appear true that Rocket can only compile with nightly as part of the toolchain. I'd suggest something in benchmark_config to not that this isn't production/realistic.

bhansconnect:
I just noticed that you can actually specify the nightly version by saying `nightly-2019-10-23`. I'll mess with that to see if I can find a specific version that works. Otherwise, I will try the suggestion above, but it shouldn't make a difference.

bhansconnect:
Also, rust is close to being into stable and is already used in production in multiple places. The important thing with using rust nightly is to only update your compiler when you have a reason to do so.

nbrady-techempower:
Yeah, I understand that. For the purposes of the test, though, you'd be grabbing the latest nightly every time the test ran so we couldn't reproduce properly. It looks like: `RUN rustup default nightly-2019-10-23` worked for me locally. I have a travis job running `RUN rustup default nightly` and curious to see if it passes.

bhansconnect:
Should be repeatable now. The dockerfile now gets a specific version of rust nightly. Should be good to merge.

nbrady-techempower:
Woo, thanks @bhansconnect

PR #5181, opened 2019-10-25
Add hyperf-raw and hyperf-micro

huangzhhui:
1. Add hyperf-raw and hyperf-micro cases
2. Upgrade Hyperf to v1.1


frameworks/PHP/hyperf/benchmark_config.json
```diff
@@ -24,6 +24,52 @@
                 "display_name": "Hyperf",
                 "notes": "",
                 "versus": "swoole"
+            },
+            "raw": {
+                "json_url": "/json",
```
jcheron: Hi @huangzhhui 
You don't need to add all these urls that are the same as the basic **hyperf** version:
raw-urls are sufficient:  `json`, `plaintext` and `fortune` are useless  and overload the results for nothing.

frameworks/PHP/hyperf/benchmark_config.json
```diff
@@ -24,6 +24,52 @@
...
+                "platform": "swoole",
+                "webserver": "None",
+                "os": "Linux",
+                "database_os": "Linux",
+                "display_name": "Hyperf",
+                "notes": "",
+                "versus": "swoole"
+            },
+            "micro": {
+                "json_url": "/json",
```
jcheron: Same thing here, only `fortune_url` is different

jcheron:
For each entry in the `benchmark-config` file, you need a corresponding docker file:
`hyperf-micro` and `hyperf-raw` docker files are missing

huangzhhui:
The useless config have been removed and the added dockerfile for hyperf-micro and hyperf-raw

PR #5183, opened 2019-10-25
Java/act update to select distinct random numbers

nbrady-techempower:
@greenlaw110 This is my proposal for updating the query/updates tests to conform with the rules about ORM caching, as this would fail after #5145 is merged. Hibernate has a first level cache that can't be turned off.

Alternatively, you could just remove the hibernate tests, since I believe eclipselink is already performing better. What are your thoughts?


greenlaw110:
@nbrady-techempower thank you! I am happy with the change. Please go ahead to merge the changes

PR #5185, opened 2019-10-27
Enable Arm64 testing with Travis CI

volyrique:
Now that Travis CI [supports](https://blog.travis-ci.com/2019-10-07-multi-cpu-architecture-support) testing in an Arm64 environment, I thought that it would be an interesting experiment to see how much the Web frameworks actually worked in that case. My expectation is that anything using C/C++ or OpenJDK will pass, but I don't have any confidence in the rest. That's why I advise against merging this change - a, frankly speaking, minor platform should not be an impediment to testing in general.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@
 #
 #
 
+arch: arm64
 sudo: required
 dist: xenial
 language: generic
--- a/toolset/benchmark/benchmarker.py
+++ b/toolset/benchmark/benchmarker.py
@@ -1,3 +1,4 @@
+#
 from toolset.utils.output_helper import log, FNULL
 from toolset.utils.docker_helper import DockerHelper
 from toolset.utils.time_logger import TimeLogger
--- a/toolset/databases/mysql/my.cnf
+++ b/toolset/databases/mysql/my.cnf
@@ -23,7 +23,6 @@ socket          = /var/run/mysqld/mysqld.sock
 port            = 3306
 skip-external-locking
 skip-name-resolve
-lower_case_table_names = 1
 
 character-set-server=utf8
 collation-server=utf8_general_ci
--- a/toolset/databases/mysql/mysql.dockerfile
+++ b/toolset/databases/mysql/mysql.dockerfile
@@ -1,11 +1,7 @@
-FROM buildpack-deps:bionic
+FROM buildpack-deps:eoan
 
 ADD create.sql create.sql
 ADD my.cnf my.cnf
-ADD mysql.list mysql.list
-
-RUN cp mysql.list /etc/apt/sources.list.d/
-RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 8C718D3B5072E1F5
 
 RUN apt-get update > /dev/null
 RUN apt-get install -yqq locales > /dev/null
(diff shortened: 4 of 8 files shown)


volyrique:
Unfortunately it is not possible to run the Arm64 and x86-64 tests simultaneously because that exceeds Travis' 200 job limit.

After some experimentation, I got results that were actually worse than expected. It seems that the main issue was that Docker couldn't run inside a LXD container, which is a [known](https://docs.travis-ci.com/user/multi-cpu-architectures/#security-and-lxd-container) [problem](https://travis-ci.community/t/debugging-docker-run-issue-on-arm64/5491). As a result, one could see error messages such as:
```
gemini-postgres: Traceback (most recent call last):
gemini-postgres:   File "/FrameworkBenchmarks/toolset/benchmark/benchmarker.py", line 131, in __run_test
gemini-postgres:     test.database.lower())
gemini-postgres:   File "/FrameworkBenchmarks/toolset/utils/docker_helper.py", line 340, in start_database
gemini-postgres:     log_config={'type': None})
gemini-postgres:   File "/usr/local/lib/python2.7/dist-packages/docker/models/containers.py", line 809, in run
gemini-postgres:     container.start()
gemini-postgres:   File "/usr/local/lib/python2.7/dist-packages/docker/models/containers.py", line 400, in start
gemini-postgres:     return self.client.api.start(self.id, **kwargs)
gemini-postgres:   File "/usr/local/lib/python2.7/dist-packages/docker/utils/decorators.py", line 19, in wrapped
gemini-postgres:     return f(self, resource_id, *args, **kwargs)
gemini-postgres:   File "/usr/local/lib/python2.7/dist-packages/docker/api/container.py", line 1093, in start
gemini-postgres:     self._raise_for_status(res)
gemini-postgres:   File "/usr/local/lib/python2.7/dist-packages/docker/api/client.py", line 263, in _raise_for_status
gemini-postgres:     raise create_api_error_from_http_exception(e)
gemini-postgres:   File "/usr/local/lib/python2.7/dist-packages/docker/errors.py", line 31, in create_api_error_from_http_exception
gemini-postgres:     raise cls(e, response=response, explanation=explanation)
gemini-postgres: APIError: 400 Client Error: Bad Request ("OCI runtime create failed: container_linux.go:345: starting container process caused "process_linux.go:430: container init caused \"write sysctl key kernel.sem: open /proc/sys/kernel/sem: permission denied\"": unknown")
gemini-postgres: Error during test: gemini-postgres
```
I suspect that it was also the cause for an earlier error message that seems harmless:
```
$ export RUN_TESTS=`./toolset/travis/travis_diff.py | tee /dev/tty | grep -oP "travis-run-tests \K(.*)"`
tee: /dev/tty: No such device or address
```

In addition to that I had to work around the fact that the test infrastructure was using the upstream repositories for the database packages instead of the distribution ones. The problem is that neither the PostgreSQL, nor the MySQL repositories contain any Arm64 builds (except MySQL binaries for RHEL, which doesn't help because everything is Ubuntu-based). I upgraded to Ubuntu 19.10 and switched to using the distribution packages, but that meant a version downgrade to PostgreSQL 11, and I had to ignore the MySQL issue with `lower_case_table_names` because the solution in the Dockefile didn't work with the distribution package. BTW it looks like using the upstream repositories also implies that the database versions are not locked down, which I thought was what we were trying to do for all dependencies, so it might be a good idea to make the switch to the distribution repositories permanent (we will probably have to wait for Ubuntu 20.04 to get PostgreSQL 12 though, and also it might be nice to have configuration files that don't depend on the version).

MongoDB was the least problematic database, but I still had to upgrade to 4.2 because that was the earliest version that had Arm64 packages for Ubuntu 18.04.

While I believe that now it should be possible to use the testing infrastructure outside of Travis CI (except for a few implementations that are trying to execute x86-64 binaries), that would still require the database fixes.

nbrady-techempower:
going to close this now @volyrique but thanks for looking into this. feel free to reopen if you want to mess with it more

volyrique:
Sure, most of the fixes have been merged as part of #5285 anyway, and at the end of the day there can't be much progress while Travis CI can't handle Docker on Arm64.

PR #5195, opened 2019-10-30
Back to MySQLSwooleDatabase

vasily-kartashov:
https://github.com/TechEmpower/FrameworkBenchmarks/commit/4a1aad25baa71e52827d6ea4d1f78dd30dfb153f#commitcomment-35708118

### DIFF
--- a/frameworks/PHP/hamlet/swoole.php
+++ b/frameworks/PHP/hamlet/swoole.php
@@ -15,6 +15,6 @@ $database = new MySQLSwooleDatabase(
 );
 $application = new Application($database);
 $onWorkerStart = function () use (&$database) {
-    $database->warmUp((int) (768 / swoole_cpu_num()));
+    $database->warmUp((int) (1024 / swoole_cpu_num()));
 };
 SwooleBootstrap::run('0.0.0.0', 8080, $application, $onWorkerStart);
--- a/frameworks/PHP/hamlet/swoole.php
+++ b/frameworks/PHP/hamlet/swoole.php
@@ -1,20 +1,15 @@
 <?php
 
 use Benchmark\Application;
-use Hamlet\Database\MySQLSwoole\MySQLSwooleDatabase;
+use Hamlet\Database\PDO\PDODatabase;
 use Hamlet\Http\Swoole\Bootstraps\SwooleBootstrap;
 
 require_once __DIR__ . '/vendor/autoload.php';
 
-$database = new MySQLSwooleDatabase(
-    'tfb-database',
+$database = new PDODatabase(
+    'mysql:host=tfb-database;dbname=hello_world',
     'benchmarkdbuser',
-    'benchmarkdbpass',
-    'hello_world',
-    4096
+    'benchmarkdbpass'
 );
 $application = new Application($database);
-$onWorkerStart = function () use (&$database) {
-    $database->warmUp((int) (1024 / swoole_cpu_num()));
-};
-SwooleBootstrap::run('0.0.0.0', 8080, $application, $onWorkerStart);
+SwooleBootstrap::run('0.0.0.0', 8080, $application);
--- a/frameworks/PHP/hamlet/swoole.php
+++ b/frameworks/PHP/hamlet/swoole.php
@@ -11,10 +11,10 @@ $database = new MySQLSwooleDatabase(
     'benchmarkdbuser',
     'benchmarkdbpass',
     'hello_world',
-    1024
+    4096
 );
 $application = new Application($database);
 $onWorkerStart = function () use (&$database) {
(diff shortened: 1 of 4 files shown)


jcheron:
Hi @vasily-kartashov 
For **Hamlet-swoole**, you have some failed transactions on ` Query` and `Update` tests.
image

You can test it yourself by cloning [this repo](https://github.com/phpMv/FrameworkBenchmarks/tree/toolset-queries-verification)

vasily-kartashov:
Hi @jcheron, I bumped up the number of connections created onWorkerStart, it seems to pass the tests now.

jcheron:
We're talking about the future verification: 
On my personal machine, I always have failed transactions. What's strange is that before your pool update, there were just some db queries missing.
Have you tried it on your side?
You should see if you can't trace any potential errors.

However, it passes on Travis.
maybe it will also pass on Citrine, I don't know...
You can wait and see

frameworks/PHP/hamlet/swoole.php
```diff
@@ -15,6 +15,6 @@
 );
 $application = new Application($database);
 $onWorkerStart = function () use (&$database) {
-    $database->warmUp((int) (768 / swoole_cpu_num()));
+    $database->warmUp((int) (1024 / swoole_cpu_num()));
```
jcheron: The number of connections to the database doesn't change anything. You only need 512 db connections.
If you have failed transactions, it is because the Http server does not respond to some requests, and you'll have to determine why.

vasily-kartashov: And yet it helps for some reason. My impression is that some of them don't return to the pool either on time or ever. Hopefully I'll get more time to take a closer look at this in the following weeks.

jcheron: In principle:
For `Query` or `Update` Test:
- I get a connection from the pool.
- I make the requests with (20 select / update)
- I put the connection back into the pool.
That's why I only need 512 connections.
Hamlet seems to work differently, **Hamlet** seems to work differently, since I don't see in your test code any access to the pool.

vasily-kartashov: The pool code is in the https://github.com/hamlet-framework/db-mysql-swoole

The idea of `Database` abstraction is to keep the creation of new connections, transactions, exception handling and pooling inside of this class, so that database access code stays the same between swoole and no-swoole environments.

jcheron: It's a nice goal,
but when (or where) are you getting a new connection from the pool (which part of the code)?

vasily-kartashov: https://github.com/hamlet-framework/db/blob/fbff8ec3b8fb9c254edec61e090beeb349741a96/src/Database.php#L86

jcheron: so there's a fetch from the pool for each request?

vasily-kartashov: It will call `$pool->pop()` 20 times. Do you think this is the issue?

vasily-kartashov: The alternative would be to expose pool management directives to the caller, something like 
```
$database->withSameConnection(function () {
     // do 20 calls within
});
```
not sure I like it, as I am not sure what's the use case of it outside of benchmarking.

jcheron: > It will call `$pool->pop()` 20 times. Do you think this is the issue?

Yes, it's huge.
That's why it works with `Fortunes` and `Db` tests, and not with the others.
You would need `512*20` ready connections in your pool => 10240

vasily-kartashov: They executors release connections back into the pool immediately after, so should the 512*20 be the upper boundary?

jcheron: I think so, yes.
so `512*20/worker_count` to create for each worker...
you can't control concurrent accesses, especially in asynchronous mode.
But anyway, it's too many connections (too many resources), and too many manipulations on the pool to be performant.
A blocking PDO connection would be more efficient in this case.

vasily-kartashov: Can you confirm, even with 1024 connections you still see 5xx when running 
```
./tfb --mode benchmark --test hamlet-swoole --type update
```

jcheron: Confirmation with 1024 db connexions:
The results are unpredictable.

image

image

jcheron:
For me, all is green (no more fails on transactions or db queries).
All you have to do is put the `--quiet` back on the `composer require`.

PR #5196, opened 2019-10-30
Use bulk update in workerman

joanhey:
Later I will try to add to the rest of php frameworks using raw db connection.


joanhey:
I will try later, now there are a problem with Postgres in travis.
Locally that changes work correctly.

PR #5199, opened 2019-10-30
Update postgres to v12

joanhey:
Now it's downloading posgres v12, but still harcoded to v11.
So my last PRs are failing.


nbrady-techempower:
Thanks, I've added it to the notes for R19!

PR #5208, opened 2019-11-02
Fix verification procedure freezing on Citrine

jcheron:
#### Fix verification blocking during `siege` execution
The verification process may be blocked due to errors issued during the execution of  `siege` requests.

This is currently the case with the verification of **akka-http-slick-postgres** which emits errors:

>[alert] HTTP: unable to determine chunk size
[alert] HTTP: unable to determine chunk size
[alert] HTTP: unable to determine chunk size
[alert] HTTP: unable to determine chunk size
[alert] HTTP: unable to determine chunk size
[alert] HTTP: unable to determine chunk size

 and crashes the verification process. -> see https://github.com/TechEmpower/FrameworkBenchmarks/pull/5145#issuecomment-549007885

The reason why the run on Citrine is currently blocked:

image

### DIFF
--- a/toolset/databases/abstract_database.py
+++ b/toolset/databases/abstract_database.py
@@ -90,8 +90,8 @@ class AbstractDatabase:
         cls.reset_cache(config)
         #Start siege requests
         path = config.db_root
-        process = PopenTimeout(shlex.split("siege -c %s -r %s %s -R %s/.siegerc" % (concurrency, count, url, path)), stdout = subprocess.PIPE, stderr = subprocess.STDOUT)
-        output, _ = process.communicate(timeout=15)#timeout 15s if socket errors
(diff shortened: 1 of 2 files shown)


jcheron:
I think the problem is solved.
Before, the verification of `ffead-cpp-apache-mongo` was blocking (see [previous Travis run](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/605281623#L6064)).
It is no longer in the [last Travis run](https://travis-ci.org/TechEmpower/FrameworkBenchmarks/jobs/606443376#L6056).

nbrady-techempower:
I've just restarted with these changes.

jcheron:
we passed the damn 25th

nbrady-techempower:
Yeah looks like we got over the hump. Now I'm curious how many tests are failing versus our average runs.

jcheron:
Yes, I'm curious too.
Not benchmarking frameworks that don't pass the check may give more stability to the run, it's a wish.

nbrady-techempower:
> Not benchmarking frameworks that don't pass the check may give more stability to the run, it's a wish.

Yeah, I have a note of this because it's something I want, too. I'm probably going to work on this soon by way of extracting the tests out from the core toolset in the same way I did databases. Soon!

PR #5213, opened 2019-11-06
update finagle and finatra versions

mgodave:


mgodave:
> I'm not sure why the checks are failing, I don't understand the output.

There was an issue with our release, I need to exclude a dependency for the benchmark to be able to run.

PR #5215, opened 2019-11-07
PHP adjust emulated prepares default in the databases

joanhey:
### Different defaults in mysql and postresql
By default MySQL use always emulated prepares and PostreSQL native server-side prepares if the cursor is not scrollable.

#### php-fpm
In the updates and multiple queries using prepared statements, Postres it's faster as it's using native prepares.
Using native prepared statements for a single query, send 2 round trips to the database, and will be slower. We can't reuse the prepared statements in different requests.
In the single and fortune both use query() without prepared statements.

#### Cli-servers (workerman, php-ngx, ... )
When created the prepared statement once, PostreSql is always faster for using native prepares.
MySQL send a new query every time without the benefit of the server-side prepares.

When created the statement in each request, PostreSQL is slower (~50%) as it need always 2 round trips to database (1 prepare query and 1 execute query).
And MySQL is faster as only send one query.

Now will use both no emulated prepared statements in the first case.
We will check which database is faster using native prepared statements.

### TODO in next PRs

- Change Swoole no async (swoole async don't have this problem)
- Remake updates and multiple queries in all cases.

### DIFF
--- a/frameworks/PHP/php-ngx/app.php
+++ b/frameworks/PHP/php-ngx/app.php
@@ -1,7 +1,9 @@
 <?php
 
 $pdo = new PDO('mysql:host=tfb-database;dbname=hello_world', 'benchmarkdbuser', 'benchmarkdbpass',
-            [PDO::ATTR_DEFAULT_FETCH_MODE => PDO::FETCH_ASSOC]);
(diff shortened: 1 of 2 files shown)


joanhey:
The php-ngx-async update test fail is fixed in the next PR.

PR #5217, opened 2019-11-07
Change updates tests in workerman and php-ngx

joanhey:
Now use prepared statements.


joanhey:
The php-ngx-async update fail is fixed in the previous PR.
Ready to be merged.

Issue #5222, opened 2019-11-08
Enhancement of the verification procedure


jcheron:
The objective is to identify possible improvements of the [new verification procedure](https://github.com/TechEmpower/FrameworkBenchmarks/pull/5145):

### Too strict checking for db queries?
For **ninja-standalone** on the `Update` test: see [Update verification](https://tfb-status.techempower.com/unzip/results.2019-11-08-05-13-52-474.zip/results/20191104142847/ninja-standalone/update/verification.txt)
- The test **fails** for `1 missing query out of 20480`.
- The other statistics pass the tests

I haven't seen yet if other frameworks are in this case, It would be nice to list them.

As the statistics on the number of db queries seemed reliable, we had not accepted a margin on this number.

I'd be interested to know why there are 1 or 2 queries missing (in Ninja)...
Perhaps we could add a margin of acceptance of `1/10000`?

### Failures due to the internal processing of the Framework
The case has already been identified for [java/hibernate](https://github.com/TechEmpower/FrameworkBenchmarks/pull/5145#issuecomment-544735202), which caches queries already made (due to the random aspect of the test), which distorted the results for the `Query` and `Update` tests.
The solution was to randomly select unique numbers.

It is possible that other frameworks, java or not, may be in this case, it would be nice to list them as well.

zloster:
1. About `Hibernate`:
AFAIR most of the Java projects that are using `Hibernate` were updated to use the `JPA` interfaces. If the project is on Hibernate version 5.4.1 this should be true. `Hibernate` and `EclipseLink` are implementations of the `JPA` interface/standard. I see that [`Java/act` with `EclipseLink` doesn't have a problem](https://tfb-status.techempower.com/unzip/results.2019-11-08-05-13-52-474.zip/results/20191104142847/act-eclipselink-mysql). I haven't inspected the code to check what's going on there.

So it should be possible and relatively easy the projects to be migrated to `EclipseLink`* as JPA provider. I don't see this option to be mentioned in the previous discussions.

_*I haven't done such migration._

Also Hibernate has the `StatelessSession` but it is not possible to use it with the JPA interfaces. The difference from the normal `Session` is exactly the absence of `Level 1` object cache. It's an option but I'll not advise for it.

2. About the usage of 512 concurrency:
[...]

PR #5224, opened 2019-11-09
Fix grizzly-jersey cache pb

jcheron:
#### Fixed
- **Hibernate** first-level cache problem (with the random selection of ids to be loaded from the database) see https://github.com/TechEmpower/FrameworkBenchmarks/issues/5222 and https://github.com/TechEmpower/FrameworkBenchmarks/pull/5145#issuecomment-544735202

see [failure](https://tfb-status.techempower.com/unzip/results.2019-11-08-05-13-52-474.zip/grizzly-jersey) in last run

@nbrady-techempower @michaelhixson  @zloster 
I left the `atomicCounter` version for now. We will have to try the [primitiveSet](https://github.com/michaelhixson/jmh-benchmark-tfb-world-ids/blob/18421109bad30bf8d7b2cc4bf317739ea2f8c768/src/main/java/rnd/WorldIdsBenchmark.java#L109) version on another run.

### DIFF
--- a/frameworks/Java/grizzly/src-jersey/main/java/hello/WorldResource.java
+++ b/frameworks/Java/grizzly/src-jersey/main/java/hello/WorldResource.java
@@ -59,7 +59,7 @@ public class WorldResource {
 			try {
 		        //Pick unique random numbers 
 		        final AtomicInteger i = new AtomicInteger(0);
-		        ThreadLocalRandom.current().ints(1, 10000).distinct().limit(queries).forEach(
+		        ThreadLocalRandom.current().ints(1, 10001).distinct().limit(queries).forEach(
 		            (randomValue)->worlds[i.getAndAdd(1)] = (World) session.byId(World.class).load(randomValue)
 		        );
 				return worlds;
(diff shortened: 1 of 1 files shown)


frameworks/Java/grizzly/src-jersey/main/java/hello/WorldResource.java
```diff
@@ -56,9 +57,11 @@ public Object queries(@QueryParam("queries") String queriesParam) throws Interru
 			Session session = emf.createEntityManager().unwrap(Session.class);
 			session.setDefaultReadOnly(true);
 			try {
-				for (int i = 0; i < queries; i++) {
-					worlds[i] = (World) session.byId(World.class).load(randomWorld());
-				}
+		        //Pick unique random numbers 
+		        final AtomicInteger i = new AtomicInteger(0);
+		        ThreadLocalRandom.current().ints(1, 10000).distinct().limit(queries).forEach(
```
michaelhixson: It should be 10001 instead of 10000 here and below since [the upper bound is exclusive](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ThreadLocalRandom.html#ints(int,int)).

PR #5229, opened 2019-11-10
Re-add Lwan to the suite

lpereira:
Source code for the benchmark is now included in this repository. Since
there's no recent version of Lwan, however, a specific Git hash is
obtained from the repository so that the framework can be built.

At the moment, only the JSON and Plaintext tests are enabled; the others,
including Fortunes, DB, and Queries, have bitrotted and need to be
adapted to conform to current rules.


.travis.yml
```diff
@@ -14,6 +14,7 @@ env:
   matrix:
     - "TESTLANG=C"
     - "TESTLANG=CSharp"
+    - 'TESTDIR="C/lwan"'
```
nbrady-techempower: You can remove this line. The framework will be tested with `"TESTLANG=C"`

lpereira: Done.

frameworks/C/lwan/lwan.dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+RUN apt update -yqq
+RUN apt install -yqq \
+	git pkg-config build-essential cmake zlib1g-dev \
+	libsqlite3-dev libmariadbclient-dev wget
+
+ADD ./ /lwan
+WORKDIR /lwan
+
+RUN git clone https://github.com/lpereira/lwan --depth 1 && \
```
volyrique: A shallow clone looks quite fragile, and I think that you can achieve the same effect simply by downloading `https://github.com/lpereira/lwan/archive/${commit hash}.zip`.

lpereira: Thanks.  That's probably the reason it's failing in CI.  I've just pushed a new version using your suggestion; seems to be working locally.

PR #5233, opened 2019-11-12
Fix laravel & Laravel-swoole cache pbs

jcheron:
#### Fixed
- **Eloquent** first-level cache problem (with the random selection of ids to be loaded from the database) see https://github.com/TechEmpower/FrameworkBenchmarks/issues/5222 and https://github.com/TechEmpower/FrameworkBenchmarks/pull/5145#issuecomment-544735202

see [missing queries](https://tfb-status.techempower.com/unzip/results.2019-11-08-05-13-52-474.zip/results/20191104142847/laravel/update/verification.txt) for Update test in last run

#### Optimization
 - remove double quote usage for strings
 - Force global namespace usage
- Replace `random_int` with `mt_rand` (10x faster)

ping @fturmel @joanhey

### DIFF
--- a/frameworks/PHP/laravel/app/Http/Controllers/Controller.php
+++ b/frameworks/PHP/laravel/app/Http/Controllers/Controller.php
@@ -19,7 +19,7 @@ class Controller extends BaseController {
 
 	public function queries($queries = 1) {
 		$rows = [];
-		$numbers = $this->getUniqueRandomNumbers($this->clamp($queries), 1, 10000);
+		$numbers = $this->getUniqueRandomNumbers($this->clamp($queries));
 		foreach ($numbers as $id) {
 			$rows[] = World::find($id);
 		}
@@ -45,7 +45,7 @@ class Controller extends BaseController {
 	public function updates($queries = 1) {
 		$rows = [];
 
-		$numbers = $this->getUniqueRandomNumbers($this->clamp($queries), 1, 10000);
+		$numbers = $this->getUniqueRandomNumbers($this->clamp($queries));
 		foreach ($numbers as $id) {
 			$row = World::find($id);
 			$row->randomNumber = \mt_rand(1, 10000);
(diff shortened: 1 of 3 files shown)


fturmel:
The only thing that caught my eye is the _getUniqueRandomNumbers_ method not behaving in a predictable way performance-wise. It may not have a large impact here because of the inputs, but I think a O(n) solution would be preferable especially in a benchmarking context.

Have you considered something like the following?

```
private function getUniqueRandomNumbers($count, $min, $max) {
    $res = \range($min, $max);
    \shuffle($res);
    return \array_slice($res, 0, $count);
}
```

With inputs (500, 1, 1000) to me they seem to perform the same in the best case scenario, but the guessing approach in this pull request has a worst case of O(infinity) which will often make it slower and by an unpredictable margin.

Everything else looks fine to me. Cheers!

jcheron:
Thanks @fturmel for this review.
The call of the `getUniqueRandomNumbers` function is performed with the following count values: `1`,`5`,`10`,`15` and `20`
`min` and `max` are always `1` and `10000` respectively.
With such low values, the original version of `getUniqueRandomNumbers` is about 200 times faster.
And it would require a count greater than 3333 for your proposal to be more effective.
It is the fault of the `shuffle` function, on an array of 10.000 values.

#### Benchmark with count=20, min=1, max=10000
image

fturmel:
Fair enough, for low count and high max values that makes sense. I misread 10000 as 1000 as the default max, and expected the count to sometimes be 500 since we were clamping to that value. Sorry about that!

Just one other thought... From what I understand the primary goal of the pull request is to make sure the _updates_ test is always writing to the database and bypassing whatever cache is in the ORM. I wonder what happens if the random number generated on line 55 is the same as the one that was just fetched from the database. I did not get the chance to confirm, but I believe the model property won't be flagged as dirty and thus there won't be an update query being sent to the database.

See:
https://github.com/laravel/framework/blob/1bbe5528568555d597582fdbec73e31f8a818dbc/tests/Database/DatabaseEloquentModelTest.php#L88

https://github.com/laravel/framework/blob/869f02783a49cd0e16c66c0d362d2e4ea371f4b6/src/Illuminate/Database/Eloquent/Model.php#L744

jcheron:
The tfb verification module considers this possibility of choosing an identical randomNumber.
A 10% margin is accepted on the modification:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/53e465a460d54b1337f5649e9f89475217057a7a/toolset/benchmark/test_types/verifications.py#L267

PR #5234, opened 2019-11-12
Update ubiquity swoole & workerman

jcheron:
#### Changed
- Swoole worker_num (x1->x2)
- Workerman worker_num (x3->x4)

### DIFF
--- a/frameworks/PHP/ubiquity/.ubiquity/_swoole.php
+++ b/frameworks/PHP/ubiquity/.ubiquity/_swoole.php
@@ -1,18 +1,20 @@
 #!/usr/bin/env php
 <?php
 // swoole.php
-
-if (! defined ( 'DS' )) {
-	define ( 'DS', DIRECTORY_SEPARATOR );
-	define ( 'ROOT', __DIR__ . \DS .'..'.\DS. 'app' . \DS );
+if (! defined('DS')) {
+	define('DS', DIRECTORY_SEPARATOR);
+	define('ROOT', __DIR__ . \DS . '..' . \DS . 'app' . \DS);
 }
-$config=include ROOT.'config/config.php';
-$sConfig= include __DIR__.\DS.'swoole-config.php';
-$config["sessionName"]=null;
-$address=$sConfig['host'].':'.$sConfig['port'];
-$config ["siteUrl"] = 'http://'.$address;
+$config = include ROOT . 'config/config.php';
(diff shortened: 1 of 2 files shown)


frameworks/PHP/ubiquity/.ubiquity/_swoole.php
```diff
@@ -1,18 +1,20 @@
...
-$swooleServer=new \Ubiquity\servers\swoole\SwooleServer();
+$swooleServer = new \Ubiquity\servers\swoole\SwooleServer();
 $swooleServer->init($config, __DIR__);
-require ROOT.'config/swooleServices.php';
-$swooleServer->run($sConfig['host'],$sConfig['port']);
\ No newline at end of file
+require ROOT . 'config/swooleServices.php';
+$swooleServer->run($sConfig['host'], $sConfig['port'], [
+	'worker_num' => \swoole_cpu_num() * 2,
+	'reactor_num' => \swoole_cpu_num() * 2
```
vasily-kartashov: @jcheron, do you see any performance improvements with this or is it speculation?

jcheron: > @jcheron, do you see any performance improvements with this or is it speculation?

I think it's better, to be confirmed...

joanhey: If it is faster, then it is possible to have better coroutine implementation.
And theoretically will be slower in the plaintext and json tests.

joanhey:
I want to test workerman x4 too, and aslo x5.
But I don't want it mixed with other changes in the same run.
Later we will try it for the platform.

vasily-kartashov:
@jcheron quite impressive results on update test.
https://www.techempower.com/benchmarks/#section=test&runid=e0e5d81b-7177-43a3-8e71-72680544dd50&hw=ph&test=update&l=zik073-v
Do you mind sharing the secret of what you're doing differently from the rest?

jcheron:
Hi @vasily-kartashov 
It's no secret, open source helps us to improve our productions.
https://github.com/TechEmpower/FrameworkBenchmarks/blob/227880491e47718f4d6a0ba380948ee2b0df5b7c/frameworks/PHP/ubiquity/app/controllers/SwooleDb.php#L33-L44

**Ubiquity** makes a bulk-update for each http request.
Only one db query is required to update, which is different from several queries in a transaction.
Note that at this high concurrency level, it only works with pgsql, and not at all with Mysql.

PR #5237, opened 2019-11-13
Add symfony-swoole

jderusse:
ping @jcheron 

www techempower com_benchmarks_

### DIFF
--- a/frameworks/PHP/symfony/benchmark_config.json
+++ b/frameworks/PHP/symfony/benchmark_config.json
@@ -43,6 +43,29 @@
       "display_name": "symfony",
       "notes": "",
       "versus": "php"
+    },
+    "swoole": {
+      "plaintext_url": "/plaintext",
+      "json_url": "/json",
+      "db_url": "/db",
+      "update_url": "/updates?queries=",
+      "query_url": "/queries?queries=",
+      "fortune_url": "/fortunes",
+      "port": 8080,
+      "approach": "Realistic",
+      "classification": "Fullstack",
+      "database": "MySQL",
+      "framework": "symfony",
+      "language": "PHP",
+      "flavor": "None",
+      "orm": "Full",
+      "platform": "swoole",
+      "webserver": "none",
+      "os": "Linux",
+      "database_os": "Linux",
+      "display_name": "symfony-swoole",
+      "notes": "",
+      "versus": "swoole"
     }
   }]
 }
--- a/frameworks/PHP/symfony/src/Controller/DbController.php
+++ b/frameworks/PHP/symfony/src/Controller/DbController.php
@@ -26,9 +26,7 @@ class DbController
      */
     public function db(): JsonResponse
     {
-        $world = $this->worldRepository->find(mt_rand(1, 10000));
-
-        return new JsonResponse($world);
+        return new JsonResponse($this->worldRepository->find(mt_rand(1, 10000)));
     }
 
     /**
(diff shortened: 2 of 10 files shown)


frameworks/PHP/symfony/benchmark_config.json
```diff
@@ -43,6 +43,29 @@
       "display_name": "symfony",
       "notes": "",
       "versus": "php"
+    },
+    "swoole": {
+      "plaintext_url": "/plaintext",
+      "json_url": "/json",
+      "db_url": "/db",
```
jderusse: Could be `/raw/db` for better performance.
Should we provide version that just provide swoole? Or a version that provide better performances with less features?

jcheron: I'll keep it as is for now,
The implementation with an ORM is more realistic with Symfony.
we'll see how it goes.

frameworks/PHP/symfony/src/Controller/DbController.php
```diff
@@ -41,10 +43,10 @@ public function queries(Request $request): JsonResponse
 
         // possibility for enhancement is the use of SplFixedArray -> http://php.net/manual/de/class.splfixedarray.php
         $worlds = [];
-
-        for ($i = 0; $i < $queries; ++$i) {
-            $worlds[] = $this->worldRepository->find(mt_rand(1, 10000));
+        foreach ($this->getUniqueRandomNumbers($queries) as $id) {
+            $worlds[] = $this->worldRepository->find($id);
         }
+        $this->entityManager->clear();
```
jderusse: I'll add a Swoole Handler to perform this only for swoole env

jcheron: You just need to make sure that this operation is performed before sending the Http response.

jderusse: this is a clear (remove internal cache). not a flush

frameworks/PHP/symfony/src/Controller/DbController.php
```diff
@@ -58,18 +57,27 @@ public function update(Request $request): JsonResponse
...
-
-        for ($i = 0; $i < $queries; ++$i) {
-            $world = $this->worldRepository->find(mt_rand(1, 10000));
-            if ($world) {
-                $randomNumber = mt_rand(1, 10000);
-                $world->setRandomNumber($randomNumber);
-                $worlds[] = $world;
-            }
+        $ids = $this->getUniqueRandomNumbers($queries);
+        sort($ids);
```
jcheron: The difference is marginal, it's more for harmonization, but I used the `ksort` in the `getUniqueRandomNumbers` function.
https://github.com/TechEmpower/FrameworkBenchmarks/blob/22998dd14c1badfe591f48156abc75ce96483001/frameworks/PHP/symfony/src/Controller/DbController.php#L26

jcheron:
I close the PR https://github.com/TechEmpower/FrameworkBenchmarks/pull/5230 if you update `DbController`

jderusse:
As you want. Is #5230 is merged first, this PR will just be about swoole.

frameworks/PHP/symfony/src/Controller/DbController.php
jderusse: the method is used in `/queries` without need to be ordered.

jcheron:
> As you want. Is #5230 is merged first, this PR will just be about swoole.

If https://github.com/TechEmpower/FrameworkBenchmarks/pull/5230 is merged before, you'll have to merge origin master in your branch before.

jderusse:
yes, a simple rebase.

jcheron:
The current failures on the verification of `Query` and `Update` tests are similar to what can happen when the connection to the database is not performed in the `onWorkerStart` event of the **Swoole** server.

jderusse:
should be Fixed now. 2 Issues :
- EntityManager were not cleared between each request
- the loop to generate a new random value should check the previous value instead of the ID
- `mt_srand` were not called, which make all calls to `mt_rand` returning always the same values

jcheron:
Okay, that's good work!
Travis is a little blocked, but the check should pass.

I'm closing the PR https://github.com/TechEmpower/FrameworkBenchmarks/pull/5230.
If you can add in your PR description that it solves the problem of the cache and missing queries.

for info: https://github.com/k911/swoole-bundle/issues/66

PR #5239, opened 2019-11-13
Update mysql.cnf

nbrady-techempower:
Per discussions in #5222

### DIFF
--- a/toolset/databases/mysql/my.cnf
+++ b/toolset/databases/mysql/my.cnf
@@ -57,6 +57,8 @@ innodb_use_native_aio   = 1
 innodb_flush_log_at_trx_commit = 2
 innodb_flush_method=O_DIRECT
 innodb_buffer_pool_instances=14
+innodb-deadlock-detect=OFF
+innodb_lock_wait_timeout=1
 
 sync_binlog=0
 


toolset/databases/mysql/my.cnf
```diff
@@ -57,6 +57,8 @@ innodb_use_native_aio   = 1
 innodb_flush_log_at_trx_commit = 2
 innodb_flush_method=O_DIRECT
 innodb_buffer_pool_instances=14
+innodb-deadlock-detect=OFF
```
jderusse: that does not reflect a real application. 
A real application should handle such situation, or disable transactions

nbrady-techempower: Thanks @jderusse; We were just having that discussion internally as well. We are leaning toward not committing this change but going to keep the discussion going for a bit.

@zloster Any response to that?

jderusse: see https://github.com/TechEmpower/FrameworkBenchmarks/issues/5222#issuecomment-553637377

zloster: @nbrady-techempower I've misunderstand the cause of the issue. This should not be needed. I've missed the fact that [`go/fasthttp` is sorting the `updates`](https://github.com/TechEmpower/FrameworkBenchmarks/blob/062105536fc9f5b3bf5ca79b515c579d8687f84a/frameworks/Go/fasthttp/src/server-mysql/server.go#L148).

PR #5261, opened 2019-11-20
Improve EF changes for query verification

sebastienros:
If think the issue comes from the fact that the new value, even though being random, can match the previous one, and in this case Entity Framework might detect there is no change and thus not generate the update command for this object.

Other frameworks might face the same issue unintentionally. In the PR that @jcheron submitted the lines that actually fixed the issue were in the `do while` loop. The other changes didn't matter AFAIK.


jcheron:
Considering the random aspect of the ids to be loaded, it may happen that sometimes it passes, and that others do not, given the margin of error accepted...
So it seems that there is a cache issue and that the uniqueness of the ids is also required.

nbrady-techempower:
@sebastienros Looks like this didn't pass. Did you want to try anything else or is it ok to close this?

PR #5268, opened 2019-11-24
Update mysql odbc connector url

sumeetchhetri:


nbrady-techempower:
@sumeetchhetri There are still a number of failing tests for ffead and they're clogging up the test environment. Can you remove the failing tests until you're able to tackle them. You can add: 
`"tags": ["broken"]` to each permutation that's failing and it will skip those.

nbrady-techempower:
Closed for #5300

PR #5274, opened 2019-11-29
Update php version for Symfony

jderusse:
Update to PHP 7.4.

mysqlnd driver uses `caching_sha2_password`, that's why I added a 2nd user in the database.

### DIFF
--- a/toolset/databases/mysql/create.sql
+++ b/toolset/databases/mysql/create.sql
@@ -31,7 +31,7 @@ declare v_counter int unsigned default 0;
     SET v_counter=v_counter+1;
   end while;
   commit;
-END 
+END
 #
 
 DELIMITER ;
--- a/frameworks/PHP/symfony/composer.json
+++ b/frameworks/PHP/symfony/composer.json
@@ -2,7 +2,7 @@
     "type": "project",
     "license": "proprietary",
     "require": {
-        "php": "^7.3.7",
+        "php": "^7.4",
         "ext-ctype": "*",
         "ext-iconv": "*",
         "ext-mbstring": "*",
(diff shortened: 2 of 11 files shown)


jderusse:
Failling tests are not related to that PR

toolset/databases/mysql/create.sql
```diff
@@ -14,8 +14,12 @@ CREATE TABLE  world (
 ENGINE=INNODB;
 CREATE USER 'benchmarkdbuser'@'%' IDENTIFIED WITH mysql_native_password BY 'benchmarkdbpass';
 CREATE USER 'benchmarkdbuser'@'localhost' IDENTIFIED WITH mysql_native_password BY 'benchmarkdbpass';
+CREATE USER 'benchmarkdbuser_sha2'@'%' IDENTIFIED WITH caching_sha2_password BY 'benchmarkdbpass';
```
nbrady-techempower: @jderusse If all users of mysql can benefit from this change without any code changes, than let's apply this to the `benchmarkdbuser` otherwise, I don't want to make this change for one framework.

jderusse: This is not about 1 framework, this is about all PHP frameworks using mysqlnd driver and the latstest version of PHP (7.4).

As far as I know, the authentication method is not configurable (nor in framework, nor in php.ini itself), the password have to use the plugin understood by the driver.
mysql_native_password for libmysql or mysqlnd with (PHP<7.4 and many other languages)
caching_sha2_password for mysqlnd (PHP>=7.4)

nbrady-techempower: @jderusse `mysqlnd` supports `mysql_native_password`

Please add 

```
default_authentication_plugin = mysql_native_password
```

to toolset/databases/mysql/my.cnf and remove the new user. That should work.

jderusse: much better. Thank you

PR #5281, opened 2019-12-04
enable string deduplication in java for wildfly

codylerum:


nbrady-techempower:
Does this do anything with the Z Garbage Collector? Some light reading tells me this may only work with G1GC?

codylerum:
Yeah that is what I'm seeing as well. I think we can just dump thins.

PR #5285, opened 2019-12-07
Various toolset changes

volyrique:
* Update MongoDB to version 4.2
* Improve the wrk build - extra optimizations and parallel compilation are enabled. Also, the system LuaJIT and OpenSSL versions are used, which reduces the build time.
* Make the Travis CI tests more robust - there is no longer an assumption that MySQL and PostgreSQL are installed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -131,8 +131,8 @@ before_script:
   - if [ "$RUN_TESTS" ]; then docker build -t techempower/tfb - < ./Dockerfile; fi
 
   # Stop services that would claim ports we may need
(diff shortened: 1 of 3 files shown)


volyrique:
@nbrady-techempower This pull request needs to be mentioned in the list of changes for the next round - the MongoDB upgrade and the wrk optimizations are most probably going to affect the results.

nbrady-techempower:
@volyrique Done. (Note MongoDB has already been mentioned bc it's had a major version update since last round)

PR #5292, opened 2019-12-10
Update MySqlConnector to 0.62.0-beta5

bgrainger:
Testing performance improvements in the latest beta of this .NET library.


bgrainger:
Looks like a spurious failure (in a "pg" test, which is unrelated). Pushing an update to start another CI build.

PR #5296, opened 2019-12-12
Set keep-alive to 0 for rocket and make the randomizer output unique numbers from a shuffled array

vegai:
Rocket devs in https://github.com/SergioBenitez/Rocket/issues/1038 suggest that the default keep-alive -setting might cause problems for Rocket.

### DIFF
--- a/frameworks/Rust/rocket/Cargo.toml
+++ b/frameworks/Rust/rocket/Cargo.toml
@@ -13,6 +13,7 @@ serde = "1.0.101"
 serde_json = "1.0.41"
 serde_derive = "1.0.101"
 yarte = "0.3.5"
+lazy_static = "1.4.0"
 
 [dependencies.rocket_contrib]
 version = "*"
--- a/frameworks/Rust/rocket/rocket.dockerfile
+++ b/frameworks/Rust/rocket/rocket.dockerfile
@@ -15,7 +15,7 @@ RUN set -eux; \
     url="https://static.rust-lang.org/rustup/dist/x86_64-unknown-linux-gnu/rustup-init"; \
     wget "$url"; \
     chmod +x rustup-init; \
-    ./rustup-init -y --no-modify-path --default-toolchain nightly-2019-10-14; \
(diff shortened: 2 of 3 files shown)


vegai:
`   FAIL for http://tfb-server:8000/updates?q=20

     Only 10106 rows updated in the database out of roughly 10240 expected.`

on the Rocket test. I guess this wasn't so simple after all :) Everything else passed though, so this is slightly weird.

vegai:
So I closed/reopened the PR to trigger a rebuild and the 2nd one passed. Slightly weird.

nbrady-techempower:
@vegai it's not consistent. I reran and it failed again. You may need to make sure that the random number you generate is unique, a solution that some of the Java frameworks use.

vegai:
I merely changed the keep-alive -setting though. Do you think that the Rocket test was broken also before this change and something else needs to change to stabilize it? Or that the keep-alive change alone made it unstable?

nbrady-techempower:
It was probably failing verification before this when we added stricter verification testing to the toolset. Probably not a result of your changes but we don’t want to merge broken tests. You could remove that single test to get this merged or if you want to take a shot at fixing it.

vegai:
I can give it a shot. I see that 

```
fn random_number() -> i32 {
    rand::thread_rng().gen_range(1, 10_001)
}
```

You mean this test fails if the same number is generated more than once here? I guess it would be against the rules to just pick sequential numbers? :)

nbrady-techempower:
Haha, yeah no sequential numbers. My guess, from seeing it happen with a couple of other frameworks, is that there is some caching mechanism somewhere that stops a request from hitting the database if it's a number it has already seen.

vegai:
Meh, too difficult, I think I'll give up on this :)

vegai:
Meh, too difficult, I think I'll give up on this :)

vegai:
Seems to have worked now several times. Hopefully I didn't crap on rocket's performance by my code too badly.

fraunos:
@vegai thanks for your input - I'd love to see how rockey performs, can't wait for another round

PR #5297, opened 2019-12-12
Add some tweaks for the Lwan framework

lpereira:
This adds a few tweaks:

- Uses a new Lwan, with some optimizations to reduce memory allocations during fast paths
- Uses a new third-party memory allocator
- Builds everything with profile-guided-optimization data obtained from local runs (requires container running in new Ubuntu version due to a linker bug)
- A variety of tweaks in the benchmark harness

### DIFF
--- a/frameworks/C/lwan/Makefile
+++ b/frameworks/C/lwan/Makefile
@@ -8,6 +8,7 @@ CFLAGS = -mtune=native -march=native -O3 -flto -ffat-lto-objects -DNDEBUG \
 	-fauto-profile=/lwan/src/gcda/techempower.gcov
 
 LDFLAGS = -mtune=native -march=native -O3 -flto -ffat-lto-objects \
+	/usr/local/lib/mimalloc-1.2/libmimalloc.a \
 	-Wl,-whole-archive /lwan/build/src/lib/liblwan.a -Wl,-no-whole-archive \
 	`pkg-config mariadb --libs` \
 	`pkg-config sqlite3 --libs` \
--- a/frameworks/C/lwan/Makefile
+++ b/frameworks/C/lwan/Makefile
@@ -4,7 +4,8 @@ CFLAGS = -mtune=native -march=native -O3 -flto -ffat-lto-objects -DNDEBUG \
 	-include /lwan/build/lwan-build-config.h \
 	-I /lwan/src/lib \
 	`pkg-config mariadb --cflags` \
-	`pkg-config sqlite3 --cflags`
+	`pkg-config sqlite3 --cflags` \
+	-fauto-profile=/lwan/src/gcda/techempower.gcov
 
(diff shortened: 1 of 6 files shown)


nbrady-techempower:
@lpereira can you remove the `.gcov` file

lpereira:
I can't, it's needed by the build.

lpereira:
I can host it somewhere and download it during the build, though.

nbrady-techempower:
@lpereira That would be preferable, because we don't want to merge binaries, but I'm also curious as to what its main function is. Looks like it's used for profiling/debugging, so can it be generated from the source code in the dockerfile?

lpereira:
It can be generated during build time, however it requires running the
program to capture profiling events, passing the data over to another
program, etc. It's kind of a convoluted process. It's generated by AutoFDO,
and used to guide the compiler to perform better optimizations.

I'll host it in the Lwan repo; I'll update the PR tonight.

volyrique:
> It can be generated during build time, however [...] It's kind of a convoluted process.

FYI that's precisely what `h2o` is [doing](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/C/h2o/h2o.sh#L79), and it wasn't particularly painful to implement.

lpereira:
I didn't realize h2o was using PGO; thanks for letting me know, Anton.
However, the method it uses doesn't work with Lwan reliably (maybe I should
revisit this method, because it didn't work at all with multithreaded
programs a few years back).

I'm using AutoFDO, which uses CPU event counters from perf instead, which
requires some fiddling and might not work well in a containerized
environment. Ultimately that's what I'll do, as the CPU in my home PC isn't
the same generation as the one that's running the servers for the benchmark
and the compiler might be optimizing for the wrong things.

volyrique:
> ... maybe I should revisit this method, because it didn't work at all with multithreaded programs a few years back...

AFAIK that's still the case, and that's why `h2o` is run single-threaded while gathering the profile. IMHO this doesn't really make the profile less representative because by design each thread is almost totally independent; only the cached queries test implementation requires a significant amount of synchronization, and I don't think it is affected that much either.

lpereira:
@nbrady-techempower It has been updated to use the `.gcov` file from the Lwan repository.

nbrady-techempower:
@lpereira I still see the `.gcov` file in this PR

lpereira:
I apologize, Nate.  I had removed the references to the .gcov file but
ended up forgetting removing it.  Double-checked and pushed; should be
gone now.

On 12/16/19, Nate <notifications@github.com> wrote:
> @lpereira I still see the `.gcov` file in this PR
>
> --
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/TechEmpower/FrameworkBenchmarks/pull/5297#issuecomment-566086855


-- 
  Leandro

PR #5316, opened 2019-12-16
[new c++ framework] Lithium.

matt-42:
Just added a new framework. It's inspired from silicon but is based on a custom async event loop. It also uses the async mysql api, which makes it run a lot faster.


matt-42:
By the way congrats for the new architecture ! it's gotten way more simple to add a framework :)

nbrady-techempower:
@matt-42 Thank you! It has come a long way. More improvements to come!

matt-42:
All tests are passing in vagrant ! Let's see what travis says.

PR #5323, opened 2019-12-19
Update to PHP 7.4

joanhey:
And with Ubuntu 19.10.

Later I'll send more PRs to update the rest of frameworks.

The new 7.4 have a new `preload` in the OPCache, for faster framework run. But that will be specific for every framework. So we will update with `preload` when are ready the frameworks.


joanhey:
It's ready to be merged.
This PR don't touch the php-unit because fail with php 7.4, but it's also failing now without the change.
Later I'll send a PR changing the dockerfile to fix it.

PR #5357, opened 2019-12-30
Update nginx.conf in php

joanhey:


joanhey:
when is ready I'll send it again

PR #5361, opened 2019-12-31
Update drogon

### LINKED ISSUE #5352: Drogon core benchmarks
i noticed drogon-core does not use load results into memory and does not use json serializer.
as i remember this is against rules.

https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/C%2B%2B/drogon/drogon_benchmark/controllers/QueriesCtrlRaw.cc#L40

an-tao:
resolve #5352


nbrady-techempower:
Thank you for updating this so quickly!

an-tao:
Thank you for answering my questions on vacation!

PR #5366, opened 2020-01-04
Optimize some tests of drogon

an-tao:

### DIFF
--- a/frameworks/C++/drogon/drogon_benchmark/views/fortune.csp
+++ b/frameworks/C++/drogon/drogon_benchmark/views/fortune.csp
@@ -4,6 +4,7 @@ using namespace drogon_model::hello_world;
 %>
 <%c++
 auto &rows=@@.get<std::vector<Fortune>>("rows");
+$$.reserve(1300);
 %>
 <!DOCTYPE html><html><head><title>Fortunes</title></head><body><table><tr><th>id</th><th>message</th></tr>
 <%c++for(auto const &row:rows){%>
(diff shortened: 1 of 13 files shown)


frameworks/C++/drogon/drogon_benchmark/views/fortune.csp
```diff
@@ -4,6 +4,7 @@ using namespace drogon_model::hello_world;
 %>
 <%c++
 auto &rows=@@.get<std::vector<Fortune>>("rows");
+$$.reserve(1300);
```
nbrady-techempower: @an-tao Can you clarify what this addition does? Is it reserving space? (We want to make sure there's no prior knowledge of the size of the fortunes array)

an-tao: Thanks for clearify this, I've removed it.
How if I calculate the size of array like follows?
```
size_t size=0;
for(auto &row : rows)
{
     size += row.getValueOfMessage().length();
}
$$.reserve(size);
```

nbrady-techempower: That looks fine to me though I don’t understand what the 800 is for?

nbrady-techempower: That looks fine to me though I don’t understand what the 800 is for?

an-tao: I didn't calculate it carefully, it represents the size of the fixed part of this page, but I don't plan to change it this way, I think these actually have little effect on the result :) You could merge it now, thanks!

Issue #5368, opened 2020-01-04
Need clarification for the multi-query test


snluu:
For the multi-query test, is the expectation that:
a) The application only fires off the next DB query **after** the previous query returns, or
b) The application is free to fire off the queries in parallel?

Looking through the different frameworks, I noticed that some implement “a” (even though they can do “b”), and some implement “b”. Frameworks that go with the “b” approach definitely has the better competitive edge. 

For example: https://www.techempower.com/benchmarks/#section=data-r18&hw=ph&test=query&l=zdhf5r-f

Crystal outperforms Dart in the single-query test, but performs significantly slower in the multi-query test because Crystal implements “a” while Dart implements “b”.

nbrady-techempower:
Yes, this is allowed.

https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#multiple-database-queries

> This test is designed to exercise multiple queries with each resulting row selected individually. It is not acceptable to execute multiple SELECTs within a single complex query. It is not acceptable to retrieve all required rows using a SELECT ... WHERE id IN (...) clause. **_However, note General Requirement #7 which permits pipelining of network traffic between the application and database._**

snluu:
@nbrady-techempower Thanks for the clarification, Nate! I will work on a patch for the Crystal frameworks to run the queries concurrently. Feel free to close this task, or we can close it once my patch is marged.

snluu:
PR here: #5370

PR #5370, opened 2020-01-05
Add a test for crystal-kemal to query the database concurrently in `/queries`

snluu:
Previous discussion: #5368 

By calling `/queries` with a `concurrent=x` query string, the caller can control how many concurrent queries are issued to the database.

The added test defaults to 2 for now.

Other changes in this PR:

* Fix the query to fetch both the id and randomNumber from the row, as per specification
* Updated to the latest Crystal and dependency versions (Crystal v0.32.1)


snluu:
What does it mean that one passed and one failed here?

```
VERIFYING QUERY COUNT FOR http://tfb-server:8080/queries?concurrency=1&queries=20
ERROR:root:Terminating process 534 by timeout of 20 secs.
   PASS for http://tfb-server:8080/queries?concurrency=1&queries=20
     Executed queries: 10227/10240
     See https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#specific-test-requirements

   FAIL for http://tfb-server:8080/queries?concurrency=1&queries=20
     Only 10115 rows read in the database out of roughly 10240 expected.
     See https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#specific-test-requirements
```

nbrady-techempower:
The number of rows read from the database on the second attempt were only 10115 which is outside the acceptable range.

snluu:
@nbrady-techempower Nate, any idea why it might be that way? The other tests verified that it should return the correct number of rows right? Looking at the implementation, I'm not sure how it would return the correct number of rows without making the same number of queries.

nbrady-techempower:
Two possibilities are some built in caching mechanism for random numbers that have already been queried (this seems unlikely as I’d expect the number of queries to be a lot less) or more likely some concurrent queries were held up / timed out.  

Earlier in the process we verify that a single request returns the appropriate number of queries. This failure is probably related to the application not being able to process 512 concurrent requests without some failures.

snluu:
Thanks for the explanation @nbrady-techempower! Interestingly, the “/update” tests are also failing without any changes (besides updating to the latest version of the language). 

One possibility I can think of is the test might be overloading the database provided by TravisCI, and causing timeouts/errors. Right now (and previously), the docker file spawns N processes, and each process opens 56 persisted-connections to the DB, with N being the number of CPU’s (using `nproc`). In your opinion, is that too much or you’ve seen other frameworks done worse?

Let me try out a few things to test out my theory.

snluu:
Looks like the old implementation still has this issue; might have something to do with the new language version and new dependency version. Let me kill this PR for now, will start over with just upgrading the language.

PR #5371, opened 2020-01-05
Fix dropwizard postgres

jcheron:
#### Some modifications
- Removing `transactional=false` on `@UnitOfWork` for **db**, **query** and **Fortunes** urls
- Adding a random selection of unique and ordered ids for **query** and **update**
- Assigning a value different from the original on **update**

The tests seem to be going better.
But I'm not sure about the optimization...
I didn't touch **jdbi-postgres**.

ping @zloster for checking

see https://github.com/TechEmpower/FrameworkBenchmarks/issues/5222#issuecomment-570947164


zloster:
Hmm, interesting.
I wonder about `transaction=false`: exactly the same code is used by the `MySQL` variant. It didn't showed problems. I don't understand. But if it is working - OK.
I'll add it to my list with puzzles needing attention and inspection.

PR #5375, opened 2020-01-06
Update Crystal/kemal to use Crystal 0.32.1 and newer shards

snluu:
Removed the “max pool size” parameter from the DB URL since the new shard version has a different behavior.


vlazar:
Hey @snluu do you have plans to update other Crystal implementations as well? It would make sense to have all updated, otherwise comparing different Crystal frameworks performance makes less sense with non latest Crystal/framework versions.

PR #5376, opened 2020-01-07
Added concurrent queries & test for Crystal/kemal

snluu:
This patch also fixed up the existing implementation to adhere by the rule to select the whole row instead of just selecting the random number column.

### DIFF
--- a/frameworks/Crystal/kemal/server-postgres.cr
+++ b/frameworks/Crystal/kemal/server-postgres.cr
@@ -1,5 +1,6 @@
 require "kemal"
 require "pg"
+require "commander"
 
 # Compose Objects (like Hash) to have a to_json method
 require "json/to_json"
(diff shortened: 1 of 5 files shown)


frameworks/Crystal/kemal/server-postgres.cr
```diff
@@ -37,14 +38,19 @@ private def fortunes
...
 
+private def sanitized_concurrency(request)
+  concurrency = request.params.query["concurrency"]? || "1"
+  concurrency.to_i? || 1
+end
+
 before_all do |env|
   env.response.headers["Server"] = "Kemal"
-  env.response.headers["Date"] = HTTP.format_time(Time.now)
+  env.response.headers["Date"] = HTTP.format_time(Time.utc)
```
snluu: `Time.now` is deprecated in favor of `Time.utc`

PR #5381, opened 2020-01-09
Add new Go framework: Fiber

Fenny:
Hello,

I would like to add [Fiber](https://github.com/fenny/fiber) to the FrameworkBenchmarks.
Local benchmark passed all tests, let me know if I'm missing something.

Thank you!

PS: *@nbrady-techempower*, what an awesome project!


nbrady-techempower:
Hi @Fenny !

Welcome and thanks for your contribution and your kind words. It's a ton of effort and great work by the whole team here at TechEmpower and an incredible community of contributors!

This looks like a good start to me. Could add a `README.md` similar [to this one](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/dropwizard/README.md) with some information about the framework. Also, could you `go get` a specific git hash, either a release or a recent commit. We want to make sure that test results are in a reproducible state as much as possible. When you create new releases or updates, you can then update that here. This also helps ensure the test implementation doesn't break here randomly.

Fenny:
Something went wrong adding to this pull request, I had to create a new one [#5382](https://github.com/TechEmpower/FrameworkBenchmarks/pull/5382)

PR #5383, opened 2020-01-09
Add a benchmark of wizzardo-http with inline-java.

facundominguez:
Some benchmarks of an hybrid http server. wizzardo-http provides the server framework, but the handlers are written in Haskell and bridged to Java with inline-java.

These benchmarks are interesting to compare to warp, which is the Haskell framework for http servers.


facundominguez:
The source code of the benchmark lives in the inline-java [github repository](https://github.com/tweag/inline-java/tree/5b2552fae1beae88bebe2144da317e7182d759fc/benchmarks/wizzardo-http). They are easier to keep in sync with changes to inline-java if they are there. I could include a verbatim copy of the sources in this PR, if necessary.

nbrady-techempower:
It is necessary to keep the test implementation source code in this repo and that the test uses that source code. Any given run from a specific git hash in this repo should be reproducible and any changes to the test implementation need to be reviewed and pass travis before it can be merged. Also, could you please shorten the "framework name." `wizzardo-haskell` or `wizzardo-inline` would be preferred. I'll review again when the source is here. Thanks and welcome!

facundominguez:
Thanks for the feedback @nbrady-techempower 

Please, take another look.

nbrady-techempower:
This isn't passing travis.

```
wizzardo-inline: Step 11/12 : RUN ./keep_travis_alive.sh bash -c ". env-linear.sh; stack build wizzardo-http-benchmark"
wizzardo-inline:  ---> Running in be234b650e71
wizzardo-inline: ./keep_travis_alive.sh: 9: ./keep_travis_alive.sh: Syntax error: Bad for loop variable
wizzardo-inline: Removing intermediate container be234b650e71
wizzardo-inline: Docker build failed; terminating
wizzardo-inline: Traceback (most recent call last):
wizzardo-inline:   File "/FrameworkBenchmarks/toolset/utils/docker_helper.py", line 57, in __build
wizzardo-inline:     raise Exception(token['errorDetail']['message'])
wizzardo-inline: Exception: The command '/bin/sh -c ./keep_travis_alive.sh bash -c ". env-linear.sh; stack build wizzardo-http-benchmark"' returned a non-zero code: 2
```

facundominguez:
Yes. There is one package which takes more than 10 minutes to build.

facundominguez:
I found a way to avoid building the problematic library in the docker container. @nbrady-techempower, travis fixed :)

PR #5391, opened 2020-01-12
Elixir plug ecto

jtwebman:
I like Phoenix but it is overkill for simple APIs when all you need is Cowboy Plug. This also includes all the tests. This replaced the elixir-cowboy tests as they were also using cowboy plug but only had the JSON and plain text tests. Also looks like the Phoenix tests are failing. I will open a different PR to fix those.


.travis.yml
```diff
@@ -30,7 +30,7 @@ env:
     - "TESTLANG=D"
     - "TESTLANG=Dart"
     - "TESTLANG=Dylan"
-    - "TESTLANG=Elixir"
+    - "TESTDIR=Elixir/plug"
```
nbrady-techempower: Can you undo this change? Phoenix still lives in this folder and needs to be tested. The removed line will test all frameworks inside the Elixir folder so no need to update.

jtwebman: The issue is the Phoenix tests are failing at least on my local they are.

nbrady-techempower: I get that. Phoenix shouldn't be tested in this PR. I've written a script that only tests the frameworks of the files being touched so your PR will still pass. We need to make sure the travis.yml file covers all possible tests though as it's a master list.

jtwebman: I was about to open another PR to fix those. Would you rather it just be all one PR?

nbrady-techempower: That works fine for me

jtwebman: Ok will switch it to the folder only.

jtwebman:
When I switch it back my fork TravisCI only tested Phoenix. Am I missing something for it to test both?

nbrady-techempower:
The script can be a little finicky. I'll see what it does here. On your fork are you making a PR to master or are you just testing on a branch?

jtwebman:
Ok, no problem. I forced it to run again and it did the correct thing.

PR #5396, opened 2020-01-13
Fix Elixir Phoenix tests passing again.

jtwebman:
The tests were not always passing locally in docker and the update test was removed a month ago because it was failing. This PR fixes all the issues I found running locally.


nbrady-techempower:
Thanks for doing these @jtwebman !

PR #5397, opened 2020-01-13
Fix mysql start issues with mysql docker container vs build one.

jtwebman:
Maybe there is a reason you are using the built container vs the stock docker hub MySQL. Let me know and I can look into it more but this word for a few tests that I tested it on and makes it much simpler.

### DIFF
--- a/toolset/databases/mysql/mysql.dockerfile
+++ b/toolset/databases/mysql/mysql.dockerfile
@@ -1,44 +1,11 @@
-FROM buildpack-deps:bionic
+FROM mysql:8
 
-ADD create.sql create.sql
-ADD my.cnf my.cnf
-ADD mysql.list mysql.list
+EXPOSE 3306
 
-RUN cp mysql.list /etc/apt/sources.list.d/
-RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 8C718D3B5072E1F5
+ENV MYSQL_DATABASE=hello_world
+ENV MYSQL_USER=benchmarkdbuser
+ENV MYSQL_PASSWORD=benchmarkdbpass
+ENV MYSQL_ALLOW_EMPTY_PASSWORD=1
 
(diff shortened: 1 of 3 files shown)


nbrady-techempower:
I'll be thrilled if this works. A lot of the script to build this came from the legacy installation. Happy to help troubleshoot this if it looks viable.

toolset/databases/mysql/mysql.dockerfile
```diff
@@ -1,44 +1,11 @@
-FROM buildpack-deps:bionic
+FROM mysql:8
```
nbrady-techempower: If this does work, let's use the full patch version for reproducible results

jtwebman: I can make it 8.0.18 but 8 is running 8.0.18. When you use 8 it will keep up with the patches each time they release.

nbrady-techempower: Each commit in this repo needs to have reproducible results. If at some point, you want to rerun results for Round 18, the MySQL version used during the run and the version used when trying to reproduce results would be different. As much as possible we want to lock down specific versions. This is true across framework dependencies as well.

jtwebman:
@nbrady-techempower You might want to cancel the builds on Travis now as they are taking a long time because of the MySQL issue.

jtwebman:
@nbrady-techempower there were failures but I couldn't find any that was because they couldn't talk to the MySQL DB. Looks like a bunch of them fail.

jtwebman:
I found one issue but from what I can tell it is in the DB correctly with utf8_general_ci encoding. Seems like they are not handling correctly but seems strange.

jtwebman:
I am going to close this PR and open a new one with just some tweaks from the old file to make it work again.

nbrady-techempower:
@jtwebman I've fixed the issue with #5399 so it's not currently a pressing matter

PR #5400, opened 2020-01-15
Fix mysql start issue.

jtwebman:
Looks like this fixes the issue with start though this is not pinned to a specific version of MySQL like before so maybe we should work on getting the official MySQL versioned container working.

The current MySql version being installed is 8.0.19.


jtwebman:
Oh looks like you already fixed it, better than I.

PR #5402, opened 2020-01-15
add V

S-YOU:
Adding V, Thanks.

### DIFF
--- a/frameworks/V/pico.v/main.v
+++ b/frameworks/V/pico.v/main.v
@@ -1,0 +1,36 @@
+import json
+import syou.picoev
+import syou.picohttpparser as hp
+
+struct Message {
+	message string
+}
+
+[inline]
+fn json_response() string {
+	msg := Message{message: "Hello, World!"}
+	return json.encode(msg)
+}
+
+[inline]
+fn hello_response() string {
+	return "Hello, World!"
+}
+
+pub fn callback(req hp.Request, res mut hp.Response) {
+	if hp.cmp(req.method, "GET") {
+		if hp.cmp(req.path, "/t") {
+			res.http_ok().header_server().header_date().plain().body(hello_response())
+		} else if hp.cmp(req.path, "/j") {
+			res.http_ok().header_server().header_date().json().body(json_response())
+		} else {
+			res.http_404()
+		}
+	} else {
+		res.http_405()
+	}
+}
+
+pub fn main() {
+	picoev.new(8080, &callback).serve()
+}
--- a/frameworks/V/pico.v/pico.v.dockerfile
+++ b/frameworks/V/pico.v/pico.v.dockerfile
@@ -1,0 +1,15 @@
+FROM debian:sid
+
+WORKDIR /app
+
+RUN apt-get update -yqq && apt-get install wget unzip build-essential ca-certificates git --no-install-recommends --yes -yqq
+RUN wget -q https://github.com/vlang/v/releases/download/0.1.24/v_linux.zip && unzip -q v_linux.zip -d ./v/ && rm v_linux.zip
+
+COPY ./main.v run.sh ./
+
+RUN git clone https://github.com/S-YOU/pico.v src && cd src && git checkout v0.0.3
+RUN cd src/picoev && git clone https://github.com/S-YOU/picoev src && cd src && git checkout 311fb2c2a148baca7b7fad33afa94e4558c63478
+RUN cd src/picohttpparser && git clone https://github.com/h2o/picohttpparser src && cd src && git checkout 81fe3d99fd90a55cafb993e53fd3000dbc4d564c
+RUN mkdir -p /root/.vmodules/ && ln -s /app/src /root/.vmodules/syou && ./v/v -prod -cflags '-std=gnu11 -Wall -O3 -march=native -mtune=native -flto' main.v
+
+CMD sh run.sh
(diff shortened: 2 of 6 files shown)


S-YOU:
If someone cloud take a look and merge, that would be great, thanks. @michaelhixson, @nbrady-techempower?

dhonx:
+1

frameworks/V/pico.v/pico.v.dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM debian:sid
+
+WORKDIR /app
+
+RUN apt-get update && apt-get install wget unzip build-essential ca-certificates --no-install-recommends --yes
```
nbrady-techempower: can you add -yqq to update and install to quiet them a bit

frameworks/V/pico.v/pico.v.dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+RUN wget https://github.com/vlang/v/releases/download/0.1.24/v_linux.zip && unzip v_linux.zip -d ./v/
```
nbrady-techempower: a `-q` here as well for the noise

frameworks/V/pico.v/main.v
```diff
@@ -0,0 +1,37 @@
+import json
+import syou.picoev
+import syou.picohttpparser as hp
+
+struct Message {
+	message string
+}
+
+[inline]
```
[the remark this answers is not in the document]
S-YOU: it is static inline equivalent of C, it does inline function but not JSON encoding part.

S-YOU:
> Also, it seems like you're adding the source for some packages here. Can you pull them in from github instead and checkout a specific git hash
> 
> edit:
> `git clone` these items:
> 
> * [picoev](https://github.com/kazuho/picoev)
> * [picohttpparser](https://github.com/h2o/picohttpparser/)

Unfortunately, those are C libraries I have based on, but C does not work directly out of the box with V.
And V packages are not yet ready to released to be standalone. So, I have to keep those like that.

frameworks/V/pico.v/pico.v.dockerfile
S-YOU: done

frameworks/V/pico.v/pico.v.dockerfile
S-YOU: done

nbrady-techempower:
I'm sorry, I'm not understanding. Is this file included in your PR from this repo? https://github.com/kazuho/picoev/blob/master/picoev_epoll.c

nbrady-techempower:
What im asking for is for you to `RUN git clone https://github.com/kazuho/picoev` and `RUN git checkout ff85d9e` instead of including those files directly in this PR

S-YOU:
> What im asking for is for you to `RUN git clone https://github.com/kazuho/picoev` and `RUN git checkout ff85d9e` instead of including those files directly in this PR

ok, I will try. I modified some of those files, so I was not sure it will work out of the box.

nbrady-techempower:
> ok, I will try. I modified some of those files, so I was not sure it will work out of the box.

Oh, ok then. I thought this was part of the framework you were adding. Is there a repo ready for your framework? It might be better to fork that repo with your changes and then pull it in here.

S-YOU:
> > ok, I will try. I modified some of those files, so I was not sure it will work out of the box.
> 
> Oh, ok then. I thought this was part of the framework you were adding. Is there a repo ready for your framework? It might be better to fork that repo with your changes and then pull it in here.

I can now git clone those C libraries. Seems working. @nbrady-techempower 
Regarding the framework, it is not yet ready to be released as standalone, I am adding this as stripped.

S-YOU:
Oh, stripped will not shown by default on the benchmarks? what is the guideline of stripped and non-stripped?

S-YOU:
Changing to realistic, I didn't know it will not shown on the benchmark list by default. Let me know if there is any issue with that.

S-YOU:
> Is there a repo ready for your framework? It might be better to fork that repo with your changes and then pull it in here.

This is done. Please take a look again. Thanks @nbrady-techempower

S-YOU:
Did a little mistake. rebasing. Hopefully this should be ok

S-YOU:
I have done updating @nbrady-techempower 
Please take a look again. Thanks

S-YOU:
seems like I was not using main.v from this PR. updated again.

S-YOU:
Ah, I forgot that I have to modified one file of picoev, because it segfault with pipelining, I will modify docker file again. sorry for taking your time.

nbrady-techempower:
@S-YOU It's no problem. Take your time. I won't be able to review again until later today.

S-YOU:
@nbrady-techempower I think I've done updating. Please take a look again when you have time.
I am here at early morning, so probably I will reply later if anything I need to update. Thanks.

Fenny:
I'm excited about `V`!

S-YOU:
Did some tuning and rebased.

frameworks/V/pico.v/main.v
```diff
@@ -0,0 +1,36 @@
...
+fn hello_response() string {
+	return "Hello, World!"
+}
+
+pub fn callback(req hp.Request, res mut hp.Response) {
+	if hp.cmp(req.method, "GET") {
+		if hp.cmp(req.path, "/t") {
+			res.http_ok().header_server().header_date().plain().body(hello_response())
+		} else if hp.cmp(req.path, "/j") {
```
vlazar: I wonder if that's the reason why pico.v tops the benchmark.

> The recommended URI is /json.

Not a must, but other implementations use `/json` URL.

PR #5407, opened 2020-01-16
warp-rust 0.2

xfix:
Upgrading warp to 0.2.0. I plan adding more paths with addition of `async`/`await` support in warp 0.2.0, but I figured I may as well start with upgrading the version, as it's a big change.

### DIFF
--- a/frameworks/Rust/warp-rust/src/main.rs
+++ b/frameworks/Rust/warp-rust/src/main.rs
@@ -7,15 +7,16 @@ struct Message {
     message: &'static str,
 }
 
-fn main() {
-    let json = warp::path("json").map(|| {
+#[tokio::main]
+async fn main() {
+    let json = warp::path!("json").map(|| {
         warp::reply::json(&Message {
             message: "Hello, world!",
         })
     });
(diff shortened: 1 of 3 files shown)


frameworks/Rust/warp-rust/src/main.rs
```diff
@@ -7,15 +7,16 @@ struct Message {
     message: &'static str,
 }
 
-fn main() {
-    let json = warp::path("json").map(|| {
+#[tokio::main]
+async fn main() {
+    let json = warp::path!("json").map(|| {
```
nbrady-techempower: What's the difference between the method `path` and the macro `path!` in this context? I'd like to know what rust code gets generated from this macro. Will the resulting code still instantiate and serialize the JSON response on every call?

xfix: In short, it's a syntactic sugar for `warp::path("json").and(warp::end())`. Nothing to do with performance, just a correctness fix (previous implementation allowed paths like `/json/abc/def`.

nbrady-techempower: No, that's fine. Thanks for the explanation.

PR #5420, opened 2020-01-21
Update V deps

S-YOU:
Update V deps and did some more tuning including following.
- code formatted by new v fmt
- changed the listening port
- using base image with upcoming v already in it
- updated dependencies

### DIFF
--- a/frameworks/V/pico.v/pico.v.dockerfile
+++ b/frameworks/V/pico.v/pico.v.dockerfile
@@ -1,12 +1,12 @@
-FROM syou/v-dev:latest
+FROM syou/v-dev:0.1
 
 WORKDIR /app
 
 COPY ./main.v run.sh ./
 
-RUN git clone https://github.com/S-YOU/pico.v src && cd src && git checkout dev \
-    && cd /app/src/picoev && git clone https://github.com/S-YOU/picoev src && cd src && git checkout dev \
-    && cd /app/src/picohttpparser && git clone https://github.com/S-YOU/picohttpparser src && cd src && git checkout dev \
+RUN git clone https://github.com/S-YOU/pico.v src && cd src && git checkout v0.0.4 \
+    && cd /app/src/picoev && git clone https://github.com/S-YOU/picoev src && cd src && git checkout v0.0.1 \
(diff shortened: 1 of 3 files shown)


S-YOU:
Please take a look when you have time @nbrady-techempower . Thanks

frameworks/V/pico.v/pico.v.dockerfile
```diff
@@ -1,15 +1,13 @@
-FROM debian:sid
+FROM syou/v-dev:0.1
```
nbrady-techempower: We prefer official bases so we can see in the dockerfile here all the steps taken to get your framework built and running. Also, it helps with the docker cache and testing speed in our main environment.

S-YOU: @nbrady-techempower , Framework is built under that in pico.v.dockerfile, docker image is the compiling V language itself with some commits picked from master.

S-YOU: But just let me know if you insist @nbrady-techempower . I could just go ahead and copy paste parent docker commands.

nbrady-techempower: This is ok for now, then. If part of updating the framework means you'll be updating this base, when you do that, can you please copy and paste the commands in a comment here for quick review. We want to make sure that it's not always pulling the latest commit, etc. There's just some things we need to audit.

S-YOU: here is the base docker file on docker hub - https://hub.docker.com/r/syou/v-dev/dockerfile
and it is tagged as 0.1, so unless there is V language update, there will be no addition pull - https://hub.docker.com/r/syou/v-dev/tags

nbrady-techempower: > when there is framework update, line 7 or 8 or 9 in pico.v.dockerfile will be updated to specific tag. like, git checkout v0.0.4 -> git checkout v0.0.5

Yes, this is my point. We want to make sure that the base you're pulling from isn't pulling the latest from master inside the docker container; we clear the docker cache every few runs. When I talk about the docker cache, I mean it in the context of a single run, such that a lot of frameworks share the same base like `buildpack-deps:xenial` and the first few lines are often the same, like apt updating.

PR #5429, opened 2020-01-24
Add Scala framework: Scalene

DanSimon:
Hi, this adds tests for the Scala framework Scalene.

### DIFF
--- a/frameworks/Scala/scalene/src/main/scala/Benchmark.scala
+++ b/frameworks/Scala/scalene/src/main/scala/Benchmark.scala
@@ -1,0 +1,83 @@
+package scalene.benchmark
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.scala.DefaultScalaModule
+import scalene.actor.Pool
+import scalene.routing._
+import scalene.http.{BodyData, BodyFormatter, ContentType}
+import scalene.sql._
+import BasicConversions._
+
+object Main extends App {
+
+  trait JsonMessage
+  case class JsonRouteMessage(message: String) extends JsonMessage
+  case class DBRouteMessage(id: Int, randomnumber: Int) extends JsonMessage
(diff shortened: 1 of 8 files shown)


frameworks/Scala/scalene/src/main/scala/Benchmark.scala
```diff
@@ -0,0 +1,83 @@
...
+
+  val multiRoute = GET / "queries" / QueryNum to {num =>
+    worldClient.query{session =>
+      MultiDBRouteMessage(Array.fill(num)(randomWorld(session).get)).ok
+    }
+  }
+
+  val routes = Routes(
+    GET / "plaintext" as "Hello, World!".ok,
+    GET / "json"      as JsonRouteMessage("Hello, World!").ok,
```
nbrady-techempower: The JSON serialization needs to happen on every request for this endpoint. Here, it is being serialized when the server starts.

DanSimon: ah good catch.  Fixed it!

frameworks/Scala/scalene/src/main/scala/Benchmark.scala
```diff
@@ -0,0 +1,83 @@
...
+    .recover{_ => 1}
+
+  val multiRoute = GET / "queries" / QueryNum to {num =>
+    worldClient.query{session =>
+      MultiDBRouteMessage(Array.fill(num)(randomWorld(session).get)).ok
+    }
+  }
+
+  val routes = Routes(
+    GET / "plaintext" to {_ => "Hello, World!".ok},
```
nbrady-techempower: For the record, this one was fine.

DanSimon: Good to know, I'll leave it for now as I don't think it will make much of a difference anyway.

PR #5432, opened 2020-01-26
Improvements to Elixir benchmarks

josevalim:
We increase the pool_size to 40. The original pool_size was
 way too low, which was proved by the fact we had to increase
 the queue_target and queue_interval, which means we were often
 waiting more than 5s to get hold of a connection. Also see
 the discussion in #5153. We also rollback the original queue
 values. If we are getting more timeouts, the best solution is
 most likely to bump pool size again.

### DIFF
--- a/frameworks/Elixir/phoenix/config/prod.exs
+++ b/frameworks/Elixir/phoenix/config/prod.exs
@@ -15,7 +15,7 @@ config :hello, Hello.Repo,
   database: "hello_world",
   hostname: "tfb-database",
   pool_size: 40,
-  queue_target: 500,
+  queue_target: 5000,
   log: false
 
 config :logger,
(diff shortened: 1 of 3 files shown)


nbrady-techempower:
@josevalim The phoenix update test looks like it's possibly doing some db caching (or another issue) as it's not passing verification. Would you be able to take a look at that?

josevalim:
@nbrady-techempower I have just revert those changes to the code.

nbrady-techempower:
@josevalim Just a reminder about those update tests before I can merge this.

josevalim:
@nbrady-techempower I have reverted the changes to update. Unless the caching was introduced in another PR and is being shown here.

josevalim:
@nbrady-techempower please ignore my previous comment. :) The issue was not caching but there was an error because the queue was too busy. Is CI actually running the benchmarks? If so, we may want to increase the pool size.

nbrady-techempower:
@josevalim Travis isn't running the full benchmarks, just a verification before getting into our continuous benchmarking environment here: https://tfb-status.techempower.com

josevalim:
@nbrady-techempower it should be good now!

PR #5436, opened 2020-01-28
Update aspnetcore-corert to .NET Core 3.1

jkotas:


jkotas:
cc @MichalStrehovsky

jkotas:
FWIW, it was not the 3.0 image that had the problem. The problem was that the self-contained runtime that comes with the compiler did not work on newer Linux versions. Picking up latest servicing version of the self-contained runtime in https://github.com/dotnet/corert/pull/7932 fixed it.

MichalStrehovsky:
> Picking up latest servicing version of the self-contained runtime in dotnet/corert#7932 fixed it.

Ah, missed that PR! Thanks!

PR #5455, opened 2020-02-02
Lithium: update and tune sql connections.

matt-42:


jcheron:
Hi @matt-42 
The **Lithium-postgres** approach on the update test is more **Raw** than **Full-orm**. It is not instances that are updated, but queries that are executed.
https://github.com/TechEmpower/FrameworkBenchmarks/blob/cc209d91295273252f7e0ecface4d9da3a8aa67c/frameworks/C%2B%2B/lithium/lithium.cc#L159-L172

This does not correspond to the classification chosen for this test:
image

matt-42:
@jcheron this will be fixed in my next PR.

jcheron:
For the time being, all you have to do is change this line...
https://github.com/TechEmpower/FrameworkBenchmarks/blob/c6c77f5e79d9decd71715262ac4744655c863fcf/frameworks/C%2B%2B/lithium/benchmark_config.json#L42

PR #5461, opened 2020-02-03
Phalcon update to v4.0 and php 7.4

joanhey:
And changes to work with v4.0.

In another PR I will try to fix the micro and mongodb variants.


joanhey:
Phalcon is working, only fail micro and mogodb.
When I have time I will change the code to v4.0 in micro and mongodb.
Actually Phalcon in all variants, fail to start in the last 3 runs.

PR #5467, opened 2020-02-04
[add] Implementation of JSON/plaintext for Servicetalk

arthur-techempower:
Initial implementation of PlainText and JSON tests for Servicetalk framework. 
Changes to reflect Nitesh's implementation suggestions.


frameworks/Java/servicetalk/src/main/java/servicetalk/Server.java
```diff
@@ -0,0 +1,100 @@
...
+
+public final class Server {
+
+
+    public static void main(String[] args) throws Exception {
+
+        /**
+         * Disable  AsyncContext
+         */
+        AsyncContext.disable();
```
michaelhixson: For comments within Java methods, use `// comments like this` or `/* comments like this */`, not `/** javadoc like this */`.  The double asterisk is for javadoc, which can be applied at the class level or member (field or method) level, but not within a method.

frameworks/Java/servicetalk/src/main/java/servicetalk/Server.java
```diff
@@ -0,0 +1,100 @@
...
+                      return succeeded(responseFactory.ok()
+                        .payloadBody("Hello, World!", HttpSerializationProviders.textSerializer())
+                        .addHeader("Date", GetCurrentTime())
+                        .addHeader("Server", "TFB"));
+                    };
+                    return null;
+                })
+                .awaitShutdown();
+        } catch (Exception e) {
+//            logging is not allowed
```
michaelhixson: You might as well not have the `try` and `catch` then.  Your main method declared `throws Exception`, so no good comes out of catching this.  The only thing it can do is hide errors from us.

frameworks/Java/servicetalk/src/main/java/servicetalk/Server.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public static String GetCurrentTime() {
+        SimpleDateFormat formatter = new SimpleDateFormat("EE, dd MMM yyyy kk:mm:ss z  ");
+        Date date = new Date();
+        return formatter.format(date);
+    }
+
+    private static int GetRandomNumber() {
+        Random random = new Random();
+        return random.nextInt(10001) + 1;
```
michaelhixson: This can return 10001, which is an invalid world id.  Also, you should use `ThreadLocalRandom`, which shares one `Random` instance with each thread instead of constructing a new `Random` instance for every number.  Do something like this instead:  https://github.com/TechEmpower/FrameworkBenchmarks/blob/f510e947b5480b83cdda66f397d31be7166a40d6/frameworks/Java/undertow/src/main/java/hello/HelloWebServer.java#L255

frameworks/Java/servicetalk/src/main/java/servicetalk/Server.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+
+    public static String GetCurrentTime() {
+        SimpleDateFormat formatter = new SimpleDateFormat("EE, dd MMM yyyy kk:mm:ss z  ");
+        Date date = new Date();
+        return formatter.format(date);
+    }
+
+    private static int GetRandomNumber() {
```
michaelhixson: The near-universal convention in Java is that methods are `lowerCamelCase`, not `UpperCamelCase`.

frameworks/Java/servicetalk/src/main/java/servicetalk/Server.java
```diff
@@ -0,0 +1,100 @@
...
+                })
+                .awaitShutdown();
+        } catch (Exception e) {
+//            logging is not allowed
+        }
+    }
+
+    public static String GetCurrentTime() {
+        SimpleDateFormat formatter = new SimpleDateFormat("EE, dd MMM yyyy kk:mm:ss z  ");
+        Date date = new Date();
```
michaelhixson: `Date` and `SimpleDateFormat` are part of the legacy Java time libraries, which were effectively replaced by the new `java.time` libraries in Java 8.  Here, what you're looking for is `DateTimeFormatter`.  Other frameworks show how to use it.  Example:  https://github.com/TechEmpower/FrameworkBenchmarks/blob/b843c9ea4f9096435ae52bf99b0182737a2c1edc/frameworks/Java/armeria/src/main/java/hello/App.java#L25-L27

Note that `DateTimeFormatter` is immutable, so a single instance can be reused every time you want to format a date.  And luckily there is already a built-in `DateTimeFormatter` constant that has the format you want here.  One of `SimpleDateFormat`'s problems is that it is mutable -- formatting a date modifies the `SimpleDateFormat` instance -- so you have to construct a new instance for each date like this.

frameworks/Java/servicetalk/src/main/java/servicetalk/Server.java
```diff
@@ -0,0 +1,100 @@
...
+                        .payloadBody(obj, serializer.serializerFor(Map.class))
+                        .addHeader("Date", GetCurrentTime())
+                        .addHeader("Server", "TFB"));
+                  }
+
+                  if(request.path().equals("/plaintext")) {
+                      return succeeded(responseFactory.ok()
+                        .payloadBody("Hello, World!", HttpSerializationProviders.textSerializer())
+                        .addHeader("Date", GetCurrentTime())
+                        .addHeader("Server", "TFB"));
```
michaelhixson: `"TFB" --> "ServiceTalk"`.  The convention we use is that the`Server` header contains the name of the framework.

frameworks/Java/servicetalk/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+            </plugin>
+        </plugins>
+    </build>
+
+    <repositories>
+        <repository>
+            <id>maven-central</id>
+            <url>https://repo1.maven.org/maven2/</url>
+        </repository>
+    </repositories>
```
michaelhixson: This `<repositories>` block looks unnecessary.  Does it build without this?  If so, please remove it.

frameworks/Java/servicetalk/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+    <artifactId>servicetalk</artifactId>
+    <version>1.0-SNAPSHOT</version>
+    <packaging>jar</packaging>
+
+
+    <dependencies>
+        <dependency>
+            <groupId>io.servicetalk</groupId>
+            <artifactId>servicetalk-serialization-api</artifactId>
+            <version>0.20.0</version>
```
michaelhixson: Usually when we have a bunch of related dependencies that are supposed to all have the same version like this, we extract that version number into a property. Please do that here.

```xml
<properties>
  <servicetalk.version>0.20.0</servicetalk.version>
</properties>

<dependencies>
  <dependency>
    <groupId>io.servicetalk</groupId>
    <artifactId>servicetalk-http-api</artifactId>
    <version>${servicetalk.version}</version>
  </dependency>
  ...
</dependencies>
```

Often times projects will extract all dependency version numbers into properties even if each of those version numbers is only referenced once in the pom file.

frameworks/Java/servicetalk/pom.xml
arthur-techempower: I would need to double check this tomorrow, as I believe it points to a non-default repository.

NiteshKant:
@arthur-techempower apologies for not being able to provide feedback on this.
I finally have permissions to engage into discussions here 🎉 , are the benchmark results published for ServiceTalk yet?

nbrady-techempower:
@NiteshKant Results can be viewed from our continuous runs here https://tfb-status.techempower.com/

NiteshKant:
Thanks @nbrady-techempower !

PR #5469, opened 2020-02-06
Bumping version of Finagle and Finatra to 20.1.0.

mosesn:
We released a new version, let's bump the version in the benchmarks!


mosesn:
thank you!

PR #5470, opened 2020-02-07
Actix: use snmalloc allocator

fafhrd91:
* use snmalloc allocator


frameworks/Rust/actix/actix-core.dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM rust:1.40
+FROM rust:1.41
+
+RUN apt update -yqq && apt install -yqq cmake g++
```
davidchisnall: Note that most of our snmalloc perf work has been done assuming that it is compiled with clang or Visual Studio (and some of it is susceptible to inlining decisions).  I have no idea how this changes when you compile with gcc instead.  If you have comparison numbers for snmalloc compiled with clang vs gcc, I would be interested in them.

davidchisnall: It looks as if the fast path on `rust_alloc` and `rust_dealloc` are one instruction longer with gcc (16 vs 15, 26 vs 25 x86-64 instructions), but with a quite different instruction mix so this is not necessarily slower.

fafhrd91: i dont think we would notice 16 vs 15 instructions difference for io workload. but from what i've read snmalloc perfectly match actix, actix is single threaded.

PR #5475, opened 2020-02-10
Symfony : Ubuntu 19.04 -> 19.10

jcheron:
### Fixed
**Symfony** and **Symfony-raw** not starting...
 - no `php-7.4` package in **Ubuntu 19.04** (without adding ppa:ondrej/php repo)


joanhey:
All use ppa:ondej/php repo.
But it's better to all use the same OS version too.

PR #5479, opened 2020-02-10
Revert "Php use 2048 child processes"

joanhey:
Reverts TechEmpower/FrameworkBenchmarks#5457

Some changes in some stacks are worst for perfomance, after the last patches for the "TCP SACK PANIC".


joanhey:
Anyway, I hope to have almost 6-7 php platforms in the top 20 fortunes tests.
Who said that php is slow ??

Later I will try with the frameworks using that platforms.

PR #5490, opened 2020-02-16
Bumb fiber v1.0.1

Fenny:


nbrady-techempower:
These tests never ran @Fenny - would you mind taking a look?

Fenny:
@nbrady-techempower, there was a bug that only happens in a docker container.
```
Verification Summary
| fiber
|       plaintext     : PASS
|       query         : PASS
|       json          : PASS
|       db            : PASS
|       update        : PASS
```

PR #5497, opened 2020-02-18
Upgraded version of ffead-cpp with latest changes

sumeetchhetri:


sumeetchhetri:
@nbrady-techempower, can you please merge this?

frameworks/C++/ffead-cpp/te-benchmark-um/src/TeBkUm.cpp
```diff
@@ -0,0 +1,312 @@
...
+	try {
+		sqli->startSession(&WORLD);
+		for (int c = 0; c < queryCount; ++c) {
+			int rid = rand() % 10000 + 1;
+			GenericObject id;
+			id << rid;
+			TeBkUmWorld w = sqli->get<TeBkUmWorld>(id);
+			int newRandomNumber = rand() % 10000 + 1;
+			if(w.getRandomNumber() == newRandomNumber) {
+				newRandomNumber -= 1;
```
nbrady-techempower: If the randomNumber is 1, this will set it to 0 and cause an error in the verification. This would be rare, but it could happen during an official run.

PR #5503, opened 2020-02-20
Add nestjs framework with fastify

werdes72:
I noticed that some frameworks are not listed in .travis.yml but their languages are. I skipped adding this framework there, hoping that everything under `- "TESTLANG=TypeScript"` will run automatically. Please let me know if I'm wrong.


nbrady-techempower:
That is correct @werdes72, however, since this is just a variation of an existing framework, I would rather this go in the `Typescript/nest` folder. You can add a new entry in the current benchmark_config called `fastify` and the tests will show up as `nest-fastify`. The dockerfile would be `nest-fastify.dockerfile`. Let me know if you need any other help doing that.

werdes72:
I was able to merge those frameworks with some changes and updates.

PR #5537, opened 2020-03-09
Bump Finagle and Finatra versions

yufangong:
Finagle to 20.3.0
Finatra to 20.3.0


yufangong:
There are some conflicts with the cached content. Does it make sense to clean the cache and rebuild it?

nbrady-techempower:
Which cache are you referring to? I've rerun this on travis a couple of times and it's not passing. Will need an update.

yufangong:
@nbrady-techempower Thanks Nate, it turns out the diverged dependency is valid, I've changed some merge strategy accordingly, PTAL.

PR #5544, opened 2020-03-13
Yii2: fix a bug in the parsing of integer GET parameters

mytskine:
Without this fix, the yii2 and yii2-raw applications accept values like "2.5" or "+9.876e2". Non-integer floats greater than 1 triggered infinite loops in routes /queries and /updates.


nbrady-techempower:
@mytskine Query and Update tests are failing. Would you mind looking into those before this gets merged?

PR #5551, opened 2020-03-17
Fix phalcon-micro to work with v4 and php7.4

joanhey:
Still fail phalcon-mongo, when I have time I'll fix it.

Ready to be merged.


joanhey:
It's ready to be merged.

PR #5561, opened 2020-03-21
Refine the HTTP parser of gnet

panjf2000:


panjf2000:
Why didn't it start a CI automatically?

@nbrady-techempower

nbrady-techempower:
@panjf2000 Not sure; sometimes it just misses. Looks good now.

PR #5567, opened 2020-03-25
Lwan: Add Lua benchmarks

lpereira:
This adds Lua versions of the JSON and Plaintext benchmarks in Lwan.


frameworks/C/lwan/techempower.conf
```diff
@@ -1,2 +1,28 @@
...
+    &plaintext /plaintext
+    &json /json
+    &db /db
+    &queries /queries
+    &fortunes /fortunes
+
+    # For Lua version of TWFB benchmarks
+    lua /lua. {
+	default_type = text/plain
+	cache period = 1h
```
nbrady-techempower: Hey @lpereira! What's this doing? Is it caching the lua script?

lpereira:
It is caching the Lua state for 1 hour, so that the work performed by
LuaJIT isn't thrown away after it does its thing. (It's not caching
responses or anything like that.)

PR #5577, opened 2020-03-30
actix: revert last change

fafhrd91:


fafhrd91:
I am not sure what is wrong with status check, but travis job is green

PR #5578, opened 2020-04-02
Updated the single and mulitple db queries.

muraray:
Just amending the existing Nim-Jester framework to include the single and multiple DB queries with Postgres database.

Have updated the necessary Reademe.md file as well for better understanding.

### DIFF
--- a/frameworks/Nim/jester/README.md
+++ b/frameworks/Nim/jester/README.md
@@ -16,3 +16,11 @@ http://localhost:8080/json
 ### PLAINTEXT
 
 http://localhost:8080/plaintext
+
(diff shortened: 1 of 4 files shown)


nbrady-techempower:
@muraray Not sure what you're asking but looking at the logs:

```
jester: could not load: libpq.so(.5|)
jester: compile with -d:nimDebugDlOpen for more information
jester: ERROR: Framework is not accepting requests from client machine
```

Seems you need to add the package for `libpq`.

nbrady-techempower:
@muraray I'm going to close this for now. Please re-open if you have time to work on it.

PR #5579, opened 2020-04-03
ASP.NET Core Platform Sync

benaadams:
The two repositories had become desynchronized

Corresponding ASP.NET Core sync https://github.com/aspnet/Benchmarks/pull/1472

/cc @adamsitnik @sebastienros @halter73

### DIFF
--- a/frameworks/CSharp/aspnetcore/PlatformBenchmarks/AsciiString.cs
+++ b/frameworks/CSharp/aspnetcore/PlatformBenchmarks/AsciiString.cs
@@ -12,6 +12,8 @@ namespace PlatformBenchmarks
 
         public AsciiString(string s) => _data = Encoding.ASCII.GetBytes(s);
 
+        private AsciiString(byte[] b) => _data = b;
+
         public int Length => _data.Length;
 
         public ReadOnlySpan<byte> AsSpan() => _data;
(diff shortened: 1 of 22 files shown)


sebastienros:
It failed for me on the aspnet repos as the MySql assembly is used in `RawDb.cs` but the conditional directive is not used.

Also IMO kind of risky to change at that point since it will only improve plaintext which is already saturated on network on TE machines.

benaadams:
> It failed for me on the aspnet repos as the MySql assembly is used in RawDb.cs but the conditional directive is not used.

They aren't conditionally included here; aspnet repo has a more complex reference include policy that I may have not got right. All tests pass for TE including the MySql ones.

> it will only improve plaintext 

Also will effect JSON from one of the earlier changes https://github.com/aspnet/Benchmarks/pull/1464

> which is already saturated on network on TE machines.

YMMV

image

sebastienros:
Off topic: pico.v results is a mistery. Ulib can reach 15M on our machines with two clients. Maybe there is something in the payload that differs, like less headers, smaller url (`/p`, `/j`) and at this rate it matters.

sebastienros:
OK for the IsDatabase, that's what i ended up doing in aspnet.

sebastienros:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/V/pico.v/benchmark_config.json#L6-L7

Header sizes might also explain the order of the first 10 results.

PR #5586, opened 2020-04-05
Fix kumbiaphp

joanhey:


joanhey:
Will be ready for Round 19 ?

joanhey:
I didn't notice the stop for Round 19.

The next time could you add a message in https://tfb-status.techempower.com/ at the top.

Thank you

nbrady-techempower:
It's not a hard stop. We're trying to look into another issue before we completely lock it down. Your best bet is to subscribe to that pinned issue but I'll try and get a notice up shortly.

PR #5599, opened 2020-04-13
Clean Fat free code

joanhey:
Use argument unpacking to have less code and be faster.
Format the code.

### DIFF
--- a/frameworks/PHP/fat-free/index.php
+++ b/frameworks/PHP/fat-free/index.php
@@ -25,7 +25,7 @@ $f3->set('DBS', array('mysql:host=tfb-database;port=3306;dbname=hello_world', 'b
 $f3->route('GET /json', function ($f3) {
     /** @var Base $f3 */
     header('Content-type: application/json');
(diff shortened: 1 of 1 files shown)


joanhey:
@nbrady-techempower  exist a problem creating the mysql docker.

https://travis-ci.org/github/TechEmpower/FrameworkBenchmarks/jobs/674552551#L772

And the same error when php try to connect:
https://travis-ci.org/github/TechEmpower/FrameworkBenchmarks/jobs/674552551#L772

joanhey:
I'll try again.

joanhey:
Now passed.
But the error still show creating the mysql docker.
Also in older builds.

nbrady-techempower:
@joanhey 
```
techempower/mysql:latest: ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)
```

Comes from the loop before where we test every second whether or not the MySQL server is up and running until we no longer get an error. Shouldn't have any affect on other errors you enouncter.

PR #5602, opened 2020-04-15
Add 60 seconds sleep to clear host connections

nbrady-techempower:
tl;dr We believe that some frameworks are hitting the OS limit for open connections and, if the next test spins up too fast, it may have trouble creating network connections. Because we're using `--host`, these connections are lingering on the host and stuck in a `TIME_WAIT` state that automatically close after a default of 60 seconds. Here, we're giving enough time for the host to close these connections before the next test starts. 

🤞 that this solves #4092 

Edit: Changed my mind about where to put this. Rather than before a test container starts, I put it in the `__exit_test` cleanup. This makes sure building database containers and all the other things that need to happen at the start of a test can reach the interwebs.

### DIFF
--- a/toolset/benchmark/benchmarker.py
+++ b/toolset/benchmark/benchmarker.py
@@ -141,6 +141,9 @@ class Benchmarker:
                 self.time_logger.mark_started_database()
 
             # Start webapp
+            # Wait 60 seconds before starting webapp to ensure all host connections are closed
+            log("Sleeping 60 seconds...", prefix=log_prefix)
+            time.sleep(60)
             container = test.start()
             self.time_logger.mark_test_starting()
             if container is None:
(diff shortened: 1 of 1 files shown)


toolset/benchmark/benchmarker.py
```diff
@@ -92,6 +92,9 @@ def __exit_test(self, success, prefix, file, message=None):
                 file=file,
                 color=Fore.RED if success else '')
         self.time_logger.log_test_end(log_prefix=prefix, file=file)
+        # Sleep for 60 seconds to ensure all host connects are closed
+        log("Clean up: Sleep 60 seconds...", prefix=prefix, file=file)
+        time.sleep(60)
```
msmith-techempower, line 97: Even if it's the only one to run?

nbrady-techempower: Feel free to crtl-C if it's the only one run. Really, I just care about getting this on citrine and seeing if it solves the issues. Then I'll worry about QoL stuff.

msmith-techempower: Fair enough.

PR #5604, opened 2020-04-15
Some QoL Stuff

nbrady-techempower:
- [x] Only sleep for 60 seconds if we're in `benchmark` mode. This should fix some travis timeout issues
- [x] Fix the "no default test for lithium" warnings
- [x] `apt-get` better suited for scripts than `apt` 
- [x] tag some frameworks that are broken in travis and citrine

### DIFF
--- a/frameworks/PHP/peachpie/benchmark_config.json
+++ b/frameworks/PHP/peachpie/benchmark_config.json
@@ -22,7 +22,8 @@
       "database_os": "Linux",
       "display_name": "peachpie",
       "notes": "",
-      "versus": "php"
+      "versus": "php",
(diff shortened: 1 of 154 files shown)


frameworks/PHP/peachpie/benchmark_config.json
```diff
@@ -22,7 +22,8 @@
       "database_os": "Linux",
       "display_name": "peachpie",
       "notes": "",
-      "versus": "php"
+      "versus": "php",
+      "tags": ["broken"]
```
jakubmisek, line 26: May I ask what was broken? We're maintaining this framework for tfb

PR #5613, opened 2020-04-18
actix: update json benchmarks

fafhrd91:
use simd-json for json serialization

### DIFF
--- a/frameworks/Rust/actix/Cargo.toml
+++ b/frameworks/Rust/actix/Cargo.toml
@@ -44,6 +44,7 @@ bytes = "0.5.3"
 num_cpus = "1.0"
 futures = "0.3.1"
 http = "0.2"
+simd-json = "0.3.8"
 diesel = { version = "1.4.3", features = ["postgres"] }
 url = { version="1.7", features=["query_encoding"] }
 log = { version = "0.4", features = ["release_max_level_debug"] }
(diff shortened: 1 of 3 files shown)


frameworks/Rust/actix/src/main.rs
```diff
@@ -20,7 +20,7 @@ async fn json() -> HttpResponse {
         message: "Hello, World!",
     };
     let mut body = BytesMut::with_capacity(SIZE);
-    serde_json::to_writer(Writer(&mut body), &message).unwrap();
+    simd_json::to_writer(Writer(&mut body), &message).unwrap();
```
Licenser, line 23: it would be better to use `simd-json-derive::Serialize` and `message.json_write` it is 2-3x faster then serde::Serialize.

`simd_json::to_writer` is a compatibility method that implements the slower serde-trait for drop-in replacement w/o code changes. Given it's a benchmark code changes are OK I think especially if it's about 2 lines :D.

That said the `simd-json-derive::Serialize` is still young and only exists as a git dependency, not as a crate yet.

fafhrd91: @Licenser thanks for the tip!

Licenser:
Would it make sense to update the other Rust frameworks as well?

fafhrd91:
I'd see result first then ping maintainers.

Licenser:
makes sense :)

frameworks/Rust/actix/Cargo.toml
```diff
@@ -44,6 +44,8 @@ bytes = "0.5.3"
 num_cpus = "1.0"
 futures = "0.3.1"
 http = "0.2"
+simd-json = { git = "https://github.com/simd-lite/simd-json.git" }
```
Licenser, line 47: This should work with the crate :) it's compatible with the derive

```suggestion
simd-json = "0.3"
```

PR #5615, opened 2020-04-18
Lithium: Use only one thread for postgres.

matt-42:


frameworks/C++/lithium/benchmark_config.json
```diff
@@ -2,7 +2,7 @@
   "framework": "lithium",
   "tests": [
     {
-      "default": {
+      "mysql": {
```
nbrady-techempower, line 5: The first test needs to be called `default`. Can you swap this back, please.

matt-42: Fixed in my last commit.

PR #5616, opened 2020-04-19
CoreRt eagerly create SequenceReader

benaadams:
From https://github.com/aspnet/Benchmarks/pull/1490

/cc @MichalStrehovsky


MichalStrehovsky:
Should this also be applied to the CoreCLR-based benchmark to get a bit more apples-to-apples comparison?

(I know the CoreRT benchmark already suffers/benefits from the fact that it consumes unstabilized bits from the .NET runtime master branch instead of having a stable 3.1-era release branch, but there's nothing we can do about that aspect.)

benaadams:
> Should this also be applied to the CoreCLR-based benchmark

Yes, but since that's the "official" MS version I need to get agreement in aspnet/Benchmarks before making a PR for it here. It's a process 😉

benaadams:
Though hopefully not too long of a process 😃

> Round 19 testing will begin on or around May 1st. Last call for changes to be reviewed and merged is April 29th.

PR #5617, opened 2020-04-20
all golang framework update to 1.14 of golang version

TonyPythoneer:


TonyPythoneer:
How does it trigger Travis CI? I haven't seen that.

PR #5621, opened 2020-04-20
Update benchmark_config.json

helyho:
change framework name from voovan to voovan-mvc


frameworks/Java/voovan/benchmark_config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "framework": "voovan",
+  "framework": "voovan-mvc",
```
nbrady-techempower, line 2: I'm confused about this name change. If the framework is called `voovan`, we'd prefer this to match the directory `Java/voovan` but you could change the `display_name` attribute to say `voovan-mvc`. If the framework name is changing, then you can change the directory to `voovan-mvc` as well.

TonyPythoneer: I agree with what @nbrady-techempower talks about.

If the lib calls `voovan` in Java ecosystem, please keep the origin name. Don't confuse the other people who will contribute to this project in the future.

helyho: ok i will fixed it

helyho:
please cancel this travis job. because this pr is droped. thanks.

PR #5623, opened 2020-04-21
add scratch for go-chi that can prove having better performance with the smallest dockerfile

TonyPythoneer:


nbrady-techempower:
If the only difference here is a `scratch` dockerfile, let's just add one new test case to compare and then if that looks good to you, change the base of the rest. I don't want to add this many permutations. Trying to get the other frameworks to cap at 10 at the most.

TonyPythoneer:
@nbrady-techempower Update. I have left only one scratch for chi.

PR #5625, opened 2020-04-21
improve django coding

TonyPythoneer:

### DIFF
--- a/frameworks/Python/django/hello/world/views.py
+++ b/frameworks/Python/django/hello/world/views.py
@@ -1,5 +1,5 @@
 import random
-from operator import attrgetter
+from operator import itemgetter
 from functools import partial
 from ujson import dumps as uj_dumps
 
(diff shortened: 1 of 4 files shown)


frameworks/Python/django/hello/world/views.py
```diff
@@ -26,60 +29,48 @@ def plaintext(request):
...
+        )
 
 
 def db(request):
-    r = random.randint(1, 10000)
-    world = uj_dumps({'id' : r, 'randomNumber' : World.objects.get(id=r).randomnumber})
+    r = _random_int()
+    world = uj_dumps({
+        'id' : r,
+        'randomNumber' : World.objects.only("randomnumber").get(id=r).randomnumber
```
nbrady-techempower, line 42: I realize that you're basically copying existing code, but it looks like this framework was in violation of the rules to begin with. [Rule number one](https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#requirements-1)

If I understand this correctly, what's happening here is equivalent to
```
SELECT randomNumber from World where id=...
```

TonyPythoneer: @nbrady-techempower updated

Issue #5582, opened 2020-04-03
Clarify plaintext urls specifications


sebastienros:
The specs currently suggest routes like `/plaintext`, `/json`, ...
However since Plaintext is network-bound right now, using smaller ones makes a difference, which can be seen with pico.v and that I have been able to confirm on ASP.NET. There is currently a PR for Actix that does the same, and I would love it to be merged to confirm the impact.

I think it would be fair and be explicit that it's either recommended to use `/p` for perf, or otherwise required to use `/plaintext`.

nbrady-techempower:
@sebastienros I did merge this. I'd like to see the impact as well. After discussing internally, we're going to make the routes required after this round. I'll make a change to the rules here shortly for round 20 and beyond.

talawahtech:
@nbrady-techempower I understand not wanting to go through wholesale and enforce this rule for every framework in round 19, but I feel like at the very least it should should be enforced for the to 15 plaintext frameworks since they are all fast enough to hit the bandwidth limitations.

Allowing pico.v, actix and aspcore to use this technicality sends an unnecessarily misleading signal to anyone who quickly looks at the top 10 list.

nbrady-techempower:
@talawahtech I appreciate your input but we’ve already made the decision to enforce this for round 20 and beyond. The round is closed next week.

NateBrady23:
Closing as part of bulk cleanup.

PR #5635, opened 2020-04-22
bumb v1.9.1

Fenny:
bumb fiber to `v1.9.1` and add `fortune` test for round `19`
```
| fiber
|       fortune       : PASS
|       plaintext     : PASS
|       db            : PASS
|       update        : PASS
|       json          : PASS
|       query         : PASS
```


frameworks/Go/fiber/benchmark_config.json
```diff
@@ -1,49 +1,53 @@
...
-      "versus": "go"
+  "tests": [
+    {
+      "default": {
+        "json_url": "/j",
+        "db_url": "/d",
+        "query_url": "/q?q=",
+        "fortune_url": "/f",
+        "update_url": "/u?q=",
+        "plaintext_url": "/p",
```
nbrady-techempower, line 11: Just so you know, I'm only approving this now because some frameworks are already doing it, but we're updating the rules for the next round. This will need to be changed after round 19. #5582

Fenny: Understood, I'm interested if this really improves results. Will change when results come back 👍

PR #5645, opened 2020-04-24
[skip ci] Add platform nodejs

joanhey:
to the benchmark_config in the js frameworks.

Now we will see the frameworks when filtering by nodejs platform.


joanhey:
No changes in the code, so they fail the same as before.

Ready to be merged.

PR #5646, opened 2020-04-24
Add libsniper HTTP client/server library

OlegRomanenko:
libsniper is a small, very fast C++ library for writing highload HTTP services (f.e. RTB services).
I wrote only plaintext bench. The main goal of the library is very fast processing of incoming requests.


frameworks/C++/libsniper/libsniper_bench/src/main.cpp
```diff
@@ -0,0 +1,101 @@
...
+        http::server::Config config;
+        config.max_conns = 100000;
+        config.connection.keep_alive_timeout = 10min;
+        config.connection.request_read_timeout = 10s;
+
+        http::Server http_server(loop, config);
+        check(http_server.bind("", 8090), "[Server] cannot bind to localhost:8090");
+
+        string date = gen_date();
+        http_server.set_cb_request([&, this](const auto& conn, const auto& req, const auto& resp) {
```
nbrady-techempower, line 65: Thanks for your contribution @OlegRomanenko 

I don't see any type of routing here. Please see rule 6 in the [General Requirements](https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#general-test-requirements)

> Some form of request routing is required, even if only a single test type (e.g., Plaintext) is implemented. In most cases, the framework's canonical router or a mainstream router library will be used. In some cases, it is considered normal and sufficiently production-grade to use hand-crafted minimalist routing using control structures such as if/else branching. This is acceptable where it is considered canonical for the framework.

OlegRomanenko: Oh sorry! I forgot about routing. We do not use routing in production.
Fixed.

PR #5647, opened 2020-04-24
Gemini Firenio Performance Tweaks

msmith-techempower:
This is the reason why we don't normally allow for `SNAPSHOT` dependencies (or dependencies that aren't version locked) - I deployed a new snapshot on which of course broke the API that the implementation currently merged relies.

I will remove `gemini-firenio` after its continuous run and will re-add when that module has an official release.


msmith-techempower:
https://travis-ci.org/github/msmith-techempower/FrameworkBenchmarks/jobs/679216966 Worked - merging.

PR #5648, opened 2020-04-24
Toolset updates for Round 20

nbrady-techempower:
** This is a work in progress **

This update includes reworking test types such that they're never hard-coded in core toolset utilities. This modularizes test types in the same way I did for databases. If we were to add a test type called `routing` for complex routing, creating the `toolset/test-types/routing` folder and adding the proper methods will automagically allow for a `routing` key to be added in a framework's `benchmark_config.json` without adding anything to other toolset code. There's still some work to be done here with cleaning up the abstract test type which I will address before this PR is ready to be merged.

This PR also includes a verification check that each test meet a specific length requirement for that test's url. Plaintext should be at least `/plaintext` length, etc.

More updates to come before this is ready.

- [x] Test type names removed from toolset
- [x] Verify test urls
- [x] Update `cached_query` keys to `cached-query`
- [x] Update vagrant
- [x] Verify the results output renders correctly


benaadams:
Might also want to add a min Server name length in the Server header?

nbrady-techempower:
@benaadams Yeah, good point. We talked about that, too.

benaadams:
Might also want to add a min Server name length in the Server header? Since there are now some 1 char headers out there...

PR #5653, opened 2020-04-26
[Go/fasthttp] Upgrade to v1.12.0 and add test cases

savsgio:


frameworks/Go/fasthttp/benchmark_config.json
```diff
@@ -85,9 +169,87 @@
...
+      "os": "Linux",
+      "database_os": "Linux",
+      "display_name": "fasthttp",
       "notes": "",
       "versus": "go"
     }
   }]
-}
\ No newline at end of file
+}
```
nbrady-techempower, line 255: @savsgio Thanks for your work here. Could you please cut this down to 10 test permutations. I know the other framework you just updated had 12, but we're going to be asking that frameworks cut these down and not accepting any past 10 in the future to try and keep the continuous benchmarking to a reasonable amount of time. Feel free to add and remove as you'd like, so long as it stays at a max of 10.

savsgio: Done!

savsgio:
Thanks!!

PR #5659, opened 2020-04-27
Lithium: remove monothread tests.

matt-42:


nbrady-techempower:
@matt-42 Pulling in the latest master should fix this now

PR #5674, opened 2020-04-28
Redhat transport and corert port changes

benaadams:
Change ports for the non "official" versions; which will reduce the size of the "Host" header in the request

Only changing the non-official versions as it may break in deployment as port 80 is a privileged port; and so close to actual round kicking off that would be a bad thing to do for the "official" version.

/cc @tmds @MichalStrehovsky 

@nbrady-techempower another issue for rule normalisation  😉


nbrady-techempower:
I don't think anyone is currently using a port less than 4 digits. I'm going to scan for that real quick and if not, I think we'll block this here.

nbrady-techempower:
Looks like no one is doing this, so we're going to add something like

> All requests must be made over a 4-digit port (8080, 3000, etc).

to the general requirements.

benaadams:
Seems good 🙂

PR #5678, opened 2020-04-30
Update libreactor

talawahtech:
This PR updates and re-enables libreactor. 

1) Uses latest version of libreactor and its dependencies
2) Dockerfile updated to ubuntu:18.04 and gcc-8
3) Makes use of the reactor_server abstraction (in my tests it is just as fast and a lower level approach)
4) Uses all CPU cores (previous version didn't take advantage of hyper-threading)
5) Misc docker file improvements
6) Updated README


frameworks/C/libreactor/libreactor.dockerfile
```diff
@@ -29,7 +29,7 @@ RUN wget -q https://github.com/fredrikwidlund/libclo/releases/download/v1.0.0/li
 RUN wget -q https://github.com/fredrikwidlund/libreactor/releases/download/v1.0.1/libreactor-1.0.1.tar.gz && \
     tar xfz libreactor-1.0.1.tar.gz && \
     cd libreactor-1.0.1 && \
-    ./configure --prefix=/usr && \
+    ./configure --prefix=/usr CFLAGS="-Wall -Wextra -Wpedantic -O3" && \
```
h3ct0rjs, line 32: nice

PR #5681, opened 2020-05-04
Libreactor: Add optimized platform implementation as new default

talawahtech:
I have renamed the previous implementation that uses the `reactor_server` abstraction libreactor-server. This new optimized approach is now the default and it has been labeled as a "platform" implementation. Here are the highlights:

1) Uses the lower level `reactor_http` abstraction to parse and handle incoming requests, and uses the even lower level `reactor_stream` abstraction to send responses directly.

2) Uses predefined "preamble" variables for json and plaintext responses. This is similar to how aspcore does it, but in our case the content length is still calculated/set for every request.

3) Uses reactor_timer to update the date header in the preambles once every second.

4) Uses a custom `reactor_net_bind` function that avoids using reactor_resolver/reactor_pool to resolve the server address and just calls `getaddrinfo()` directly. This allows us to skip starting a `reactor_pool` and compile the program without -pthread. Profiling showed that the unused pool was still using up cpu cycles unnecessarily.


talawahtech:
@nbrady-techempower my sincere apologies for the 11th hour PR. It took waaay longer than expected to get it done and life kept on intervening. 

I know that the Citrine run for round 19 is only a few hours away, and I know I am way past the deadline for submissions.  Just hoping you will have a little time to merge this before it begins. Let me know if there are any questions.

nbrady-techempower:
@talawahtech sorry, I was afk for the rest of the day! R19 is running now and citrine and azure will be using the same commits so we’ll get this in once those runs are complete.

talawahtech:
@nbrady-techempower ok understood.

PR #5704, opened 2020-05-25
Update versions

pmlopes:
Signed-off-by: Paulo Lopes <pmlopes@gmail.com>

Upgrade versions of es4x dependencies, test code is untouched.


pmlopes:
It turns out the postures driver did change the api between versions. This has been fixed.

Issue #5709, opened 2020-05-28
Changes / Updates / Dates for Round 20


nbrady-techempower:
I'll be editing this pinned issue to link to other issues or pull requests that may affect tests for this round.

See: https://github.com/TechEmpower/FrameworkBenchmarks/issues/4966 for details about the previous round released on 5/28/2020

nbrady-techempower:
** Placeholder **

bhauer:
Round 20 will be adding visualization of the [cached queries test type](https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#caching-work-in-progress).  In fact, it's very likely that we'll push this visualization out soon since Round 19 already has some results, and doing so will make these render for [continuous runs](https://tfb-status.techempower.com/).

If your framework doesn't yet include an implementation of this new test type, please consider adding one!

![image](https://user-images.githubusercontent.com/777068/85637996-99538480-b639-11ea-81ca-f9a32838394d.png)

volyrique:
The visualization of the cached queries results is a great addition, especially given that the test has been defined (and the first implementations have appeared) 3 years ago. I think the lesson here is that there won't be a significant number of implementations of a future test type unless the results are being shown properly on the results dashboard.

There is just one small issue right now - while a partial set of cached queries results (i.e. for a run that hasn't completed yet) is shown properly, the visualization doesn't work for runs that have completed - I just get "No data available for this test".

As for the changes for the next round, are there any plans to upgrade to Ubuntu 20.04? Updating the kernel version has had a significant performance impact in the past.

michaelhixson:
> There is just one small issue right now - while a partial set of cached queries results (i.e. for a run that hasn't completed yet) is shown properly, the visualization doesn't work for runs that have completed - I just get "No data available for this test".

I think this is a side effect of renaming the test type from `cached_query` to `cached-query`.  Runs that started after [this change in the toolset](https://github.com/TechEmpower/FrameworkBenchmarks/commit/0b8a38273c98782fca3cc098efd36cfcac3e8ee8) will have visualizable data, runs from before that will not.  The TFB website only understands the new name.  For Round 19, the new name was manually edited into the website's copy of the results.

nbrady-techempower:
> As for the changes for the next round, are there any plans to upgrade to Ubuntu 20.04? Updating the kernel version has had a significant performance impact in the past.

[...]

PR #5710, opened 2020-05-29
ntex framework

fafhrd91:
Add ntex framework benchmark


joanhey:
I think that we need to enforce all the url names.
Check 
https://github.com/TechEmpower/FrameworkBenchmarks/pull/5710/files#diff-1cd02081fa7857523980ad89891563d4

@nbrady-techempower  What do you think ?

nbrady-techempower:
@joanhey Agreed.

This change will be coming in #5648 and will break frameworks not doing it.

I'm going to get that merged in this week so @fafhrd91 could you please update this to use `/json`, `/plaintext`, etc.

This change for Round 20 is also noted here: #5709

fafhrd91:
updated

frameworks/Rust/ntex/benchmark_config.json
```diff
@@ -0,0 +1,61 @@
...
+      "database_os": "Linux",
+      "display_name": "ntex",
+      "notes": "",
+      "versus": ""
+    },
+    "db": {
+      "fortune_url": "/f",
+      "db_url": "/d",
+      "query_url": "/q?q=",
+      "update_url": "/u?q=",
```
joanhey, line 26: Also /fortune, /db, /query and /update.

fafhrd91: this should be reflected in ticket

fafhrd91: @nbrady-techempower are those mandatory paths as well?

nbrady-techempower: @fafhrd91 Sorry, yes. I should have made that clearer. The verification changes in #5648 account for minimum required lengths for all routes.

fafhrd91: sure, np. Updated

botika:
@fafhrd91 uses `yarte@0.9.3` which integrates the rendering of numbers by `itoa` and `dtoa` with a considerable performance improvement.

fafhrd91:
I will, but I need base bench numbers first

botika:
Sorry, [benchs](https://github.com/botika/template-bench-rs) and [results](https://travis-ci.org/github/botika/template-bench-rs). The difference in Big table bench is around `60us`

fafhrd91:
I need base numbers for this bench

fafhrd91:
@botika I will switch to yarte in next pr. is there way to render yarte template to `io::Write`?

botika:
@fafhrd91  I suppose you mean something similar to [this](https://github.com/botika/yarte/blob/master/benches/src/all.rs#L216-L271), I can do it without problem and set the maximum size for the template at compile time, if you mean something else, open an issue or a PR and I will gladly integrate it, even if it is not entirely safe.

PR #5714, opened 2020-05-29
Initial commit for Comet PHP framework

gotzmann:
Initial commit with ALL Techempower Benchmarks for Comet PHP framework: https://github.com/gotzmann/comet


gotzmann:
I'm not sure why Travis fails on this string:

TESTDIR="PHP/php PHP/comet"

Maybe that's because folder "frameworks/PHP/comet" doesn't exist in master branch?

.travis.yml
```diff
@@ -62,8 +62,9 @@ env:
     - "TESTLANG=Nim"
     - "TESTLANG=Perl"
     - 'TESTDIR="PHP/php"'
+    - 'TESTDIR="PHP/comet"'
```
joanhey, line 65: Don't create a new job for only 1 framework.
Please, revert that change.

joanhey:
It's the same where you place comet, always will show you that fail the job.
If you click in that job, you will see why it is failing.
https://travis-ci.org/github/TechEmpower/FrameworkBenchmarks/jobs/692627577#L1292

.travis.yml
gotzmann: Done

gotzmann:
I've fixed Travis warnings. All checks have passed. Woohoo!

PR #5726, opened 2020-06-02
Add benchmark for Frank

panesofglass:
This adds the [Frank](https://github.com/frank-fs/frank) micro-framework written in [F#](https://fsharp.org/).


panesofglass:
@forki did I skip a step in submitting this PR? How did you get the other F# frameworks into the benchmark?

forki:
Looks good to me, but tbh I have absolutely no idea. Sorry

frameworks/FSharp/frank/src/App/Program.fs
```diff
@@ -0,0 +1,126 @@
...
+            new MemoryStream(capacity)
+
+    static member Release(ms:MemoryStream) = 
+        if ms.Capacity <= MaxBuilderSize then
+            MemoryStreamCache.instance <- ms
+
+let inline contentLength x = new Nullable<int64> ( int64 x )
+
+let json' data : HttpContext -> Task<unit> =
+    let bytes = Utf8Json.JsonSerializer.Serialize(data)
```
nbrady-techempower, line 48: @panesofglass Apologies for taking so long to review.

The object for the json test needs to be instantiated and serialized on every request.

`struct {| message = "Hello, World!" |}` and this serialization should be moved into the `ctx` function below.

nbrady-techempower: Also, I just looked at some other F# implementations and I see that Zebra has this problem since 2018

panesofglass: Ah, thank you! I'll update.

panesofglass: Hmm, this is actually correct. The object is created on each request and passed to the `json'` function where it is serialized. This is not cached anywhere ahead of time. This matches the F# Giraffe implementation, as well.

nbrady-techempower, line 48: @panesofglass  if I put a `printfn "test"` here and then hit the json route over and over again, it will not print. If it is moved into the function, it will, so this is not currently meeting the requirements.

panesofglass: You are correct. I thought it was under the `fun ctx -> ...` bit. I have updated it so that it serializes on each request.

frameworks/FSharp/frank/src/App/Program.fs
```diff
@@ -45,21 +45,21 @@ type MemoryStreamCache =
...
             do! ctx.Response.Body.WriteAsync(bytes, 0, bytes.Length)
         }
 
 let text' (msg:string): HttpContext -> Task<unit> =
-    let bytes = System.Text.Encoding.UTF8.GetBytes(msg)
     fun ctx ->
         ctx.Response.ContentLength <- contentLength bytes.Length
         ctx.Response.ContentType <- "text/plain"
         ctx.Response.StatusCode <- 200
+        let bytes = System.Text.Encoding.UTF8.GetBytes(msg)
```
nbrady-techempower, line 62: Just to be clear, the plaintext route does not have the same requirements.

panesofglass: Ah, thanks!

frameworks/FSharp/frank/src/App/Program.fs
```diff
@@ -0,0 +1,88 @@
...
+let plaintext = 
+    resource "/plaintext" {
+        name "Plain text"
+        get (text' "Hello, World!")
+    }
+
+let json =
+    resource "/json" {
+        name "JSON"
+        get (json' struct {|message="Hello, World!"|})
```
nbrady-techempower, line 66: Sorry to be a pain here, but this `struct` creation still needs to be moved into the function at line 24. My understanding of what is happening here is that this struct is only being created once and then being serialized on each request. It also needs to be instantiated on each request.

> i. For each request, an object mapping the key message to Hello, World! must be instantiated.

panesofglass: No worries. Sorry that I misunderstood.

panesofglass:
Thank you, @nbrady-techempower, for the help on getting this ready to merge.

nbrady-techempower:
@panesofglass happy to help!

PR #5727, opened 2020-06-03
racket: tweak responses, leverage multiple cores

Bogdanp:


nbrady-techempower:
I noticed you and @fafhrd91 discussing this in another thread so just want to wait for that approval. Thanks!

nbrady-techempower:
@Bogdanp Way too many tabs open. That's absolutely what I did.

PR #5733, opened 2020-06-05
Add yarte@0.9.8 in ntex-db bench

botika:
@fafhrd91  It is a test without commitment.
In this new implementation, there is no possibility of buffer overruns, since before writing it is checked that it is within the limits.

[Unit bench, 50 elem, 100KK < id](https://travis-ci.org/github/botika/template-bench-rs/jobs/694909228#L1003)
```
TechEmpower Fortunes/Markup time:        [12.413 us 12.419 us 12.427 us]
TechEmpower Fortunes/Yarte time:         [11.315 us 11.323 us 11.332 us]
TechEmpower Fortunes/Yarte Fixed time:   [5.4315 us 5.4438 us 5.4576 us]
```

### DIFF
--- a/frameworks/Rust/ntex/Cargo.toml
+++ b/frameworks/Rust/ntex/Cargo.toml
@@ -18,7 +18,6 @@ path = "src/main_raw.rs"
 [dependencies]
 ntex = "0.1.18"
 snmalloc-rs = "0.2.6"
-markup = "0.4.1"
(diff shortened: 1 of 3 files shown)


botika:
@fafhrd91 It has been optimized by eliminating unnecessary checks and working directly with `memcopy`. It could be further optimized by using byte-by-byte writing for shorter literal cases. At the moment it does the minimum so that the implementation on `memcopy` is safe.

PR #5736, opened 2020-06-06
Set Agoo to realistic

ohler55:
I was not able to see any results for Agoo. I changed the `approach` in the configuration file in hopes that would cause results to be displayed. If that is not the correct way to do that please let me know what is. Thanks you.

### DIFF
--- a/frameworks/Ruby/agoo/benchmark_config.json
+++ b/frameworks/Ruby/agoo/benchmark_config.json
@@ -8,7 +8,7 @@
       "update_url": "/updates?queries=",
       "plaintext_url": "/plaintext",
       "port": 8080,
-      "approach": "Stripped",
+      "approach": "Realistic",
       "classification": "Micro",
       "database": "Postgres",
       "framework": "rack",


nbrady-techempower:
`stripped` approaches are hidden by default as these types of tests do things that are considered unrealistic. Marking a framework as stripped usually involves a discussion with the implementer of the test. Changing this would be incorrect.

To see the results, click on the filters panel

image

and click on `Stripped` to enable it

image

ohler55:
Right, I am the implementor of Agoo and would like to know what to do to make it realistic.

nbrady-techempower:
@ohler55 I mean the implementer of the test. That was @GertThiel. If you're the owner of the Agoo framework, and you think the approach is realistic as it is currently implemented, I'm fine with reopening this.

ohler55:
Thank you. I'm not sure what constitutes realistic but the test implementer did include the use of a database so that seems more than enough to qualify. It does skew the results since the database has to be accessed for each request though. Do most of the frameworks include a database query in the implementation? If not I can stub that part out. Otherwise it looks pretty reasonable.

nbrady-techempower:
It actually looks like it may have been a copy/paste of a few other benchmark configs that were using the "stripped" approach. I'm fine with getting this in until someone objects with a valid reason.

ohler55:
Great, thanks!

PR #5737, opened 2020-06-07
racket: tune nginx conf

Bogdanp:
I spent a little more time tuning the nginx config today. This improves throughput by about 4-5x on the plaintext benchmark compared to the previous PR.


sebastienros:
Made the same discoveries last week while benchmarking proxies. Nginx uses http 1.0 by default. And enabling it is really not simple as many flags need to be set, like you did.

PR #5738, opened 2020-06-07
Upgrade to version ffead-cpp v4.0

sumeetchhetri:
Added tests for multiple server backends


sumeetchhetri:
@nbrady-techempower, any idea how can I surpass the [50min](https://docs.travis-ci.com/user/customizing-the-build#build-timeouts) travis-ci build timeout limitation?

sumeetchhetri:
So that would mean I would need multiple framework folders for all my server backends then or maybe like 2 different framework folders, will that work?

nbrady-techempower:
> So that would mean I would need multiple framework folders for all my server backends then or maybe like 2 different framework folders, will that work?

No. One framework/project should only have 10 permutations at a time. That takes up a lot of benchmarking time. You can keep the src for the permutations that you're not currently benchmarking and swap them out when you want, but we need to conserve benchmarking time for others. We don't expect you to have every possible backend, etc. That's just not feasible for this project.

sumeetchhetri:
@nbrady-techempower, there should be a work-around for this, it is very difficult to keep track of all the performance metrics behind each server backend across tfb runs, I agree 10 should be enough, but exceptions should be possible, anyways I'll come up with a PR to fix the top 10 tests as a new field (string array) in the benchmarking_config.json file, that way it would be easier to track which 10 tests you may want to run across tfb runs with a single line change instead of making like say 10 changes across the file as is possible currently.

nbrady-techempower:
@sumeetchhetri You could just make a couple of backup benchmark configs. like `benchmark_config2.json`, `benchmark_config3.json`. I'm fine with keeping those here and that way you can just rename when you're ready to try the next set of 10.

sumeetchhetri:
> @sumeetchhetri You could just make a couple of backup benchmark configs. like `benchmark_config2.json`, `benchmark_config3.json`. I'm fine with keeping those here and that way you can just rename when you're ready to try the next set of 10.

👍

PR #5758, opened 2020-06-11
ntex: reduce memory allocations in fortunes bench

fafhrd91:


frameworks/Rust/ntex/src/db.rs
```diff
@@ -161,31 +175,49 @@ impl PgConnection {
...
-
+    pub fn tell_fortune(&self) -> impl Future<Output = Result<Bytes, io::Error>> {
         let fut = self.cl.query_raw(&self.fortune, &[]);
 
         async move {
             let mut stream = fut
                 .await
                 .map_err(|e| io::Error::new(io::ErrorKind::Other, format!("{:?}", e)))?;
 
+            let mut rows = SmallVec::<[_; 32]>::new();
```
nbrady-techempower, line 186: The changes here and setting the body length capacity violate rule 5. in the fortunes test:

fafhrd91: SmallVec is dynamic, it can contain any number of elements

nbrady-techempower: I don't see any documentation on the `TemplateFixed` macro, but it makes me think that using that with these initial capacities is optimizing for a known size of memory allocation. Can you help me understand what `TemplateFixed` does? Also,

```Rust
let mut body = BytesMut::with_capacity(2048);
            unsafe {
                // Maybe uninit
                body.set_len(2048);
```

fafhrd91: this is just uninitialized buffer, yarte engine write to that memory directly.

https://docs.rs/yarte/0.9.9/yarte/derive.TemplateFixed.html

nbrady-techempower: If the fortune body is more than 2048 bytes:
```
ntex-db: thread 'ntex-rt:worker:9' panicked at 'called `Option::unwrap()` on a `None` value', src/db.rs:211:28
```
This can be replicated by adding a couple more rows to the Fortunes table.

fafhrd91: fair point. it is fixed.

nbrady-techempower: @fafhrd91 thanks!

PR #5771, opened 2020-06-18
Libreactor Refactor

talawahtech:
Code refactoring and simplification. No performance impacting changes.

### DIFF
--- a/frameworks/C/libreactor/libreactor.dockerfile
+++ b/frameworks/C/libreactor/libreactor.dockerfile
@@ -1,11 +1,11 @@
-FROM ubuntu:18.04 as builder
+FROM ubuntu:20.04 as builder
 
 RUN apt-get update -yqq
-RUN apt-get install -yqq wget make automake libtool file gcc-8 g++-8
+RUN apt-get install -yqq wget make automake libtool file gcc-9 g++-9
 
 WORKDIR /libreactor
 
-ENV CC=gcc-8 AR=gcc-ar-8 NM=gcc-nm-8 RANLIB=gcc-ranlib-8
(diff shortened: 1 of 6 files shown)


frameworks/C/libreactor/libreactor.dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM ubuntu:18.04 as builder
+FROM ubuntu:20.04 as builder
 
 RUN apt-get update -yqq
-RUN apt-get install -yqq wget make automake libtool file gcc-8 g++-8
+RUN apt-get install -yqq wget make automake libtool file gcc-9 g++-9
```
volyrique, line 4: GCC 9 is the default version in Ubuntu 20.04, so there is no need to specify it. In fact, you really shouldn't hardcode the version (keeping the dependency versions fixed is served by hardcoding the base image version) unless you have a very good reason to. Just think about the moment when Ubuntu 22.04 comes out - ideally that would be a one line change, but in your case it won't.

talawahtech: @volyrique @nbrady-techempower I wanted the dependency to be explicit. I don't want the gcc version to be automatically upgraded. For example gcc-10 has a backwards incompatible change that I know will cause this code to no longer build.

volyrique: It won't be upgraded as long as you keep the base image the same.

talawahtech: @volyrique right, but I don't want the gcc version to be updated automatically when I update the base image. I am deliberately pinning the dependency on that specific version of gcc.

PR #5800, opened 2020-06-25
[C++/cuehttp]improve

xcyl:


frameworks/C++/cuehttp/cuehttp.dockerfile
```diff
@@ -9,7 +9,7 @@ WORKDIR /
 RUN git clone https://github.com/xcyl/cuehttp.git
 
 WORKDIR /cuehttp
-RUN git checkout b24ccd101edcaa1d21f9b1ffbbb97e3ac926250a
+RUN git checkout tfb
```
nbrady-techempower, line 12: Please checkout a specific hash, not a branch. We don't want this to be automatically updated.

xcyl: OK

PR #5807, opened 2020-06-27
Aspnet platform updates

benaadams:
Adds platform caching scenario; syncs with current aspnet benchmarks, also updates urls from round 20 as per https://github.com/TechEmpower/FrameworkBenchmarks/issues/5582

```diff
- Json = "/j";
+ Json = "/json";
- Plaintext = "/p";
+ Plaintext = "/plaintext";
```

/cc @sebastienros


frameworks/CSharp/aspnetcore/PlatformBenchmarks/appsettings.postgresql.json
```diff
@@ -0,0 +1,4 @@
+﻿{
+  "ConnectionString": "Server=tfb-database;Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=256;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4;Multiplexing=true;Write Coalescing Delay Us=500;Write Coalescing Buffer Threshold Bytes=1000",
```
benaadams, line 2: @roji is this the correct connection string for Platform? I've kept updates the same (one below)

roji: Yeah that's the good one for multiplexing (though have we turned that on for all database scenarios, or just fortunes?)

benaadams: Would be turned on for fortunes, single query and multi-query (but not updates). Should they use different ones?

benaadams:
If that's all the feedback might as well merge and see the results and iterate from there

frameworks/CSharp/aspnetcore/PlatformBenchmarks/appsettings.postgresql.json
roji: Sorry for not answering this, have been underwater. I plan a thorough pass comparing 5.0 multiplexing, 5.0 non-multiplexing and 4.1.3 perf (i.e. currentl released version of Npgsql) for platform and non-platform benchmarks... I should be able to look into this in a couple of weeks.

PR #5815, opened 2020-06-29
Temporarily disable ffead

michaelhixson:
The last two runs in the Citrine environment failed because of problems induced by ffead.  We don't know that ffead is doing anything wrong -- even if it is, the TFB toolset is to blame for letting that ruin the run -- but we would like to continue benchmarking other frameworks while we figure this out.

### DIFF
--- a/frameworks/C++/ffead-cpp/benchmark_config.json
+++ b/frameworks/C++/ffead-cpp/benchmark_config.json
@@ -23,7 +23,7 @@
 			"display_name": "ffead-cpp-mongo",
 			"notes": "mongodb redis",
 			"versus": "",
-			"tags": []
+			"tags": ["broken"]
 		},
 		"nginx-mysql": {
 			"json_url": "/te-benchmark-um/json",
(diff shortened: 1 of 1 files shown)


nbrady-techempower:
@sumeetchhetri just an FYI on this

sumeetchhetri:
@nbrady-techempower, I have been experimenting with a few custom allocators and upgrading a few libraries in the master branch, maybe that caused the issue, will create a new PR with the fixes and the passing verify steps to fix this, in the meanwhile, this should anyways be handled in the tfb toolset for sure to avoid possible future issues with test runs.

nbrady-techempower:
@sumeetchhetri I must have missed this in the review, but your dependencies (including the framework itself) should be a specific commit or version such that breaking changes aren't introduced without going through a PR. Please make sure all your dependencies are locked down as best as possible and I'll review more closely next time.

> this should anyways be handled in the tfb toolset for sure to avoid possible future issues with test runs

We understand this: "the TFB toolset is to blame for letting that ruin the run" We're looking into the issue.

nbrady-techempower:
@sumeetchhetri Also, if you happen to know what the offending commit was that you think may have caused this, that would be helpful in tracking down the bug in the the toolset.

PR #5824, opened 2020-07-01
Increase number of Python/flask workers

leontrolski:
As per: http://calpaterson.com/async-python-is-not-faster.html

The number of `gunicorn` workers is probably too low for Flask. Cal recommends overshooting then dialling back _(and also that the cost of too many workers is higher than the cost of too few)_, so I've upped the number of workers to a (seemingly crazy) 8 to see what happens.

### DIFF
--- a/frameworks/Python/flask/gunicorn_conf.py
+++ b/frameworks/Python/flask/gunicorn_conf.py
@@ -5,7 +5,7 @@ import sys
 _is_pypy = hasattr(sys, 'pypy_version_info')
 _is_travis = os.environ.get('TRAVIS') == 'true'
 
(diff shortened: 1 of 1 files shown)


leontrolski:
Is it possible to easily run this new configuration on equivalent hardware to the public runs - I guess the CI run is not equivalent? (So as to tune the number of workers).

nbrady-techempower:
@leontrolski You can set up your own environment following the instructions in the wiki, or you can watch our continuous benchmarking on https://tfb-status.techempower.com; once this passes travis and is merged, it will be part of the next run.

leontrolski:
Fantastic, what a smooth set up! Thanks @nbrady-techempower

PR #5827, opened 2020-07-02
[Go/fasthttp] Add Cached Queries test

savsgio:


nbrady-techempower:
@savsgio Please pull from latest; the test urls are now required, not recommended. For instanced: `/fortune` needs to be `/fortunes`, etc. the cached queries route must be at least `/queries` (8 characters).

savsgio:
> @savsgio Please pull from latest; the test urls are now required, not recommended. For instanced: `/fortune` needs to be `/fortunes`, etc. the cached queries route must be at least `/queries` (8 characters).

@nbrady-techempower Done! 😉

PR #5839, opened 2020-07-06
KumbiaPHP fix urls

joanhey:


joanhey:
Passed travis test, sometimes ge stuck the message.

PR #5840, opened 2020-07-06
Comet fix fortunes url

joanhey:


joanhey:
Passed travis test, sometimes get stuck the message.

PR #5847, opened 2020-07-07
Update firenio to 1.3.4

wangkaish:


frameworks/Java/firenio/pom.xml
```diff
@@ -14,12 +14,22 @@
...
+        </repository>
+    </repositories>
+
     <dependencies>
 
         <dependency>
             <groupId>com.firenio</groupId>
             <artifactId>firenio-all</artifactId>
-            <version>1.3.2</version>
+            <version>1.3.5-SNAPSHOT</version>
```
nbrady-techempower, line 32: I realize that some frameworks are currently using `SNAPSHOT`, but we're going to start prohibiting this. I was actually unaware that snapshots could be updated at will. Could you update this to a more recent stable version and I will let others know to do the same. Thanks!

wangkaish: Agree with you, updated.

PR #5852, opened 2020-07-08
aspnetcore: multiplexing for updates

n-stefan:
Enable multiplexing for updates.

Locally it improved rps from
[1925](https://www.techempower.com/benchmarks/#section=test&shareid=28d909d6-3050-4391-afc3-eb0fec4dece7&hw=ph&test=update&a=2)
to
[3058](https://www.techempower.com/benchmarks/#section=test&shareid=b0834491-2485-43df-b5da-5bbdd6e0c5fa&hw=ph&test=update&a=2)
in one instance.


roji:
Thanks @benaadams. As long as it produces better results I'm fine with it :)

Am planning mapping out perf for all benchmarks, for Npgsql v5.0 with multiplexing, without multiplexing, and with the current release (v4.1) to get a good overview of where we are. But it will be a while before I can get around to that.

roji:
@n-stefan can you provide more information on what "in one instance" means? Are there any regressions ?

n-stefan:
@roji by "in one instance" I meant that the Updates test was run first with multiplexing off and immediately afterwards with multiplexing on. In another instance the boost from multiplexing was about 65%.
As for regressions, unless I'm missing something, Npgsql 5.0 without multiplexing seems a bit slower than 4.1 in the Updates test (~14.2k vs ~15.0-15.1k on Citrine).

roji:
@n-stefan thanks, that makes sense. I'll hopefully have time to do all the comparisons soon.

PR #5863, opened 2020-07-11
Lwan: Address performance issues with the cached queries benchmark

lpereira:
This updates Lwan to the latest version (where a critical performance regression was fixed), and fix the test harness so that all worker threads are kept busy.

### DIFF
--- a/frameworks/C/lwan/src/techempower.c
+++ b/frameworks/C/lwan/src/techempower.c
@@ -52,9 +52,9 @@ static struct db_connection_params {
 
 static const char hello_world[] = "Hello, World!";
 static const char random_number_query[] =
-    "SELECT randomNumber FROM world WHERE id=?";
+    "SELECT randomNumber, id FROM world WHERE id=?";
 static const char cached_random_number_query[] =
-    "SELECT randomNumber FROM world WHERE id=?";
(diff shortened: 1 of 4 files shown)


frameworks/C/lwan/src/techempower.c
```diff
@@ -53,6 +53,8 @@ static struct db_connection_params {
 static const char hello_world[] = "Hello, World!";
 static const char random_number_query[] =
     "SELECT randomNumber FROM world WHERE id=?";
+static const char cached_random_number_query[] =
+    "SELECT randomNumber FROM world WHERE id=?";
```
nbrady-techempower, line 57: hey @lpereira I've missed this in the past but for the query tests, we're looking for both `randomNumber` and `id` to be read from the db.

> i. For every request, a single row from a World table must be retrieved from a database table. It is not acceptable to read just the randomNumber.

lpereira: Fair.  Rebased and made a commit making this change.

lpereira:
@nbrady-techempower The requested change has been made; could you please take a second look?

nbrady-techempower:
Travis got borked here, but this passes my local testing.

PR #5867, opened 2020-07-13
🐦[FSharp] Falco - Update benchmark for Falco to v2.0.0

pimbrouwers:
This PR updates the Falco benchmark from v1.2.x to v2.0.0


frameworks/FSharp/falco/src/App/Value.fs
```diff
@@ -3,14 +3,12 @@
...
 
-    let handleJson : HttpHandler =
-        fun next ctx ->
-            let json = { message = defaultMsg }
-            (jsonOut json) next ctx
+let defaultMsg = "Hello, World!"
+
+let handleJson : HttpHandler =
+    { message = defaultMsg }
+    |> Response.ofJson
```
nbrady-techempower, line 11: Can you explain how this works? It doesn't look to me like this would be instantiating and serializing a new object on every request. It looks like it's doing it once and the route `/json` is just getting this constant value.

pimbrouwers: Great question. It does look like a static value. But the HttpContext is thread through it implicitly, so it is recomputed each request. You can see it being transformed into a `RequestDelegate` [here](https://github.com/pimbrouwers/Falco/blob/d613fa40b6aee7124cd75612e30f80f4883882f4/src/Falco/Core.fs#L36).

nbrady-techempower: @pimbrouwers That doesn't appear to be working the way you say. To prove this out, I've changed the `handleJson HttpHandler` to the following:

```fsharp
let handleJson : HttpHandler =
    {
        message =  System.DateTime.Now.ToLongTimeString()
    }
    |> Response.ofJson
```

Starting the server in debug mode `./tfb --mode debug --test falco` and hitting the following endpoint over the course of a few minutes:

`curl http://localhost:8080/json`

```
{"message":"00:13:40"}
{"message":"00:13:40"}
{"message":"00:13:40"}
```

Of course, I would expect to see a new system date on every request.

pimbrouwers: Fascinating! Will update this PR today!

PR #5868, opened 2020-07-13
Lithium update & fix mysql version.

matt-42:
This should fix the crash of the last update on pgsql.

### DIFF
--- a/frameworks/C++/lithium/lithium-postgres.dockerfile
+++ b/frameworks/C++/lithium/lithium-postgres.dockerfile
@@ -5,7 +5,7 @@ RUN apt-get install -yqq g++-9 libboost-dev postgresql-server-dev-all libpq-dev 
 
 COPY ./ ./
 
-ENV COMMIT=537d651735f0ee25b2ab21aa65519ecf39ba025b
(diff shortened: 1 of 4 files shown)


nbrady-techempower:
Yeah, actually, I'm going to update the `mysql.dockerfile` to remove the pinned down patch version. Mysql removes the minor version from apt repos causing these failures, so I'm going to update this to:

```
RUN echo "Installing mysql-server version: $(apt-cache policy mysql-server | grep -oP "(?<=Candidate: )(.*)$")"
RUN DEBIAN_FRONTEND=noninteractive apt-get -y install mysql-server > /dev/null
```

Because of the sources list, this will keep us at the proper major version we have defined, but always get the correct patch version.

nbrady-techempower:
@matt-42 This is an interesting error:

one of your fortunes returned less than the full text:
`fortune: <tr><td>10</td><td>Computers make very fast, very</td></tr>`
versus:
`fortune: <tr><td>10</td><td>Computers make very fast, very accurate mistakes.</td></tr>`

I'm wondering what it is about your current implementation that would cause that?

matt-42:
Yes I'm  investigating this issues. I still don't know why the string are
truncated.
You can do a separate pr to upgrade mysql and I'll do another PR with
lithium specific upgrades.

PR #5879, opened 2020-07-16
Add ubiquity memory cache

jcheron:
### Added
- Ubiquity memory cache (first version)
### Updated
- Swoole pecl (no version)
- Ubuntu 19.10 -> 20.04
### Fixed
- Mysql version in docker file

**Note that this is a full Travis run**


nbrady-techempower:
@jcheron Thanks for getting the mysql version, too!

jcheron:
I just saw that @matt-42 had done it before me.
see PR https://github.com/TechEmpower/FrameworkBenchmarks/pull/5868

nbrady-techempower:
Yeah, actually, I'm going to update the `mysql.dockerfile` to remove the pinned down patch version. Mysql removes the minor version from apt repos causing these failures, so I'm going to update this to:

```
RUN echo "Installing mysql-server version: $(apt-cache policy mysql-server | grep -oP "(?<=Candidate: )(.*)$")"
RUN DEBIAN_FRONTEND=noninteractive apt-get -y install mysql-server > /dev/null
```

Because of the sources list, this will keep us at the proper major version we have defined, but always get the correct patch version.

nbrady-techempower:
`sw-fw-less` is filling up the logs but ubiquity passes locally

jcheron:
Thanks @nbrady-techempower !
I'll wait for the docker images (Memcached and Redis) for the rest, so as not to make the runs too heavy.

Issue #5874, opened 2020-07-15
Loosen Update verification


benaadams:
Since the update is random values; perhaps lower the threshold a little (e.g. 95%)

e.g. the last tfb run failed Rails for updating too few rows (1 short) https://tfb-status.techempower.com/unzip/results.2020-07-12-19-02-35-234.zip/results/20200708175354/rails/update/verification.txt
```
--------------------------------------------------------------------------------
VERIFYING UPDATE
--------------------------------------------------------------------------------
   FAIL for http://10.0.0.1:8080/update?queries=20
     Only 20479 executed queries in the database out of roughly 20480 expected.
     See https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#specific-test-requirements
   PASS for http://10.0.0.1:8080/update?queries=20
     Rows read: 10240/10240
     See https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#specific-test-requirements
   PASS for http://10.0.0.1:8080/update?queries=20
     Rows updated: 10237/10240
     See https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#specific-test-requirements
```

nbrady-techempower:
Hi @benaadams,

We're currently reviewing the verification steps for all database tests.

For the record, this looks like a failure to execute all the required queries:
`FAIL: Only 20479 executed queries in the database out of roughly 20480 expected.`
But the updating of rows does allow for a margin of error:
`PASS: Rows updated: 10237/10240`

msmith-techempower:
This may be a bug to the MySQL implementation.

I have checked the code and we *do* have a margin of 1.5% for the solo-update check (the last `PASS`), but it is not applied to the first check where both `SELECT` and `UPDATE` statements are counted. 

What this verification output suggests to me is that one of either an update or select was not executed (or counted by MySQL). We can confirm that at least all 10240 `SELECT` statements were executed as confirmed by the second `PASS`, and this should *not* have the margin applied to it but does.

Ignoring for the moment the third `PASS`'s counts, what I don't understand is how the first `FAIL` has the numbers it does. I tested updating the same row with the same value, and saw that `Show global status where Variable_name = 'Com_update';` does indeed increment, so it isn't that a double-update occurred.

Why, then, would we give a margin at all?

jcheron:
[...]

PR #5882, opened 2020-07-17
Rails fix update failure

jcheron:
### updated
- Generates a random number different from that of the instance loaded from the database (force update).

see https://github.com/TechEmpower/FrameworkBenchmarks/issues/5874


msmith-techempower:
I do not believe this is necessary once #5883 is merged.

nbrady-techempower:
It looks like Rails still failed the update for postgres in the latest fix so im going to get this in.

PR #5883, opened 2020-07-17
Addresses issue with MySQL query counting

msmith-techempower:
See #5874 for details.


msmith-techempower:
It turns out my original attempt used a query that "worked" but did not count the right things. I needed to experiment with some negative cases to see why.

So, I reverted the query back to counting selects and updates, and simply gave the 1.5% margin rather than trying to be super accurate. This should be good enough.

PR #5891, opened 2020-07-21
Update VoovanTFB.java

helyho:

### DIFF
--- a/frameworks/Java/voovan/src/main/java/org/voovan/VoovanTFB.java
+++ b/frameworks/Java/voovan/src/main/java/org/voovan/VoovanTFB.java
@@ -11,6 +11,7 @@ import org.voovan.tools.TEnv;
 import org.voovan.tools.TObject;
 import org.voovan.tools.json.JSON;
 import org.voovan.tools.log.Logger;
+import org.voovan.tools.TString;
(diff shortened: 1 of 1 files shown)


frameworks/Java/voovan/src/main/java/org/voovan/VoovanTFB.java
```diff
@@ -65,7 +66,7 @@ public void process(HttpRequest req, HttpResponse resp) throws Exception {
                 webServer.get("/json", new HttpRouter() {
                         public void process(HttpRequest req, HttpResponse resp) throws Exception {
                                 resp.header().put(HttpStatic.CONTENT_TYPE_STRING, HttpStatic.APPLICATION_JSON_STRING);
-                                resp.write(JSON.toJSON(msg, false, false));
+                                resp.write(TString.toAsciiBytes(JSON.toJSON(msg, false, false)));
```
nbrady-techempower, line 69: `msg` needs to be instantiated on each request. I missed this before, but that doesn't appear to be happening here. 

`Message msg = new Message(HELLO_WORLD_STR)` needs to happen here.

helyho: ok, i fixed it, thanks

PR #5898, opened 2020-07-24
Disabled v-vweb test which was causing tfb tests to stall

sumeetchhetri:

### DIFF
--- a/frameworks/C++/ffead-cpp/benchmark_config.json
+++ b/frameworks/C++/ffead-cpp/benchmark_config.json
@@ -23,7 +23,7 @@
 			"display_name": "ffead-cpp-mongo-redis",
 			"notes": "mongodb redis",
 			"versus": "",
-			"tags": ["broken"]
+			"tags": []
 		},
 		"lithium": {
 			"json_url": "/te-benchmark-um/json",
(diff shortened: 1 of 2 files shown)


sumeetchhetri:
The vweb tests were causing docker to go into an unresponsive state, the only way the processes could be killed was to restart the docker engine itself, this was the root cause to the stalled tfb tests on multiple occasions.

sumeetchhetri:
ping @nbrady-techempower

nbrady-techempower:
How did you reproduce to be sure it was vweb? What about the vweb test was causing this to happen? I'm hesitant to merge any of these tests back in right now until I have a better understanding of what happened. I also want to better manage the the amount of memory docker uses, possibly with control groups. I'll need some time to set that up. The Docker-Py API we're currently using in the toolset didn't stop this from happening.

sumeetchhetri:
Yes, I just ran a simple test
`
./tfb --mode benchmark --test ffead-cpp-v-vweb
`
Running this blocks docker completely, you cannot even kill the docker processes after this test.
The only resolution is to restart the docker service, also I have disabled 2 other tests which did not complete.

PR #5920, opened 2020-08-05
add one

lizhichao:


frameworks/PHP/one/App/Controllers/IndexController.php
```diff
@@ -0,0 +1,80 @@
...
+
+    public function fortunes()
+    {
+        $arr   = Fortune::findAll()->toArray();
+        $rs    = [];
+        $rs[0] = 'Additional fortune added at request time.';
+        foreach ($arr as $item) {
+            $rs[$item['id']] = $item['message'];
+        }
+        asort($rs);
```
joanhey, line 42: Please, use objects to maintain the full ORM.
Now it's using arrays.
https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#fortunes

lizhichao: `Fortune::findAll()`  is objects to maintain the full ORM

joanhey: For me, I don't see any object.
All is managed like array and the `asort` only order arrays.

lizhichao: `0e78a78` commits , The parameter of `asort` is not `Fortune::findAll()`

matt-42: Using usort directly on the $arr array of objects and removing the $rs key value map would fit the requirement. Here is an example:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/PHP/kumbiaphp/bench/app/controllers/fortunes_controller.php#L10

lizhichao:
done

PR #5931, opened 2020-08-12
Bump finatra and finagle versions to 20.8.0

bonniee:
Bump finagle to 20.8.0
Bump finatra to 20.8.0

### DIFF
--- a/frameworks/Scala/finagle/build.sbt
+++ b/frameworks/Scala/finagle/build.sbt
@@ -1,4 +1,4 @@
-lazy val finagleVersion = "20.7.0"
+lazy val finagleVersion = "20.8.0"
 
 name := "finagle-benchmark"
 scalaVersion := "2.12.8"
--- a/frameworks/Scala/finatra/build.sbt
+++ b/frameworks/Scala/finatra/build.sbt
@@ -1,4 +1,4 @@
-lazy val finatraVersion = "20.7.0"
+lazy val finatraVersion = "20.8.0"
 
 name := "techempower-benchmarks-finatra"
 organization := "com.twitter"


bonniee:
Re-opening to re-run CI.

bonniee:
The previous builds have failed due to issues finding `util-security_2.12/20.8.0`, however it shows up in Maven Central: https://mvnrepository.com/artifact/com.twitter/util-security_2.12/20.8.0

Let's try again. Apologies for the noise.

msmith-techempower:
I have restarted the jobs, though I don't expect that will resolve anything.

Travis, for reasons, caches stuff even when it claims to have no cached objects.

msmith-techempower:
Okay, nuts to figuring Travis out - it will invalidate that cache *someday*. I pulled your branch locally and am testing; I'll let you know if it passes.

msmith-techempower:
Looks like the download error is real:
image

msmith-techempower:
Hmm, it [does exist on Maven](https://repo1.maven.org/maven2/com/twitter/util-security_2.12/20.8.0/)... looks like they got a bad hash in there somehow.

It's pulling the jar and computing the hash, and it always gets a different one(`c3951e4999498c95e659bdccf1b01ddebd29b251`) from the [published one](https://repo1.maven.org/maven2/com/twitter/util-security_2.12/20.8.0/util-security_2.12-20.8.0.jar.sha1)(`e90666a3e5fd9454606988954d0899ec8d6ae6d6`).

image

msmith-techempower:
Basically, I expect everyone to be running into this problem, not just us. Anyone relying on that jar and a java package manager should be computing the hash and getting a different one than the published one, meaning it cannot be trusted.

I don't think we can do anything until the author is made aware and publishes a fix or a new version.

bonniee:
Thank you for investigating that! I’ll investigate what went wrong with the Maven release.

PR #5933, opened 2020-08-14
aspnetcore: ef updates

n-stefan:
Try to improve Entity Framework Updates.


frameworks/CSharp/aspnetcore/Benchmarks/Data/EfDb.cs
```diff
@@ -54,26 +54,25 @@ public async Task<World[]> LoadMultipleQueriesRows(int count)
         public async Task<World[]> LoadMultipleUpdatesRows(int count)
         {
             var results = new World[count];
-            var random = new Random();
-            int i = 0;
+            int currentValue, newValue;
 
-            var ids = Enumerable.Range(1, 10000).OrderBy(x => random.Next()).Take(count);
+            var ids = Enumerable.Range(1, 10000).Select(x => _random.Next(1, 10001)).Distinct().Take(count).ToArray();
```
Kryptos-FR, line 59: In this version, you have a couple of issues:
- no guarantee that enough distinct numbers will be created to satisfy `count`
- it iterates twice over the whole  range (with `Select()` and `Distinct()`)
- it does 3 allocations: 1 for the buffer in `Distinct()`, on for the temporary buffer in `ToArray()` and one for the final array.

The previous implementation does have some allocations implied by `OrderBy()` but it also has the guarantee of having sufficient elements.

That said, both propositions do too many allocations and are not efficient. Instead we can use a simple Fisher-Yates to generate a randomize range in **O(n)** with only one allocation. See [my own implementation](https://github.com/TechEmpower/FrameworkBenchmarks/pull/5933/files) as an example.

```csharp
var ids = ShuffleUtility.RandomRange(1, 10000).Take(count);

foreach (int id in ids)
{
    // etc.
}
```

n-stefan: _count_ is 20 at most so enough distinct numbers should be generated, else there may be a problem with Random.Next(). All implementations are required to generate distinct Ids in order to pass test verification, I believe, though it may allow for a margin of error.
Regarding allocations, there may be a few but I doubt it matters, the database is likely the performance bottleneck.
Feel free to open a PR, if it really improves rps, great.

PR #5937, opened 2020-08-18
Update index-micro.php

diplopito:
Add root entry point and modify `$url` source for handler.

PR #5938, opened 2020-08-18
Update benchmark_config.json

diplopito:
Add `plaintext` test


nbrady-techempower:
Can you keep all phalcon changes to #5937 while it's still open? Thanks!

PR #5950, opened 2020-08-25
Upgrade http4k, add Cached test, add backends, tidy configs for clarity

daviddenton:


daviddenton:
(after a bit of a battle with apache server and hostnames, this is ready to merge now whenever is convenient! 🙃)

PR #5955, opened 2020-08-26
Crax framework added to benchmarks

ephmann:


nbrady-techempower:
Hi @ephmann, 

Thanks for your first contribution! This will need to be added to the [.travis.yml file](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/.travis.yml#L71-L73) to be tested properly

PR #5960, opened 2020-08-28
Lithium optimizations.

matt-42:

### DIFF
--- a/frameworks/C++/lithium/lithium.cc
+++ b/frameworks/C++/lithium/lithium.cc
@@ -37,7 +37,7 @@ void siege(int port) {
 }
 #endif
 
-thread_local lru_cache<int, decltype(mmm(s::id = int(), s::randomNumber = int()))> world_cache(10000);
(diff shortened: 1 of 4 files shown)


frameworks/C++/lithium/lithium.cc
```diff
@@ -191,7 +190,7 @@ int main(int argc, char* argv[]) {
     std::sort(table.begin(), table.end(),
               [] (const fortune& a, const fortune& b) { return a.message < b.message; });
 
-    char b[100000];
+    char b[2000];
```
nbrady-techempower, line 193: This isn't acceptable to cover the dynamic sizing rule of the fortunes test and won't pass the new verification step we're working on where we verify dynamic sizing by adding a large number of fortunes during that step

matt-42: Ok. I'll work on a dynamic sizing implementation.

matt-42: @nbrady-techempower This is fixed in my last commit.

nbrady-techempower: thanks @matt-42!

PR #5962, opened 2020-08-29
op mysql update

lizhichao:
op mysql update


lizhichao:
hi @joanhey `Using bulk updates—batches of update statements or individual update statements that affect multiple rows—is acceptable but not required. `  in specific test requirements。
Can I use `replace into` to update in batches?
The data is updated, but the prompt did not pass ：https://travis-ci.org/github/TechEmpower/FrameworkBenchmarks/jobs/722209086 
What does this mean?

lizhichao:
I found that `PostgreSQL` requires 10752 expected, but why does it require 20480 times for `Mysql` ?

PR #5969, opened 2020-09-01
Add new JavaScript framework: Just-JS

billywhizz:


billywhizz:
Hi. I am not sure of the process for submitting a new framework. This is a new JS platform based on v8 I have been working on for a while. Please let me know if there is anything else you need.

nbrady-techempower:
Hi @billywhizz,

Thanks for submitting. I'll try to review this shortly. In the meantime, all of those `/lib` files... were they built specifically for this benchmark or do they exist somewhere where you can clone them into the the test in the dockerfile. We'd rather not keep those kinds of things here.

billywhizz:
Hi @nbrady-techempower. Thanks for getting back. They are libraries written specifically for the benchmark. I have them in a repo here: https://github.com/just-js/bench. I had seen most other projects have the source checked into your repo so was following that. i can easily move this to the dockerfile and remove everything but the techempower.js script from the PR if that is what you are looking for?

nbrady-techempower:
> I had seen most other projects have the source checked into your repo so was following that.

Yeah, I'm trying to avoid this as much as possible. The ideal scenario is that just the source for the actual test implementation is what should exist in here. Ideally, in a "realistic" approach, users would be installing the necessary http libraries via some package manager instead of writing their own libraries. With a newer project like yours, I understand. Either way is fine for now.

billywhizz:
> > I had seen most other projects have the source checked into your repo so was following that.
> 
> Yeah, I'm trying to avoid this as much as possible. The ideal scenario is that just the source for the actual test implementation is what should exist in here. Ideally, in a "realistic" approach, users would be installing the necessary http libraries via some package manager instead of writing their own libraries. With a newer project like yours, I understand. Either way is fine for now.

That makes sense. If ok with you I would be happy to leave as is for now. I am working on a module/build system for the platform at moment. Will hopefully write it all up soon and get some community feedback. Looking forward to seeing how it does in your environment. If I can help with your project in any way please let me know.

PR #5973, opened 2020-09-01
Added swoole-postgres

doubaokun:
* Added swoole-postgres
* Updated the PHP version to be 7.4
* Updated Swoole postgres version to be 4.5.1


frameworks/PHP/swoole/benchmark_config.json
```diff
@@ -24,6 +24,29 @@
       "notes": "",
       "versus": "php"
     },
+    "postgres": {
+      "json_url": "/json",
+      "plaintext_url": "/plaintext",
```
joanhey, line 29: Delete the tests not using db.

joanhey (changes requested):

frameworks/PHP/swoole/swoole-server-postgres.php
```diff
@@ -0,0 +1,247 @@
...
+    private function createDbInstance()
+    {
+        if ($this->type === 'postgres') {
+            $db = new Swoole\Coroutine\PostgreSql;
+            if ($db->connect("host={$this->server['host']} port=5432 dbname={$this->server['database']} user={$this->server['user']} password={$this->server['password']}")){
+                return $db;
+            }
+        }
+        
+        return false;
```
joanhey, line 235: Why not add in `swoole-server.php` for both mysql and postgres?
So will be only one codebase to maintain.

frameworks/PHP/swoole/benchmark_config.json
doubaokun: Removed now.

doubaokun:
@joanhey Thanks for the review. The code is updated based on suggestion.

joanhey (approved):

PR #5980, opened 2020-09-03
http4k - tidy, adding virtual hosting to apache server implementation

daviddenton:
... as the TFB build is actually inconsistent with the way tests are run in the benchmarking

### DIFF
--- a/frameworks/Kotlin/http4k/apache/src/main/kotlin/Http4kApacheServer.kt
+++ b/frameworks/Kotlin/http4k/apache/src/main/kotlin/Http4kApacheServer.kt
@@ -1,6 +1,41 @@
-
-import org.http4k.server.ApacheServer
+import org.apache.hc.core5.http.impl.bootstrap.ServerBootstrap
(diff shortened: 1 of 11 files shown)


daviddenton:
After a tiny fight with gradle and docker, this is now ready to merge. It should fix the http4k and http4k-apache benchmarks :)

BogdanJava:
Hi @daviddenton after upgrading our http4k version I stumbled upon this issue while trying to deploy our app to k8s - liveness probe was failing with 421 "Not authoritative". Is there a proper way to configure my app to avoid this error or I should simply use **Apache4Server** instead?

daviddenton:
I'm not really an expert on the configuration of the Apache backend tbh.... I tend to use undertow as a first option whenever I can (it's easier to configure). Feel free to switch obvs but Apache 4 isn't being developed any more so if there were cves from it I'm not sure it would get updated by the Apache team.

Sorry for not being more helpful - please let us know if you fix it in v5 🙃

PR #6000, opened 2020-09-15
Update .NET frameworks to .NET 5.0

benaadams:
.NET 5.0 has moved to rc 1 which means it's now officially supported in production


sandeep-eroad:
> .NET 5.0 has moved to rc 2 which means it's now officially supported in production

should it be rc1?

benaadams:
> should it be rc1?

Yes, clearly I'm living in the future ;)

sebastienros:
We checked this morning to see if you had done it already. Now I understand why it was not here yet 😅

benaadams:
Is now :)

nbrady-techempower:
Woo! Congrats!

PR #6009, opened 2020-09-17
Swoole workers

joanhey:
Revert workers in async mode, after the bad results.
Add more workers in no async mode.

Ping: @doubaokun


doubaokun:
@joanhey LGTM. Thanks.

PR #6012, opened 2020-09-18
Use SmallRng in warp-rust

xfix:
I'm not sure whether this is fine. See, tests should be "production-grade" (I probably would never use `SmallRng` in production), but at the same time I feel like using a worse RNG algorithm will make the benchmark more comparable with implementations in other programming languages, and in fact there are already implementations written in Rust that use `SmallRng`.

Currently `SmallRng` uses a [permuted congruential generator](https://en.wikipedia.org/wiki/Permuted_congruential_generator) with 128-bit state (albeit this will change in the near future - `SmallRng` was replaced with xoshiro256** on `rand` crate master).

Figured I would send that anyway and ask you for opinion whether this okay.

### DIFF
--- a/frameworks/Rust/warp-rust/Cargo.toml
+++ b/frameworks/Rust/warp-rust/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2018"
 
 [dependencies]
 futures = "0.3.1"
-rand = "0.7.3"
+rand = { version = "0.7.3", features = ["small_rng"] }
 serde = { version = "1.0.103", features = ["derive"] }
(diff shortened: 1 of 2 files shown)


nbrady-techempower:
This isn't something we have an issue with right now, I think. If you can just fix the conflict, I'll get this in.

xfix:
Hm, weird, I think I accidentally pushed code using `Cell` (was just experimenting, but that was ugly), when I wanted to use `RefCell`.

xfix:
Okay, it's now using `RefCell` as I initially meant to.

PR #6020, opened 2020-09-20
cutelyst: Update to newer version, add async SQL, use Cutelee templating

dantti:


frameworks/C++/cutelyst/src/databaseupdatestest.cpp
```diff
@@ -13,6 +17,47 @@ DatabaseUpdatesTest::DatabaseUpdatesTest(QObject *parent) : Controller(parent)
...
+        int id = (qrand() % 10000) + 1;
+
+        int randomNumber = (qrand() % 10000) + 1;
+
+        array.append(QJsonObject{
+                         {QStringLiteral("id"), id},
+                         {QStringLiteral("randomNumber"), randomNumber}
+                     });
+
+        db.execPrepared(APreparedQueryLiteral("SELECT randomNumber FROM world WHERE id=$1"),
```
nbrady-techempower, line 42: Should be selecting `id` and `randomNumber` from the db on each query

dantti: Well I re-read the requirements and was unsure if that was needed, after all the where is already selecting the id, if it ok if I do select id, random from... but doesn't use the id from the query?

dantti: What I mean is that as you can see I'm building the json before because I already know how it's going to be, doesn't make much sense to build it later.

nbrady-techempower: > where is already selecting the id

What we mean by selecting the `id` is that it will be returned from the database. Without the `id` in the select clause, fewer bytes are returned:

```
hello_world=# select randomNumber from world where id=1;
 randomnumber
--------------
         5703
(1 row)

hello_world=# select id, randomNumber from world where id=1;
 id | randomnumber
----+--------------
  1 |         5703
(1 row)
```

dantti: Ok, look:
At least the randomNumber field must be read from the database result set.
That's on the wiki requirements page, so I'm ok right?

dantti: vi -> https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#database-updates

nbrady-techempower: No. That is "read from the result set" meaning you must access that field of the object that's returned from the database.

> Each row must be selected randomly using one query in the same fashion as the single database query test.

And then from the single database query test:

> 1. For every request, a single row from a World table must be retrieved from a database table. It is not acceptable to read just the randomNumber.

nbrady-techempower: If you feel like this is unclear, that is fine. I will update all of the rules to make it clearer that is not acceptable to read just the `randomNumber` field from the World table for any of the database query / update tests.

dantti: >No. That is "read from the result set" meaning you must access that field of the object that's returned from the database.

hmm how should I "access" it, because if I don't have an use for it the compiler will simply optimize the code and remove the retrieval.

nbrady-techempower: The compiler is smart enough to remove the "id" from the prepared select statement? that's impressive, if that's what's happening. this is fine for now, but something i'd be interested in looking into further at some point.

dantti: Nope, what I was saying was about the access the current random number returned by the database.

PR #6021, opened 2020-09-21
Fix dotnet docker images (every .NET server now failing)

benaadams:
Just started breaking because `core/` has been dropped from the path


benaadams:
Thank you

PR #6023, opened 2020-09-21
Lithium update

matt-42:

### DIFF
--- a/frameworks/C++/lithium/lithium.cc
+++ b/frameworks/C++/lithium/lithium.cc
@@ -67,10 +67,10 @@ struct json_cache {
     std::string json = "[";
     json.append(buffer, positions[ids[0]], positions[ids[0]+1] - positions[ids[0]]);
     for (int i = 1; i < ids.size(); i++) {
-      json.append(',', 1);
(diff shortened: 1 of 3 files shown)


frameworks/C++/lithium/lithium.cc
```diff
@@ -54,7 +54,31 @@ void siege(int port) {
...
+    }
+    json.append(1, ']');
+    return json;
+  }
+
+  std::string buffer;
+  std::vector<int> positions;
+};
+
+json_cache world_cache;
```
nbrady-techempower, line 81: It's too early on Monday morning for me to figure out exactly what this cache is doing or why it's needed but if it's caching anything so that subsequent serializations require less work, it's a violation of the last rule in the JSON requirements.

> The serialization to JSON must not be cached

matt-42: Sorry @nbrady-techempower  I forgot about this rule. I removed the json caching in my last commit.

nbrady-techempower: Thanks! Will merge when it passes travis!

PR #6026, opened 2020-09-22
Remove Red Hat Linux Transport benchmarks.

tmds:
There are no plans to maintain the Red Hat Linux transport for .NET 5 and later.

@benaadams @sebastienros ptal


benaadams (approved):
😢 🥃

PR #6031, opened 2020-09-24
fix update del transaction

lizhichao:

### DIFF
--- a/frameworks/PHP/one/App/Controllers/IndexController.php
+++ b/frameworks/PHP/one/App/Controllers/IndexController.php
@@ -57,15 +57,9 @@ class IndexController extends Controller
         while ($count--) {
             $row       = World::repeatStatement()->find(mt_rand(1, 10000));
             $list[]    = $row;
-//            $old       = intval($row->randomNumber);
-            $new       = mt_rand(1, 10000);
-            $updates[] = 'update world set randomNumber=' . $new . ' where id=' . $row->id;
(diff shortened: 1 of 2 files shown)


frameworks/PHP/one/App/Controllers/IndexController.php
```diff
@@ -57,15 +57,9 @@ public function updates($count = 1)
...
-            $new       = mt_rand(1, 10000);
-            $updates[] = 'update world set randomNumber=' . $new . ' where id=' . $row->id;
+            $updates[] = 'update world set randomNumber='.mt_rand(1, 10000).' where id='.$row->id;
         }
-        World::beginTransaction();
-        foreach ($updates as $sql) {
-            $row->exec($sql);
-        }
-        World::commit();
+        $row->exec(implode(';',$updates));
```
joanhey, line 62: Fix to meet the requirements.

 https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#database-updates

vii. Each World object must have its randomNumber field updated to a new random integer between 1 and 10000.
This code never update de World object only send the new randomnumber to the bd.

viii. Each World object must be persisted to the database with its new randomNumber value.
World->update();  
Now it's using SQL to change the values.
Using SQL for that simple action, break to be a Full ORM.

joanhey (changes requested):

joanhey (approved):

PR #6035, opened 2020-09-25
add cached test

lizhichao:

### DIFF
--- a/frameworks/PHP/one/App/Controllers/IndexController.php
+++ b/frameworks/PHP/one/App/Controllers/IndexController.php
@@ -74,6 +74,17 @@ class IndexController extends Controller
         return $this->json($list);
 
     }
+
+    public function cachedWorlds($count = 1)
+    {
+        $count = max(min(intval($count), 500), 1);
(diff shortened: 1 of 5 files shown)


frameworks/PHP/one/App/Controllers/IndexController.php
```diff
@@ -74,6 +74,17 @@ public function queries($count = 1)
...
+    public function cachedWorlds($count = 1)
+    {
+        $count = max(min(intval($count), 500), 1);
+        $list  = [];
+        while ($count--) {
+            $list[] = $this->server->caches[mt_rand(1,10000)];
+        }
+        return $this->json($list);
+    }
+
```
nbrady-techempower, line 87: I realize there are a couple of cache implementations that are not following the rules. I think I'm even responsible for one of them while I was adding the test to make sure things worked, and those are going to be removed before the next round. This violates rule `vii`: https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#caching

PR #6040, opened 2020-09-26
Update http4s

wjoel:
- Replace doobie with quill and async drivers.
- Update http4s, Scala, etc. to newest versions.
- Update to JDK 15.
- Improve JVM flags.
- Tweak thread pool sizes.

### DIFF
--- a/frameworks/Scala/http4s/src/main/scala/http4s/techempower/benchmark/DatabaseService.scala
+++ b/frameworks/Scala/http4s/src/main/scala/http4s/techempower/benchmark/DatabaseService.scala
@@ -1,0 +1,73 @@
+package http4s.techempower.benchmark
+
+import java.util.concurrent.{Executor, ThreadLocalRandom}
+
+import scala.concurrent.{ExecutionContext, ExecutionContextExecutor}
+import cats.effect.{ContextShift, IO => CatsIO}
+import cats.syntax.all._
+import io.getquill._
+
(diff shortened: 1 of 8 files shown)


frameworks/Scala/http4s/src/main/scala/http4s/techempower/benchmark/DatabaseService.scala
```diff
@@ -0,0 +1,73 @@
...
+  // Update the randomNumber field with a random number
+  def updateRandomNumber(world: World): CatsIO[World] =
+    for {
+      randomId <- randomWorldId()
+    } yield world.copy(randomNumber = randomId)
+
+  // Select a World object from the database by ID
+  def selectWorld(id: Int): CatsIO[World] =
+    CatsIO.fromFuture(
+      CatsIO.pure(
```
Daenyth, line 33: Shouldn't this be `delay`? You don't generally use `fromFuture` with `pure`

wjoel: I wanted the eager evaluation, but that was a bit of premature optimization. Using `delay` is more correct, as you say, and doesn't seem to make any difference for the benchmark results. Changed, thanks!

rossabaker (approved):
:+1: from a maintainer

PR #6044, opened 2020-09-27
Optimise swoole-pg

doubaokun:
Optimise `Swoole-postgres`.

### DIFF
--- a/frameworks/PHP/swoole/swoole-server.php
+++ b/frameworks/PHP/swoole/swoole-server.php
@@ -38,15 +38,15 @@ $db_postgres = function (int $queries = 0) use ($pool): string {
     // Create an array with the response string.
     $arr = [];
 
-    $db->prepare('s', 'SELECT id, randomnumber FROM World WHERE id = $1');
+    $db->s = $db->s ?? $db->prepare('s', 'SELECT id, randomnumber FROM World WHERE id = $1');
 
     // For each query, store the result set values in the response array
     while ($query_count--) {
         $id = mt_rand(1, 10000);
         $res = $db->execute('s', [$id]);
-        $ret = $db->fetchAll($res);
+        $ret = $db->fetchAssoc($res);
         // Store result in array.
-        $arr[] = ['id' => $id, 'randomnumber' => $ret[0]['randomnumber']];
+        $arr[] = ['id' => $id, 'randomnumber' => $ret['randomnumber']];
     }
 
     // Use the PHP standard JSON encoder.
@@ -72,7 +72,7 @@ $fortunes_postgres = function () use ($pool): string {
 
     $fortune = [];
 
-    $db->prepare('f', 'SELECT id, message FROM Fortune');
+    $db->f = $db->f ?? $db->prepare('f', 'SELECT id, message FROM Fortune');
     $res = $db->execute('f', []);
     $arr = $db->fetchAll($res);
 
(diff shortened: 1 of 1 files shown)


frameworks/PHP/swoole/swoole-server.php
```diff
@@ -38,15 +38,15 @@
     // Create an array with the response string.
     $arr = [];
 
-    $db->prepare('s', 'SELECT id, randomnumber FROM World WHERE id = $1');
+    $db->s = $db->s ?? $db->prepare('s', 'SELECT id, randomnumber FROM World WHERE id = $1');
```
jcheron, line 41: `$db->s` is useless and could be removed.

This is enouth, no?
```php
$db->prepare('s', 'SELECT id, randomnumber FROM World WHERE id = $1');
```

frameworks/PHP/swoole/swoole-server.php
```diff
@@ -72,7 +72,7 @@
 
     $fortune = [];
 
-    $db->prepare('f', 'SELECT id, message FROM Fortune');
+    $db->f = $db->f ?? $db->prepare('f', 'SELECT id, message FROM Fortune');
```
jcheron, line 75: Same thing:
```php
$db->prepare('f', 'SELECT id, message FROM Fortune');
```

frameworks/PHP/swoole/swoole-server.php
```diff
@@ -113,16 +113,16 @@
 
     $arr = [];
 
-    $db->prepare('us', 'SELECT randomnumber FROM World WHERE id = $1');
-    $db->prepare('uu', 'UPDATE World SET randomnumber = $1 WHERE id = $2');
+    $db->us = $db->us ?? $db->prepare('us', 'SELECT randomnumber FROM World WHERE id = $1');
+    $db->uu = $db->uu ?? $db->prepare('uu', 'UPDATE World SET randomnumber = $1 WHERE id = $2');
```
jcheron, line 117:
```php
$db->prepare('us', 'SELECT randomnumber FROM World WHERE id = $1');
$db->prepare('uu', 'UPDATE World SET randomnumber = $1 WHERE id = $2');
```

jcheron (changes requested):
Either we use the named prepared queries of **Swoole_postgresql**:
```php
$db->prepare('us', 'SELECT randomnumber FROM World WHERE id = $1');
$res = $db->execute('us', [$id]);
```
Either we store in a variable the statement:
```php
$db->s ??= $db->prepare('SELECT randomnumber FROM World WHERE id = $1');
$res = $db->s->execute([$id]);
```
but not both mixed

frameworks/PHP/swoole/swoole-server.php
doubaokun: @jcheron Thanks for the review, but`$db->s` is used to store states, it is initialized only once across multiple requests and added by purpose. 

There is no such API `$db->s->execute([$id]);`.

Same thing for the following suggestions.

jcheron: I can' t see where are used `$db->us`, `$db->uu` and `$db->s`...
The initialization of `$db->s` and others seems to be unused afterwards.

we can assume that this code below already does this initialization work for each worker:
```php
$db->prepare('s', 'SELECT id, randomnumber FROM World WHERE id = $1');
```

And that it is possible to run the prepared query `s` several times this way:
```php
$res = $db->execute('s', [$id]);
```

doubaokun: Hi @jchannon, these variable is for checking if the `prepare` query is executed for once. For performance reason, these `prepare` query are only executed once for all the requests after the change. Hope this make sense?

jcheron: Hi @doubaokun 
I modified my previous comment after testing

doubaokun:
@jcheron updated to the shorter form.

jcheron (approved):

PR #6051, opened 2020-10-03
Update PostgresQL to version 13

### LINKED ISSUE #6050: postgres.dockerfile doesn't build
`toolset/databases/postgres/postgres.dockerfile` doesn't build anymore for me

Reporting before investigating

```
[12:48:56.471] Step 19/35 : RUN mv postgresql.conf /etc/postgresql/${PG_VERSION}/main/postgresql.conf
[12:48:56.943]  ---> Running in 0e1a5fb95a7f
[12:48:57.270] mv: cannot move 'postgresql.conf' to '/etc/postgresql/12/main/postgresql.conf': No such file or directory
[12:48:57.501] [STDERR] The command '/bin/sh -c mv postgresql.conf /etc/postgresql/${PG_VERSION}/main/postgresql.conf' returned a non-zero code: 1
[12:48:57.502]
```

sebastienros:
Fixes #6050 

Not sure this is the fix you want to take since it might impact benchmarks.

### DIFF
--- a/toolset/databases/postgres/postgres.dockerfile
+++ b/toolset/databases/postgres/postgres.dockerfile
@@ -14,7 +14,7 @@ RUN wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-k
 RUN apt-get -yqq update > /dev/null
 RUN apt-get -yqq install locales
 
-ENV PG_VERSION 12
+ENV PG_VERSION 13
 RUN locale-gen en_US.UTF-8
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US:en
--- a/toolset/databases/postgres/postgresql.conf
+++ b/toolset/databases/postgres/postgresql.conf
@@ -40,13 +40,13 @@
 
 data_directory = '/ssd/postgresql'		# use data in another directory
 					# (change requires restart)
-hba_file = '/etc/postgresql/12/main/pg_hba.conf'	# host-based authentication file
+hba_file = '/etc/postgresql/13/main/pg_hba.conf'	# host-based authentication file
 					# (change requires restart)
-ident_file = '/etc/postgresql/12/main/pg_ident.conf'	# ident configuration file
+ident_file = '/etc/postgresql/13/main/pg_ident.conf'	# ident configuration file
 					# (change requires restart)
 
 # If external_pid_file is not explicitly set, no extra PID file is written.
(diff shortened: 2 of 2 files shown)


volyrique:
Shouldn't we stop hardcoding version numbers? At the very least not inside the configuration files and not in the paths.

toolset/databases/postgres/postgres.dockerfile
```diff
@@ -14,7 +14,7 @@ RUN wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-k
 RUN apt-get -yqq update > /dev/null
 RUN apt-get -yqq install locales
 
-ENV PG_VERSION 12
+ENV PG_VERSION 13
```
jcheron, line 17: maybe add a `PG_VERSION` mask in `postgresql.conf` and replace it from here:
```bash
RUN sed -i "s|PG_VERSION|${PG_VERSION}|g" postgresql.conf
```

toolset/databases/postgres/postgresql.conf
```diff
@@ -40,13 +40,13 @@
 
 data_directory = '/ssd/postgresql'		# use data in another directory
 					# (change requires restart)
-hba_file = '/etc/postgresql/12/main/pg_hba.conf'	# host-based authentication file
+hba_file = '/etc/postgresql/13/main/pg_hba.conf'	# host-based authentication file
```
jcheron, line 43:
```
hba_file = '/etc/postgresql/PG_VERSION/main/pg_hba.conf'
```

sebastienros:
@jcheron I think this PR solves it in a better way by pinning PG 12 at the source: https://github.com/TechEmpower/FrameworkBenchmarks/pull/6052

And in PG13 the default conf file is different, so I don't think making it a constant here makes sense, since bumping to13 should also change the conf file completely.

jcheron:
PR https://github.com/TechEmpower/FrameworkBenchmarks/pull/6052 generates a 404 error on package downloads...
It is necessary to modify the list of packages without using the provided list file (which has the disadvantage of being also hard-coded).
```bash
RUN echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release  -sc || cat /etc/*-release|grep -oP  'CODENAME=\K\w+$'|head -1)-pgdg main" > /etc/apt/sources.list.d/pgdg.list
```

PR #6041, opened 2020-09-27
Add ocaml webmachine

rbjorklin:
If accepted this PR will add an initial implementation of OCaml webmachine to the test suite. This first implementation only covers tests 1, 2, 3 and 6 as per the [project recommendations](https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview).

There is one warning present around the formatting of the contents of the `Date` header. I was unable to find to find an OCaml library that can produce date strings in the `RFC1123` format and I have therefore settled upon using the `RFC3339` (ISO8601) format. I have however created a [draft PR adding the RFC1123 format](https://github.com/dbuenzli/ptime/pull/21).

PR #6060, opened 2020-10-09
fix: significantly improve webmachine performance

rbjorklin:
* installing libev-devel and conf-libev enables epoll instead of select
* multithreading haproxy improves throughput
* GC tweaks
* use upstream docker image with ocaml preinstalled to save some time

These changes _triple_ (:exclamation:) performance on my local machine.

See #6041 for my first PR that introduced webmachine.


rbjorklin:
@nbrady-techempower would really appreciate a merge before the weekend so my changes can get in before the next run :sun_with_face:

Issue #6065, opened 2020-10-10
Running verify fails


ulrikstrid:
### OS (Please include kernel version)
macOS Cataline (10.15.7)

### Expected Behavior

Running `./tfb --mode verify --test morph` should work

### Actual Behavior

It fails with the bellow output

### Other details and logs

```
./tfb --mode verify --test morph
[+] Building 0.1s (10/10) FINISHED                                                                                                                                    
 => [internal] load .dockerignore                                                                                                                                0.0s
 => => transferring context: 2B                                                                                                                                  0.0s
 => [internal] load build definition from Dockerfile                                                                                                             0.0s
 => => transferring dockerfile: 1.02kB                                                                                                                           0.0s
 => [internal] load metadata for docker.io/library/buildpack-deps:bionic                                                                                         0.0s
 => [1/6] FROM docker.io/library/buildpack-deps:bionic                                                                                                           0.0s
 => CACHED [2/6] RUN apt-get -yqq update                                                                                                                         0.0s
 => CACHED [3/6] RUN apt-get -yqq install -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold"   git-core   cloc dstat                      0.0s
 => CACHED [4/6] RUN pip install colorama==0.3.1 requests MySQL-python psycopg2-binary pymongo docker==4.0.2 psutil                                              0.0s
 => CACHED [5/6] RUN apt-get install -yqq siege                                                                                                                  0.0s
 => CACHED [6/6] RUN cp -r /usr/local/lib/python2.7/dist-packages/backports/ssl_match_hostname/ /usr/lib/python2.7/dist-packages/backports                       0.0s
 => exporting to image                                                                                                                                           0.0s
 => => exporting layers                                                                                                                                          0.0s
 => => writing image sha256:7b51c315aa5702f8f4d4eb1a3acbb37337a43430516d0b04c9b5fcf04f029923                                                                     0.0s
 => => naming to docker.io/techempower/tfb                                                                                                                       0.0s
================================================================================
Running Tests...
================================================================================
wrk: Docker build failed; terminating
wrk: Traceback (most recent call last):
wrk:   File "/FrameworkBenchmarks/toolset/utils/docker_helper.py", line 52, in __build
wrk:     token = json.loads(token)
wrk:   File "/usr/lib/python2.7/json/__init__.py", line 339, in loads
wrk:     return _default_decoder.decode(s)
[...]

PR #6067, opened 2020-10-10
Fix parsing of docker build output in docker_helper

matthewmcgarvey:
Fixes: #6065 

When parsing the list of docker build output, some of the lines had multiple json objects. In the tests I ran, it was always a newline character.

The python docker library has a parameter to automatically parse the output into dicts. Switching to that solved the problem. https://docker-py.readthedocs.io/en/4.0.2/api.html?highlight=APIClient#module-docker.api.build


nbrady-techempower:
Hey @matthewmcgarvey Thanks for looking into and solving this!

PR #6069, opened 2020-10-11
[Crystal] Add Lucky Framework

matthewmcgarvey:
Adding [Lucky](https://luckyframework.org/) which is a Crystal framework.

Was kept from running tests by #6065 unless using changes in #6067


frameworks/Crystal/lucky/src/actions/db/index.cr
```diff
@@ -0,0 +1,6 @@
+class Db::Index < BaseAction
+  get "/db" do
+    world = WorldQuery.find(rand(1..ID_MAXIMUM))
```
nbrady-techempower, line 3: I vaguely remember taking a dive into the Crystal codebase and seeing that the `rand(range)` wasn't inclusive on the upper bounds. I could be wrong, but want to confirm that this could return 10,000.

nbrady-techempower: Yeah, I remember now. The documentation wasn't clear so I had to dig through the code: https://crystal-lang.org/api/0.24.2/Random.html#rand%28range%3ARange%28Int%2CInt%29%29%3AInt-instance-method

https://github.com/crystal-lang/crystal/blob/4f9ed8d03208dd0db33993c5a6fa6753bc1cf91e/src/random.cr#L219 Seems in this case it's basically doing:

`range.begin + rand(range.end - range.begin)` and `rand(int)` is not inclusive, so the max of : `1 + rand(9999)` is 9999

matthewmcgarvey: admittedly I copied it from the other crystal projects so it definitely could have been wrong but I ran a little test in a repl calling `rand(1..3)` and was able to get a 3. https://play.crystal-lang.org/#/r/9toe

nbrady-techempower: wow that is interesting. the source code in the crystal repo reads to me like that would be doing:

`puts (1..10_000).map { 1 + rand(3 - 1) }.uniq`

But the REPL says otherwise lol, thanks!

PR #6084, opened 2020-10-18
[New Framework] Nim/Prologue

rishavs:
Hi. I have added a basic plaintext and json benchmarks for the new Nim framework [Prologue](https://github.com/planety/prologue). More details about it can be seen here https://planety.github.io/prologue/

I will be adding the database benchmarks later.


frameworks/Nim/prologue/prologue.dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM nimlang/nim:latest
```
nbrady-techempower, line 1: Please tie this down to a specific version for reproducability/stability

rishavs: I started with `nimlang/nim:1.14` but that doesn't seems to be supported.
What's the current latest nim version that TFB supports?

nbrady-techempower: It's coming from dockerhub. We don't choose the support for these https://hub.docker.com/r/nimlang/nim/tags

.travis.yml
```diff
@@ -63,7 +63,7 @@ env:
     - "TESTLANG=Lua"
     - "TESTLANG=Mumps"
     - "TESTLANG=Nim"
-    - 'TESTDIR="Nim/httpbeast Nim/jester Nim/basolato"'
+    - 'TESTDIR="Nim/httpbeast Nim/jester Nim/basolato Nim/prologue"'
```
nbrady-techempower, line 66: Also, please remove this line. The line above it tests all Nim frameworks

rishavs: It was there previously as well. But I will remove the 2nd line.

rishavs: Done

msmith-techempower (commented):
This is one of the greatest commit messages ever. I applaud your resolve!

PR #6085, opened 2020-10-19
opium: restructure for easier reuse

mudrz:
- updates the readme since it had missing libraries
- restructures the library and removed a dependency on Opium, so that the core logic can be more easily reused/copy pasted by other frameworks


frameworks/OCaml/.gitignore
```diff
@@ -2,3 +2,4 @@
 **/_build
 .merlin
 .ocamlformat
+!**/bin
```
mudrz, line 5: in OCaml /bin is a common naming convention for putting executables, so removing it from the global `.gitignore`
https://dune.build/examples

nbrady-techempower:
@mudrz would you mind updating this conflict, or removing the gitignore altogether. I don't think there's a real need to push it up here for anyone cloning, as all the generated files would be contained to the docker container.

mudrz:
hey @nbrady-techempower resolved the conflict

I added it to the parent, because in OCaml projects `bin` is a common convention for code, not generated executables

so anyone adding additional OCaml frameworks would be surprised to find out that `bin` is globally ignored

PR #6086, opened 2020-10-20
[Crystal] Run Lucky per core on machine

matthewmcgarvey:


CaDs:
Nice! 
Looking forward to seeing the results 👍

PR #6094, opened 2020-10-25
Add json benchmark support to dylan framework.

abeaumont:


nbrady-techempower:
Awesome! Thanks for adding that @abeaumont !

PR #6097, opened 2020-10-27
Updating to OfficeFloor 3.28.0

sagenschneider:
We have deprecated support for Rapidoid, as found unstable.   However, have added support for Undertow (used by JBoss).


sagenschneider:
Tests are failing because OfficeFloor deliberately sends 503 responses to gracefully shed load when under stress.

Explanation:
Typically, web servers will leave queued requests on socket buffers to let TCP/network handle load.  OfficeFloor finds that this can cause unresponsive applications due to network escalation back offs going into the seconds.  To avoid network delays, when networks are just getting faster, OfficeFloor provides gracefully back off by 503 responses.  This means that when OfficeFloor detects it is under load, it can immediately respond with 503 to avoid clients waiting.  The client can then provide responsive feedback to the user (rather than just hanging).

However, this means in the benchmark performance tests, there will be 503 responses for the heavier load tests (queries and updates).  We at OfficeFloor find this an actually success, however given the nature of expecting all requests to pass, this means OfficeFloor may not meet the always 200 responses defined by the tests of this build.


This has been detailed in the read me, which we've copied below for convenience:

 Within the performance tests (particularly the query and update tests), there are 503 responses from OfficeFloor.  These are deliberate and make OfficeFloor significantly more responsive than typical web servers.  To understand why, we need to explain the servicing change introduced by OfficeFloor.
 
 **Typical web servers** manage load by only processing one request at a time per socket.  This means on pipelined requests over a single connection, the first request needs to be fully processed before the next request is started.  As requests then queue on the network buffer, it allows the network's natural TCP throttling to manage load.
 
 **Web browser request profiles have changed** to an increased number of asynchronous requests pipelined over a few number of connections.  Days of old (possibly very old), web pages loaded and then did a post to reload the page again.  There would be one page submission with a few resource requests (e.g. images).  These days, single page applications are a lot more prevalent.  The nature of single page applications is to avoid page reloads and run many asynchronous requests.  This means more logic submission requests in parallel.  As browsers re-use connections to multiplex these requests to the server, the result is pipelining requests over a few number of connections.
 
 **OfficeFloor processes pipelined requests in parallel**.  To cater to this situation and allow more responsive single page web applications, OfficeFloor processes the pipeline requests in parallel.  This means the second request in the pipeline no longer has to wait for the first one to complete.  Both the first, second, third, etc will be processed immediately by OfficeFloor.  This means better responsiveness to the single page applications, as there is no queuing of requests in the pipelines.
 
 **Load throttling managed by 503 responses.**  As requests are no longer queued on the network buffer, there is no TCP throttling.  Requests will be accepted in parallel into OfficeFloor.  To avoid denial of service attacks, OfficeFloor has built in load throttling.  It is beyond this overview discussion to explain how this works, but basically when thread pools are overloaded they reject new jobs.  OfficeFloor captures these rejections and translates them into 503 HTTP responses (indicating temporarily unavailable).
 
 **This improves responsiveness of single page applications**.  The problem with TCP throttling is that requests start to slow down before you get rejections.  This means your single page application ends up waiting for responses.  The wait time is noticeable by the user, as can be upwards of 10's of seconds to minutes.  This then results in the user's perception of the single page application being slow.  OfficeFloor by parallel processing everything and responding immediately with 503 responses when overloaded, reduces these long wait times.  Rather than than hanging, the single page application can deal with the 503 by calling another server or providing user friendly messages back to the user that system is under load and try back in a few moments.  This provides an overall more responsive and what we find is a better experience for the user.
 
 However, this does mean in some of the heavier load tests, you will see errors for OfficeFloor.  But we here at OfficeFloor are proud of these errors, as they avoid hanging applications and provide better user experience under extreme loads.

sagenschneider:
Tidied up PR to deal with avoiding:
 - overload conditions
 - caching of entities
 - not updating entity if same random values
 - adding DB query to raw

sagenschneider:
Adding support for all remaining tests to officefloor-raw

sagenschneider:
Will look to put a PR together soon for cached queries.  But will do separately and get the upgrades in first.

PR #6099, opened 2020-10-27
[cutelyst] Update to v2.13.0 and add an async update in batched mode

dantti:


nbrady-techempower:
hey @dantti -- looks like a few tests didn't start

nbrady-techempower:
I'll get this in but please update the missing tests. Going to kick of Round 20

dantti:
sure, I just didn't find time to checkout why the tests worked here and crashed there..

PR #6110, opened 2020-11-01
Use nightly esy when building

ulrikstrid:
Fixes the ocamlfind build by using nightly esy that has fixed the bug


nbrady-techempower:
@ulrikstrid does `nightly` here imply that they'll be replacing this image with updated nightly images? like `latest` for this version? if that's the case, we don't want to do that here.

ulrikstrid:
@nbrady-techempower it's just the package manager that is nightly. It has a bugfix that stopped the build in the last run.

I will update to not nightly as soon as it's possible.

I control the base images and won't do any weird changes.

PR #6131, opened 2020-11-11
add OCaml httpaf

blandinw:
Hi, this adds a server using [httpaf](https://github.com/inhabitedtype/httpaf).
Regarding `.travis.yml`, I'm assuming I don't need to update it since [the line below](https://github.com/TechEmpower/FrameworkBenchmarks/blob/5b19c803cb46840a550311c6885a532df69ba00e/.travis.yml#L66) is already specified.
```
- "TESTLANG=OCaml"
```
Thanks!


nbrady-techempower:
That's correct @blandinw 

Thanks for your first contribution!

frameworks/OCaml/httpaf/httpaf_unix.ml
```diff
@@ -0,0 +1,180 @@
...
+(* Dates *)
+
+let get_date () = Unix.(gettimeofday () |> gmtime)
+
+let dow = function
+  | 0 -> "Sun"
+  | 1 -> "Mon"
+  | 2 -> "Tue"
+  | 3 -> "Wed"
+  | 4 -> "Tue"
```
rbjorklin, line 14: @blandinw While adopting your changes to the webmachine test suite I found this typo. Should be `Thu`, right?

blandinw: Yes, good catch!

PR #6147, opened 2020-11-15
fix: opium is built on top of the httpaf platform

rbjorklin:
opium is built on top of httpaf so I think these changes should show how much overhead is introduced by opium itself as httpaf was added in #6131.

### DIFF
--- a/frameworks/OCaml/httpaf/benchmark_config.json
+++ b/frameworks/OCaml/httpaf/benchmark_config.json
@@ -13,7 +13,7 @@
         "language": "OCaml",
         "flavor": "None",
         "orm": "None",
-        "platform": "None",
+        "platform": "httpaf",
         "webserver": "None",
         "os": "Linux",
         "database_os": "Linux",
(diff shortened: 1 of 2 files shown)


blandinw:
I agree, although it's not an apple to apple comparison since the httpaf benchmark does not use haproxy but instead shares a socket among children, saving precious CPU cycles. In the latest run, the other OCaml benchmarks also seem to configure HAProxy in `http` mode (as opposed to `tcp`) which is probably very inefficient since it forces 2 HTTP parses per request.

rbjorklin:
Yeah that was introduced in #6133 by me. I was hoping being able to use take advantage of http-keepalive would offset that but it does indeed seem to have had a significant negative impact. [`tcp` mode](https://www.techempower.com/benchmarks/#section=test&runid=1e6b74c0-4672-41f5-be44-13d2a019a74d&hw=ph&test=json&l=zik0zj-v&a=2) compared to [`http` mode](https://www.techempower.com/benchmarks/#section=test&runid=032630e0-3a86-4eac-ae2d-517e8b9586ac&hw=ph&test=json&l=zik0zj-1b&a=2) looks like a ~40% performance loss.

PR #6157, opened 2020-11-22
fix: make OCaml webmachine multiprocess

rbjorklin:
* revert haproxy http mode + keep alive as it incurred a 40% performance penalty
* spawned child processes is configurable with CORE_COUNT to still allow
  a fair Haproxy comparison
* this is an adaptation of @blandinw's solution in #6131

### DIFF
--- a/frameworks/OCaml/webmachine/haproxy.cfg
+++ b/frameworks/OCaml/webmachine/haproxy.cfg
@@ -7,10 +7,10 @@ global
     nbthread    2
 
 defaults
-    mode                    http
+    mode                    tcp
     log                     global
     option                  dontlognull
-    option http-keep-alive
+    option http-server-close
     option forwardfor       except 127.0.0.0/8
     option                  redispatch
     retries                 3
@@ -21,7 +21,7 @@ defaults
     timeout server          1m
     timeout http-keep-alive 10s
     timeout check           10s
-    maxconn                 3000
+    maxconn                 32768
 
 frontend main
     bind *:8080
--- a/frameworks/OCaml/webmachine/src/src/bin/tfb.ml
+++ b/frameworks/OCaml/webmachine/src/src/bin/tfb.ml
@@ -1,16 +1,17 @@
 (* https://github.com/ocsigen/lwt/blob/d7fabaa077389a0035254e66459a6a366c57576e/src/core/lwt_result.ml#L88-L91 *)
 (* >>= is Lwt.Infix equivalent to Lwt.bind:
-   https://ocsigen.org/lwt/5.2.0/api/Lwt#3_Callbacks *)
+   https://ocsigen.org/lwt/5.3.0/api/Lwt#3_Callbacks *)
 (* >|= is Lwt.Infix equivalent to Lwt.map:
-   https://ocsigen.org/lwt/5.2.0/api/Lwt#2_Convenience *)
+   https://ocsigen.org/lwt/5.3.0/api/Lwt#2_Convenience *)
 open Lwt.Infix
+open Lwt.Syntax [@@ocaml.warning "-33"]
 open Cohttp_lwt_unix
 
 module Wm = struct
   module Rd = Webmachine.Rd
 
   module UnixClock = struct
-    let now () = int_of_float (Unix.gettimeofday ())
+    let now () = 0
   end
 
   include Webmachine.Make (Cohttp_lwt_unix__Io) (UnixClock)
@@ -138,7 +139,46 @@ class queries =
       Wm.continue (`String (Lib.Db_j.string_of_queries json)) rd
   end
 
+let dump_lwt () =
+  let options =
+    [
+      ("fd_passing", `fd_passing);
+      ("fdatasync", `fdatasync);
+      ("get_affinity", `get_affinity);
+      ("get_cpu", `get_cpu);
+      ("get_credentials", `get_credentials);
+      ("libev", `libev);
+      ("madvise", `madvise);
+      ("mincore", `mincore);
+      ("recv_msg", `recv_msg);
+      ("send_msg", `send_msg);
+      ("set_affinity", `set_affinity);
+      ("wait4", `wait4);
+    ]
+  in
+  ignore @@ Lwt_io.eprintf "Lwt:\n";
+  List.iter
+    (fun (str, opt) ->
+      ignore @@ Lwt_io.eprintf "  %s = %b\n" str (Lwt_sys.have opt))
+    options
+
 let main () =
+  (* https://github.com/mirage/ocaml-cohttp/issues/328#issuecomment-222583580 *)
+  Lwt_io.set_default_buffer_size 0x10000;
+  let nproc =
+    match Sys.getenv "CORE_COUNT" with
+    | x -> int_of_string x
+    | exception Not_found ->
+        Unix.open_process_in "getconf _NPROCESSORS_ONLN"
+        |> input_line
+        |> int_of_string
+  in
+  ignore @@ Lwt_io.eprintf "Detected %d cores\n" nproc;
+  let ulimit_n =
+    Unix.open_process_in "ulimit -n" |> input_line |> int_of_string
+  in
+  ignore @@ Lwt_io.eprintf "Detected %d max open files\n" ulimit_n;
+  dump_lwt ();
   let port =
     match Sys.getenv "PORT" with
     | x -> int_of_string x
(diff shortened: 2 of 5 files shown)


rbjorklin:
@blandinw would you mind reviewing this?

I've been doing a lot of `ignore @@ Lwt_io.flush_all ();` and similar and it feels like code smell. Do you have any better ideas?

frameworks/OCaml/webmachine/haproxy.cfg
```diff
@@ -7,10 +7,10 @@ global
     nbthread    2
 
 defaults
-    mode                    http
+    mode                    tcp
```
blandinw, line 10: We should probably remove haproxy from the OCaml benchmarks, now that we know that sharing a socket yields better throughput

frameworks/OCaml/webmachine/src/src/bin/tfb.ml
```diff
@@ -138,7 +139,46 @@ class queries =
...
+      ("libev", `libev);
+      ("madvise", `madvise);
+      ("mincore", `mincore);
+      ("recv_msg", `recv_msg);
+      ("send_msg", `send_msg);
+      ("set_affinity", `set_affinity);
+      ("wait4", `wait4);
+    ]
+  in
+  ignore @@ Lwt_io.eprintf "Lwt:\n";
```
blandinw, line 159: We probably shouldn't ignore the return value (promise) of all the async Lwt functions.
Alternatively, you can use regular Printf when you're not serving requests yet.

frameworks/OCaml/webmachine/src/src/bin/tfb.ml
```diff
@@ -138,7 +139,46 @@ class queries =
...
+  List.iter
+    (fun (str, opt) ->
+      ignore @@ Lwt_io.eprintf "  %s = %b\n" str (Lwt_sys.have opt))
+    options
+
 let main () =
+  (* https://github.com/mirage/ocaml-cohttp/issues/328#issuecomment-222583580 *)
+  Lwt_io.set_default_buffer_size 0x10000;
+  let nproc =
+    match Sys.getenv "CORE_COUNT" with
```
blandinw, line 169: why not use all cores?

frameworks/OCaml/webmachine/src/src/bin/tfb.ml
```diff
@@ -138,7 +139,46 @@ class queries =
...
+    | x -> int_of_string x
+    | exception Not_found ->
+        Unix.open_process_in "getconf _NPROCESSORS_ONLN"
+        |> input_line
+        |> int_of_string
+  in
+  ignore @@ Lwt_io.eprintf "Detected %d cores\n" nproc;
```
blandinw, line 176: See above, don't discard returned promise, or use regular Printf

frameworks/OCaml/webmachine/src/src/bin/tfb.ml
```diff
@@ -138,7 +139,46 @@ class queries =
...
+  let ulimit_n =
+    Unix.open_process_in "ulimit -n" |> input_line |> int_of_string
+  in
+  ignore @@ Lwt_io.eprintf "Detected %d max open files\n" ulimit_n;
```
blandinw, line 180: See above, don't discard returned promise, or use regular Printf

frameworks/OCaml/webmachine/src/src/bin/tfb.ml
```diff
@@ -160,35 +200,36 @@ let main () =
...
-  Lwt_io.set_default_buffer_size 0x10000;
-  Lwt_main.run (main ())
+  Lwt_unix.setsockopt socket Unix.SO_REUSEADDR true;
+
+  Lwt_main.run
+  @@ ( Lwt_unix.bind socket sockaddr >|= fun () ->
+       Lwt_unix.listen socket (Lwt_unix.somaxconn () [@ocaml.warning "-3"]) );
+
+  for i = 1 to nproc do
+    ignore @@ Lwt_io.flush_all ();
```
blandinw, line 219: See above, don't discard returned promise, or use regular Printf

frameworks/OCaml/webmachine/src/src/bin/tfb.ml
```diff
@@ -160,35 +200,36 @@ let main () =
...
+    if Lwt_unix.fork () = 0 then (
+      (* child *)
+      Lib.Time.refresh_date ();
+      ignore
+      @@ Lwt_io.eprintf "Listening on %s:%d (child %d)\n"
```
blandinw, line 224: See above, don't discard returned promise, or use regular Printf

frameworks/OCaml/webmachine/src/src/bin/tfb.ml
```diff
@@ -160,35 +200,36 @@ let main () =
...
+           (Unix.string_of_inet_addr ipaddr)
+           port i;
+      let config = Server.make ~callback () in
+      ignore @@ Server.create ~mode:(`TCP (`Socket socket)) config;
```
blandinw, line 228: See above, don't discard returned promise, or use regular Printf

frameworks/OCaml/webmachine/src/src/bin/tfb.ml
```diff
@@ -160,35 +200,36 @@ let main () =
...
+      let forever, _ = Lwt.wait () in
+      Lwt_main.run forever;
+      exit 0 )
+  done;
+  Unix.pause ()
```
blandinw, line 233: You need to loop here, because any signal is gonna wake up the process

frameworks/OCaml/webmachine/webmachine-haproxy.dockerfile
```diff
@@ -8,6 +8,8 @@ ENV TZ  :/etc/localtime
 # https://linux.die.net/man/1/ocamlrun
 # https://blog.janestreet.com/memory-allocator-showdown/
 ENV OCAMLRUNPARAM a=2,o=240
+# This makes the program only spawn one child process to serve requests
+ENV CORE_COUNT 1
```
blandinw, line 12: why not use all cores?

blandinw (changes requested):
Hey! Yeah, we should not discard the returned promises, or use non-async functions when we can.
Also, we should probably get rid of haproxy, as it offers scalability but hinders thoughput, which is what matters in this benchmark.

frameworks/OCaml/webmachine/haproxy.cfg
rbjorklin: In the long term yes but I would like to have one run completed to see how much overhead it adds. Locally the plaintext test actually produces a higher RPS behind haproxy than with the multiprocess approach.

frameworks/OCaml/webmachine/src/src/bin/tfb.ml
rbjorklin: I agree I just don't know how to make it work without either ignoring the promise or using the regular Printf. My Lwt foo is pretty weak so any pointers on how to make this work with `Lwt_io` without `ignore` would be appreciated.

frameworks/OCaml/webmachine/src/src/bin/tfb.ml
rbjorklin: I wanted to compare the forking shared socket approach with a multiprocess multisocket approach while still using your tweaks to memoize the `Date` header.

frameworks/OCaml/webmachine/src/src/bin/tfb.ml
rbjorklin: It seems to work without a loop, what could happen?

With the loop all signals are effectively ignored, right?

frameworks/OCaml/webmachine/src/src/bin/tfb.ml
blandinw: Lwt exposes a monadic API (monadic just means that you chain things together using the `bind`, aka. `>>=` or `>|=` operators) to allow you to do just that.

Currently, you're starting up async computations in a fire and forget fashion instead of executing them sequentially. If you're familiar with Javascript promises, you can think of `>>=` and `>|=` as `.then()`. You can look at my code in `httpaf_unix.ml` for an example of using the monadic API. There's a lot of information encoded in the type signatures of the Lwt monadic functions, do not hesitate to scrutinize them, as well as the type signatures of `>>=`, `>|=` and `Lwt_main.run`.

frameworks/OCaml/webmachine/src/src/bin/tfb.ml
blandinw: Well, if you want to keep the (parent) process running (for example if it's your monitored Docker process), you should loop to avoid exiting.

If you don't care about the parent process (as currently, since haproxy is actually the monitored Docker process), you can just remove `Unix.pause` altogether. You'd need to update it the code if you get rid of haproxy.

My main point was that the current version (`Unix.pause` without loop) looks unintended.

PR #6159, opened 2020-11-23
Add EmbedIO

Kaliumhexacyanoferrat:
https://github.com/unosquare/embedio


frameworks/CSharp/embedio/benchmark_config.json
```diff
@@ -0,0 +1,22 @@
+{
+  "framework": "embedio",
+  "tests": [{
+    "default": {
+      "plaintext_url": "/plaintext",
+      "json_url": "/json",
+      "port": 8080,
+      "approach": "Realistic",
+      "classification": "Fullstack",
+      "database": "Postgres",
```
nbrady-techempower, line 10: You can change this to "None" or just remove it. This will spin up the database container, which doesn't look necessary here.

nbrady-techempower (changes requested):

frameworks/CSharp/embedio/benchmark_config.json
Kaliumhexacyanoferrat: That's right - too much copy and paste. Thanks, I have updated the configuration.

PR #6163, opened 2020-11-25
ocaml opium process forking

mudrz:
uses https://github.com/TechEmpower/FrameworkBenchmarks/pull/6131 to create an additional comparison variant that uses process forking


frameworks/OCaml/opium/src/bin/main_forks.ml
```diff
@@ -0,0 +1,69 @@
...
+  Lwt_unix.setsockopt socket Unix.SO_REUSEADDR true;
+
+  Lwt_main.run (
+    let+ () = Lwt_unix.bind socket listen_address in
+    Lwt_unix.listen socket (Lwt_unix.somaxconn () [@ocaml.warning "-3"])
+  );
+
+  let rec accept_loop socket handler instance =
+    let* (socket', sockaddr') = Lwt_unix.accept socket in
+    Lwt.async (fun () -> handler sockaddr' socket');
```
mudrz, line 21: @blandinw I notice that you have this function wrapped both in `async` and `catch`
I'm not that familiar with `Lwt`, is it necessary?

blandinw: Probably an oversight on my end!
If I recall correctly, `catch` allows customized error handling, while `async` defers to `async_exception_hook`.
If that's correct, they're redundant and only one should be used.

mudrz: okay, thanks for the clarification 👍 

assuming that the rest of the implementation is correct, we'll be able to compare httpaf with higher level libraries

blandinw: Yes! The implementation looks good at a high level. It may be worth running a local test to compare the haproxy solution and this one, I'd expect a 5x to 10x speed up

mudrz:
this PR should be ready to merge

PR #6168, opened 2020-11-29
Update actix-* yarte fortunes 

botika:
Update yarte implementation in `actix-*` fortunes, less `actix-diesel` (here yarte has no effect), for compare different actix implementations with the same behavior.

### DIFF
--- a/frameworks/Rust/actix/Cargo.toml
+++ b/frameworks/Rust/actix/Cargo.toml
@@ -33,8 +33,7 @@ actix-server = "1.0.4"
 actix-service = "1.0.6"
 snmalloc-rs = "0.2.6"
 askama = "0.8"
-markup = "0.4.1"
-yarte = "0.12"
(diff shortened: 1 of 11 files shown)


botika:
It seems to me that @robjtede is in charge of this? 

You can confirm this that I want to see the difference between the implementation in an actor and the direct implementation.

robjtede (commented):
not 100% familiar with yarte or what the implications of this change is but *looks* sensible

robjtede (approved):

botika:
Yarte now is using direct `tokio-bytes` implementation through the `buf-min` trait giving it a higher performance by implementing all the runtime at a low level with the use of raw pointers and simd instructions .

sebastienros:
+10% RPS, -10% mean latency on my tests, good job!

sebastienros:
image

PR #6198, opened 2020-12-11
Updated Benchmarks to use latest version of Giraffe

dustinmoris:
Updated The F# Giraffe project to the latest version and changed the application to run all three supported JSON serializers as part of the benchmarks.


dustinmoris:
@slang25 @benaadams @forki I saw you all contributing on behalf of F# here so would love to get your input as part of the review here.

forki:
Cool

PR #6200, opened 2020-12-11
Update finagle and finatra to version 20.12.0

mosesn:
We released a new version, let's bump the version in the benchmarks!


mosesn:
@nbrady-techempower PTAL!

mosesn:
Thanks @nbrady-techempower!

PR #6209, opened 2020-12-14
change concurrency level for ocaml/tiny_httpd

c-cube:
see if it affects latency

### DIFF
--- a/frameworks/OCaml/tiny_httpd/src/src/bin/tiny.ml
+++ b/frameworks/OCaml/tiny_httpd/src/src/bin/tiny.ml
@@ -1,7 +1,7 @@
 module S = Tiny_httpd
 
 let () =
-  let server = S.create ~addr:"0.0.0.0" () in
+  let server = S.create ~addr:"0.0.0.0" ~max_connections:256 () in
   let headers = [ ("Server", "tiny_httpd") ] in
   (* say hello *)
   S.add_route_handler ~meth:`GET server


rbjorklin:
Hi @c-cube!

Thanks for taking an interest and trying to improve my first draft implementation! I tried running a few tests locally by doing `./tfb --mode benchmark --test tiny_httpd --type json --concurrency-levels 512` and throughput decreases while latency increases with a higher `max_connections` value. The same may not hold in a real test with actual network latency though. I would suspect that a value in the range of 32-64 would work better though given that the tests are run on a local 1Gbit network.

rbjorklin:
I gave [this solution to keeping the up-to-date](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/OCaml/webmachine/src/src/lib/time.ml) a try with `tiny_httpd`. Unfortunately I'm seeing `Fatal error: exception Unix.Unix_error(Unix.EINTR, "accept", "")` which makes sense as I'm not handling the signal and it seems Lwt does it for you so no special handling was required for the other OCaml frameworks on here. This is a bit unfortunate as this tweak had a very substantial impact on performance.

The other OCaml frameworks on here [fork and bind multiple times](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/OCaml/webmachine/src/src/bin/tfb.ml#L218-L232) which I think isn't possible with `tiny_httpd` at the moment as it doesn't take socket when creating the server and [SO_REUSEADDR](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/OCaml/webmachine/src/src/bin/tfb.ml#L209-L212) would need to be set for that.

c-cube:
- for the fork: I think the server could take a `?setsockopts:(socket -> unit)` optional argument to be able to tweak what you need. Note that the following is already set:
  ```ocaml
    Unix.setsockopt sock Unix.SO_REUSEADDR true;
    Unix.setsockopt_optint sock Unix.SO_LINGER None;
  ```

- for the date, could the server store `Unix.gettimeofday()` (+ a string representation of it), and at every query check if the timestamp rounded to the second changed? The string version would only be updated if `floor (Unix.gettimeofday())` is not the same as the one stored. I expect the overhead of that to be pretty low.

rbjorklin:
Ah, I had missed that `SO_REUSEADDR` was already set and I was thrown off by getting `EADDRINUSE` last night but it sounds like I might have been doing something else wrong.

Regarding your suggestion around the handling of the `Date` header if I understood that correctly it should already [work the way you described it.](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/OCaml/tiny_httpd/src/src/lib/time.ml#L36-L43)

c-cube:
I think more in the way of:

```patch
- if _now -. !last_time >= 1. then (
+ if floor _now <> floor !last_time then (
```

nbrady-techempower:
@c-cube is this PR ready to be merged?

rbjorklin:
I'd say you can go ahead and merge this @nbrady-techempower, I might create another PR for the things discussed above.

PR #6215, opened 2020-12-16
http-jl performance updates

mcmcgrath13:
* Use JSON3 instead of JSON
* Run server with multiple processes and threads
* Better reproducibility with Project.toml
* Rename README to match standards


frameworks/Julia/Http.jl/Project.toml
```diff
@@ -0,0 +1,3 @@
+[deps]
+HTTP = "cd3eb016-35fb-5094-929b-558a96fad6f3"
+JSON3 = "0f8b85d8-7281-11e9-16c2-39a750bddbf1"
```
oxinabox, line 3: Maybe include a `[compat]` section here?
So can use `]up` to update manifest while respecting semver

oxinabox:
> Run server with multiple processes and threads

I think this is just multiple threads.
Not multiple processes.
Which is good.

mcmcgrath13:
> > Run server with multiple processes and threads
> 
> I think this is just multiple threads.
> Not multiple processes.
> Which is good.

The run.sh spins up 4 servers on different processes pointing at the same port.  Each server is only single process.

frameworks/Julia/Http.jl/run.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 for i in $(seq 0 $(($(nproc --all)-1))); do
-	julia server.jl
+	julia --threads auto server.jl &
```
cmcaine, line 4: Does this make a difference in this benchmark, btw?

mcmcgrath13: The running multiple in processes in the background makes a big difference.  The threads might make some difference, but it's less pronounced if at all (no harm in having them though?).

cmcaine: I would have expected that either would do (threads or multiprocess) if HTTP.jl is multithreaded.

As far as I can see in the code, tho, I think HTTP.jl doesn't start its own threads, but is expected to work if you call `listen` in multiple threads, so I wouldn't expect `--threads auto` to do much unless you change `server.jl` to use threads. OTOH, there's no point launching loads of threads if you're doing multiprocessing anyway (you probably want one thread or process  per core, not more).

frameworks/Julia/Http.jl/server.jl
```diff
@@ -1,7 +1,13 @@
...
+Pkg.activate(@__DIR__)
+
 using HTTP
-using JSON
+import JSON3
 using Dates
+using Sockets
 
+@info "starting listener"
 HTTP.listen("0.0.0.0", 8080, reuseaddr=true) do http
```
cmcaine, line 11: FWIW, I think this is all you need to make this multithreaded:

```jl
Threads.@threads for _ in 1:Threads.nthreads()
    HTTP.listen("0.0.0.0", 8080, reuseaddr=true) do http
        # blah
    end
end
```

frameworks/Julia/Http.jl/server.jl
```diff
@@ -1,7 +1,13 @@
+using Pkg
+
+Pkg.activate(@__DIR__)
+
 using HTTP
-using JSON
+import JSON3
 using Dates
+using Sockets
```
cmcaine, line 8:
```suggestion
```

This import is unused. Maybe you intended to use the `@ip_str` macro below?

frameworks/Julia/Http.jl/server.jl
mcmcgrath13: Thanks! I tried this both with the multiple processes and without and both were significantly slower than the current state - any theories as to why?

cmcaine: Sorry for the bad steer! I'm just skimming the code in HTTP.jl, but nothing is screaming out to me.

If you're doing multiprocess with SO_REUSEPORT (which you are if `reuseaddr=true` and you're on a \*nix), then the kernel can handle all the shared state (working out which stream is for which listener) for you, and maybe whatever Julia is doing in threaded mode is getting in the way of that?

Or it could be something to do with your OS. Maybe the process ulimits aren't high enough, so 8 threads sharing the same limits are losing out when compared to 8 processes with independent ulimits. No real idea. Just guesses.

PR #6220, opened 2020-12-18
feat: make tiny_httpd fork to increase throughput

rbjorklin:
Hi @c-cube!

I tried forking tiny_httpd on startup but I'm getting met with this:

```
❯ ./_build/default/src/bin/tiny.exe
Listening on 0.0.0.0:8080 (child 2)
Fatal error: exception Unix.Unix_error(Unix.EADDRINUSE, "bind", "")
Listening on 0.0.0.0:8080 (child 3)
Fatal error: exception Unix.Unix_error(Unix.EADDRINUSE, "bind", "")
Listening on 0.0.0.0:8080 (child 4)
Fatal error: exception Unix.Unix_error(Unix.EADDRINUSE, "bind", "")
Listening on 0.0.0.0:8080 (child 5)
Fatal error: exception Unix.Unix_error(Unix.EADDRINUSE, "bind", "")
Listening on 0.0.0.0:8080 (child 6)
Fatal error: exception Unix.Unix_error(Unix.EADDRINUSE, "bind", "")
Listening on 0.0.0.0:8080 (child 7)
Fatal error: exception Unix.Unix_error(Unix.EADDRINUSE, "bind", "")
Listening on 0.0.0.0:8080 (child 8)
Fatal error: exception Unix.Unix_error(Unix.EADDRINUSE, "bind", "")
```

I'm not sure here but could this be because each forked process [creates its own socket](https://github.com/c-cube/tiny_httpd/blob/21f474332071fa07ce5b89b2e460f3e0962ddbcc/src/Tiny_httpd.ml#L947-L950) within the `run` function?

### DIFF
--- a/frameworks/OCaml/tiny_httpd/src/src/bin/tiny.ml
+++ b/frameworks/OCaml/tiny_httpd/src/src/bin/tiny.ml
@@ -1,9 +1,8 @@
 module S = Tiny_httpd
 
 let () =
-  let server = S.create ~addr:"0.0.0.0" () in
+  let server = S.create ~addr:"0.0.0.0" ~port:8080 () in
   let headers = [ ("Server", "tiny_httpd") ] in
-  (* say hello *)
   S.add_route_handler ~meth:`GET server
     S.Route.(exact "plaintext" @/ string @/ return)
     (fun _ _req ->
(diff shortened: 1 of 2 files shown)


c-cube:
I'm afraid so. Seems like the only possibilities are:

- create the socket before forking (which I think would need to be supported directly in tiny-httpd, possibly with a `?fork:int` optional argument)
- use the linux (non posix) socket option `SO_REUSEPORT`, which is not provided by the OCaml stdlib (but it's present in [extunix](https://github.com/ygrek/extunix/blob/master/src/extUnix.mlpp#L854) apparently; we'd still need a way to provide a way to set specific socket options in tiny-httpd).

None of these can be done right now but PRs/discussions are welcome.

nbrady-techempower:
Going to close this for now @rbjorklin; Please re-open when you're ready and GitHub Actions should pick it up for testing.

PR #6225, opened 2020-12-18
Add SNUnit: Scala Native micro framework

lolgab:
SNUnit is a Scala Native HTTP library built on top of NGINX Unit
external process interface. It allows users to build binaries that
can be run by NGINX Unit to serve HTTP requests.


frameworks/Scala/snunit/benchmark_config.json
```diff
@@ -0,0 +1,26 @@
...
+        "port": 8080,
+        "approach": "Realistic",
+        "classification": "Micro",
+        "database": "None",
+        "framework": "SNUnit",
+        "language": "Scala",
+        "flavor": "Native",
+        "orm": "None",
+        "platform": "None",
+        "webserver": "NGINX Unit",
```
joanhey, line 17: In the benchmark already exist `Nginx unit`, but like platform. 
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/PHP/php/benchmark_config.json#L109-L110

Nginx is a web server (front-end server).
But `Nginx unit` is an application server with web server included, and for me it's better to be in platform, like `vert.x` and others. 

We need to try to have it in only one classification and with the same name `unit nginx`.

@nbrady-techempower What do you think ?

lolgab: Didn't notice that, thank you @joanhey.
So, being consistent with the PHP example like this would work?
```suggestion
        "platform": "Unit Nginx",
        "webserver": "None",
```
Then they can be changed later in a follow up PR.

lolgab:
Thank you very much for reviewing and merging!
It is really appreciated!

PR #6245, opened 2020-12-28
OfficeFloor 3.30.2

sagenschneider:
Bumping to OfficeFloor 3.30.2 and also using Parallel GC (as gets a little higher throughput).

Currently, running validation and will publish results here shortly.


sagenschneider:
Have successful runs on 2 separate machines locally.   Will post links when https://tfb-status.techempower.com/share is back available.

PR #6246, opened 2020-12-28
[Java] Add ActiveJ framework

eduard-vasinskyi:


nbrady-techempower:
Hi @eduard-vasinskyi,

Thanks for your first contribution! We've moved away from `.travis.yml`, so your changes there can be removed. When you rebase off the main branch, that file will be deleted. Also, side-by-side with the `benchmark_config.json` is a new `config.toml`. You should see several examples in all the other frameworks (sorry, the documentation isn't ready yet, but it's straightforward). If you could create a new PR based of the main branch, and also add a `config.toml` that matches the `benchmark_config.json` we can get this merged.

Thanks!

PR #6247, opened 2020-12-29
[aiohttp] Updates

Dreamsorcerer:
Minor updates to bring it up-to-date with best practices.

### DIFF
--- a/frameworks/Python/aiohttp/app/main.py
+++ b/frameworks/Python/aiohttp/app/main.py
@@ -42,7 +42,7 @@ def pg_dsn() -> str:
     ))
 
 
-async def startup(app: web.Application):
+async def db_ctx(app: web.Application):
     dsn = pg_dsn()
     # number of gunicorn workers = multiprocessing.cpu_count() as per gunicorn_conf.py
     # max_connections = 2000 as per toolset/setup/linux/databases/postgresql/postgresql.conf:64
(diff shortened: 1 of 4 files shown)


Dreamsorcerer:
@samuelcolvin In case you want to check it over.

frameworks/Python/aiohttp/app/main.py
```diff
@@ -80,14 +80,13 @@ def setup_routes(app):
...
 
 
-def create_app(loop):
-    app = web.Application(loop=loop)
+def create_app():
+    app = web.Application()
 
     jinja2_loader = jinja2.FileSystemLoader(str(THIS_DIR / 'templates'))
-    aiohttp_jinja2.setup(app, loader=jinja2_loader)
+    aiohttp_jinja2.setup(app, loader=jinja2_loader, enable_async=True)
```
samuelcolvin, line 87: I thought async Jinja was very slow. Is this really a performance improvement?

samuelcolvin (commented):
Otherwise looks good.

frameworks/Python/aiohttp/app/main.py
Dreamsorcerer: I'll revert it just in case, but I think the performance issue is only with using `await` etc. in the template.

My theory was that this might be able to improve parallelisation in future, if jinja is able to return control to the event loop while rendering or something (as this would now be doing `await render_template_async()` instead of `render_template()`). Though I don't think it has this benefit currently.

Note that async wasn't supported at all with aiohttp-jinja2<1.4.

PR #6248, opened 2020-12-29
[aiohttp] Use SQLAlchemy directly (rather than aiopg)

Dreamsorcerer:
One more update. This replaces aiopg with modern sqlalchemy. Because we are using asyncio this requires 1.4+.
This should obviously become the standard approach as 1.4 matures. It's also changed the API usage to actually use ORM, the previous code is only Core.

I've also marked with TODOs very minor improvements that can be made once another beta release is made available.

@samuelcolvin


frameworks/Python/aiohttp/app/main.py
```diff
@@ -52,16 +53,14 @@ async def db_ctx(app: web.Application):
     min_size = max(int(max_size / 2), 1)
     print(f'connection pool: min size: {min_size}, max size: {max_size}, orm: {CONNECTION_ORM}')
     if CONNECTION_ORM:
-        app['pg'] = await aiopg.sa.create_engine(dsn=dsn, minsize=min_size, maxsize=max_size, loop=app.loop)
```
Dreamsorcerer, line 55: I'm not sure how these min/max size values translate into sqlalchemy directly..

nbrady-techempower:
Closing and re-opening for new CI

nbrady-techempower:
Looks like a lot of errors here for the default test

Dreamsorcerer:
Thanks, I'll take a look. I tested the web app itself, but struggled to get the tests working locally.

Dreamsorcerer:
OK, that's the tests fixed.

I can try and play with the pool size thing, but I don't see any pool being used in (for example) the Flask code, so I'm not sure it's a fair comparison if we use a pool and other tests don't..

nbrady-techempower:
@Dreamsorcerer I'll merge this in but feel free to work on that. If it improves performance here, the goal would be for the other frameworks to also implement the same strategy and lift the group up as a whole.

samuelcolvin:
This looks great, thank you @Dreamsorcerer. Sorry not to reply earlier.

Do we have any provisional or indicative results for this? Or feeling about how it will perform relative to aiopg?

Dreamsorcerer:
Going by some results in the CI tests, it looks pretty similar. The one thing we've lost at the moment is the connection pool params. I've opened a PR to set the max size again, now I have some understanding of what I'm looking at. But, the min_size in aiopg appears to be an option to pre create a number of connections, so the aiopg code appears to be creating 100s of connections before the benchmark is actually run.

I'm not seeing any obvious way to pre create them in sqlalchemy, so we're probably losing a little bit of time as they get created during the benchmark.

PR #6252, opened 2021-01-01
test martian

yuyenews:


nbrady-techempower:
Thanks for your first contribution @yuyenews 

We won't be merging in any new frameworks for a few days, while we continue to work on getting the new toolset up and running. In the meantime, would you please add a README.md here similar to: https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Kotlin/hexagon/README.md with more information about your framework

PR #6255, opened 2021-01-03
Update hyperf benchmark

cexll:


nbrady-techempower:
Can you open a new PR so that this can be tested with our new GH actions workflow? thank you!

PR #6256, opened 2021-01-05
OfficeFloor fortune raw

sagenschneider:
Focus of officefloor-raw is raw performance of the OfficeFloor HTTP server.  Using raw fortune to remove mustache overheads.

Not sure if too late and can run individually, but would appreciate this update for officefloor-raw fortunes to be included in round 20


sagenschneider:
Again, would provide link to locally run success results.  But seems 2020 just wants to keep  breaking things :(   Hope things get sorted and looking forward to round 20 being published :)

nbrady-techempower:
@sagenschneider Would you mind creating a new PR so this can be tested with GH Actions.

sagenschneider:
Yep, no worries :+1:

PR #6257, opened 2021-01-05
Switch to GitHub Actions for CI

ajohnstonTE:
Switches the CI service from Travis to Github Actions. Functional demonstrations of this can be found below:
- This is a fake merge into master (the code was updated to pretend a different branch is master):
  - https://github.com/ajohnstonTE/FrameworkBenchmarks/actions/runs/462417858
- This is an example of using `[ci skip]` in a commit:
  - https://github.com/ajohnstonTE/FrameworkBenchmarks/actions/runs/462212487
  - This is an earlier build so it still has some debugging in there. I'll start a new build demonstrating this once the full build above completes (it'll pretty much block other builds from starting until it's done, at least that's my experience).
- This is an example of a single framework being tested in a PR based on files changed:
  - TODO. While this has already been verified, it was in a _much_ earlier build and I'd rather verify it against the latest.
  - Fwiw, here is that early build: https://github.com/ajohnstonTE/FrameworkBenchmarks/actions/runs/453719137

The remaining TODOs before undrafting the PR are genuinely just to update these links, and wait for the builds to complete. A lot of frameworks don't pass the tests. Maybe those are issues on my end, but so many of them pass that it _seems_ unlikely that that's the case. Some wiki updates will also be needed, not sure what the best approach is for that.


.github/workflows/build.yml
```diff
@@ -0,0 +1,195 @@
...
+          echo "RUN_TESTS<<EOF" >> $GITHUB_ENV
+          echo "$(grep -oP "github-actions-run-tests \K(.*)" <<< $DIFF || true)" >> $GITHUB_ENV
+          echo "EOF" >> $GITHUB_ENV
+      - name: Log status
+        run: |
+          if [ "$RUN_TESTS" ]; then echo "Proceeding to run tests."; else echo 'Skipping test verification.'; fi
+      - name: Build tfb dockerfile
+        if: ${{ env.RUN_TESTS }}
+        run: docker build -t techempower/tfb - < ./Dockerfile;
+      - name: Stop services
```
nbrady-techempower, line 184: This may not be needed anymore?

.github/workflows/build.yml
```diff
@@ -0,0 +1,195 @@
...
+    env:
+      TESTLANG: ${{ matrix.TESTLANG }}
+      TESTDIR: ${{ matrix.TESTDIR }}
+      COMMIT_MESSAGE: ${{ needs.setup.outputs.commit_message }}
+      BRANCH_NAME: ${{ needs.setup.outputs.branch_name }}
+      PR_NUMBER: ${{ github.event.pull_request.number }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 10
```
nbrady-techempower, line 154: Is there a reason for this specific number?

.github/workflows/build.yml
```diff
@@ -0,0 +1,195 @@
+name: build
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+jobs:
+  setup:
+    runs-on: ubuntu-18.04
```
nbrady-techempower, line 9: is 20.04 available?

msmith-techempower: We are still using 18.04 in all benchmark environments, so I think it makes sense to have parity here until we upgrade.

nbrady-techempower: We're upgrading.

ajohnstonTE:
Closing this PR. Will reopen as a new PR to see if GH actions triggers on the new PR.

.github/workflows/build.yml
ajohnstonTE: Not sure. This was originally from when the setup job was still included in this one, so it's actually from the link in that job's comment: https://github.community/t/accessing-commit-message-in-pull-request-event/17158/2. I already tried removing the fetch-depth but it doesn't work without that, so I left it in.

TL;DR: It makes it work, so it's there. I could have investigated further but it was working and I didn't want to keep delaying the PR.

ajohnstonTE: Though I'm realizing now that the website it links to dynamically updates the URL by scroll so I didn't notice the URL changed before I copied it. The comment should link to https://github.community/t/accessing-commit-message-in-pull-request-event/17158/8. I'll fix that in a follow-up PR.

.github/workflows/build.yml
ajohnstonTE: It is available, as a "preview". Not entirely sure what that means in the context of reliability.

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on

PR #6262, opened 2021-01-05
Github Actions follow-up

ajohnstonTE:
This is a follow-up PR to #6260, doing the following:

- Switches all matrix options to `TESTLANG`
- Renames the `test` job to `verify`
- Updates an incorrect link in a comment


msmith-techempower (approved):
🥳

PR #6270, opened 2021-01-07
edit edap-http config

luyongshou:


nbrady-techempower:
We won't be merging any PR's for a couple of days, until the new toolset is ready. After that, I'll ask you to re-open this PR using the new config file. Very little will change from an implementation stand point, just the config will be different. I've merged config.toml's into master already, if you'd like to take a look at what the new config should look like.

PR #6273, opened 2021-01-07
Add Watson Webserver framework (C#)

Kaliumhexacyanoferrat:


nbrady-techempower:
Hi @Kaliumhexacyanoferrat --

We won't be merging any PR's for a couple of days, until the new toolset is ready. After that, I'll ask you to re-open this PR using the new config file. Very little will change from an implementation stand point, just the config will be different. I've merged `config.toml`'s into master already, if you'd like to take a look at what the new config should look like.

Kaliumhexacyanoferrat:
@nbrady-techempower : There is already a `config.toml` in this PR (but also a `benchmark_config.json`). Nevertheless I can look over the changes on master once again as soon as the config changes are stable.

nbrady-techempower:
I missed it the first time. Thanks!

PR #6274, opened 2021-01-08
Fix CI issue

ajohnstonTE:
The CI is silently failing to run tests for fork:master -> base:master PRs.

Demonstration of difference: Compare the "Get all changes vs master" step's log for the following workflows.

no changes, pre-fix: https://github.com/TechEmpower/FrameworkBenchmarks/runs/1666054554?check_suite_focus=true
with fix: https://github.com/TechEmpower/FrameworkBenchmarks/runs/1666081021?check_suite_focus=true

Without the fix, it fails to see the master branch. With the fix it proceeds normally.

### DIFF
--- a/toolset/github_actions/github_actions_diff.py
+++ b/toolset/github_actions/github_actions_diff.py
@@ -63,13 +63,10 @@ is_master = os.getenv("BRANCH_NAME") == "master"
 
 if is_PR:
     curr_branch = "HEAD"
-elif not is_master:
-    curr_branch = os.getenv("GITHUB_SHA")
-
-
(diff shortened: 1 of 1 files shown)


ajohnstonTE:
A short-ish explanation of the issue. The python script previously was would see that a branch was already on master and skip grabbing master, but GH actions in PRs checks out the code in a detached HEAD state with no branch, so you always need to fetch master in a PR. Without the fetch of master, the diff step was silently failing resulting in an empty string diff, leaving the rest of the CI steps to assume there were no changes and to move on.

PR #6284, opened 2021-01-08
Master diff-only builds

ajohnstonTE:
Updates the CI to not force a full rebuild on push to a branch, and instead only build the changes. PR builds/diffs will still work as they did before.

Also removes the branch filter for build triggers, so that any git flow still can have this same CI flow.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,9 +1,5 @@
 name: build
-on:
-  push:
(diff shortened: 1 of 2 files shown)


ajohnstonTE:
Closing this in favor of a PR off of my master branch (same changes, it's to verify/demonstrate behavior)

PR #6289, opened 2021-01-12
[aiohttp] Eliminate some overhead in jinja environment

Dreamsorcerer:
I noticed that uvicorn is practically cheating by pre-loading a jinja template, without using the rest of the jinja environment like a regular application would.

Rather than copying that cheat, I had a look at what unneeded overheads existed in the jinja environment. The cache_size causes LRUCache to be replaced with a simple dict. auto_reload ensures we only check the disk on the first load.

I'm not sure exactly how the benchmarks are run, but it would probably be a good idea if the benchmarks did a dummy run first, which allows connections to be created and caches to be populated. Then do the real benchmark. Or maybe a best of 3 runs or similar (as long as the app is not reloaded between runs). This would ensure the comparisons are fairer without needing every framework to implement logic to ensure things are prepopulated before the benchmark runs.

### DIFF
--- a/frameworks/Python/aiohttp/app/main.py
+++ b/frameworks/Python/aiohttp/app/main.py
@@ -1,6 +1,5 @@
 import os
 import multiprocessing
-from pathlib import Path
 
 import asyncpg
 from aiohttp import web
@@ -23,8 +22,6 @@ from .views import (
 )
 
 CONNECTION_ORM = os.getenv('CONNECTION', 'ORM').upper() == 'ORM'
-
-THIS_DIR = Path(__file__).parent
 
 
 def pg_dsn() -> str:
--- a/frameworks/Python/aiohttp/app/main.py
+++ b/frameworks/Python/aiohttp/app/main.py
@@ -2,9 +2,7 @@ import os
 import multiprocessing
 from pathlib import Path
 
-import aiohttp_jinja2
 import asyncpg
-import jinja2
 from aiohttp import web
 from sqlalchemy.engine.url import URL
 from sqlalchemy.ext.asyncio import AsyncSession, create_async_engine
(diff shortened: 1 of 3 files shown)


Dreamsorcerer:
Hmm, test failure appears that by random chance number of DB queries was 1 less than allowed in the updates.

This looks like random chance, as it picks a series of random ids. Any duplicate ids will result in 1 less DB update (as everything is prepared and deduplicated by sqlalchemy before commit).

Dreamsorcerer:
I see there are others using the same cheat as well. Would it be better to do the same thing, even if it doesn't really match how an application would actually be written?

nbrady-techempower:
@Dreamsorcerer Yes, that would be ideal. We do expect that if the same ID be chosen twice at random, two updates would happen.

Dreamsorcerer:
I don't think any of the Python examples do 2 updates for the same ID, they are all based on the same logic using SQLAlchemy. I assume this is why the tests match an approximate number in the first place (it appears to allow 112 updates to be missing, and this time we had 113).

They all use code something like:
```
with sess.begin():
  for id_ in ids:
     world = sess.get(World, id_)
     world.randomnumber = rand_new
```

Because the commit doesn't happen until the loop is finished, I'm sure sqlalchemy is smart enough to deduplicate the unneeded update before it sends the command to the server. It probably also caches the select in there too.

nbrady-techempower:
Some implementations make sure that each random ID is unique, another way to solve. Up to you.

Dreamsorcerer:
I mean, if there's an example, we can do this for all of them, but the logic appears the same (allowing duplicates) in all the ones I looked at (uvicorn, flask, aiohttp, fastapi).

nbrady-techempower:
Sorry, I'm not talking about Python specifically. But I imagine that each of the ones you looked at with the same logic will also have this problem, and may encounter it on official round runs.

Dreamsorcerer:
Would it be reasonable to generate the random ids in a global variable that is referenced in the view function?

This means that multiple calls to the same endpoint would update the same objects, but would make it easier to ensure no random-related timings affect the benchmark making it fairer. That should be easy for me to implement quickly.

Dreamsorcerer:
Or, actually, maybe it simple solution would be:
```
all_ids = list(range(10000))
random.shuffle(all_ids)
ids = all_ids[:num_ids]
```

I'm guessing random.shuffle should be pretty consistent in runs.

Dreamsorcerer:
I had a look through a couple of other languages, and didn't see any others to base the logic on, they all seem to do the same thing. Except this one, which appears to not be random at all, just iterates from 0: https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/JavaScript/nodejs/handlers/mongoose.js#L82

Dreamsorcerer:
Switched to using the same cheat as others.

I'll propose a fix for randint separately.

PR #6302, opened 2021-01-18
Fix random IDs

Dreamsorcerer:
Avoid duplicate IDs in Python tests.

Is it permittable to randomise it once at the top of the file, and avoid doing any random functions within the handlers?

### DIFF
--- a/frameworks/Python/aiohttp/app/views.py
+++ b/frameworks/Python/aiohttp/app/views.py
@@ -1,6 +1,6 @@
 from functools import partial
 from operator import attrgetter, itemgetter
-from random import randint
+from random import randint, shuffle
 
 from aiohttp_jinja2 import template
 from aiohttp.web import Response, json_response
@@ -11,6 +11,8 @@ from sqlalchemy import select
 from .models import sa_fortunes, sa_worlds, Fortune, World
 
 json_response = partial(json_response, dumps=ujson.dumps)
+world_ids = list(range(1, 10000))
+shuffle(world_ids)
 
 
 def get_num_queries(request):
@@ -61,7 +63,7 @@ async def multiple_database_queries_orm(request):
     """
     num_queries = get_num_queries(request)
 
-    ids = [randint(1, 10000) for _ in range(num_queries)]
+    ids = world_ids[:num_queries]
     ids.sort()
 
     result = []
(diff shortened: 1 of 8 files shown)


Dreamsorcerer:
Appears this fails to fix the issue. Looks like all the python examples consistently get to:
`Rows updated: 10128/10240`

But, sometimes it comes out at 10126, which then fails the test. I suspect this is because of the random value being set. When we update the values, we pick a random number between 0-10000 and update the object. If that number is the same as the previously set number, then I'd expect sqlalchemy to drop the update statement again.
With over 10000 updates happening, it's normal to have anywhere from 0-2 duplicates like this.

Dreamsorcerer:
@nbrady-techempower ^^^

Dreamsorcerer:
Running the maths, there should be only a 36% chance that 0 collisions happen and the test succeeds.

nbrady-techempower:
> Is it permittable to randomise it once at the top of the file, and avoid doing any random functions within the handlers?

There needs to be a mechanism in place that replenishes this array with a new set of randomly ordered numbers in the event is it exhausted. 

> Appears this fails to fix the issue. 

It may be worthwhile to wait a few days. We're still in the process of switching over to the new toolset and this might give you more insight into what's happening.

Dreamsorcerer:
> > Is it permittable to randomise it once at the top of the file, and avoid doing any random functions within the handlers?
> 
> There needs to be a mechanism in place that replenishes this array with a new set of randomly ordered numbers in the event is it exhausted.

The current code just reuses the beginning of the array every time. So, multiple requests would get the same set of IDs. So, my question is more does it have to be shuffled on every request to give a different set of ids?

> It may be worthwhile to wait a few days. We're still in the process of switching over to the new toolset and this might give you more insight into what's happening.

Given the frequency, my theory that it is caused by the random number matching the existing value in the DB seems pretty likely. If a single update is missing, the test fails.

nbrady-techempower:
> The current code just reuses the beginning of the array every time. So, multiple requests would get the same set of IDs. So, my question is more does it have to be shuffled on every request to give a different set of ids?

Yes, it needs to be randomized on each request.

nbrady-techempower:
Hey @Dreamsorcerer ! It's been a bit since we talked about this. Any time to take a look at it or should I close it up for now?

Dreamsorcerer:
Well, I don't know how to fix the issue, as I mentioned before the tests should probably fail with about a 64% chance. I'm not sure if there's maybe some option to sqlalchemy or something that could disable the optimisations, otherwise I don't see a fix.

Feel free to close the PR given that it doesn't resolve the issue.

PR #6309, opened 2021-01-20
Update dependencies of warp-rust

xfix:
warp 0.3.0 was released which is a good excuse to update all dependencies.


xfix:
May as well have "Update yarte in warp-rust to 0.14.1" commit from #6310 so that both pull requests could merge in cleanly

PR #6310, opened 2021-01-20
Enable LTO for warp-rust

xfix:
This makes compilation slower, but this isn't a compilation time benchmark, so that's fine.


xfix:
Seems like warp-rust currently doesn't build, so this will update yarte as well.

joanhey:
Aside:
The github-actions need to delete the label `PR: Please Update` when another commit pass the tests.
@ajohnstonTE

ajohnstonTE:
@joanhey This was actually discussed on our end at the time, and we decided to have it require manual review to remove the label. @nbrady-techempower Anything to add to that?

nbrady-techempower:
Yeah, I prefer to leave the label. We could do a separate label, but if I add the label manually for a different reason, I wouldn't want the action to delete it on the next passing test. The label at least gives me, the person who is normally doing all the review, the callout to take a closer look.

PR #6323, opened 2021-01-26
ntex: upgrade ntex

fafhrd91:
* update ntex
* drop sailfish bench


pickfire:
```rust
        let mut updates = Vec::new();
        for num in 1..=500u16 {
            let mut pl: u16 = 1;
            let mut q = String::new();
            q.push_str("UPDATE world SET randomnumber = CASE id ");
            for _ in 1..=num {
                let _ = write!(&mut q, "when ${} then ${} ", pl, pl + 1);
                pl += 2;
            }
            q.push_str("ELSE randomnumber END WHERE id IN (");
            for _ in 1..=num {
                let _ = write!(&mut q, "${},", pl);
                pl += 1;
            }
            q.pop();
```

Should we specify the capacity there to prevent multiple allocations?

fafhrd91:
> ```rust
> ```rust
>         let mut updates = Vec::new();
>         for num in 1..=500u16 {
>             let mut pl: u16 = 1;
>             let mut q = String::new();
>             q.push_str("UPDATE world SET randomnumber = CASE id ");
>             for _ in 1..=num {
>                 let _ = write!(&mut q, "when ${} then ${} ", pl, pl + 1);
>                 pl += 2;
>             }
>             q.push_str("ELSE randomnumber END WHERE id IN (");
>             for _ in 1..=num {
>                 let _ = write!(&mut q, "${},", pl);
>                 pl += 1;
>             }
>             q.pop();
> ```
> 
> 
> Should we specify the capacity there to prevent multiple allocations?
> ```

doesnt matter, this code get executed at startup only

PR #6325, opened 2021-01-28
WIP: Initial addition of cfml-lucee

bdw429s:
This is a work in progress pull for the CFML language running on Lucee Server.


bdw429s:
@nbrady-techempower Hi, can I get a review on this pull please?  I posted on the mailing list asking for feedback when I created it, but no one has chimed in.  I have more frameworks I'd like to add on top of this, but I'm waiting to see if this is on the right track first.  Thanks.

frameworks/CFML/cfml-lucee/cfml-lucee.dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM ortussolutions/commandbox:latest
```
nbrady-techempower, line 1: Can you pin this down to a specific version, like 3.0.1

frameworks/CFML/cfml-lucee/cfml-lucee.dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+COPY ./src/server.json /app/server.json
+COPY ./src/.cfconfig.json /app/.cfconfig.json
+
+RUN box install commandbox-cfconfig --verbose --force --debug
+
+RUN ${BUILD_DIR}/util/warmup-server.sh
+
+RUN export FINALIZE_STARTUP=true;$BUILD_DIR/run.sh;unset FINALIZE_STARTUP
+
```
nbrady-techempower, line 11: Though it's not needed yet, can you `EXPOSE 8080` here. When we move to the new toolset, this will be necessary.

nbrady-techempower (changes requested):
Just a couple of notes in the dockerfile, otherwise looks great! Thanks for the contributions!

frameworks/CFML/cfml-lucee/cfml-lucee.dockerfile
bdw429s: Pinned to 3.2.1

frameworks/CFML/cfml-lucee/cfml-lucee.dockerfile
bdw429s: Exposed

bdw429s:
@nbrady-techempower Thanks for the feedback.  I have implemented the changes you requested.

I'd also like some quick feedback on whether the directory structure is correct and if what I'm planning in the future (as explained in the Google group) is correct.  Lucee is an implementation of the CFML language.  Adobe ColdFusion is another so my next planned step is `/CFML/cfml-adobe` so both vendors are represented.  Then, I have some MVC frameworks written in CFML to add (which can be run on either CF engine), so I should end up with

```
/CFML/cfml-lucee
/CFML/cfml-adobe
/CFML/cfml-lucee-coldbox
/CFML/cfml-adobe-coldbox
/CFML/cfml-lucee-fw1
/CFML/cfml-adobe-fw1
etc...
```
Does that seem reasonable?

nbrady-techempower:
Hmm, well I don't think we need 6 different directories. Is there a better way to group these? I'd be fine with lucee and adobe directories and if there's only going to be one implementation of each framework, doing those there. So a lucee directory might have the default test that you have here and then permutations for each framework. 

Another option is single lucee and adobe directories, and then framework directories that have permutations for each. I'm fine with either option.

nbrady-techempower (approved):

bdw429s:
@nbrady-techempower Thanks for the feedback on the directory structure.  This is what I was looking to iron out before the pull when I posted to the mailing list a week ago :)  

I'm still not entirely sure what you're suggesting as far as folder structure.  I looked at several other languages such as Clojure but I can't find an example of another language that has more than one vendor implementation (such as Ruby and jRuby).  It does seem now like it's superfluous for me to include `cfml` again in the sub directories.  I did that based on how they show up in the benchmarks (ex: `php-ngx-mysql` ) but upon further inspection, I guess those hyphenated strings are composed on the fly by the benchmarking tool, they aren't represented in the directly in the folder structure.

I guess if the goal is to reduce duplicate code- the test implementations SHOULD be pretty much the same for both adobe and lucee code-wise.  Just another docker file.  Does it make sense to share the same folder of source code and just use the `benchmark_config.json` file to declare two separate versions using two different docker files that run the same code, but on different engines?  

The source code for the base CFML tests should be the same across lucee/adobe engines and the source code for the ColdBox framework should be the same across lucee/adobe engine, etc.  Maybe something like this?

```
/CFML/CFML
   - src/ <-- shared source code for both CFML engines
   - benchmark_config.json  <--declare base tests in both CFML engines
   - cfml-lucee.dockerfile
   - cfml-adobe.dockerfile
/CFML/coldbox
   - src/ <-- shared coldbox source code for both CFML engines
   - benchmark_config.json  <--declare coldbox tests in both CFML engines
   - cfml-lucee-coldbox.dockerfile
   - cfml-adobe-coldbox.dockerfile
/CFML/fw1
   - src/ <-- shared fw1 source code for both CFML engines
   - benchmark_config.json  <--declare fw1 tests in both CFML engines
   - cfml-lucee-fw1.dockerfile
   - cfml-adobe-fw1.dockerfile
```

nbrady-techempower:
> This is what I was looking to iron out before the pull when I posted to the mailing list a week ago :)

As you can probably tell, we don't keep up much with that. We're going to be moving exclusively to the discussions tab in this repo shortly. Will send notice on that soon.

>  Maybe something like this?

The structure you're proposing here is great.

bdw429s:
Looking forward to a new discussion forum.  I actually have a handful of unanswered questions sitting in the Google group right now.

I'll re-arrange the folder structure and send it with the Adobe engine.  I do have a question-- is it a requirement to have a `default` test configuration in the JSON file or is it ok just to have two JSON entries, one for each engine?  i.e.

```js
{
  "framework": "cfml",
  "tests": [{
    "lucee": {
      ...
    },
    "adobe": {
      ...
    }
  }]
}
```

nbrady-techempower:
> Looking forward to a new discussion forum. I actually have a handful of unanswered questions sitting in the Google group right now.

Feel free to move them here and I'll do what I can to get them addressed https://github.com/TechEmpower/FrameworkBenchmarks/discussions/categories/q-a

> is it a requirement to have a default test configuration

Yes. The key of the first test needs to be "default" but you can have the display name be "lucee". There are changes to this in the new toolset that's about to drop, but nothing you need to worry about atm.

bdw429s:
Thanks again for all the help :+1:   I mentioned this in the Google group, but I looked at getting CFML into this stack back in 2013 and gave up after 8+ hours of fiddling with the test setup.  When I found the project again recently I was pleased to see it based on Docker.  I had the base tests running in about 20 minutes and I had a proof of concept for CFML running in an hour.  Great work on the improvements to this over the years :)

nbrady-techempower:
That's so great to hear! No problem at all :)

PR #6342, opened 2021-02-05
Add unic framework

rajkumardusad:
Add [unic framework](https://github.com/unicframework/unic) benchmark


joanhey (changes requested):
Please install the framework with composer or git.
Only the files used for the benchmark would be in this repo.

rajkumardusad:
> Please install the framework with composer or git.
> Only the files used for the benchmark would be in this repo.

I have installed this framework with composer

joanhey:
You need to install it in the dockerfile.

Example:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/PHP/cakephp/cakephp.dockerfile#L21

rajkumardusad:
> You need to install it in the dockerfile.
> 
> Example:
> https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/PHP/cakephp/cakephp.dockerfile#L21

I have done it this way.

joanhey:
The core files don't need to be in this repo.

rajkumardusad:
> The core files don't need to be in this repo.

But when we use `composer create-project unicframework/unic myproject` it will create a project structure that includes the core files.
I mean all core files comes with the framework itself, it cannot be installed explicitly.

nbrady-techempower:
> But when we use composer create-project unicframework/unic myproject it will create a project structure that includes the core files.

This happens inside the dockerfile. None of those files need to be included in this PR. Can you remove the unnecessary files please?

PR #6360, opened 2021-02-12
Enable HTTP/1.1 pipelining for warp-rust

xfix:
This is using an undocumented method that has a following comment.

```rust
// Generally shouldn't be used, as it can slow down non-pipelined responses.
//
// It's only real use is to make silly pipeline benchmarks look better.
```

I think it's fine to enable HTTP/1.1 pipelining for warp-rust because Framework Benchmarks wiki says the following.

> Server support for HTTP/1.1 pipelining is assumed. Servers that do not support pipelining may be included but should downgrade gracefully. If you are unsure about your server's behavior with pipelining, test with the wrk load generation tool used in our tests.


joanhey:
Pipelining is only permited in the plaintext test.

nbrady-techempower:
Pipelining is allowed for plaintext and pipelining is allowed for server <-> database per no. 7 of the general requirements. 

>  However, it is permissible to pipeline traffic between the application and database as a network-level optimization.

PR #6367, opened 2021-02-14
[c#/beetlex] update

beetlex-io:
[c#/beetlex] update


frameworks/CSharp/beetlex/PlatformBenchmarks/DBRaw.cs
```diff
@@ -161,7 +161,7 @@ private async Task<World[]> LoadMultipleRows(int count, DbConnection db)
 
         public async Task<List<Fortune>> LoadFortunesRows()
         {
-            var result = new List<Fortune>();
+            var result = new List<Fortune>(20);
```
nbrady-techempower, line 164: Please remove this. The data structure must be dynamically sized according to the test rules.

nbrady-techempower (changes requested):

frameworks/CSharp/beetlex/PlatformBenchmarks/DBRaw.cs
beetlex-io: Specify the initial size of the list to reduce the memory copy problem caused by insufficient capacity. Other frameworks initialize the list in this way。
## java\wizzardo
 ``` java
 public void fortunes() {
        response.async();
        dbService.getClient().preparedQuery("SELECT * FROM fortune", res -> {
            try {
                if (res.succeeded()) {
                    PgRowSet result = res.result();
                    ArrayList<Fortune> fortunes = new ArrayList<>(result.size() + 1);
                    for (Row row : result) {
                        fortunes.add(new Fortune(row.getInteger(0), row.getString(1)));
                    }

                    fortunes.add(new Fortune(0, "Additional fortune added at request time."));
                    fortunes.sort(Comparator.comparing(fortune -> fortune.message));

                    model().append("fortunes", fortunes);
                    response.setBody(renderView("fortunes").renderReadableData());
                    response.appendHeader(Header.KV_CONTENT_TYPE_HTML_UTF8);
                } else {
                    res.cause().printStackTrace();
                    response.status(Status._500).body(res.cause().getMessage());
                }

                commitAsyncResponse();
            } catch (Exception e) {
                e.printStackTrace();
            }
        });
    }
```
## aspnetcore\PlatformBenchmarks
``` csharp
        public async Task<List<Fortune>> LoadFortunesRows()
        {
            var result = new List<Fortune>(20);

            using (var db = new NpgsqlConnection(_connectionString))
            {
                await db.OpenAsync();

                using (var cmd = new NpgsqlCommand("SELECT id, message FROM fortune", db))
                using (var rdr = await cmd.ExecuteReaderAsync())
                {
                    while (await rdr.ReadAsync())
                    {
                        result.Add(new Fortune
                        (
                            id:rdr.GetInt32(0),
                            message: rdr.GetString(1)
                        ));
                    }
                }
            }

            result.Add(new Fortune(id: 0, message: "Additional fortune added at request time." ));
            result.Sort();

            return result;
        }

```

nbrady-techempower: Those frameworks need to be fixed. Thank you for pointing it out.

frameworks/CSharp/beetlex/PlatformBenchmarks/DBRaw.cs
```diff
@@ -161,7 +162,7 @@ private async Task<World[]> LoadMultipleRows(int count, DbConnection db)
 
         public async Task<List<Fortune>> LoadFortunesRows()
         {
-            var result = new List<Fortune>(20);
+            var result = new List<Fortune>(ListDefaultSize);
```
nbrady-techempower, line 165: Again, this is not acceptable. Please remove this. You can fix the other two you pointed out in this PR if you'd like, or I can get to them tomorrow. It was missed.

beetlex-io: Can I use Array buffer, dynamically-sized returned ArraySegment?

PR #6372, opened 2021-02-16
Avoid deadlocks in CFML updates test

bdw429s:
The cfml tests for Lucee and Adobe appear to be getting deadlocks in the updates test.  I'm not sure why, but I refactored them to loop over and make separate updates to individual rows instead of a single update statement that hits all rows at once.  I'm afraid this new way may perform more slowly, but I'm not sure how to avoid the deadlocks with the original method.


bdw429s:
Thanks @nbrady-techempower !

PR #6373, opened 2021-02-16
These files were meant to be removed in pull #6337

bdw429s:
Pull https://github.com/TechEmpower/FrameworkBenchmarks/pull/6337 which renamed the cfml-lucee framework to `cfml` and added a `cfml-adobe` variant was meant to remove these files, but I just realized somehow they persisted.  This should hopefully also clear up my question here:

https://github.com/TechEmpower/FrameworkBenchmarks/discussions/6371


bdw429s:
Thanks @nbrady-techempower !

Issue #6184, opened 2020-12-01
PHP 8 update [info]


joanhey:
## PHP 8
PHP 8 was released on November 26.
Updated all the frameworks using `php-fpm` that don't have problems.

## Need to fix
This list will be updated as we found problems or solutions.
* **:ok: duckphp**  @dvaknheo  ~Problem using curly braces, send a [PR](https://github.com/dvaknheo/duckphp/pull/6) to duckphp repo to fix it~ #6311
* **:ok: ubiquity** @jcheron  ~`phpmv/ubiquity` dev-master requires php ^7.4  in composer~ #6312
* **:ok: symfony**  ~Fatal error: Out of memory  in composer install
      symfony-swoole still with 7.4~ #6386
      Doctrine DBAL fatal error https://tfb-status.techempower.com/unzip/results.2021-05-12-03-16-44-950.zip/results/20210507120805/symfony/run/symfony.log 
      https://github.com/doctrine/dbal/pull/3770  https://github.com/symfony/symfony/pull/36895
* **zendframework**  Need to update to laminas framework to use PHP 8
* **:ok: cakephp** ~`aura/intl` 3.0.0 requires php ^5.6|^7.0~ #6281
* **:ok: hamlet** @vasily-kartashov `hamlet-framework/http` dev-master requires php ^7
              `hamlet-framework/db-mysql` dev-master requires php ^7
              `cache/array-adapter` 1.1.x-dev requires php ^5.6 || ^7.0

## Swoole, php-ngx, workerman, phalcon, ...
* :ok:  **swoole** is ready, ~waiting for the php:8.0 image docker update
     working with mysql, but `swoole/ext-postresql` fail at compile time https://github.com/swoole/ext-postgresql/pull/49~ #6296
* :ok: **php-ngx** @rryqszq4 ~updated in dev branch, but still fail one permutation~ #6282 
* :ok: **workerman** use event extension and it's ready
* **phalcon** [is not ready](https://github.com/phalcon/cphalcon/issues/15186) for PHP 8
* **:ok: amp** using 'libevent',  `uv` extension fail at compile time https://github.com/bwoebi/php-uv/commit/afe868812897660841148991d09e61c5731341cf  https://github.com/bwoebi/php-uv/commit/444acfb9636094c2c96d1e7b43332027f464efc5
     :ok: now using `event` in the bench
*  :ok: **roadrunner**  ~not ready~
[...]

PR #6386, opened 2021-02-19
[symfony] Update to PHP 8

joanhey:
Symfony swoole bundle update to 0.9 and php8.

#6184


joanhey:
@k911 Please update the `beberlei/assert 3.0` to a new version that work with php8.
https://github.com/k911/swoole-bundle/blob/v0.9.0/composer.json#L24

PR #6387, opened 2021-02-19
cfml performance tweaks & add missing test

bdw429s:
There are a few small tweaks in this pull from my local rounds of performance tuning the CFML and ColdBox frameworks.

* Added `Application.cfc` stub to CFML framework to avoid necessary disk hits in Lucee
* Added config to disable all file logging in Lucee (as well as docker command to import it into the web context)
* Added default password for Lucee and Adobe admin for convenience when testing
* Added missing caching test for CFML and ColdBox

There is additional performance gains to be had, but it will have to come from the CF engine vendors when they fix the tickets I've entered for them :)
https://luceeserver.atlassian.net/browse/LDEV-3288
https://luceeserver.atlassian.net/browse/LDEV-3287
https://tracker.adobe.com/#/view/CF-4211113
https://luceeserver.atlassian.net/browse/LDEV-3290


bdw429s:
Thanks @nbrady-techempower !

PR #6404, opened 2021-02-25
[EmbedIO] Use shortcut methods

rdeago:
This PR simplifies EmbedIO testing code by using shortcut methods.

I had to create a static variable to hold a custom `Encoding`; otherwise, plain-text responses would start with a UTF-8 byte order mark. This will be no longer necessary with the upcoming EmbedIO version 3.5.0.


rdeago:
The failed check seems to be due to some bug in EmbedIO. Version 3.5.0 will allow for buffered plain-text / JSON responses, thereby working around it, but the lack of a `Transfer-Encoding` header in responses deserves some investigation. I'll get back to this PR as soon as the problem is solved in EmbedIO.

nbrady-techempower:
If you have a chance to take a look at this please feel free to re-open!

PR #6411, opened 2021-02-27
Use SO_REUSEPORT and listen per child

frejsoya:
Each process has its own backlog. No test on performance changes yet... (need a linux vm)


nbrady-techempower:
Looks like it's been a bit since this was worked on. Feel free to re-open when it's ready!

frejsoya:
I was also not able to label this with ocaml.

PR #6431, opened 2021-03-10
Update 2.3.0

redkale:


frameworks/Java/redkale/src/main/java/org/redkalex/benchmark/Service.java
```diff
@@ -34,8 +39,8 @@
     }
 
     @RestMapping(name = "json")
-    public Message getHelloMessage() {
-        return new Message("Hello, World!");
+    public Object getHelloMessage() {
+        return JSON_TEST ? msgjson : new Message("Hello, World!");
```
nbrady-techempower, line 43: Please revert this. This function should always instantiate a new object. Code for testing something different shouldn't be committed here.

redkale: This is just a debugging code. The reason why the test data used to locate JSON is not high depends on whether it is IO problem or JSON parsing problem

redkale: The next update will delete this code

nbrady-techempower: Unfortunately I’m not going to merge debugging code that breaks the test rules. It’s too easy to forget about that with the number of tests here. You’ll have to test these types of things locally.

redkale: OK, I'll take it out now

nbrady-techempower: Thanks!

PR #6432, opened 2021-03-10
change server tag

beetlex-io:
change server tag


joanhey (commented):
This change is going in the wrong direction.

Check the discusion:
https://github.com/TechEmpower/FrameworkBenchmarks/discussions/5673

PR #6440, opened 2021-03-13
Update dockerfile

redkale:


frameworks/Java/redkale/config.toml
```diff
@@ -14,36 +14,35 @@ platform = "Redkale"
 webserver = "Redkale"
 versus = "Redkale"
 
-[postgres]
-urls.db = "/db"
-urls.fortune = "/fortunes"
-urls.query = "/queries?q="
-urls.update = "/updates?q="
+[cache]
```
nbrady-techempower, line 17: This config doesn't match the one above. Hard to have a cache test without a database 😆

redkale: added， thanks

PR #6443, opened 2021-03-14
Fix Phalcon installation and code base update

Jeckerson:
Hello!

`psr` extension is required for phalcon work. I also updated code base...


frameworks/PHP/phalcon/public/index-micro.php
```diff
@@ -1,114 +1,109 @@
...
-try {
+use Phalcon\Db\Adapter\Pdo\Mysql;
+use Phalcon\Db\Enum;
+use Phalcon\Mvc\Micro;
+use Phalcon\Mvc\View;
+use Phalcon\Mvc\View\Engine\Volt;
+use Phalcon\Exception as PhalconException;
+
+use function header;
+use function json_encode;
```
joanhey, line 11: Please delete these two lines, they are useless.
It is trying to add non-namespaced imports to non-namespaced class, so by default it automatically inherits the global symbol table.

Check the error:
https://github.com/TechEmpower/FrameworkBenchmarks/pull/6443/checks?check_run_id=2104369951#step:9:646

joanhey:
Could you try to update the mongodb variant, if the bug is fixed ?

Good PR !!

Jeckerson:
@joanhey Can't understand how it works in terms of MongoDB...

In the docs rules it has:
> The schema for World is id (int, primary key) and randomNumber (int), except for MongoDB, wherein the identity column is _id, with the leading underscore.

So, when I keep response as is it will send warning message:
> WARN for http://tfb-server:8080/mongodb/queries?queries=2
     An extra key(s) is being included with the db object: _id

Meanwhile, when I disable `id` key:
>  WARN for http://tfb-server:8080/mongodb/queries?queries=501
     Response object was missing required key: id

> FAIL for http://tfb-server:8080/mongodb/queries?queries=foo
     Response object was missing required key: id

frameworks/PHP/phalcon/public/index-micro.php
Jeckerson: Done.

PR #6446, opened 2021-03-14
Update to sqlalchemy 1.4b2

Dreamsorcerer:


Dreamsorcerer:
Not sure why the async engine is not working now, filed a bug at:
https://github.com/sqlalchemy/sqlalchemy/issues/6044

Dreamsorcerer:
So, I think the issue with deduplicating SELECT statements is fixed using `populate_existing`.

Regarding UPDATE statements, the suggestion is that the benchmarks should be changed:
https://github.com/sqlalchemy/sqlalchemy/discussions/6045#discussioncomment-480552

Though there is also another suggestion that using `update()` statements directly could work.

PR #6447, opened 2021-03-14
feat(zio-http): add plaintext and json benchmarks for zio-http

amitksingh1490:

### DIFF
--- a/frameworks/Scala/ZIO-Http/build.sbt
+++ b/frameworks/Scala/ZIO-Http/build.sbt
@@ -1,0 +1,13 @@
+scalaVersion := "2.13.3"
+lazy val root = (project in file("."))
+  .settings(
+    name := "helloExample",
+    libraryDependencies ++= Seq(
+      "com.github.plokhotnyuk.jsoniter-scala" %% "jsoniter-scala-core"   % "2.6.4",
+      // Use the "provided" scope instead when the "compile-internal" scope is not supported
+      "com.github.plokhotnyuk.jsoniter-scala" %% "jsoniter-scala-macros" % "2.6.4" % "compile-internal",
+      "io.d11" % "zhttp" % "1.0.0-RC1"
+
+    ),
+    testFrameworks += new TestFramework("zio.test.sbt.ZTestFramework")
+  )
--- a/frameworks/Scala/ZIO-Http/project/build.properties
+++ b/frameworks/Scala/ZIO-Http/project/build.properties
@@ -1,0 +1,1 @@
+sbt.version = 1.4.7
--- a/frameworks/Scala/ZIO-Http/src/main/scala/Main.scala
+++ b/frameworks/Scala/ZIO-Http/src/main/scala/Main.scala
@@ -1,0 +1,24 @@
+import zhttp.http._
+import zhttp.service.Server
+import zio.{App, ExitCode, URIO}
+
+object WebApp extends App {
+
+  val message: String = "Hello, World!"
+
+  import com.github.plokhotnyuk.jsoniter_scala.macros._
+  import com.github.plokhotnyuk.jsoniter_scala.core._
+
+  case class Message(message: String)
+  implicit val codec: JsonValueCodec[Message] = JsonCodecMaker.make
+
+  val jsonstr = new String(writeToArray(Message(message)))
+
+  val app = Http.collect[Request] {
+    case Method.GET -> Root / "plaintext" => Response.text(message)
+    case Method.GET -> Root / "json" => Response.jsonString(jsonstr)
+  }
+
+  override def run(args: List[String]): URIO[zio.ZEnv, ExitCode] = Server.start(8080, app).exitCode
+
+}


frameworks/Scala/zio-http/src/main/scala/Main.scala
```diff
@@ -0,0 +1,41 @@
...
+    case Method.GET -> Root / "json" => Response.http(
+      headers = List(
+        Header(JHttpHeaderNames.SERVER, "zio-http"),
+        Header(JHttpHeaderNames.DATE, s"$createDate"),
+        Header(JHttpHeaderNames.CONTENT_TYPE, JHttpHeaderValues.APPLICATION_JSON),
+      ),
+      content = HttpContent.Complete(jsonstr))
+  }
+
+  override def run(args: List[String]): URIO[zio.ZEnv, ExitCode] = Server.start(8080, app).exitCode
```
kubukoz, line 39: Suggestion: disable zio tracing :)

Also, thread count should probably match the one in other frameworks (https supposedly has 32: https://github.com/TechEmpower/FrameworkBenchmarks/blob/a3771b148f434b007fe6d63f97007cdbc2c145ff/frameworks/Scala/http4s/src/main/scala/http4s/techempower/benchmark/WebServer.scala#L98)

frameworks/Scala/zio-http/src/main/scala/Main.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  val message: String = "Hello, World!"
+def createDate: String = DateTimeFormatter.RFC_1123_DATE_TIME.format(ZonedDateTime.now)
+  import com.github.plokhotnyuk.jsoniter_scala.macros._
+  import com.github.plokhotnyuk.jsoniter_scala.core._
+
+  case class Message(message: String)
+  implicit val codec: JsonValueCodec[Message] = JsonCodecMaker.make
+
+  val jsonstr = new String(writeToArray(Message(message)))
```
kubukoz, line 19: Maybe it's just me, but I'd say manually encoding the value to a byte array defeats the purpose of testing the framework's performance, since this code is part of the test and not the framework.

frameworks/Scala/zio-http/src/main/scala/Main.scala
```diff
@@ -0,0 +1,41 @@
...
+import zio.{App, ExitCode, URIO}
+import io.netty.handler.codec.http.{HttpHeaderNames => JHttpHeaderNames, HttpHeaderValues => JHttpHeaderValues}
+
+import java.time.ZonedDateTime
+import java.time.format.DateTimeFormatter
+
+object WebApp extends App {
+
+  val message: String = "Hello, World!"
+def createDate: String = DateTimeFormatter.RFC_1123_DATE_TIME.format(ZonedDateTime.now)
```
kubukoz, line 12: To stay in the idiomatic realm of using ZIO-related libraries this should be suspended in ZIO (`IO.effectTotal` on `ZonedDateTime.now`, or `zio.clock.currentDateTime` etc.)

frameworks/Scala/zio-http/src/main/scala/Main.scala
amitksingh1490: We wanted to not have circe or jsoniter dependency in zio-http. We are reevaluating this will get back

frameworks/Scala/zio-http/src/main/scala/Main.scala
tusharmath: We will move certain "basic" headers to the core. For eg: `Date` and `Server`. Almost all libraries do it this way for performance reasons. Also, the code above will look a lot more concise.

frameworks/Scala/zio-http/src/main/scala/Main.scala
tusharmath: By default AFAIK it's going to be 2x `availableCPU()`. Need to check if the test machine already has virtualization because in that case it should be equal to `availableCPU()`

frameworks/Scala/zio-http/zio-http.dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM alpine as builder
```
lolgab, line 1: You could use a fixed alpine version here to avoid that future updates can break the build

frameworks/Scala/zio-http/src/main/scala/Main.scala
```diff
@@ -0,0 +1,37 @@
...
+        Header(JHttpHeaderNames.SERVER, "zio-http"),
+        Header(JHttpHeaderNames.DATE, s"$createDate"),
+        Header(JHttpHeaderNames.CONTENT_TYPE, JHttpHeaderValues.APPLICATION_JSON),
+      ),
+      content = HttpContent.Complete(writeToString(Message(message))))
+  }
+
+  override def run(args: List[String]): URIO[zio.ZEnv, ExitCode] = Server.start(8080, app).exitCode
+
+}
```
tusharmath, line 37: @amitksingh1490 can you just run it thru scalafmt also.

tusharmath (approved):

PR #6487, opened 2021-03-24
Update 2.3.0

redkale:


nbrady-techempower:
@redkale Is there a reason you keep making these somewhat meaningless pull requests? Are you trying to trigger something?

redkale:
redkale 2.3.0-SNAPSHOT is updated

nbrady-techempower:
This PR doesn't update your snapshot? Are you changing snapshots while keeping the same version number?

PR #6491, opened 2021-03-26
Add FaF (rust) framework

errantmind:
Add FaF (rust) framework


fafhrd91:
to me this doesnt look like framework

errantmind:
I suppose that depends on your criteria for a framework. It is minimal, but I think it would satisfy basic criteria. It it an event loop which parses the complete HTTP request and passes everything to the 'user' in a callback who can define the response how they want.  I'd like to get this merged ASAP for testing so if there is a concrete issue please let me know.

fafhrd91:
sure, you can call framework anything that provides callback. but to me this still looks like bunch of low level code.

errantmind:
I generally appreciate feedback but I don't see how that is a useful criticism.  The goals / values for this framework are speed and simplicity of design.  If it is `low level`, it is because it has very few dependencies: it implements its own event loop and manages its own threads, which is going to be 'low level' by definition. Is this because it is a source distribution and not a regular library?

nbrady-techempower:
Hi @errantmind,

Is it possible to get a lot of this low level code in a lib/crate so it's not all committed here?  Could you also tie this down to a specific version, not latest: `FROM errantmind/debian-faf`

errantmind:
@nbrady-techempower Sure, I made the changes

PR #6503, opened 2021-04-01
Updatet may-minihttp(rust)

Xudong-Huang:
update rust stable version 1.51
update db update sql
update fortunes template engine
uddate json serialize backend

### DIFF
--- a/frameworks/Rust/may-minihttp/src/main.rs
+++ b/frameworks/Rust/may-minihttp/src/main.rs
@@ -6,11 +6,11 @@ use std::io;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
 
-use may_minihttp::{BodyWriter, HttpService, HttpServiceFactory, Request, Response};
-use may_postgres::{self, Client, RowStream, Statement};
+use may_minihttp::{HttpService, HttpServiceFactory, Request, Response};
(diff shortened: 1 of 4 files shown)


frameworks/Rust/may-minihttp/src/main.rs
```diff
@@ -159,47 +158,50 @@ impl PgConnection {
...
-        self.client.simple_query(&update)?;
+        self.client.query(&self.updates[num - 1], &params)?;
         Ok(worlds)
     }
 
-    fn tell_fortune(&self) -> Result<Vec<Fortune>, may_postgres::Error> {
-        let mut items = Vec::with_capacity(80);
-        items.push(Fortune {
+    fn tell_fortune(&self) -> Result<SmallVec<[Fortune; 32]>, may_postgres::Error> {
+        let mut items: SmallVec<[_; 32]> = smallvec::smallvec![Fortune {
```
nbrady-techempower, line 201: This is going to fail with the new toolset. The array that holds fortunes needs to have dynamic sizing. We test thousands of fortunes during verification process to make sure this is the case.

Xudong-Huang: Thanks. I think SmallVec would fall back to the heap for larger allocations. The doc says " If the data exceeds this limit, the SmallVec will "spill" its data onto the heap, allocating a new buffer to hold it."

PR #6519, opened 2021-04-06
salvo webframework update

chrislearn:


frameworks/Rust/salvo/src/main_db.rs
```diff
@@ -91,31 +92,43 @@ async fn updates(req: &mut Request, res: &mut Response) -> Result<(), Error> {
...
-        Err(e) => Err(io::Error::new(io::ErrorKind::Other, e)),
-    }?;
-    let tmpl = FortuneTemplate { items: &rows };
+    let mut items = fortune::table.get_results::<Fortune>(&conn)?;
+    items.push(Fortune {
+        id: 0,
+        message: "Additional fortune added at request time.".to_string(),
+    });
+    items.sort_by(|it, next| it.message.cmp(&next.message));
+    let mut body = String::with_capacity(2048);
```
nbrady-techempower, line 104: This will fail before the next round starts. The new toolset checks to make sure that the application is doing dynamic sizing for fortunes by testing thousands of fortune messages.

PR #6530, opened 2021-04-10
Update crystal benchmarks for 1.0 release

matthewmcgarvey:
This benchmark hasn't really been updated in 3 years.

Updated:
- crystal
- crystal-radix


matthewmcgarvey:
Failing to make enough queries because it's terminating one of the processes early. `ERROR:root:Terminating process 1049 by timeout of 20 secs.`

Not sure how to fix this or why it's happening

PR #6531, opened 2021-04-11
Gin gorm

absinthetized:
HI,

I've added a gin+gorm test suite. Neither cache nor fortunes are implemented. This code is here to test how gorm behaves. While Gin is a web framework for Go, Gorm is probably the most used go ORM.

I'm not the author of the other gin test. This PR is implemented by a totally different person, therefore even the limited business logic inside the tests can diverge from the existent gin test suite.


nbrady-techempower:
Thanks for the contribution @absinthetized 

> I'm not the author of the other gin test. This PR is implemented by a totally different person, therefore even the limited business logic inside the tests can diverge from the existent gin test suite.

This is totally fine, but we'd still want this to be included in the `gin` directory instead of making a new directory for a different test permutation with the same framework. Could move this into that directory, add the test in that benchmark_config as a separate permutation, and just put this code in a subfolder there.

absinthetized:
@nbrady-techempower I've moved the test under the common gin folder, I hope I've not misunderstood and this layout is the one you suggested!

PR #6537, opened 2021-04-14
Remove execute bit from top-level files

kevmoo:


kevmoo:
CC @nbrady-techempower

Issue #6538, opened 2021-04-15
Framework fails to start containers on systems using runc >=1.0.0.rc93


johnaohara:
Running tfb across multiple machines, using host networking, fails on systems that have `runc >= 1.0.0.rc93` installed

Setting any `net.` kernel options in a hosts default network namespace is bug which was fixed in runc-1.0.0.rc93;

https://github.com/opencontainers/runc/commit/09523b79d01d032d2faeeb7eef4ad5a31eda528f#diff-7f34ebfb08bb04cf9313703ad9d8e4ceee634a5ad8a412326c72333680b588feR164

### OS (Please include kernel version)
Linux ******.redhat.com 4.18.0-240.el8.x86_64 #1 SMP Wed Sep 23 05:13:10 EDT 2020 x86_64 x86_64 x86_64 GNU/Linux

### Expected Behavior

It is expected that the following command will execute the tfb benchmark;

` ./tfb --mode benchmark $CPUSET_CPUS --type plaintext --network-mode=host --server-host 192.168.0.1 --database-host 192.168.0.2 --client-host 192.168.0.3 --test quarkus`

### Actual Behavior

Any containers that are started with `sysctl=net.*` kernel configuration fail with the following error on the host:

e.g.  https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/utils/docker_helper.py#L322

```
docker: Error response from daemon: OCI runtime create failed: sysctl "net.core.somaxconn" not allowed in host network namespace: unknown.
ERRO[0000] error waiting for container: context canceled 
```

The same error is displayed in the tfb output;

```
--------------------------------------------------------------------------------
Running Test: quarkus
--------------------------------------------------------------------------------
quarkus: Traceback (most recent call last):
quarkus:   File "/FrameworkBenchmarks/toolset/benchmark/benchmarker.py", line 139, in __run_test
quarkus:     test.database.lower())
quarkus:   File "/FrameworkBenchmarks/toolset/utils/docker_helper.py", line 338, in start_database
quarkus:     log_config={'type': None})
quarkus:   File "/usr/local/lib/python2.7/dist-packages/docker/models/containers.py", line 809, in run
quarkus:     container.start()
quarkus:   File "/usr/local/lib/python2.7/dist-packages/docker/models/containers.py", line 400, in start
quarkus:     return self.client.api.start(self.id, **kwargs)
[...]

PR #6547, opened 2021-04-22
Do not set  kernel parameters when running in host network mode

johnaohara:
As mentioned on #6538 :

This change has the potential for impacting results, by changing the kernel params set depending on network type, I wanted to be sure that there were no negative side effects before opening a PR.

I have looked in the contributing docs and the github action for running tests when a PR is opened, but can not see anywhere where host network mode is tested. Are there any tests that run specifically in host network mode, or is running in this mode manually validated?

Any help in validating these changes would be greatly appreciated

### DIFF
--- a/toolset/utils/docker_helper.py
+++ b/toolset/utils/docker_helper.py
@@ -184,7 +184,11 @@ class DockerHelper:
                 }
                 name = None
 
-            sysctl = {'net.core.somaxconn': 65535}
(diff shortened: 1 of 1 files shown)


nbrady-techempower:
This looks good to me. We'll merge it in and keep an eye on the results when the next full run starts. Thanks for doing this!

PR #6560, opened 2021-05-02
Gin gorm - revision

absinthetized:
hello,

after the outcome of the continuous tests I've made a couple of changes to the code to improve it. Namely:
- disabled any middleware in gin as per original gin submission
- disabled gorm logging
while I would not define these as "real life impl. for production" they are more aigned with the other gin codebase and represent better the overhead of the gorm ORM.

If you think this makes sens please merge the PR!


absinthetized:
Nice of github to not notify conflicts by mail... Lets fix!

PR #6564, opened 2021-05-02
add magician framework

yuyenews:
add magician framework


yuyenews:
add Magician framework  and update Martian framework

PR #6565, opened 2021-05-03
Add ASP.NET Core support for EF Core scenarios with Pomelo and MySQL

lauxjpn:
Adds ASP.NET Core tests to support EF Core scenarios with the [Pomelo.EntityFrameworkCore.MySql](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql) provider and MySQL as the backend database.

* Tests were implemented analog to the ones for Npgsql/PostgreSQL.
* The warning for the updates tests is expected and a result of the `SELECT ROW_COUNT()` statement executed after each UPDATE.
* Updates `MySqlConnector` to its latest version, since this is a requirement for Pomelo 5.0.0.
* I will add tests for MariaDB in another PR later as well.

(I removed the ASP.NET Core on Mono tests from this PR, which where part of an earlier version, since their implementation depends on the approach choosen for #6566, and I don't even know yet, if the ASP.NET on Mono tests in general are being kept at all, or if they are just a relict of the past; maybe someone can comment on that).


nbrady-techempower:
@lauxjpn Just a couple of conflicts here and not sure if you still want this in. Going to close for now since it's been open so long, but if you re-open I'll get it in right away.

PR #6566, opened 2021-05-03
Add ASP.NET Core RAW MySQL test

lauxjpn:
Adds support for an ASP.NET Core MySQL RAW benchmark, that was already partially setup.


lauxjpn:
@roji The UPDATE statement counter for PostgreSQL seems flaky (as it seems to be for MySQL without the margin that was added to it in the past). Also fails from time to time in my local TFB verification runs. Is that a known issue?

Made the [checks failing](https://github.com/TechEmpower/FrameworkBenchmarks/pull/6566/checks?check_run_id=2509951116) here.

This PR should not have had any impact on the Npgsql/PostgreSQL tests at all.
I'll just trigger the Action again.

lauxjpn:
> @roji The UPDATE statement counter for PostgreSQL seems flaky (as it seems to be for MySQL without the margin that was added to it in the past). Also fails from time to time in my local TFB verification runs. Is that a known issue?

Ah, its related to https://github.com/TechEmpower/FrameworkBenchmarks/pull/6568#issuecomment-832740576. Makes sense for EF not to update, if no change has happened.

roji:
For the platform benchmarks, one of the goals is to get rid of the virtual dispatches involved in using the System.Data abstraction (DbConnection, DbCommand) so this PR would slow that down a bit (this also introduces boxing by switching to DbParameter, so no more use of the Npgsql-specific generic `NpgsqlParameter<T>`). I know it sounds a bit extreme, but the platform benchmarks are extreme (e.g. see the totally stripped-down ASP.NET, which sane applications generally won't do). So it may be better to duplicate the benchmarks rather than switch to the abstraction. Note the difference with the regular, non-platform benchmark which indeed looks much more like a regular application.

I don't really know anything about any UPDATE statement counter - are you referring to a TechEmpower mechanism for verifying how many actual updates were performed on the database side? If so, would that be flaky for *all* TechEmpower tests, regardless of .NET?

roji:
Thanks for the clarification - I admit I'm not familiar with that aspect of the TE infrastructure. So this would be a global TE problem as opposed to something specific to .NET right?

lauxjpn:
> So this would be a global TE problem as opposed to something specific to .NET right?

I suspect this to be a database server implementation bug, that counts incorrectly in high concurrency situations, and not a .NET specific or TE specific issue.

It could still be a TE specific issues, I some databases like Postgres or MySQL might update their update statements counters in a delayed fashion. Then, TE would need to accommodate for that (e.g. query the counter again after some time, if it is less than expected the first time).

To be a .NET specific issue, it would need to be some very weird async related concurrency issue on the ADO.NET layer, which I think is unlikely, or an even weirder general async related issue (highly unlikely).

lauxjpn:
> I know it sounds a bit extreme, but the platform benchmarks are extreme (e.g. see the totally stripped-down ASP.NET, which sane applications generally won't do). So it may be better to duplicate the benchmarks rather than switch to the abstraction.

@roji I updated the PR in a way, that abtracts with as little impact as possible, by introducing an `IDb` interface, so that the same endpoints can call the different database provider implementations. The additional cost is a vtable call (instead of a direct call) per endpoint call, which I think is very reasonable.

Otherwise, we would already need to differentiate between database providers within the endpoint URI, and we would need to measure, whether parsing additional characters in the endpoint string is slower or faster than the vtable call.
(I suspect slower, but it could depend on, whether the newly introduced character(s) start a new DWORD/QWORD alignment block, or not).

lauxjpn:
Once someone approves the workflow run (the most recent one is enough, older ones can be skipped/cancelled), this PR can be squashed & merged.

bgrainger:
> Adds support for an ASP.NET Core MySQL RAW benchmark, that was already partially setup.

FYI, the "Raw" MySQL benchmark existed for quite some time until it was deleted (without warning or notice) here: https://github.com/TechEmpower/FrameworkBenchmarks/commit/77fd80534ab640725b6be26d527da406244fcaad#r41825500

Thanks for restoring it!

(I do see there's still an `IDb` abstraction... can the JIT devirtualize that to eliminate performance concerns? Would making a generic `BenchmarkApplication<TDb>` implementation help at all?)

frameworks/CSharp/aspnetcore/PlatformBenchmarks/Data/RawDbMySqlConnector.cs
```diff
@@ -0,0 +1,268 @@
...
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Caching.Memory;
+using MySqlConnector;
+
+namespace PlatformBenchmarks
+{
+    // Is semantically identical to RawDbNpgsql.
+    // If you are changing RawDbMySqlConnector, also consider changing RawDbNpgsql.
+    public class RawDbMySqlConnector : IDb
```
roji, line 13: Make implementations sealed - that may help the JIT devirtualize in some cases.

roji (commented):
@lauxjpn @bgrainger I think the point of the platform benchmark was a stripped-down, ultra-high-speed benchmark showing the maximum that .NET can get to; this is why it is Npgsql-only - this happens to be the setup that's fastest. The point wasn't to do an extensible benchmark across databases and technologies, but to squeeze every bit of perf out for .NET as a platform.

So I'm not sure there's a lot of value of adding a MySQL implementation of the **platform** benchmark - the regular database benchmarks should provide a good place for measuring different databases etc. I'm not saying it shouldn't be done and that this change will slow things down - although I have been seeing virtual calls slowing down perf slightly in some scenarios.

What do you think?

lauxjpn:
> I think the point of the platform benchmark was a stripped-down, ultra-high-speed benchmark showing the maximum that .NET can get to; this is why it is Npgsql-only - this happens to be the setup that's fastest.

@roji I don't so how we can know that it is the fastest, if we don't compare its benchmark to other implementations.

> So I'm not sure there's a lot of value of adding a MySQL implementation of the platform benchmark - the regular database benchmarks should provide a good place for measuring different databases etc.

I think the platform test for MySQL makes a lot of sense, as it demonstrates the contrast between what is achievable and what is typical. When looking at the locally run benchmarks referenced in https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/1408#issuecomment-839367646, it can be seen that the middleware benchmark has only 90% of the performance of the raw benchmark, but it is a _much_ smaller gap than between the PostgreSQL raw and middleware benchmarks.

> I'm not saying it shouldn't be done and that this change will slow things down - although I have been seeing virtual calls slowing down perf slightly in some scenarios.

I expect, that for the worst case, which is that the vtable is not cached at the moment the call needs to be made, the virtual call results in an additional memory read operation.
In practice, take a look at the following two samples:

* [With interface](https://sharplab.io/#v2:C4LglgNgPgAgTARgLACgYAYAEMEDoBKArgHbBgC2AprgMID25ADpJQE4DKbAbmAMaUBnANypUxAIZUBjcf0wAVQcFQBvVJg3Y4AZnWa1KTUcyNWYLuOCVsCAGzYALJgCy4sMQAUASj3HMBvz8ASQAxOjpMADNwzABeTABxSmAwum8RQ0CjXyyoulYPd2BMMDjMdCESzAAeTAc4SrAAaiafTNzjaLpcACFxAq8MrIBfHOz24wBtZ2SACzoAEyCmCA8Z4HmllYB5RjI6YgFcADk6IOIId3cAcy8AXTHNU3NLaxx7UJiklPDvR40Ah1MAB6YEucQAa2sAkIrGsG0smAAUkF5JheOJiMQ6MUIdiAO4AGkw+NmiN4B14cKsWkwGzeAHZKAt/sZQZg6AAjABWlF4xTAAg5kVwrKMMAZmGIlHxmHwmIWDG8J0oAA9gB50MSEF5MABSTBwOLxdBiwIAfilMswqT+EyBmBAVtl23prFtg1Zowm3vG4u0tNSjswnzoOUBfmmc0Wy0Yq3Wm1jEF2+0OJzOFyuxFuD3tmhgAZgTj6A1ZEdyOAAnB4AESpGuevMaX2aFsaHIF2mu2ZsINO0Ph/5RjYxlZraNbOMpsAHI6nc6XYg3e6sztFzAlu1ZctZKu17u98INoaBNuYM8dgNFNiRWTWAcTHf54v9dI5b3DIA==)
* [Without interface](https://sharplab.io/#v2:C4LglgNgPgAgTARgLACgYAYAEMEDoBKArgHbBgC2AprgMID25ADpJQE4DKbAbmAMaUBnANypUxAIZUBjcf0wAVQcFQBvVJg3Y4AZnWa1KTUcyNWYLuOCVsCAGzYALJgCy4sMQAUASj3HMBvz8AMTo6TAAzUMwAXkwAcUpgELpvEUNAo18MiLpWD3dgTDAYzHQhIswAHkwHOHKwAGoGn3Ts40i6XAAhcTyvNIyAXyzM1uMAbWdEgAs6ABMASSYIDyngWcXlgHlGMjpiAVwAOToF4gh3dwBzLwBdEc1Tc0trHHtk+MTk7weNALbsAB2TDESgAd0w336v0wwzGcNGRhg2i0kNCWX+fkmM3mS0YKzWGzxEB2ewOx1O50uxBu9zGSJRMCcPT6MMx2RwAE4PAAiZI86H0zQI4WoQZAA===)

For some reason, the generated IL contains a `virtual` call in both cases. So when switching to `JIT Asm`, both have generated the virtual call (in line 11/13). (No sure why, maybe you can spot what I did wrong.)

In general however, if a vtable call does matter, than the current implementation is likely to loose a lot of time with its endpoints calling other methods in the first place (and even multiple times). In which case, all those method calls should either be inlined by the compiler (unless they already are by a command line switch or something) or manually by flattening the call tree in the code.

> I do see there's still an `IDb` abstraction... can the JIT devirtualize that to eliminate performance concerns? Would making a generic `BenchmarkApplication<TDb>` implementation help at all?

@bgrainger As far as I can see, the `TDb` would still need to represent an abstract base class or an interface, so we can call its methods independent of its implementation.

We _could_ duplicate the endpoints or just manually branch between database implementations within them, if the vtable call turns out to be significant.

---

I think its much easier to just run the benchmark, and see what the results are. We can always revert the PR or optimize it after the run, when we know what is what.

frameworks/CSharp/aspnetcore/PlatformBenchmarks/Data/RawDbMySqlConnector.cs
```diff
@@ -0,0 +1,268 @@
...
+
+        public RawDbMySqlConnector(ConcurrentRandom random, AppSettings appSettings)
+        {
+            _random = random;
+            _connectionString = appSettings.ConnectionString;
+        }
+
+        public async Task<World> LoadSingleQueryRow()
+        {
+            using (var db = new MySqlConnection(_connectionString))
```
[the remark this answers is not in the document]
lauxjpn: I kept the code is similar as possible to the Npgsql implementation, so that general changes to one class can be easily copied over to the other implementation, as we are doing copy-paste-inheritance here.

So if we modernize one class, we should do the same to the other, but I did not want to introduce anything unnecessary in regards to the point of this commit, so it is clear what is responsible for a potential performance impact on PostgreSQL.

roji:
>> I think the point of the platform benchmark was a stripped-down, ultra-high-speed benchmark showing the maximum that .NET can get to; this is why it is Npgsql-only - this happens to be the setup that's fastest.
>
> @roji I don't so how we can know that it is the fastest, if we don't compare its benchmark to other implementations.

One way is to do a one-off comparison, rather than merging a permanent change. 

I'll just say that I noticed virtual dispatch making a different recently in another benchmark, and as this specific benchmark pretty much represents .NET on the global scale of these benchmark, I'd be wary of making changes here. Once again, the question is exactly what this is meant to achieve; if the point is to get a number at a point in time, I'll gladly help you by running any benchmarks on the ASP.NET platform (https://github.com/aspnet/Benchmarks/) - that's a great place to do any experimentation without affecting .NET's place in the rankings etc.

lauxjpn:
> One way is to do a one-off comparison, rather than merging a permanent change.

Sounds good to me. How is this process being done on Citrine?

> I'll just say that I noticed virtual dispatch making a different recently in another benchmark, and as this specific benchmark pretty much represents .NET on the global scale of these benchmark, I'd be wary of making changes here.

I am not. That is what benchmarks are for. We should just measure it. If its significant, we can just revert it, or optimize something else along the hot path, that compensates for it. We should be still far off from the next official Round.

> Once again, the question is exactly what this is meant to achieve [...]

The point is to demonstrate what is possible with a minimalist approach on ASP.NET using MySQL without any approaches like MW or MVC in its way. It is of interest to anyone who needs to squeeze everything out of the platform and database they are using, which is pretty much in line with what TFB measures in general (concurrent web access). It is the extreme ASP.NET with MySQL usage.

> I'll gladly help you by running any benchmarks on the ASP.NET platform (https://github.com/aspnet/Benchmarks/) - that's a great place to do any experimentation without affecting .NET's place in the rankings etc.

We can run the benchmarks there first and measure it there if you think this is really necessary. If I remember correctly, they use the same hardware etc. as the TFB machines. (I did asked about the ASP.NET platform benchmarks a couple of days ago, but the general vibe seemed to be, that they are MySQL unrelated, at least at this point in time.)

---

Now, what I am interested in is: What are your actual concerns about testing this PR on Cetrine in one of its ongoing runs?

roji:
> What are your actual concerns about testing this PR on Cetrine in one of its ongoing runs?

I guess I'm not too used to the approach of trying out and reverting things in the official TechEmpower benchmark repo. This creates a lot of churn, result history is going to be messy and difficult to interpret, you're asking for TE people's time to review and merge, and there's a chance we'd accidentally forget and leave something in that regresses performance (not that I mistrust you specifically - I mistrust myself after a long and painful experience with perf experimentation), etc.

To summarize, I see this repo as a production repo, where stability is important, and not as a playground for experiments. But I do definitely agree that you need a way to do perf experiments, and ideally on TE hardware. We should see what we can do to help you do stuff on the aspnet, but I guess until then I don't have a better solution.

lauxjpn:
The unreachable MySQL server (in general, not just for the platform tests), which made the PR fail, should not be related to anything of this PR (as it hasn't touched a single file that influences the MySQL server in any way, and only changed platform related files in general).

roji:
@lauxjpn I just discussed with @sebastienros, and I can run jobs for you on our platform if you want. You can do your tweaks in your fork of the TE repo, and then I can send a job on our platform against that fork. For example, the config for the TE benchmarks on the ASP.NET is https://github.com/aspnet/Benchmarks/blob/main/scenarios/te.benchmarks.yml - you can have your own Docker of MySQL too.

How does sound? That should hopefully provide you reactive experience (after all, we're in the same timezone :rofl:), and at the same time keep this repo cleaner from experimentation etc...

lauxjpn:
> To summarize, I see this repo as a production repo, where stability is important, and not as a playground for experiments.

Because I am wondering how non ASP.NET related projects are doing this, who cannot afford the luxury of having a duplicated setup of the TFB hardware available for internal testing, before being absolutely sure, that a change will indeed improve all possible related benchmarks.

_(I had a second paragraph here, but I decided to not post it.)_

>> One way is to do a one-off comparison, rather than merging a permanent change.
>
> Sounds good to me. How is this process being done on Citrine?

Can this even be achieved on Citrine? I was under the impression (but might remember this incorrectly) that there is only a five-minute gap between ongoing rounds (probably for restarting or resetting).

>>I'll gladly help you by running any benchmarks on the ASP.NET platform (https://github.com/aspnet/Benchmarks/) - that's a great place to do any experimentation without affecting .NET's place in the rankings etc.
>
>We can run the benchmarks there first and measure them if you think this is really necessary. If I remember correctly, they use the same hardware etc. as the TFB machines.

If this cannot be tested on Citrine and therefore has to be tested using the ASP.NET Core benchmark repo to get this merged at some point into TFB, so be it.

_(I had a second paragraph here, but I decided to not post it.)_

bgrainger:
> I think the point of the platform benchmark was a stripped-down, ultra-high-speed benchmark showing the maximum that .NET can get to; this is why it is Npgsql-only - this happens to be the setup that's fastest.

If that's the purpose, I can understand leaving a MySQL implementation off (because it complicates the code paths and we can be pretty certain from all TFB entrants that a MySQL-based solution is never going to be as fast as PG).

I'm happy to put effort into improving `aspcore-mw-ado-my` as the fastest MySQL implementation. (Or benchmarking privately if we just need to know what's possible.)

> I'll gladly help you by running any benchmarks on the ASP.NET platform (https://github.com/aspnet/Benchmarks/)

Is the best way to open an issue? I have one! https://github.com/aspnet/Benchmarks/issues/1676

lauxjpn:
> @lauxjpn I just discussed with @sebastienros, and I can run jobs for you on our platform if you want. You can do your tweaks in your fork of the TE repo, and then I can send a job on our platform against that fork. For example, the config for the TE benchmarks on the ASP.NET is https://github.com/aspnet/Benchmarks/blob/main/scenarios/te.benchmarks.yml - you can have your own Docker of MySQL too.
>
>How does sound? That should hopefully provide you reactive experience (after all, we're in the same timezone 🤣), and at the same time keep this repo cleaner from experimentation etc...

Thanks, that works for me!

lauxjpn:
> If that's the purpose, I can understand leaving a MySQL implementation off (because it complicates the code paths and we can be pretty certain from all TFB entrants that a MySQL-based solution is never going to be as fast as PG).

There are other ways. We could also just duplicate the PlatformBenchmark for MySQL/MariaDB and then link in the files from the original project (that would then still be Npgsql only) and just use our own database implementation layer instead of the one of Npgsql. Still "relatively" easy to maintain and no vtable call in the code path.

lauxjpn:
@roji BTW, shouldn't _all_ platform benchmarks be marked as `stripped` instead of `realistic`?

_(From the info icon tooltip of the filters panel:)_

> **Implementation approach** describes the test's design disposition.
> * A **realistic** implementation approach uses the framework with most out-of-the-box functionality enabled.  We consider this realistic because most applications built with the framework will leave these features enabled.
> * A **stripped** implementation approach removes or outright avoids implementing features that are unnecessary for the particulars of this benchmark exercise.  This might illuminate the marginal improvement available in fine-tuning a framework to your application's use-case.
> **Stripped** implementations are hidden by default.  See the Motivation and Question section for more information.

roji:
> shouldn't all platform benchmarks be marked as stripped instead of realistic?

I guess it's a matter of interpretation, but I personally don't think so... The same components are being used in the platform as in the non-platform benchmarks - it's the same ASP.NET, Npgsql, rather than versions where functionality has been stripped out. The **benchmark implementation** indeed uses ASP.NET in a way which is very perf-oriented and atypical, but I think that's a different matter.

For example, there are web frameworks out there which don't have MVC at all (or a middleware pipeline), but do expose a low-level API that's similar to how the platform benchmarks are using Kestrel directly. The .NET platform benchmark is about comparing against those, and avoiding using optional web components (e.g. MVC) which aren't strictly-speaking necessary. On the other hand, if I were to produce a version of Npgsql which only supports the code path needed to run Fortunes (no transactions, no parameters, no nothing), that would definitely fall into the definition of "stripped" above, as I understand it.

roji:
>> I'll gladly help you by running any benchmarks on the ASP.NET platform (https://github.com/aspnet/Benchmarks/)
>
> Is the best way to open an issue? I have one! aspnet/Benchmarks#1676

@bgrainger @lauxjpn this sounds like a good direction - I'll do my best to help. Let's move the conversation on helping you guys run experiments to aspnet/Benchmarks#1676, and keep this issue about bringing in the MySQL changes into the TechEmpower repo, does that make sense?

@lauxjpn 

> Because I am wondering how non ASP.NET related projects are doing this, who cannot afford the luxury of having a duplicated setup of the TFB hardware available for internal testing, before being absolutely sure, that a change will indeed not negatively alter all possible related benchmarks. Because if you cannot test it on TFBs hardware, you can only guess, which is bad when it comes to optimizing complex systems.

You're right, and maybe I was too hard on this approach above - apologies if so. But at least in our case, we do have a better alternative for running experiments, and I hope it would even provide you with a faster response loop; I can just launch a job with your branch and report back the results, rather than you having to submit a PR to TechEmpower, get it reviewed and merged, wait for the results to show up in the next run, and then revert it later.

> There are other ways, if the vtable call turns out to be significant. We could also just duplicate the PlatformBenchmark project for MySQL/MariaDB and then link in the files from the original project (that would then still be Npgsql only) and just use our own database implementation layer instead of the one of Npgsql. Still "relatively" easy to maintain and no vtable call in the code path.

Yeah, I'd personally prefer that approach. Another way here would be to use conditional compilation (#if) to select the database - that may help with the code duplication and maintenance. See how the DATABASE compilation symbol is being used in the current platform benchmark implementation - that should also give you an idea how strictly it gets treated wrt perf :sweat_smile:

lauxjpn:
> Yeah, I'd personally prefer that approach. Another way here would be to use conditional compilation (#if) to select the database - that may help with the code duplication and maintenance. See how the DATABASE compilation symbol is being used in the current platform benchmark implementation - that should also give you an idea how strictly it gets treated wrt perf 

Yes, that would be a fine approach too.

> Let's move the conversation on helping you guys run experiments to aspnet/Benchmarks#1676, and keep this issue about bringing in the MySQL changes into the TechEmpower repo, does that make sense?

I extracted the approach discussion into #6585.

lauxjpn:
@roji Alright, so lets run the currently existing ASP.NET Core platform benchmark (only containing Npgsql) as a baseline, and then this `fix/aspnetcore_my` branch twice, once with Npgsql as the database provider and once with MySqlConnector as the database provider.

Afterwards, I propose we compare the RPS of the Npgsql runs to each other, and also the `fix/aspnetcore_my` Npgsql run against the last 7 runs of the actual TFBs (the number 7 is arbitrary, any number higher than 4 is fine with me).

bgrainger:
> the alternative you proposed, which would use conditional preprocessor statements.

I used a `MYSQL` preprocessor definition to add my own "Platform" MySQL benchmark: https://github.com/bgrainger/FrameworkBenchmarks/commit/4e6fab769dabce3eab38aadb3348ad88c0e2740c; cf. https://github.com/aspnet/Benchmarks/issues/1676#issuecomment-840884063.

lauxjpn:
> @roji Alright, so lets run the currently existing ASP.NET Core platform benchmark (only containing Npgsql) as a baseline, and then this `fix/aspnetcore_my` branch twice, once with Npgsql as the database provider and once with MySqlConnector as the database provider (everything on [aspnet/Benchmarks](https://github.com/aspnet/Benchmarks)).
> 
> Afterwards, I propose we compare the RPS of the Npgsql runs to each other, and also the `fix/aspnetcore_my` Npgsql run against the last 7 runs of the actual TFBs (the number 7 is arbitrary, any number higher than 4 is fine with me).
> 
> If a performance degradation is being shown between the two Npgsql runs and RPS fall below the minimum of the last 7 TFB runs, then I would call the vtable call as too impactful, and I will change the PR to the alternative you proposed, which would use conditional preprocessor statements.
> 
> Otherwise, we should just merge it as is.

@roji Any update on this?

lauxjpn:
@roji Have there turned up roadblocks in regards to running this on the ASP.NET hardware?

Otherwise, it has been a week since I asked for it, and there were already two runs for the (unrelated) https://github.com/aspnet/Benchmarks/issues/1676 issue within that time frame.

So I would really appreciate a run within a timely fashion, so we finally have an accurate measurement.

roji:
@lauxjpn sorry I haven't responded - I've been busy with other things. I'll take a proper look tomorrow.

roji:
Though without taking a deeper look, any reason to not just do things with conditional compilation as @bgrainger suggests above? I'd really rather we did this, and not touched any of the existing code paths.

lauxjpn:
> Though without taking a deeper look, any reason to not just do things with conditional compilation as @bgrainger suggests above? I'd really rather we did this, and not touched any of the existing code paths.

I think at this point, with all the previous theorizing and discussion about how this might impact the existing tests, we should really just run it so we _know_ the impact.

If it has one, than performance trumps maintainability.
If it has none, than maintainability trumps spaghetti code.

roji:
> I think at this point, with all the previous theorizing and discussion about how this might impact the existing tests, we should really just run it so we know the impact.

For this benchmark specifically, I'd rather we didn't introduce more overhead - JIT code generation is constantly being improved, what doesn't have an effect may have an effect tomorrow. I understand this is a bit of an extreme position - and wouldn't it for the other benchmarks - but let's just keep this one, specific benchmark as lean and clean as possible, especially since we have another option here (conditional compilation). 

Will look at

lauxjpn:
> For this benchmark specifically, I'd rather we didn't introduce more overhead - JIT code generation is constantly being improved, what doesn't have an effect may have an effect tomorrow. I understand this is a bit of an extreme position - and wouldn't it for the other benchmarks - but let's just keep this one, specific benchmark as lean and clean as possible, especially since we have another option here (conditional compilation).

I think we are kind of moving in circles around this one. If this introduces overhead, the benchmark will tell us, in which case, I am obviously changing the PR to use a conditional.

Otherwise, there should be no need, since if the JIT deoptimizes this particular case in the future (for whatever reason), this benchmark would show it. So its a good thing.

Therefore, lets really first run the benchmark, and then draw final conclusions here. It is likely to perform worse, but we don't _know_ until we test it.

At this point, the new debate whether this benchmark should even be run on the ASP.NET Citrine hardware, has likely taken more time than just running it.

Your solution for us, to work around the apparently very tight coupling of the ASP.NET and EF Core teams to even the in-between benchmark runs of the TFB, was to let us run more experiment type PRs on the ASP.NET Citrine hardware. I think it is good solution.

roji:
@lauxjpn we're indeed moving around in circles - I've already written my opinion here, several times. Regardless of what the benchmark runs will show, I am against this change, so I don't see the point of running it.

lauxjpn:
Alright, if that's how it is, then that's how it is. Though the solution you offered to open source contributions lost a lot of their sexiness with this.

---

Anyway, then let's at least get the process down, so it is clear what the rules are and when they apply.
This is what we know so far:

* We ping you and you decide whether it is in your interest to run our experimental commit or PR on the ASP.NET Core Citrine. If it is not, we cannot contribute to the TFB. If it is, it is going to be tested.
* If it is a PR and you are happy with whatever result the ASP.NET Core Citrine tests show, we can ping someone from the TFB team to get it merged in the TFB repo. Otherwise, no contribution as is.

What is not clear so far is, to what contributions this rules apply (the scope):

* All (but only) the ASP.NET Core _platform_ benchmarks
* All ASP.NET Core benchmarks (really _anything_ that runs on ASP.NET Core)
* Only those that touch specific files that could have an influence on specific already existing ASP.NET Core benchmarks (e.g. those that touch files that could affect Npgsql/PostgreSQL benchmarks).
* All contributions to the TFB
* There are no fixed rules, all contributors need to ping you and you decide on a case by case basis.

Once this has been defined, I think it needs to be part of some official issue here, so it is clear what the rules and processes are for open source contributions (at least in regards to ASP.NET Core benchmarks) and they can be looked-up at any time.

/cc @ajcvickers @sebastienros @bhauer

roji:
> Alright, if that's how it is, then that's how it is. Though the solution you offered to open source contributions lost a lot of their sexiness with this.

@lauxjpn maybe the source of the confusion is this... You are free to create your own benchmark implementations and submit them to TechEmpower (cloning the existing ASP.NET ones, or otherwise). Various people have done that (see [ben](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/CSharp/ben), [beetlex](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/CSharp/beetlex)) and run experiments in their own corner, without affecting the official .NET ones. You wouldn't need anyone's permission for that, and you could iterate as much as you want directly with the TechEmpower folks (like in other frameworks).

What you're asking is to introduce your changes into the "official" .NET platform benchmark - the most high-perf one, where we already go to extreme lengths to avoid *any* extra processing that isn't needed. I'd rather we avoided that on principle; an extra virtual call that doesn't appear on a benchmark run today could become important at some future point, if other things are optimized by then. And no, I prefer not to wait to *see* that virtual call in a profiler, at that point - I'd rather we just kept the platform benchmark minimal, on principle.

I think the "rules and processes" here are really very simple:

1. As far I'm concerned, you can do whatever you want outside frameworks/CSharp/aspnetcore (the "official" .NET benchmarks) - create your own thing, experiment, anything. I'd be glad to help in any way I can with anything that touches on .NET data.
2. If you want to merge changes into frameworks/CSharp/aspnetcore, then yeah, I think it's important that a .NET perf person approve them. That seems reasonable - otherwise anyone can just go and make modifications which impact .NET's place.
3. I'd gladly help by running your proposed changes/experiments that attempts to improve things for .NET, as I'm already doing in https://github.com/aspnet/Benchmarks/issues/1676 and https://github.com/TechEmpower/FrameworkBenchmarks/pull/6588.
    * The problem with this specific PR, is that it doesn't attempt to optimize anything - it simply integrates MySQL into the platform benchmark in a way which potentially makes it slower. Since this isn't something I think we should do (for the reasons above), I don't see the point of running a benchmark on it. But again, I'll gladly run anything else that aims to improve things.
4. Final merging needs to be done by the TechEmpower people.

I really hope the above explains things clearly, and makes sense. I'm available to answer any further question on this.

lauxjpn:
> If you want to merge changes into frameworks/CSharp/aspnetcore, then yeah, I think it's important that a .NET perf person approve them. That seems reasonable - otherwise anyone can just go and make modifications which impact .NET's place.

Thanks, that is what was missing from the picture. I will ping you for related PRs then and wait until they get approved.

> I really hope the above explains things clearly, and makes sense. I'm available to answer any further question on this.

Yes, it's clear now and fine. I was asking about it, so issues don't get blocked unnecessarily by waiting for approval in cases where no approval is needed.

> The problem with this specific PR, is that it doesn't attempt to optimize anything - it simply integrates MySQL into the platform benchmark in a way which potentially makes it slower. Since this isn't something I think we should do (for the reasons above), I don't see the point of running a benchmark on it. But again, I'll gladly run anything else that aims to improve things.

I know. I just though since we talked in length about the potential impact and you offered to run it earlier on the ASP.NET Core Citrine hardware (before you later changed your mind), it would have been interesting to actually measure the impact, so informed decisions can be made based of those findings in the future for Citrine.

I would have been completely fine with a compromise like: "Let's run it, so we know what the impact of a virtual method call is on Citrine, even though the outcome will not change our position on not making it part of the official repo, because we think it's overall too risky."

Anyhow, I will implement a preprocessor condition based approach.

lauxjpn:
@roji Alright, take a look at it and give it a spin if you like.

lauxjpn:
Since this hasn't been merged yet, I added a commit with a very minor project file cleanup/simplification.

roji:
Sorry @lauxjpn, I've been deep into something - will definitely get around to this today or tomorrow.

frameworks/CSharp/aspnetcore/PlatformBenchmarks/Program.cs
```diff
@@ -55,13 +51,20 @@ public static IWebHost BuildWebHost(string[] args)
             Console.WriteLine($"Database: {appSettings.Database}");
             Console.WriteLine($"ConnectionString: {appSettings.ConnectionString}");
 
+    #if NPGSQL
```
sebastienros, line 54: Why not just `#if DATABASE` ?

sebastienros:
LGTM

Have you found optimal connection string arguments? There is a default pool size of 1024, isn't it impacting perf since it will never reach this amount?

roji (approved):
Finally got around to this... Thanks, this looks good to me too!

frameworks/CSharp/aspnetcore/PlatformBenchmarks/Program.cs
lauxjpn: Do you mean to get rid of the individual library preprocessor directives here and just use one check for all/both of them?
Probably simpler.
I implemented it the way it currently is, as a sanity check for developers when copy pasting docker files, so they get a message that is a bit more helpful than "cannot connect" or so, but the code already outputs the connection string, so this might be overkill.

lauxjpn:
> Have you found optimal connection string arguments? There is a default pool size of 1024, isn't it impacting perf since it will never reach this amount?

@sebastienros I have started some local perf checks in regards to connection string parameter and config values a couple of weeks ago. I will ping @roji from time to time over the next few weeks, to test some changes on the ASP.NET Core Citrine, so we can get this optimized before the next official run.

roji:
Yeah, we can definitely tweak this in future PRs if we need to.

lauxjpn:
The failing CI seems to be some non-deterministic MySQL server setup issue (we had this randomly in other PRs as well).
It should not be related to the specifics of this PR.

Does it need to be triggered again until it succeeds, or can it be merged anyway?

lauxjpn:
> The failing CI seems to be some non-deterministic MySQL server setup issue (we had this randomly in other PRs as well).

Looks like I found the root cause for this. As expected, it is unrelated to this issue here. I'll open an issue for it and post a PR to fix it.

lauxjpn:
Okay, looks like #6611 did the trick.

This PR can be squashed & merged now.

PR #6567, opened 2021-05-03
Update Ben connection string

benaadams:
Update max pool size


sebastienros:
Only if you update the other benchmarks

NinoFloris:
I'll do the others tomorrow ;)

benaadams:
> Only if you update the other benchmarks

Not merged the change in aspnet/benchmarks so seems wrong to propergate it for the others?

roji:
Have no fear - am gonna benchmark the change on the others tomorrow.

PR #6568, opened 2021-05-05
Update connection strings used for ASP.NET Core

NinoFloris:
Pending merge https://github.com/aspnet/Benchmarks/pull/1667#pullrequestreview-652264613

Updates connection strings to use a tuned maximum of db connections, and enables multiplexing for non-platform scenarios.

/cc @sebastienros @benaadams @roji


sebastienros:
When it's merge we'll need to compare all the results, since TE runs with different number of connections, while our internal ones have a fixed number (256).

sebastienros:
Update EF is broken. Pretty sure I had disabled it and filed an issue on EF with the concern. Like it's detecting an object doesn't need to change and is not sending the update. The TE CI steps counts the expected statements.

roji:
@sebastienros don't remember anything about this... We don't have update EF in aspnet/Benchmarks, right?

In any case, EF would be pretty slow here, with the entire change tracking mechanism coming into play :)

sebastienros:
@roji right, that's why I was pretty sure I had disabled it.

@NinoFloris can you remove the two ef-updates sections in the `benchmark_config.json` file so the tests pass?

sebastienros:
And probably also `aspnetcore.mw-ef-pg-up.dockerfile` and `aspnetcore.mvc-ef-pg-up.dockerfile` that are only use by these scenarios

NinoFloris:
@sebastienros done, it needs an approval to run CI

NinoFloris:
@sebastienros it's green!

sebastienros (approved):

benaadams:
What about all the others? 😉

lauxjpn:
> Update EF is broken. Pretty sure I had disabled it and filed an issue on EF with the concern. Like it's detecting an object doesn't need to change and is not sending the update. The TE CI steps counts the expected statements.

> @sebastienros don't remember anything about this... We don't have update EF in aspnet/Benchmarks, right?
> 
> In any case, EF would be pretty slow here, with the entire change tracking mechanism coming into play :)

In my benchmark run for https://github.com/TechEmpower/FrameworkBenchmarks/pull/6566 (also discussed there), it the RAW `aspcore-ado-pg-up` benchmark had the same issue. So I suspected, that PostgreSQL might count the updates inaccurately under heavy pressure (e.g. a non-interlocked increment).

We have the same issue with the MySQL tests, but someone did implement a (suboptimal) error margin for it in the past.

roji:
>> In any case, EF would be pretty slow here, with the entire change tracking mechanism coming into play :)
>
> That is more of a reason to expose it with a benchmark, so users can make informed decisions on what ORM to use for their particular needs.

Am definitely not saying we should make it work. Are you working in that area as well for MySQL?

lauxjpn:
> Am definitely not saying we should make it work. Are you working in that area as well for MySQL?

Yes, #6566 adds the raw benchmark for MySQL (including those for updates), and #6565 the ones for EF Core against MySQL.
I also got a PR ready, that adds all the benchmarks for MariaDB.

The point of my previous post is, that we _should_ keep the benchmarks, because it is vital information to know, how different ORMs compare to each other and to RAW ADO.NET in regards to update scenarios.

Here is what I propose:

### Immediate items

* Revert the removal of the update related Npgsql benchmarks.
* Implement an instrumentation inaccuracy margin as a workaround in `postgres.py`. If it was acceptable all this time as a workaround for MySQL, it should be acceptable for Postgres.
* Running the Npgsql update benchmarks suggests, that the margin needs to be higher for Postgres (counted 10114 of 10240 updates the posted benchmark run above) than for MySQL (usually counts between 10237 and 10240 of 10240 updates).

This makes all benchmarks verifications run again. It's not great, but a false positive error state should not prevent a perfectly fine benchmark from running. (It should also be mentioned, that a successful verification run does in no way garuantee a correct benchmark implementation anyhow. It's just a hint that something might be off.)

### Long-term items

* Check whether this is just a timing issue. It is possible, that the database server might have queued up the update tracking variable increments, so it does not have to lock the actual update statement. If the queue has not been completed, while we already check the update tracking variable in the `postgres.py` and `mysql.py` scripts, we could get an inaccurate result back.
To check for a timing issue, we should wait and recheck the tracking variable again, if the result is not the expected one on the first try.
  * If the result is correct the second time, it is indeed a timing issue. Than TFB needs to accommodate for that, after which the previous workaround can be removed.
  * If the result did not change the second time, this is not a timing issue, and thereby very likely a counting bug in the database server in high concurrency situations (non-interlocked increments/unlocked variable math etc.).
    * The bug for Postgres needs to be reported. For [MariaDB](https://jira.mariadb.org/browse/MDEV-25598) and [MySQL](https://bugs.mysql.com/bug.php?id=103587), I already did that.
    * The workaround should stay in place, until the bug has been fixed in the database server implementation.

PR #6584, opened 2021-05-12
[Crystal] Optimize JSON responses

matthewmcgarvey:
`response.print(...)` is WAY faster than `.to_json(response)` for some reason. Local test on /json went from 80k per second to 220k per second.

### DIFF
--- a/frameworks/Crystal/crystal/server.cr
+++ b/frameworks/Crystal/crystal/server.cr
@@ -20,7 +20,8 @@ server = HTTP::Server.new do |context|
   when "/json"
     response.status_code = 200
     response.headers["Content-Type"] = CONTENT_JSON
-    {message: "Hello, World!"}.to_json(response)
+    json = {message: "Hello, World!"}.to_json
+    response.print(json)
   when "/plaintext"
     response.status_code = 200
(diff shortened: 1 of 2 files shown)


vlazar:
@matthewmcgarvey Would it actually make sense to wait till https://github.com/crystal-lang/crystal/pull/10716 gets merged in? I hope it would make into Crystal 1.1.0, which shouldn't be too far ahead.

I believe using `.to_json(response)` is a current recommended way for performant code I nCrystal and maybe it's better to keep benchmark code as it is and just update to Crystal version where this is fixed once it's released?

matthewmcgarvey:
@vlazar if the difference in performance really is due to the multiple calling of `flush`, then this change will still be faster since calling `io.print(...)` does not flush the io. That means this way will always be faster.

I've looked into most of the major frameworks and they all do it this way as well.

Mainly, this has made me wonder why `HTTP::Server::Response` is an `IO` at all.

vlazar:
OTOH `.to_json` creates new objects while `.to_json(response)` does not? I guess the performance impact could be different depending on the size of the serialized object.

matthewmcgarvey:
@nbrady-techempower This is ready to go 👍

PR #6586, opened 2021-05-14
Upgrade ffead-cpp to v6.0

sumeetchhetri:
Upgraded ffead-cpp from v5.3 to v6.0.


nbrady-techempower:
@sumeetchhetri I think I'm counting 17 tests here. Can you remove some? We just don't have enough disk space right now and it's causing a couple of tests at the end to not be able to run. I'm going to start trimming some tests myself as well.

sumeetchhetri:
My bad, now only 15 are enabled as per your suggestion. Done

PR #6598, opened 2021-05-21
PG Client driver implementation for OfficeFloor

sagenschneider:
Providing a PG client driver under OfficeFloor's HTTP server for faster database interaction.


sagenschneider:
Note: Vertx is in process of supporting multiple databases, so is now becoming SQL Client (rather than PG Client) 

Naming is to reflect this future direction

sagenschneider:
Also, swapping OfficeFloor-async to use Vertx SQL Client

PR #6610, opened 2021-05-29
Bump Finagle benchmarks

cacoco:
We've released the latest releases of twitter/finagle and twitter/finatra. We'd like to bump each to their latest revision.

- Finagle to 21.5.0
- Finatra to 21.5.0


dotordogh (approved):
Thank you!

PR #6613, opened 2021-06-01
aiohttp improvements

Dreamsorcerer:
I was just looking at the tests where aiohttp is slower than fastapi, and wanted to remove any differences in the test setup. So, these changes just get the code to match exactly what fastapi is doing.


Dreamsorcerer:
I'm a bit confused how this deadlock is suddenly occurring...

PR #6614, opened 2021-06-01
appMpower version 1.0

LLT21:
This implementation in .NET and Kestrel is natively compiled in a reflection-free way. As the .NET ODBC interface library is written in a reflection-free way, this interface is used to PostgreSQL. In these 2 respects, this framework implementation is unique.


LLT21 (commented):
This is the first commit of a new CSharp framework that is special in the following respect: 
1. It is natively compiled with the option to disable reflection: this generates the smallest executable
2. It uses the ODBC interface instead of the .NET ngpsql library (driven by the reflection free implementation). 

Can someone review and if ok approve this as a maintainer as this is a first time contribution ?

nbrady-techempower:
Letting CI run and will get around to a review in the next couple of days. 

@sebastienros  @benaadams don't know if this is something you guys are interested in looking at.

frameworks/CSharp/appmpower/src/HttpApplication.cs
```diff
@@ -0,0 +1,93 @@
...
+            string pathStringStart = pathString.Value.Substring(1, 1);
+
+            if (pathStringLength == 10 && pathStringStart == "p")
+            {
+               PlainText.Render(httpResponse.Headers, httpResponseBody.Writer, _plainText);
+               return;
+            }
+            else if (pathStringLength == 5 && pathStringStart == "j")
+            {
+               Json.RenderOne(httpResponse.Headers, httpResponseBody.Writer, new JsonMessage { message = "Hello, World!" }, new JsonMessageSerializer());
```
benaadams, line 40: `new JsonMessageSerializer()` looks idempotent and everything is passed via parameter; so could hoist this class to a readonly static rather than allocating each time?

frameworks/CSharp/appmpower/src/HttpApplication.cs
```diff
@@ -0,0 +1,93 @@
...
+               return;
+            }
+            else if (pathStringLength == 3 && pathStringStart == "d")
+            {
+               Json.RenderOne(httpResponse.Headers, httpResponseBody.Writer, await RawDb.LoadSingleQueryRow(), new WorldSerializer());
```
benaadams, line 45: `new WorldSerializer()` also looks idempotent and everything is passed via parameter; so could hoist this class to a readonly static rather than allocating each time?

frameworks/CSharp/appmpower/src/HttpApplication.cs
```diff
@@ -0,0 +1,93 @@
...
+               {
+                  count = 1;
+               }
+               else if (count > 500)
+               {
+                  count = 500;
+               }
+
+               //Json.RenderMany(httpResponse.Headers, httpResponseBody.Writer, await RawDb.LoadMultipleQueriesRows(count), new WorldSerializer());
+               Json.RenderMany(httpResponse.Headers, httpResponseBody.Writer, await RawDb.ReadMultipleRows(count), new WorldSerializer());
```
benaadams, line 62: As above

frameworks/CSharp/appmpower/src/HttpApplication.cs
```diff
@@ -0,0 +1,93 @@
...
+               if (!Int32.TryParse(request.QueryString.Substring(request.QueryString.LastIndexOf("=") + 1), out count) || count < 1)
+               {
+                  count = 1;
+               }
+               else if (count > 500)
+               {
+                  count = 500;
+               }
+
+               Json.RenderMany(httpResponse.Headers, httpResponseBody.Writer, await RawDb.LoadMultipleUpdatesRows(count), new WorldSerializer());
```
benaadams, line 83: As above

frameworks/CSharp/appmpower/src/appMpower.csproj
```diff
@@ -0,0 +1,27 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
```
benaadams, line 4: Can run `net6.0` instead of `net5.0`; just 'cos the "official" benchmarks are laggards on version doesn't mean you need to be 😎

frameworks/CSharp/appmpower/appmpower.dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM mcr.microsoft.com/dotnet/sdk:5.0 AS build
+RUN apt-get update
+RUN apt-get -yqq install clang zlib1g-dev libkrb5-dev libtinfo5
+
+WORKDIR /app
+COPY src .
+RUN dotnet publish -c Release -o out -r linux-x64
+
+FROM mcr.microsoft.com/dotnet/aspnet:5.0 AS runtime
```
benaadams, line 9: If you did switch to .NET 6.0; up to you, would need to also change this line
```suggestion
FROM mcr.microsoft.com/dotnet/aspnet:6.0 AS runtime
```

benaadams (approved):
Some performance suggestions, but looks good

frameworks/CSharp/appmpower/appmpower.dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM mcr.microsoft.com/dotnet/sdk:5.0 AS build
```
benaadams, line 1: Also this one :)
```suggestion
FROM mcr.microsoft.com/dotnet/sdk:6.0 AS build
```

roji:
@lauxjpn thanks for the ping. Yeah, we discussed in https://github.com/dotnet/aspnetcore/issues/31561#issuecomment-854074113 about using Npgsql. I've done some work to make it reflection-free etc, though I can't say for sure it would work in this scenario. We'll give it a try after this is merged etc.

frameworks/CSharp/appmpower/src/HttpApplication.cs
LLT21: Thanks for the review Ben. This recurring suggestion is a good one. As I am new on github since April and this is my first TechEmpower Benchmark pull request, can you just let me know how I best do the change ? Do I start locally in my appMpowerv1 branch, commit there, push to my forked repository and then do a new pull request to this repository ? Just to be sure: I assume your suggestion is like adding  
 
private readonly static WorldSerializer _worldSerializer = new WorldSerializer(); 

at the top of HttpApplication.cs and using that one ? 
Luc

frameworks/CSharp/appmpower/src/appMpower.csproj
LLT21: Follow your point, but would first like to compare apples with apples: my interest is to see how the odbc compares to the "default" npgsql implementations in the official benchmarks. As these are version 5, like to keep it like that as a start. Will afterwards soon move to 6. Another beginner's question: if I "Resolve conversation", does this mean I accept the proposed change, or does it just close this comment ?

frameworks/CSharp/appmpower/src/HttpApplication.cs
benaadams: > Just to be sure: I assume your suggestion is like adding
> 
> private readonly static WorldSerializer _worldSerializer = new WorldSerializer();

Yep

> Do I start locally in my appMpowerv1 branch, commit there, push to my forked repository and then do a new pull request to this repository ?

If the PR is still open (e.g. this one) you just need to push to same fork and it will automatically update, you don't need to do another pull request.

frameworks/CSharp/appmpower/src/appMpower.csproj
benaadams: > if I "Resolve conversation", does this mean I accept the proposed change, or does it just close this comment ?

Just closes comment

benaadams (approved):

Issue #6627, opened 2021-06-10
Disable PG log subsystem


NinoFloris:
Locally I've noticed a big difference (in the order of >25% faster) between the default `log_destination='stderr'` and `log_destination=''`.

I'd like to do a run where it's off to see if it has similar effect on a larger machine. 
This would help faster clients widen the gap if they have it in them.

With a destination (note the 19.8% errfinish):
<img width="516" alt="Screenshot 2021-06-10 at 17 08 59" src="https://user-images.githubusercontent.com/4218809/121550146-c5187780-ca0e-11eb-8ad6-f2204c3c1d38.png">

Without (3.7%):
<img width="544" alt="Screenshot 2021-06-10 at 17 08 42" src="https://user-images.githubusercontent.com/4218809/121550189-ce094900-ca0e-11eb-8dbd-415b2779d9bc.png">

Most likely there are other paths in portal run not taken as a result of this being off as well.

nbrady-techempower:
@NinoFloris Sorry about taking so long to look at this; been busy on our end. Discussing with the team now.

nbrady-techempower:
Closing because it didn't look like we were seeing any real positive results from this and some frameworks and alt environments were failing. Thanks again for the test!

PR #6628, opened 2021-06-10
Implements disabled log destination

NinoFloris:
See #6627

### DIFF
--- a/toolset/databases/postgres/postgresql.conf
+++ b/toolset/databases/postgres/postgresql.conf
@@ -294,7 +294,7 @@ max_wal_senders = 0		# max number of walsender processes
 
 # - Where to Log -
 
-#log_destination = 'stderr'		# Valid values are combinations of
+log_destination = ''		# Valid values are combinations of
 					# stderr, csvlog, syslog, and eventlog,
 					# depending on platform.  csvlog
 					# requires logging_collector to be on.


nbrady-techempower:
I think we can give this a shot. A new run will be starting shortly with this change and can be compared to https://tfb-status.techempower.com/results/67c8e670-dab3-44f9-94c3-21a80074afc9

sebastienros:
This is breaking the ASP.NET perf lab. Not saying it's wrong but didn't expect that. What's happening is that now the postgresql log is only comprised of these lines

```
[STDERR] 2021-06-16 19:00:15 GMT LOG:  ending log output to stderr
[STDERR] 2021-06-16 19:00:15 GMT HINT:  Future log output will go to log destination "".
```

And we are not detecting that the server is ready to accept connections.

I might still be able to change the setting on my side, or check for this message instead.

nbrady-techempower:
@sebastienros Happy to revert after the next run finishes if you need more time to figure it out. Let me know. Still have to confirm that this is a net positive change.

sebastienros:
I am not blocked, had to change what message we look for to detect if postgres is ready.

NinoFloris:
@sebastienros I've asked to get a run to see what the effects are, I think given the amount of failing tests I noticed there are many frameworks that don't really support running their benchmarks with this disabled (yet). Really just curious what the outcome is. Like I said locally I saw *a lot* of effect, if that replicates I might start writing postgres patches to reduce the general cost of having any log destinations.

NinoFloris:
That being said, I don't see a lot of difference in the running results yet https://www.techempower.com/benchmarks/#section=test&runid=5409dd1a-7afa-4be6-a81e-8160b38cfce2 specifically actix would have absolutely shown better numbers if this had any effect so if you want @nbrady-techempower you could revert already.

nbrady-techempower:
@NinoFloris @sebastienros ok let's revert. I appreciate the effort here!

PR #6668, opened 2021-06-21
Implement caching test and refactor whole Hexagon benchmark

jaguililla:


nbrady-techempower:
Hey @jaguililla! Looks good! If you could just remove the unit test code, this should be good to go.

jaguililla:
Hi @nbrady-techempower, I just removed the tests as the other implementations, that's for pointing that out!

PR #6675, opened 2021-06-23
Adding Jam.py

platipusica:
Hello, 

I created simple Jam.py App which is using Mysql hello_world database. The App will display all Fortunes. The DB username/password/DB details are on http://127.0.0.1/builder.html under Database option. This information is stored in admin.sqlite DB within /app folder. Jam.py is different to some other Python frameworks since everything is JS, except Jam.py core.
Hope this helps.

Cheers


nbrady-techempower:
I'm sorry @platipusica but I'm a little confused about what I'm looking at here. Where is the code that fetches the fortunes and adds the additional fortune at request time? The test implementation code needs to exist in this repo. Also, we don't want anything unrelated to the tests here, like the html and css that's unrelated to benchmarking.

platipusica:
Hi Nate,
Jam.py is a portable app. It has no means to code the SQL fetches or updates in flat files. Hence I added one simple test, which is to show ALL fortunes, which Jam displays when is accessed on "/". As I see it Flask does the same when accessed with "/fortunes". However, this is the html output, Jam does JS. 

What you see is the full App with the templates and everything which makes an App as realistic as it is in here: https://demo.jam-py.com
I'm not quite sure if anything can be removed. If it can't be tested, well, we tried our best. Pls let me know.
Cheers

nbrady-techempower:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Python/flask/app.py#L130

Flask is doing work when you hit the fortunes endpoint to add the additional fortune and we need to see that work. I'm not sure what the 79kb sqlite file is. Or why any css files are present. We're not looking for a realistic demo, but a realistic benchmark. So any code that isn't needed for the benchmark shouldn't be present here. If we can't see the work that it's doing to produce the results, including the sorting and the adding of the additional fortune at request time, then it's probably not a good fit for this project.

platipusica:
Jam has no routing as Flask/Django, so "/fortunes" or any other can't be routed. I guess this is why the Test is failing:
```
FAIL for http://tfb-server:8080/
     Route for fortunes must be at least 9 characters, found '/' instead
```
Would that be the issue with **The recommended URI is /fortunes.** from the Project-Information-Framework-Tests-Overview? If that is true than it is not "recommended" but "required", so Jam can't fulfil this requirement, I'm afraid.

You can't see the work because that is not how Jam engine works.  Everything is added via Admin interface, similar to Django's Admin. To access this it is simple as http://tfb-server:8080/builder.html
The sqlite stores the table definitions from mysql DB and any other metadata by the design. Sorting is a feature of Jam enabled in builder.html for the "message" table field, which again goes into sqlite file. So, there is simply no files to see how sql is constructed. Like in ie Flask 

If "recommended URI is /fortunes" can be solved somehow I might think of how to add additional fortune. 
But if can't be done, it cant be done. 

No probs.

nbrady-techempower:
The route does have to be at least 9 characters long like `/fortunes`. If you can get this passing CI, I can further investigate how to do a review of the implementation.

platipusica:
Thanks.
I think the App can be mapped to /fortunes with Nginx. However, I am still puzzled with how the test will be performed since everything in Jam is JS? If this IS possible to test, than sure, I can spend time for Nginx, uwsgi and Docker.
Like in Jam.py Demo from above link https://demo.jam-py.com/ this is all JS. I just cant find any info about testing JS Apps.
Cheers

nbrady-techempower:
@platipusica Yeah, I don't think we can do this the way it needs to be done for this framework. Just doesn't seem like a good fit. Sorry :(

platipusica:
No worries, 
in the next major release I think Jam will be able to fulfill full html requirement. Hoping for ver6 soon.
Closing.

PR #6681, opened 2021-06-26
fix: ocaml webmachine & bump to 4.12

rbjorklin:
This PR fixes the OCaml webmachine build and updates to OCaml 4.12.

I might submit more PRs to align the Dockerfiles across all OCaml frameworks to reduce build times. Depends on time and motivation :sweat_smile:


nbrady-techempower:
> Depends on time and motivation 😅

I hear ya! Thanks for doing that. Looks like some dependency issues in here atm.

PR #6685, opened 2021-06-28
Update Service.java

redkale:


frameworks/Java/redkale/src/main/java/org/redkalex/benchmark/Service.java
```diff
@@ -86,9 +86,9 @@ public Message getHelloMessage() {
         }
         final int size = Math.min(500, Math.max(1, q));
         final CachedWorld[] worlds = new CachedWorld[size];
-        final ThreadLocalRandom random = ThreadLocalRandom.current();
+        final int random = ThreadLocalRandom.current().nextInt(10000-size);
         for (int i = 0; i < size; i++) {
-            worlds[i] = cache.findAt(random.nextInt(10000));
+            worlds[i] = cache.findAt(i+random);
```
nbrady-techempower, line 91: No. Each number has to be random. Not one random number incremented.

PR #6690, opened 2021-07-04
add xitca-web

fakeshadow:
An experimental http library written in Rust.

### DIFF
--- a/frameworks/Rust/xitca-web/Cargo.toml
+++ b/frameworks/Rust/xitca-web/Cargo.toml
@@ -1,0 +1,31 @@
+[package]
+name = "xitca-web"
+version = "0.1.0"
+edition = "2018"
+
+[dependencies]
+xitca-http = { git = "https://github.com/fakeshadow/xitca-web.git", branch = "tfb" }
+xitca-web = { git = "https://github.com/fakeshadow/xitca-web.git", branch = "tfb" }
+
+ahash = { version = "0.7.4", features = ["compile-time-rng"] }
+atoi = "0.4.0"
+bytes = "1"
+futures-util = { version = "0.3", default-features = false, features = ["alloc"] }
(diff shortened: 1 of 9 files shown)


frameworks/Rust/xitca-web/src/db.rs
```diff
@@ -0,0 +1,146 @@
...
+        for _ in 1..=num {
+            let _ = write!(&mut q, "${},", pl);
+            pl += 1;
+        }
+        q.pop();
+        q.push(')');
+
+        let st = client.prepare(&q).await.unwrap();
+        updates.insert(num, st);
+    }
```
nbrady-techempower, line 47: Is this really creating 500 prepared update statements?

fakeshadow: Yes I believe it does prepare all possible update query and store them in client object. I copied this piece of code from actix bench and made slight change to make it compile.

nbrady-techempower: Ah, now I see that xitca is actix backwards...

frameworks/Rust/xitca-web/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "xitca-web"
+version = "0.1.0"
+edition = "2018"
+
+[dependencies]
+xitca-http = { git = "https://github.com/fakeshadow/xitca-web.git", branch = "tfb" }
```
nbrady-techempower, line 7: Can you use the `rev` property here to lock all these down to specific commits? We want to make sure these aren't being updated without PRs here. Other than that, this looks ready to go.

fakeshadow: Done.

PR #6697, opened 2021-07-10
Add spliffy framework

snowbldr:
This adds the spliffy framework. 

Spliffy provides directory based routing and middleware, and is compatible with some express middleware.


snowbldr:
Ran out of connections on the last build. 

Changed to query max_connections from database and only use 90% of the max.

snowbldr:
" WARN for http://tfb-server:1420/updates?queries=20
     20257 rows read in the database instead of 10240 expected. This number is excessively high."

This is due to using a "SELECT FOR UPDATE" in order to prevent deadlocks.

PR #6714, opened 2021-07-21
adding benchmarks for Nim stdlib

rishavs:
Hi

I am adding support for the core stdlib webserver implementation in the Nim Language. 
Nim has a robust stdlib which contains all the tools and methods required to make simple web apps.

This is to gauge the perf of the stdlib web server itself. Also, I intend to add all the benchmarks (which are currently missing for other Nim frameworks) so that we can calculate the combined summary score for Nim.


rishavs:
Hi this is ready to be merged. Can you please check up on this?

Only the caches test is remaining that I am thinking of skipping due to personal bandwidth issues. Will pick it up in the future.
Other major future work is to optimize the code here as it is very naively coded without considering possible perf issues.

rishavs:
Checking back if there is anything needed from my side to get this picked up. I am not sure what does  the label "PR: Please Update" stands for

frameworks/Nim/nim-stdlib/src/handlers.nim
```diff
@@ -0,0 +1,106 @@
...
+    let headers = {"Date": now().utc.format("ddd, dd MMM yyyy HH:mm:ss") & " GMT", "Content-type": "text/plain; charset=utf-8"}
+    await req.respond(code, msg, headers.newHttpHeaders())
+
+proc handlePlaintext*(req: Request) {.async.} =
+    let headers = {"Date": now().utc.format("ddd, dd MMM yyyy HH:mm:ss") & " GMT", "Content-type": "text/plain; charset=utf-8",  "Server": "Example"}
+    await req.respond(Http200, "Hello, World!", headers.newHttpHeaders())
+
+proc handleJson*(req: Request) {.async.} =
+    let headers = {"Date": now().utc.format("ddd, dd MMM yyyy HH:mm:ss") & " GMT", "Content-type": "application/json; charset=utf-8", "Server": "Example"}
+    await req.respond(Http200, """{"message":"Hello, World!"}""", headers.newHttpHeaders())
```
nbrady-techempower, line 15: This isn't acceptable for the JSON test requirements. An object has to be created and then serialized

rishavs: @nbrady-techempower thanks. I have fixed all instances of this issue. now all json objects are created using Nim's serializer.

rishavs: To help with the review, Nim uses "%*" as the operator for creating json. More here
https://nim-lang.org/docs/json.html#creating-json

dom96:
This is cool, but it uses `db_postgres` which is not an async library. You will get really poor performance that is not representative of Nim here.

rishavs:
Thanks for the heads up @dom96 .
I think after a few rounds of reports, I will update the db drivers to async ones. Which would be the best postgres driver for nim in your opinion?

dom96:
I'm afraid I'm not aware of any. Maybe some popped up that I haven't noticed, but AFAIK there aren't any truly efficient drivers for postgres in Nim that support async. Maybe you can write the first one? :)

PR #6726, opened 2021-07-24
Fix xitca-web test

fakeshadow:
Fix build failure caused by nightly Rust regression.


fakeshadow:
Sorry there is some overlook in the PR. Would temporary close it and re-open after further fix.

PR #6731, opened 2021-07-27
use db.getSiblingDB in mongo create.js

snowbldr:
When running mongodb.dockerfile locally on windows, I noticed that it was creating a database named test instead of hello_world which caused all of the validations to fail when run.

I changed the create.js script to use db.getSiblingDb("hello_world") instead of use hello_world and that resolved the issue. This should work identically in all cases where it was already working.

https://docs.mongodb.com/manual/reference/method/db.getSiblingDB/


snowbldr:
I don't think this will ever pass CI...

PR #6740, opened 2021-07-30
Cache use RedRandom

redkale:

### DIFF
--- a/frameworks/Java/redkale/src/main/java/org/redkalex/benchmark/Service.java
+++ b/frameworks/Java/redkale/src/main/java/org/redkalex/benchmark/Service.java
@@ -90,7 +90,7 @@ public class Service extends AbstractService {
             }
         }
         final int size = Math.min(500, Math.max(1, q));
-        return cache.random(cached ? rands.get() : ThreadLocalRandom.current(), size);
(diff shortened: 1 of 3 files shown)


frameworks/Java/redkale/src/main/java/org/redkalex/benchmark/Service.java
```diff
@@ -88,7 +90,7 @@ public Message getHelloMessage() {
             }
         }
         final int size = Math.min(500, Math.max(1, q));
-        return cache.random(ThreadLocalRandom.current(), size);
+        return cache.random(cached ? rands.get() : ThreadLocalRandom.current(), size);
```
nbrady-techempower, line 93: Random numbers should never be cached? What is this doing?

youland-mason: /cached-worlds  use RedRandom.class   other use ThreadLocalRandom
look at the performance difference between the two

redkale: /cached-worlds use RedRandom.class other use ThreadLocalRandom
look at the performance difference between the two

PR #6752, opened 2021-08-09
cutelyst: Update to v3.1 and Cutelee v6

dantti:


dantti:
Interesting github-actions bot is not so smart to remove the "PR: Please Update" after I fixed it :)

nbrady-techempower:
@dantti It does do that if we want it to, but I didn't want it to. I wanted to make sure I double-checked PR's that passed but also had the "Please Update" tag. Also, I add that tag manually to some PR's that are passing. Probably just need a tag specifically for CI.

PR #6754, opened 2021-08-10
Update php to Ubuntu 21.04

joanhey:
Ubuntu 20.10 EOL

And actually Odrej  automated PHP builds, don't support EOL versions.
The last 3 runs in Citrine failed almost all PHP frameworks for that.

We can use the 20.04 LTS.
But we will test the new kernel 5.11 in Ubuntu 21.04, with the FSGSBASE support to improve context switch performance on x86 processors.
That theoretically will help Intel/AMD CPU performance that had been hurt badly by Spectre/Meltdown and other CPU vulnerability mitigations largely on the Intel side.

### DIFF
--- a/frameworks/PHP/amp/amp.dockerfile
+++ b/frameworks/PHP/amp/amp.dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:20.10
+FROM ubuntu:21.04
 
 ARG DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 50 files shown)


joanhey:
@nbrady-techempower this PR don't need to pass the tests, as it don't touch any code, only update the Ubuntu version.
Locally it's working, and actually all these frameworks are failing in the continuous runs in Citrine.

PR #6774, opened 2021-08-20
Align naming of tests with asp.net core + create less MySQL pooled connections

LLT21:
One question: what should be added for appearing in the composite scores ?


nbrady-techempower:
> One question: what should be added for appearing in the composite scores ?

I believe just one of your test permutations needs to implement all the benchmark tests.

PR #6777, opened 2021-08-21
Include Fortunes/Single/Multiple/Updates benchmarks in CoreRT

MichalStrehovsky:
Matches the platform benchmarks for vanilla Kestrel.


MichalStrehovsky:
Looks like it needs some more RD.XML work for .NET 5. Gonna wait for .NET 6. I know it works with that.

PR #6836, opened 2021-09-30
Swift/kitura adding broken tags

joanhey:
It's failing for a lot of months, and it have all 13 variants failing


nbrady-techempower:
@djones6 FYI

PR #6838, opened 2021-09-30
Hummingbird: Fortunes optimisation, plus minor DB optimisations

adam-fowler:
Don't rely on `Mirror` to extract data out of `Fortune`


adam-fowler:
This doesn't seem to be an issue with my changes. It is crashing in the middle of the Postgres setup

Issue #6841, opened 2021-10-01
Docker image `techempower/postgres:latest` build is failing 


adam-fowler:
### OS (Please include kernel version)
macOS 11.4. This is also being seen on any new PR that requires a test of postgres. See #6838 and #6840

### Expected Behavior
Docker image `techempower/postgres:latest` builds

### Actual Behavior

techempower/postgres:latest fails to build. It generates the following error
```
techempower/postgres:latest: Step 19/35 : RUN mv postgresql.conf /etc/postgresql/${PG_VERSION}/main/postgresql.conf
techempower/postgres:latest:  ---> Running in 6dfeb5297267
techempower/postgres:latest: mv: cannot move 'postgresql.conf' to '/etc/postgresql/13/main/postgresql.conf': No such file or directory
techempower/postgres:latest: Removing intermediate container 6dfeb5297267
techempower/postgres:latest: Docker build failed; terminating
```
### Steps to reproduce behavior
- Ensure you don't have the `techempower/postgres:latest` image installed
- Run `./tfb --mode verify --test hummingbird-postgres`

nbrady-techempower:
This is resolved atm, but want to look at upgrading to 14. thanks!

PR #6842, opened 2021-10-01
Fix PostgreSQL to v13

jcheron:
[PostgreSQL Version 14 is released...](https://www.postgresql.org/about/news/postgresql-14-released-2318/) which explains this problem: https://github.com/TechEmpower/FrameworkBenchmarks/issues/6841
I fixed the docker file on v13, not knowing if we're moving to v14.

### DIFF
--- a/toolset/databases/postgres/postgres.dockerfile
+++ b/toolset/databases/postgres/postgres.dockerfile
@@ -22,7 +22,7 @@ ENV LC_ALL en_US.UTF-8
 ENV DEBIAN_FRONTEND noninteractive
 
 # install postgresql on database machine
-RUN apt-get -yqq install -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" postgresql postgresql-contrib
+RUN apt-get -yqq install -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" postgresql-${PG_VERSION} postgresql-contrib-${PG_VERSION}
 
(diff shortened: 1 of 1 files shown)


nbrady-techempower:
Thanks for the quick fix. I think we would be fine moving to 14 as well, as I'm not sure when we'll be ready for the next round.

sebastienros:
This is not sufficient for me, we had the same issue with 13, the conf file needs updating too, see what was done in https://github.com/TechEmpower/FrameworkBenchmarks/pull/6051

sebastienros:
This is the error I get:

```
Step 18/35 : RUN apt-get -yqq install -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" postgresql postgresql-contrib
 ---> Using cache
 ---> 83afd5377976
Step 19/35 : RUN mv postgresql.conf /etc/postgresql/${PG_VERSION}/main/postgresql.conf
 ---> Running in 4ecaa5ca3877
[91mmv: cannot move 'postgresql.conf' to '/etc/postgresql/13/main/postgresql.conf': No such file or directory
The command '/bin/sh -c mv postgresql.conf /etc/postgresql/${PG_VERSION}/main/postgresql.conf' returned a non-zero code: 1
[0m
```

sebastienros:
Unless this error was before the fix 😅

sebastienros:
@jcheron all good, confirm it works now, should have checked the timestamps before commenting  ...

PR #6844, opened 2021-10-01
Upgrade to .NET 6.0-rc1 and PostgreSQL 14.0 + implementation of cached-worlds test in a reflection free way

LLT21:


LifeIsStrange:
I'm curious do the .NET pgconnection use batch/bulk queries?
I am interested in performance technological progress regarding SOTA performance but all the .NET 6 optimizations are moot as long as it does not use batched queries. This is the ugly secret of the techempowerup benchmark and being on the top 5-10 fastest framework necessitate batched queries.
lithium-postgres-batch vs lithium-postgres illustrate well the difference (congrats to the author for the intellectual honesty, even if it's a real **shame** that the author of the techempowerup benchmark do not enforce entrie to name wether they use batch mode or not, leading most readers to delusions..)

msmith-techempower:
@LifeIsStrange Thanks for the suggestion of adding some metadata to the config files to discern when a framework is utilizing the driver in this way. I started a rewrite of the Toolset around a year ago, but paying work has largely sidetracked me since February - your recommendation is definitely worth adding as a note to that when I can jump back into development on it.

As for the pipelining/batching, we discussed this a while back and decided that it was fine/fair. Essentially, since it's the driver doing the optimization and the developer gets it more or less for free, then it isn't a case of trying to cheat. Rather, and I feel that this is an important part of why these benchmarks are useful, it shows that deciding to use a particular database/driver can have real performance consequences, and thanks to maintainers of test implementations like `lithium`, prospective decision-makers can see that impact quantified.

Again, you have a totally valid point in that we should capture this distinction in metadata and allow the ranking website to filter on it. 

Lastly, I'm not sure where the confusion came from, but the company name is "TechEmpower" (short for Technical Empowerment).

LLT21:
@LifeIsStrange
Most of the other C#/.NET implementations which I have read on TechEmpower benchmarks do use pipelining; this appmpower implementation doesn't because of a special reason: I wanted to test native compilation without reflection, to have the best source code protection possible with C#/.NET; there are other implementations like core-rt which are also nativily compiled, but still use reflection (the interfaces are public). The reason is that the .NET pgconnection needs reflection to work. What I found is that the .NET odbc drivers do not require reflection. So as a consequence I used the odbc interfaces and not the native .NET pg library. And as I am using odbc, as far as I know that does not support pipelining/batching. I am certainly not using it. 
@msmith-techempower
What I do wonder with this pipeling/batching is whether if user A does a request he receives the correct response for his question; it seems that the TechEmpower benchmark tests do test whether all requested responses are correctly returned, but I am not sure whether it tests if an individual response is correctly matched to a specific request. If this is so, I see no issue with pipelining, but if this functional requirement is not met, then the test is not what most would expect. 

I might today remove this pull request because I have been "offline" for some 2 months and in the meantime .NET 6 has been released, so it might be easier to fix my issue in a new pull request. Do not know what will happen to these comments in that case.

LLT21:
@msmith-techempower
@nbrady-techempower 
Sorry for the very late reply but work was super busy the last 2 months. I have ran the tests again today without changing any code and these ran fine. If I still need to update something, can you give a hint what you want to be changed ? I think beginning of October it failed once because of a PostgreSQL upgrade issue, but in the meantime that seems to be resolved.

LifeIsStrange:
@msmith-techempower
Hi, I just wanted to say sorry since my last message was a bit rude and you gave me a great answer. So thank you for making this great benchmark. I was wondering wether it was still on your roadmap to mandate to specify batching use in the name as a suffix, to make it clear for end users.

PR #6877, opened 2021-11-01
Rocket async

### LINKED ISSUE #6876: Rocket async
Current implementation for the Rust rocket framework uses version 0.4, a non async version. Version 0.5 (rc1) introduces support for async. As well the database layer has to change from diesel, a non async framework to a an async one. I suggest sqlx, a well supported async library.

dragosv:
Rust rocket async using sqlx (postgresql)

This fixes #6876


frameworks/Rust/rocket/rocket.dockerfile
```diff
@@ -1,18 +1,18 @@
-FROM rust:1.46.0-slim-buster
+FROM rust:1.55-slim-buster
 
 ENV DATABASE_URL=postgres://benchmarkdbuser:benchmarkdbpass@tfb-database/hello_world
```
fakeshadow, line 3: You may want to change this variable to match the one you use in main.rs

dragosv: Fixed

PR #6890, opened 2021-11-09
[genhttp] Switch to .NET 6

Kaliumhexacyanoferrat:

### DIFF
--- a/frameworks/CSharp/genhttp/Benchmarks/Model/DatabaseContext.cs
+++ b/frameworks/CSharp/genhttp/Benchmarks/Model/DatabaseContext.cs
@@ -25,7 +25,7 @@ namespace Benchmarks.Model
         {
             var optionsBuilder = new DbContextOptionsBuilder<DatabaseContext>();
 
-            optionsBuilder.UseNpgsql("Server=tfb-database;Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=18;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4;Multiplexing=true;Write Coalescing Buffer Threshold Bytes=1000");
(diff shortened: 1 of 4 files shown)


frameworks/CSharp/genhttp/Benchmarks/Model/DatabaseContext.cs
```diff
@@ -25,7 +25,7 @@ private static DbContextOptions<DatabaseContext> GetOptions(bool tracking)
         {
             var optionsBuilder = new DbContextOptionsBuilder<DatabaseContext>();
 
-            optionsBuilder.UseNpgsql("Server=tfb-database;Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=64;NoResetOnClose=true;Enlist=false;Max Auto Prepare=3");
+            optionsBuilder.UseNpgsql("Server=tfb-database;Database=hello_world;User Id=benchmarkdbuser;Password=benchmarkdbpass;Maximum Pool Size=18;NoResetOnClose=true;Enlist=false;Max Auto Prepare=4;Multiplexing=true;Write Coalescing Buffer Threshold Bytes=1000");
```
ronnygunawan, line 28: Other frameworks set this to 256 or 512. Are you sure setting it to 18 improves the score?

Kaliumhexacyanoferrat: I actually just copied this from the ASP.NET implementation. In my opinion, almost all implementations here are full of unrealistic optimizations that do not really correspond to the idea of realistic performance determination. I have deliberately avoided such optimizations, here I was just lazy :)

Kaliumhexacyanoferrat: Removed this change for now as it does not relate to the version update anyways.

PR #6891, opened 2021-11-09
Update giraffe sample to .NET 6 alpha

forki:
This updates the Giraffe sample to .NET 6 as soon as Giraffe 6 comes out

### DIFF
--- a/frameworks/FSharp/giraffe/src/App/App.fsproj
+++ b/frameworks/FSharp/giraffe/src/App/App.fsproj
@@ -1,15 +1,15 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
     <EnableDefaultContentItems>false</EnableDefaultContentItems>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Update="FSharp.Core" Version="6.0.0" />
-    <PackageReference Include="Dapper" Version="2.0.90" />
+    <PackageReference Update="FSharp.Core" Version="6.0.1" />
+    <PackageReference Include="Dapper" Version="2.0.123" />
     <PackageReference Include="Giraffe" Version="5.0.0" />
-    <PackageReference Include="Npgsql" Version="5.0.7" />
+    <PackageReference Include="Npgsql" Version="5.0.10" />
   </ItemGroup>
 
   <ItemGroup>
--- a/frameworks/FSharp/giraffe/src/App/App.fsproj
+++ b/frameworks/FSharp/giraffe/src/App/App.fsproj
@@ -8,7 +8,7 @@
   <ItemGroup>
     <PackageReference Update="FSharp.Core" Version="6.0.1" />
     <PackageReference Include="Dapper" Version="2.0.123" />
-    <PackageReference Include="Giraffe" Version="5.0.0" />
+    <PackageReference Include="Giraffe" Version="6.0.0-alpha-1" />
     <PackageReference Include="Npgsql" Version="5.0.10" />
   </ItemGroup>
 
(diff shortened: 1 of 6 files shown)


frameworks/FSharp/giraffe/src/App/App.fsproj
```diff
@@ -1,15 +1,15 @@
...
 
   <ItemGroup>
-    <PackageReference Update="FSharp.Core" Version="6.0.0" />
-    <PackageReference Include="Dapper" Version="2.0.90" />
-    <PackageReference Include="Giraffe" Version="5.0.0" />
-    <PackageReference Include="Npgsql" Version="5.0.7" />
+    <PackageReference Update="FSharp.Core" Version="6.0.1" />
+    <PackageReference Include="Dapper" Version="2.0.123" />
+    <PackageReference Include="Giraffe" Version="6.0.0-alpha-1" />
+    <PackageReference Include="Npgsql" Version="5.0.10" />
```
slang25, line 12: We can go to version `6.0.0` now for `Npgsql`

forki: thanks

forki:
this one is ready

frameworks/FSharp/giraffe/src/App/Program.fs
```diff
@@ -128,20 +130,25 @@ module Main =
...
-            .ConfigureServices(
-                fun services ->
-                    services
-                        .AddRouting()
-                        .AddGiraffe()
-                        .AddSingleton(jsonSerializer)
-                    |> ignore)
-            .Build()
-            .Run()
+        let configureApp (appBuilder : IApplicationBuilder) =
```
dsyme, line 133: Glancing at this, is there anything here that could cause a performance change? UseKestrel being dropped? Anything else?

forki: AFAIK that's how hosting is initialized in asp.net core 6. /cc @davidfowl

davidfowl: This could be a bit more minimal overall:

```F#
let builder = WebApplication.CreateBuilder(args)

builder.Services.AddGiraffe()
                .AddSingleton(jsonSerializer) |> ignore

let app = builder.Build()

app.UseGiraffe HttpHandlers.endpoints |> ignore

app.Run()
```

forki: There is another PR coming that disables logging. Could this be a change to asp.net core 5 that affected performance? Was logging enabled before?

forki: And thanks for the help!

slang25: @davidfowl for the avoidance of doubt, is that because these things are all registered by default:
https://github.com/dotnet/aspnetcore/blob/main/src/DefaultBuilder/src/WebApplicationBuilder.cs#L238-L299

NinoFloris: @davidfowl AddRouting could indeed go but removing UseRouting errors with:

```
Unhandled exception. System.InvalidOperationException: EndpointRoutingMiddleware matches endpoints setup by EndpointMiddleware and so must be added to the request execution pipeline before EndpointMiddleware. Please add EndpointRoutingMiddleware by calling 'IApplicationBuilder.UseRouting' inside the call to 'Configure(...)' in the application startup code.
```

Are we expected to add this inside the UseGiraffe implementation?

davidfowl: Change this to MapGiraffeEndpoints instead of UseGiraffe

cc @dustinmoris

dustinmoris: When the `EndpointRouting` namespace is open then the `UseGiraffe` extension method is [doing the correct thing](https://github.com/giraffe-fsharp/Giraffe/blob/master/src/Giraffe/EndpointRouting.fs#L366-L376) already.

PR #6893, opened 2021-11-09
Routing

abrowao:
Added Routing to tests.


frameworks/Julia/Jewelia/julia_server.jl
```diff
@@ -14,25 +14,31 @@ end
...
+    
+        return HTTP.Response(200, headers, body = "Hello, World!")
+    end
+    
+    function jsonSerialization(req::HTTP.Request)
+        headers = [ "Content-Type" => "application/json",
+                    "Server" => "Julia-HTTP",
+                    "Date" => Dates.format(Dates.now(), Dates.RFC1123Format) * " GMT" ]
+        
+        json_string = "{\"Message\": \"Hello, World!\"}"
```
nbrady-techempower, line 30: In this controller, you have to instantiate an object that then gets serialized.

Issue #6894, opened 2021-11-10
PHP 8.1 update [info]


joanhey:
## PHP 8.1
As the `ppa:ondrej/php` include now `php8.1-rc5`, I'm starting to test the frameworks for the benchmark.
PHP 8.1 final will be released 25th of this month.
https://www.php.net/releases/8.1/en.php

If you want to view the php8.0 update info, check #6184, that is closed now.

This list will be updated as we found problems or solutions.

Opened #7703 for php8.2
## Need to fix

- [x] **php-ngx** ~don't compile  https://github.com/rryqszq4/ngx_php7/issues/133~
- [x]  **webman**  ~https://github.com/walkor/webman/issues/249 `exit with status 139`~
-  [x] **comet** ~https://github.com/gotzmann/comet/issues/26 `Deprecated: str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /comet/vendor/workerman/workerman/Protocols/Http/Request.php on line 166`
  And various `exit with status 139`~
-  [x] **fat-free**  ~https://github.com/bcosca/fatfree/issues/1241 `PHP Fatal error:  $GLOBALS can only be modified using the $GLOBALS[$name] = $value syntax in /fat-free/vendor/bcosca/fatfree-core/base.php on line 2533`~
- [x] ~**nginx-unit** we will  update with php 8.0, when ready to php 8.1 final we will do the same~
   #7387
  But with errors:
  https://tfb-status.techempower.com/unzip/results.2022-08-21-11-27-27-592.zip/results/20220816103420/php-unit/run/php-unit.log
- [ ] **fuel** Fail without any error
- [x] **slim** ~`PHP Fatal error:  Uncaught TypeError: Slim\App::__construct(): Argument #1 ($responseFactory) must be of type Psr\Http\Message\ResponseFactoryInterface, array given, called in /slim/index.php on line 7 and defined in /slim/vendor/slim/slim/Slim/App.php:61`~
  To work with PHP 8.1 is necessary update the code to v4  https://github.com/slimphp/Slim/pull/3186
- [ ] **phpixie**  `PHP Fatal error:  During inheritance of Iterator: Uncaught ErrorException: Return type of PHPixie\DB\Result::current() should either be compatible with Iterator::current(): mixed, or ...`
[...]

PR #6908, opened 2021-11-11
[laravel] Update to v8.70 and php 8.1

joanhey:
Only `fpm/fastcgi` #6894


joanhey:
Updated dependencies in laravel-roadrunner.
Now pass the tests, but with errors in roadrunner.
Is still using php 8.0 this variant.

PR #6917, opened 2021-11-14
Updated ffead-cpp to latest master

sumeetchhetri:
Smaller names for test benchmark applications resulting in smaller context path urls
Updated ffead-cpp to latest master


sumeetchhetri:
Hello @nbrady-techempower, requesting to please review and merge accordingly if possible (before the next test run would be really awesome). The changes are majorly related to changes in the long application/context paths for all the different framework modes

sumeetchhetri:
Thanks @nbrady-techempower, the latest results are really exciting for me, had spent lots of time over optimizing code with minimal improvements, all this while the framework was running without proper application of prepared statements with libpq. That single change along with another major event handler change just changed everything 👍

PR #6923, opened 2021-11-15
Enable Giraffe .NET 6 optimizations

slang25:
This enables some optimizations that are available in .NET 6 now that we have updated.


forki:
cool!

PR #6949, opened 2021-11-28
Added openswoole

doubaokun:


doubaokun:
@nbrady-techempower Could you please merge this PR?

nbrady-techempower:
@doubaokun I will review shortly, but not on the weekends ;)

joanhey:
From what I know, Swoole is now OpenSwoole. 
So it's the same. 
I'll be better to change the name in the Swoole, the code is exactly the same.
But remember that if you change the name, Swoole will disappear from the history. 
Well, It will be better to change only the `display_name` in `benchmark_config.json` and the `Readme`, so the history will be preserved.

doubaokun:
@joanhey I few devs from the community still like to develop on the old Swoole and we like to leave it as is, so please merge this PR. Thanks.

PR #6966, opened 2021-12-08
Updated FaF framework dependencies and made a few performance tweaks

errantmind:


errantmind:
Hello @nbrady-techempower, any chance this can make it in before the next test cycle (~24h from now)? I'm curious about some of my performance improvements.

frameworks/Rust/faf/src/main.rs
```diff
@@ -35,55 +30,43 @@ const PLAINTEXT_BASE: &[u8] = const_concat_bytes!(
...
-   _headers: &[faf::phr_header; MAX_HEADERS_TO_PARSE],
-   _num_headers: usize,
    response_buffer: *mut u8,
+   date_buff: *const u8,
 ) -> usize {
    unsafe {
       if likely(method_len >= GET_LEN && path_len >= ROUTE_PLAINTEXT_LEN) {
-         if likely(memcmp(GET.as_ptr() as *const i8, method, GET_LEN) == 0) {
+         if likely(memcmp(GET.as_ptr(), method, GET_LEN) == 0) {
             // For performance purposes, this will successfully match '/p' to '/plaintext' and '/pickle'. Use with caution
```
nbrady-techempower, line 48: I've missed this before, and I'm not sure how many other people are doing it, but I don't like this. I want it to only match `/plaintext` but we're not enforcing that.

errantmind: I updated the code to perform a 'full' comparison. It should be good to go now

errantmind:
Thanks for getting this merged before the next test cycle :+1:

Issue #6967, opened 2021-12-08
Verify routing matches exact path


nbrady-techempower:
In https://github.com/TechEmpower/FrameworkBenchmarks/pull/6966 I noticed a comment above the `/plaintext` route is:

`// For performance purposes, this will successfully match '/p' to '/plaintext' and '/pickle'. Use with caution`

We're requiring that the requests be made to `/plaintext` but we're not verifying in the toolset whether `/p` -> `/plaintex` returns `Hello, World!`. I'd like to add this verification to the new toolset @msmith-techempower

joanhey:
Will be good to verify the process priority too.

errantmind:
> Will be good to verify the process priority too. `ps al `

Is there a problem with running at a higher priority? This is nice for scheduling optimization

joanhey:
The problem with running at a higher priority, that personally I will not do it.

Is that all the frameworks will use it or none at all.

errantmind:
Priority mostly helps with mean and max latency, not throughput. I care about latency almost as much as throughput though so I attempt to set it. It is not required and no one really needs to set it unless they want more consistency across requests.

volyrique:
IMHO all OS-level optimizations should be fair game, be it process priorities, processor affinities, usage of `io_uring`, etc. (and no, personally I would not use all of them). Probably the only line that should be drawn is if something requires special privileges - for example, setting a real-time scheduling policy such as `SCHED_FIFO` for a process would be excluded. Incidentally, if you check @talawahtech's excellent blog post (Thanks a lot, mate!) in #6597, most of the significant optimizations are in fact privileged.

If a framework is not using a particular OS-level optimization, then its developers should think long and hard about that (and possibly remedy it). I think that is pretty much one of the motivations for the existence of the benchmarks in this repository.

As an aside - any usage of `sudo` and setting capabilities should be regarded with suspicion. As for what particular OS-level optimizations a framework is using, it should be straightforward to discover that with a `strace` session.

NateBrady23:
Closing as part of bulk cleanup.

PR #6976, opened 2021-12-14
FaF performance tweak: updated dockerfile

errantmind:
Small tweak to test thread prioritization


frameworks/Rust/faf/faf.dockerfile
```diff
@@ -13,4 +13,4 @@ RUN RUSTFLAGS="-Ctarget-cpu=native -Ztune-cpu=native -Zmutable-noalias=yes -Clin
    && strip ./target/x86_64-unknown-linux-gnu/release/faf-ex
 
 EXPOSE 8089
-CMD ./target/x86_64-unknown-linux-gnu/release/faf-ex
+CMD nice -19 ./target/x86_64-unknown-linux-gnu/release/faf-ex
```
joanhey, line 16: Sorry,  Will you do that in your servers ??

errantmind: Hey, I'm not sure what you mean. Do it where? In the code itself?

errantmind: Nevermind, I'll move this into my code to simplify it

joanhey:
I will be here again, for look your code in rust.

joanhey:
A lot of people don't understand what is a benchmark.

It's a really cool app to change the code in our frameworks for be faster.
NOT a list of trips for look faster, that it if make the others fw will be faster too. But nobody will do it in his servers.
Thank you.

joanhey:
And sorry but it's not your first time.
Read #6967

errantmind:
> And sorry but it's not your first time. Read #6967

I don't see how changing the priority of the process is an issue. That is normal behavior and not a problem. It isn't about looking fast, it makes sense to ensure optimal scheduling for a critical piece of infrastructure is running. 

That said, I agree with the decision about routing mentioned in #6967. That optimization went too far but I will ask you not to use that to make assumptions about the rest of my code. I'm here to be the fastest but I will obey the rules. I want to make something interesting and useful.

PR #6990, opened 2021-12-20
[genhttp] Update to version 6.3

Kaliumhexacyanoferrat:


frameworks/CSharp/genhttp/Benchmarks/Tests/JsonResource.cs
```diff
@@ -12,9 +12,10 @@ public sealed class JsonResult
 
     public sealed class JsonResource
     {
+        private static readonly JsonResult _Result = new() { Message = "Hello, World!" };
 
         [ResourceMethod]
-        public JsonResult GetMessage() => new JsonResult() { Message = "Hello, World!" };
+        public JsonResult GetMessage() => _Result;
```
nbrady-techempower, line 18: This change violates the JSON test rules. An Object needs to be instantiated and serialized on every request.

Kaliumhexacyanoferrat: Ah, sorry - will fix that right away.

PR #7003, opened 2021-12-29
racket: use places for parallelism

Bogdanp:
This change removes nginx from the Racket benchmark and uses Racket's native `place`s for parallelism, spawning one place per core instead of a process per core. It also bumps Racket to the latest release (8.3).


Bogdanp:
👋 bump

PR #7006, opened 2022-01-02
new test ktor-pgclient

tmohme:
Add new test: ktor-pgclient


tmohme:
@ShreckYe I also thought about this name for the test, but decided against it as 
- the test is not using any other feature of vert.x
- other tests (e.g. jooby-pgclient, quarkus ... + pgclient) are named similar

PR #7014, opened 2022-01-07
Add Npgsql in natively compiled mode without metadata

LLT21:
Npgsql is added to the test of natively compiled alternatives without metadata. In this case reflection cannot be eliminated completely, as the Ngpsql library uses reflection within its implementation. However, based on the feedback from Michal Strehovsky in the issue https://github.com/dotnet/runtimelab/issues/1798, it is possible to natively compile the main library and exclude all metadata from the custom implementation.


bvbbi518:
@nbrady-techempower Nate, just checking if I still have to do something on this one before being able to be merged ? Gr, Luc

nbrady-techempower:
All set. Sorry for the delay!

LLT21:
No problem. Thanks a lot. 

Luc

PR #7033, opened 2022-01-24
upgrade `fastendpoints` to v3.1 and match with `aspcore-mvc`

dj-nitehawk:


dj-nitehawk:
@nbrady-techempower would really appreciate if you could merge this. 
planning to submit another PR with latest `servicestack` as the current code in this repo seems to be based on an ancient mono platform and doesn't even run anymore. 
thanks!

PR #7043, opened 2022-01-28
Use Rails metal controller for plaintext and json

spizm:
This switches the plaintext and json endpoints to use the Rails metal controller. Metal is meant to be used for these types of endpoints where performance is important and the work itself doesn't need the full breadth of rails features. 

I could technically implement the rest of the endpoints with metal controllers to get better scores, but the other endpoints exercise more of the framework, so it doesn't seem to be in the spirit of the benchmarks here.


spizm:
@nbrady-techempower It looks like the CI tool is breaking on building the `techempower/mysql:latest` docker image

nbrady-techempower:
@spizm Thanks, it looks like it's just breaking here. We'll get it in and see what happens.

nbrady-techempower:
@spizm Fixed, thanks! You can pull in the latest master to get this to rerun.

spizm:
@nbrady-techempower It passes now!

PR #7076, opened 2022-01-29
Add faf-ext

botika:
faf-ext is just extension of faf for put a async runtime with shared state. The idea is to be able to implement all the functionality of the benchmark


botika:
The bot would also have to remove the tag if the build is done correctly. This is ready, the weekend I will put the database.

PR #7099, opened 2022-02-04
[workerman] Update to Ubuntu 21.10

joanhey:
Testing Ubuntu 21.10 in Citrine.

### DIFF
--- a/frameworks/PHP/workerman/workerman-async.dockerfile
+++ b/frameworks/PHP/workerman/workerman-async.dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM ubuntu:21.10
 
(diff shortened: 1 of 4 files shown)


joanhey:
We will test the performance of the new kernel 5.9.
With FSGSBASE support to improve context switch performance on x86 processors.

https://www.phoronix.com/scan.php?page=news_item&px=FSGSBASE-For-Linux-5.9
https://www.phoronix.com/scan.php?page=article&item=x86-fsgsbase-xeonr&num=1
This bench is with 2 x Intel Xeon Gold 5220R. Citrine is using Intel Xeon Gold 5120 CPU.
One of the big looser in performance with meldown/spectre patches.
It's nice to see the improvements in context switching performance after all the hits encountered over the past two years as a result of various CPU security vulnerability mitigations.

PR #7120, opened 2022-02-10
prepare Actix Web v4 update

robjtede:


robjtede:
Stable v4 is imminent but this looks acceptable to the TFB maintainers I'd like to get the RC version in-tree for a prelim run just to iron out any regressions observed.

PR #7158, opened 2022-02-25
Upgrade goframe version to v2.0, Update cases implements handlers for goframe

gqcn:
- upgrade `goframe` version to `v2.0`
- update cases implements handlers for `goframe`


gqcn:
@nbrady-techempower

PR #7160, opened 2022-02-27
Bump up gnet to v2

panjf2000:


panjf2000:
PTAL
@nbrady-techempower

PR #7177, opened 2022-03-05
gnet: fix a bug when reading HTTP requests

panjf2000:


panjf2000:
PTAL
@nbrady-techempower

PR #7222, opened 2022-03-16
Fix Scala/vertx-web-scala start/stop issues

nbrady-techempower:


nbrady-techempower:
Looks like some workflow issues. Passed locally.

PR #7235, opened 2022-03-22
Eliminate the buggy HTTP parser

panjf2000:


panjf2000:
PTAL
@nbrady-techempower

PR #7243, opened 2022-03-24
use homebrew postgres driver.

fakeshadow:

### DIFF
--- a/frameworks/Rust/xitca-web/Cargo.toml
+++ b/frameworks/Rust/xitca-web/Cargo.toml
@@ -17,21 +17,21 @@ xitca-http-codegen = "0.1"
 xitca-server = "0.1"
 xitca-service = "0.1"
 xitca-web = "0.1"
+xitca-postgres = "0.1"
 
 ahash = { version = "0.7.6", features = ["compile-time-rng"] }
 atoi = "0.4.0"
 core_affinity = "0.5.10"
-diesel = { git = "https://github.com/diesel-rs/diesel.git", rev = "37ec18f46ced2d6e9197414156fdb705d7a61426", default-features = false }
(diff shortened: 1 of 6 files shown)


michaelhixson:
It looks good to me in manual verification.

* Pull in the changes from this PR.
* Pull in [these other changes](https://github.com/michaelhixson/FrameworkBenchmarks/commit/3513992ac0467d029ecadf5068d50163b5c88b7c) to enable the postgres query log.
* Run `tfb --mode debug --test xitca-web`
* Run tcpdump on the database:
  ```shell
  docker run --rm -it \
    --net container:tfb-database \
    nicolaka/netshoot \
    tcpdump -i eth0 port 5432 and inbound -Xvv
  ```
* Run `curl -i -s localhost:8080/queries?q=10`
* Verify the postgres log prints a distinct virtual transaction id for each query.  (Batch queries would reuse the same virtual transaction id across multiple queries.)
* Verify the tcpdump output shows 10 "B", "E", "S" (bind, execute, sync) messages.  (Batch queries would have fewer than 10 "S" messages.)

fakeshadow:
@nbrady-techempower Thanks for informing the new rule and after a read I think my driver does not violate it.
There is some of the source code of it if you don't mind:
https://github.com/HFQR/xitca-web/blob/e1debcd4630196669923f94c7693e5519f34db52/postgres/src/prepare.rs#L230
https://github.com/HFQR/xitca-web/blob/e1debcd4630196669923f94c7693e5519f34db52/postgres/src/query.rs#L44

Every query and prepared statement is encoded with sync .

fakeshadow:
@michaelhixson Thanks for the verify. That's a nice way to reinforce the new rule.

PR #7244, opened 2022-03-24
akka-http: use latest version and optimized settings

jrudolph:


hepin1989:
continuous benchmarking results: https://tfb-status.techempower.com/.

PR #7245, opened 2022-03-25
Upgrade to Swift 5.6

adam-fowler:
Upgrade all swift projects to Swift 5.6 also use focal docker images as standard swift docker images still use bionic


0xTim (approved):
Yep LGTM. We should probably migrate to static linking but that can be done later

PR #7247, opened 2022-03-26
Ur/Web updates

sheganinans:
Hello! I am interested in getting the Ur/Web benchmark back up and running, I noticed the error at:

https://tfb-status.techempower.com/unzip/results.2022-03-21-18-55-25-452.zip/results/20220316112215/urweb-cache/cached-query/verification.txt

This is just the first step to update the code to conform to the new requirements, Travis may reveal more..


sheganinans:
I'm a bit puzzled by the the recent error we're receiving in the test. It says that the query path must be at least 8 character, and now it's over that limit and it's still not having it..

nbrady-techempower:
@sheganinans Thanks, that's actually a typo. The cached queries path needs to be at least 15 characters long: like `/cached-queries`. I've fixed in the toolset and docs.

nbrady-techempower:
@sheganinans We're starting preview runs for Round 21. Haven't looked at why this isn't passing, do you have any time to take a look? Maybe pulling in the master branch and not doing a force push to trigger a new check

sheganinans:
Yes, so the main issue is that it seems like the MySQL branch is now old
and a bit bitrotted... I suppose it might be worthwhile to just disable
those tests. If that seems reasonable, I'll do that and submit a new push.

nbrady-techempower:
@sheganinans Sure, that works!

nbrady-techempower:
@sheganinans Going to close this for now. Re-open whenever you get the chance to work on it. Thanks!

PR #7290, opened 2022-04-11
Don't reuse shared message instance

sebastienros:
I believe the specs forbid reusing the same instance for json serialization. It's also wrong code-wise since concurrent requests could assign different strings to the same instance.


nbrady-techempower:
Thank you @sebastienros !

PR #7308, opened 2022-04-26
Update Clojure/Aleph - Add DB, queries, fortunes, updates tests

arnaudgeiser:
## Description

This PR adds four missing tests for Aleph:
- /db
- /fortunes
- /queries
- /updates

It also bumps/removes dependencies and updates links:
- Aleph bumped to 0.4.7 from 0.4.5-alpha5
- Clojure bumped to 1.11.0 from 1.9.0
- jsonista bumped to 0.3.5 from 0.2.0 
- clj-tuple has been removed
- JVM bumped to Java 17
- JVM flags have been adapted to match the ones from [reitit](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Clojure/reitit/reitit.dockerfile#L6)
- Links inside this repository have been fixed
- Links to `Aleph` have been changed to reflect its new home


frameworks/Clojure/aleph/src/hello/handler.clj
```diff
@@ -1,45 +1,188 @@
...
+
+(defn update-and-persist
+  "Fetch a number of `queries` random world from the database.
+  Compute a new `randomNumber` for each of them a return a `CompletableFuture`
+  with the updated worlds."
+  [queries]
+  (d/chain' (run-queries queries)
+            (fn [worlds]
+              (let [worlds' (mapv #(assoc % :randomNumber (random)) worlds)]
+                (d/chain' (apply d/zip (mapv update-world worlds'))
```
KingMob, line 120: This updates each world independently, even if in parallel, but according to

we could group the statements, which might be faster.

PR #7337, opened 2022-05-09
[roa] Fix fortunes route

joanhey:


joanhey:
Ready to be merged.
Only fail 1 variant, that it was failing before.

PR #7376, opened 2022-05-27
Add kruize-Autotune configuration to Quarkus benchmark to improve performance.

kusumachalasani:
Update JAVA_OPTIONS to include the configuration generated by kruize-Autotune to Quarkus benchmark to improve performance.


franz1981:
Some of the Java options set are using their default value, hence could be omitted for brevity
I have spotted a problem with the techempower benchmark due to running all the tests without restarting the container, hence I suggest to address that on the autotune test methodology as well 
See

Sanne:
right, we shouldn't use autotune on this yet - the benchmark methodology needs to be fixed first.
Thanks

kusumachalasani:
Will open a new PR, once the validation tests complete.

PR #7377, opened 2022-05-27
Quarkus: fix and refactoring for Fortunes benchmark

Sanne:
Did some cleanup of the Quarkus benchmark code; the persistence layer is now better organized and separated from the webserver layer.

In particular the benchmark for `/fortunes` was doing some dump stuff compared to the benchmark we had for Quarkus 1.1, so fixed a regression there.

Finally, some additional touches to JVM tuning.


franz1981 (changes requested):
I would remove always pretouch and settings heap size

Sanne:
Ok. I'll update later and send a new one.

PR #7378, opened 2022-05-28
Plumber API for R

emilmahler:
According to the TIOBE index, R is the 13th most popular programming language, widely used by data scientists and statisticians. Plumber is the mostly popular API framework for R, developed and maintained by Rstudio. This would make a great addition to the Tech Empower Benchmarking suite. We have all 6 tests passing locally.


nbrady-techempower:
Exciting! Thanks for the contribution @emilmahler !

PR #7403, opened 2022-06-08
[C#][aspnetcore] - mark all classes sealed

unsafePtr:
That's a tiny improvement which might help for runtime to devirtualize class. 

I would better quote Stephen Toub from [this topic](https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-6/#peanut-butter).

> If the runtime can see that a given instance on which a virtual call is being made is actually sealed, then it knows for certain what the actual target of the call will be, and it can invoke that target directly rather than doing a virtual dispatch operation. Better yet, once the call is devirtualized, it might be inlineable, and then if it’s inlined, all the previously discussed benefits around optimizing the caller+callee combined kick in.

### DIFF
--- a/frameworks/CSharp/aspnetcore/Benchmarks/Configuration/AppSettings.cs
+++ b/frameworks/CSharp/aspnetcore/Benchmarks/Configuration/AppSettings.cs
@@ -3,7 +3,7 @@
 
 namespace Benchmarks.Configuration;
 
-public class AppSettings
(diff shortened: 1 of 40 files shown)


nbrady-techempower:
Closing this in favor of #7698 right now. Please ping the guys there to see if this is something they want to do.

unsafePtr:
Thank you Nate. I see the change is already in place

PR #7411, opened 2022-06-13
Switch aspnetcore and easyrpc from MemoryCache to FastCache.Cached fo…

neon-sunset:
This change switches `cached-worlds` `aspnetcore` and `easyrpc` workloads from `MemoryCache` to `FastCache.Cached`.

The main reason for this change is that `FastCache.Cached` is much more performant in high-throughput and/or multi-threaded scenarios. This is achieved via several factors:
- Providing a more limited feature set when compared to `MemoryCache` like per-entry callbacks, multiple expiration modes supported simultaneously, etc.
- Using `Environment.TickCount64` instead of `DateTime.UtcNow` as a timestamp source which has significantly lower overhead
- Reducing memory indirection cost by using `readonly struct`-based cache entries - both timestamp and entry values are fetched in a single cache line load (when not crossing cache line boundary) and then are passed in CPU registers (struct enregistration)
- Using static dispatch to a static cache store instance (caches are separated per key-value types combinations, e.g. `int + CachedWorld`, `(string, bool) + OtherType`, etc.)
- Having read and write paths lock-free and wait-free (for the most part thanks to https://github.com/VSadov/NonBlocking) provides better scaling per thread
- `MemoryCache` uses keys boxed into `object` which may allocate on every read and cannot take advantage of specialized lookup implementations available for number-based keys
- Even regardless of other techniques, the implementation itself being more limited when compared to `MemoryCache` means that CPU has to execute much less code and cache entries being much smaller (min size is only 40B for node in ConcurrentDictionary that nests 16B+ sized `CachedInner`) allows for higher read and write throughput per same amount of memory traffic

Any feedback on further possible improvements or potential issues will be much appreciated.

Also attaching TL;DR comparison of latency and throughput vs MemoryCache in a standalone benchmark suite (string-based keys, for int-based keys the difference in perf will be much higher)
|            Library | Lowest read latency | Read throughput (M/1s) | Lowest write latency | Write throughput (M/1s) | Cost per item | Cost per 10M items |
| ------------------ | ------------------- | ---------------------- | -------------------- | ----------------------- | ------------- | ------------------ |
|   **FastCache.Cached** |            **15.63 ns** | **114-288M MT / 9-72M ST** |             **33.75 ns** |    **39-81M MT / 6-31M ST** |          **40 B** |             **381 MB** |
|        MemoryCache |            56.93 ns |   41-46M MT / 4-10M ST |            203.32 ns |    11-26M MT /  2-6M ST |         224 B |           2,136 MB |

*`NonBlocking.ConcurrentDictionary` has specialized implementations for number-based keys. See diff in https://github.com/neon-sunset/fast-cache/blob/main/docs/performance/defaults-net7-win-x64.md
**As of now, the library is incompatible with AOT. Once this is fixed, I will update (here or separate PR) `CoreRT` and `Mono` workloads accordingly. See https://github.com/VSadov/NonBlocking/pull/13#issuecomment-1153598339

### DIFF
--- a/frameworks/CSharp/aspnetcore/PlatformBenchmarks/Data/Providers/RawDbMySqlConnector.cs
+++ b/frameworks/CSharp/aspnetcore/PlatformBenchmarks/Data/Providers/RawDbMySqlConnector.cs
@@ -8,6 +8,8 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
+using FastCache;
+using FastCache.Extensions;
 using Microsoft.Extensions.Caching.Memory;
 using MySqlConnector;
 
@@ -19,11 +21,6 @@ namespace PlatformBenchmarks
     {
         private readonly ConcurrentRandom _random;
         private readonly string _connectionString;
-        private readonly MemoryCache _cache = new(
-            new MemoryCacheOptions
-            {
-                ExpirationScanFrequency = TimeSpan.FromMinutes(60)
-            });
 
         public RawDb(ConcurrentRandom random, AppSettings appSettings)
         {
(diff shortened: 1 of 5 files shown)


neon-sunset:
CCing the authors of previous commits for `aspnetcore` workload since `README` has no listed area experts: @jo-ninja @ShreyasJejurkar @MichalStrehovsky

neon-sunset:
@nbrady-techempower Thank you, the build has been fixed.

davidfowl:
This should be a different benchmark

neon-sunset:
@davidfowl Right, the downside of this change is that `M.E.C.M.MemoryCache`, even if distributed separately, is effectively standard. Do you see the change as a separate benchmark in the form of `aspnetcore-{libraryname}` that exposes just `cached-worlds`?

Issue #7402, opened 2022-06-07
Investigate faf results on Plaintext


sebastienros:
It's currently over 8M RPS which is above the theorical limit of the network.

sebastienros:
I think the date is empty as the variable is never initialized: https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Rust/faf/src/main.rs#L43

The initialization seems to have been removed in this PR: https://github.com/TechEmpower/FrameworkBenchmarks/pull/6523/files#diff-e828d6a53c980b3107862cfa2530ba397b76f56d2a0bb3d555465a6f6217f4feL52-L53

That would explain the network bandwidth difference.

sumeetchhetri:
Please look at the source for the framework [here](https://github.com/errantmind/faf/blob/aae75a54637bcf0e225e1078693d5110e444f7d5/src/epoll.rs#L70), the date value is passed along from the main event loop, also the date header validation does exist in the tfb toolset.

sebastienros:
Thanks for confirming. Do you know how it's getting over the "physical" limits then? If not then I will try to check the full payload it's sending so I can understand.

fafhrd91:
could [process priority](https://github.com/errantmind/faf/blob/aae75a54637bcf0e225e1078693d5110e444f7d5/src/epoll.rs#L47) affect throughput?

sebastienros:
@fafhrd91 historically all plaintext results have been capped at 7M rps because when you look at the minimum request size and the network packets, over a 10Gb/s network, it can't be faster than that.

This was confirmed by:
- stripping the `/paintext` url to `/p` (and triggered a new rule in the specs)
- using smaller `Server: ...` headers (and triggered a new rule in the specs)
- using a faster network card (in our Microsoft's lab) which improved all plaintext results for top frameworks (up to 14M rps) without changing anything else.

I am not excluding any magic trick though, even at the network level, which could be a totally fine explanation.

billywhizz:
@sebastienros @fafhrd91 @sumeetchhetri @errantmind 

i've done a little investigation here and it seems each faf request from wrk is 87 bytes and each response back is 126 bytes.

so, if we assume for argument sake an average ethernet packet size of 8192 bytes, that would allow 64 responses per packet plus 66 bytes of protocol overhead for TCP/IP/Ethernet, giving us an overhead of ~1 byte per response for an overall average including protocol overhead of 127 bytes per response.

[...]

PR #7446, opened 2022-07-05
Change port for FAF

kant2002:
I decide to make this change after this fascinating read.
https://github.com/TechEmpower/FrameworkBenchmarks/issues/7402#issuecomment-1159342415


errantmind:
I don't think there is any reason to suspect the port is related to performance. I'm ok with this tweak if it will put some people's minds at ease but I would prefer not to have FaF further 'tweaked' without some actual evidence to support the change.

PR #7481, opened 2022-07-20
Complete mORMot2 benchmarks (Pascal language)

pavelmash:
mORMot2 is an open source framework for modern Pascal (Free Pascal or Delphi)  
This benchmark uses cross platform Free Pascal compiler

Since where is no Pascal language in suite yet, we think this would make a great addition to the Tech Empower Benchmark.

All tests are passing locally.


pavelmash:
Initial release was fails CI for /db, but we run `--verify` many time on local environment and all are success.
What is even weirder is that `/queries?queries=20` do pass - but not /db ?
It does not make much sense... they use the same ORM method (retrieve by ID)

I even setup a VM with very limited resources on Ubuntu 18.04, and `--verify` is also success.
May be there is some performance/timeout issue during the validation tests?

Also we find memory leaks for "raw" test suite and fix it in second commit. Hope it should pass verification.

pavelmash:
We found the reason why verification for previous commit fails - out application do not close sockets correctly after each `siege` execution, so they are remains in CLOSE_WAIT state. Should be fixed by [325b1e2]

pavelmash:
@nbrady-techempower - please, give us a feedback - may be we need to modify something in this PR to be merged?

PR #7482, opened 2022-07-20
Update Dart2 sample to Dart 2.17

kevmoo:
Use *new* official Dart Docker image
Migrate to newly recommended lints


kevmoo:
CC @nbrady-techempower 👍

PR #7498, opened 2022-07-27
Update the Quart dependencies

pgjones:
There are recent updates to both Quart and Werkzeug that should
materially impact the performance in a positive way.


pgjones:
Updated

PR #7536, opened 2022-08-07
Julia Genie framework benchmark

Vladislav-Zolotaryov:
Implemented benchmarks for Genie.

Without using SearchLight ORM that comes with it, because it pluralizes model names and currently there is no way to override it.

All tests except for Cache are implemented.


Vladislav-Zolotaryov:
Updated

nbrady-techempower:
@Vladislav-Zolotaryov The test logs for this are expired. Not sure what went wrong. If you have a chance to update, please re-open, make an update to trigger the tests, and we'll take a look.

PR #7563, opened 2022-08-27
Update Deno benchmark

littledivy:
Hi 👋 The benchmarks were very old - updated it to use the new `Deno.serve` API. Removed mongodb benchmarks since they cannot be updated to latest version. I'll add more cases when time permits.


littledivy:
```
./tfb --mode verify --test deno
``` 

Passes on my machine.

PR #7586, opened 2022-09-18
[Dart] Upgraded Angel3 Framework to 7.0.0

dukefirehawk:
- Bumped `Angel3` to 7.0.0
- Added `ORM for MySQL` test case


frameworks/Dart/angel3/angel3-mysql.dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM dart:latest
```
nbrady-techempower, line 1: Can you tie this down to a specific version of dart, please

nbrady-techempower (changes requested):
Just the one comment above and if you could remove the favicon if it's not required for the tests to run successfully

frameworks/Dart/angel3/angel3-mysql.dockerfile
dukefirehawk: - Removed favicon
- Updated to use official dart docker image 2.18.1

PR #7588, opened 2022-09-18
feat: updated robyn to the latest and fastest version

sansyrox:
Updated the version of Robyn.


JackThomson2:
Should we also remove `async` from the function?

sansyrox:
@JackThomson2 , I benchmarked locally for async using our own test suite , and async was faster. I can try doing the same using this command and then we can choose whichever is faster.
> ./tfb --mode verify --test 

> Also wonder how having the const plays with the rules, seems a fairer test to have it disabled.

Why do you think this is the case? I feel if Robyn has a feature that will be used by everyone without changing anything fundamentally then it should be included in the performance benchmarks. Maybe, it is a feature that is too good and hence feels illegal?

Let me know what you think.

JackThomson2:
I'm not sure, I haven't followed the design recently, however, when I last tested sync functions where significantly faster. 

With regards to the const functions, it just seems a bit disingenuous as it's all done in rust and won't benchmark the round trip to python which is likely the slowest part of the framework.

PR #7614, opened 2022-09-23
New benchmark implementation for Helidon that includes Nima support

spericas:
This new version of the Helidon benchmark includes a module for Nima, the new implementation of Helidon that is based on JDK 19 virtual threads. It also contains the previous (reactive) implementation nearly unchanged from before.

The code is organized into two Maven modules, namely, *reactive* and *nima*. Both modules implement the same set of tests outlined below. Metadata files have been updated for the two modules. For more information see the README file.

PR #7619, opened 2022-10-05
Add the helidon-nima framework.

Blquinn:
This adds a test for the [Helidon Nima](https://helidon.io/nima) framework.


barchetta:
The Helidon team has already addressed this with PR #7614

Blquinn:
Well it was fun while it lasted.

PR #7625, opened 2022-10-17
[ci skip] Fix Fastapi and Silverlining default

joanhey:
in `benchmark_config.json`

Tired of the message:

image


nbrady-techempower:
Thank you!

PR #7653, opened 2022-10-27
update netty

lneves:


franz1981:
@lneves the Citrine machine can make uses of io_uring? Just curious

lneves:
@franz1981 I'm not sure, It seems that Citrine it's running Ubuntu 20.04 and  I think updated version of it are able to use io_uring

PR #7664, opened 2022-11-02
Update atreugo's pgx from v4 to v5

azybler:


nbrady-techempower:
Thanks for your first contribution @azybler !

PR #7667, opened 2022-11-02
New parent images in Helidon docker files

spericas:
New parent images for Helidon docker files that do not require installation of Maven. Current images are failing while attempting to access host `apache.osuosl.org` in the TE environment. These new docker files avoid that problem by using a different parent image that already includes Maven.

Piggybacking a small config change related to SQL pools as well.


spericas:
@nbrady-techempower

PR #7686, opened 2022-11-10
[php] Update Codeigniter to v 4

joanhey:
It's a complete rewrite and added update test.


joanhey:
### Performance info

CI3 vs CI4

Framework | JSON | 1-query | 20-query | Fortunes | Updates | Plaintext
-- | -- | -- | -- | -- | -- | --
CI3 |
CI4 |

PR #7687, opened 2022-11-11
[php] Add JIT in php-fpm

joanhey:
https://github.com/TechEmpower/FrameworkBenchmarks/issues/6894


joanhey:
Ready to be merged !
Only fail php-unit, that also fail in the last runs.

PR #7696, opened 2022-11-13
[may_minihttp] refine tell fortune code

Xudong-Huang:
also update rust edition to 2021

### DIFF
--- a/frameworks/Rust/may-minihttp/src/main.rs
+++ b/frameworks/Rust/may-minihttp/src/main.rs
@@ -195,8 +195,8 @@ impl PgConnection {
         Ok(worlds)
     }
 
-    fn tell_fortune(&self) -> Result<Vec<Fortune>, may_postgres::Error> {
-        let mut items = vec![Fortune {
+    fn tell_fortune(&self) -> Result<SmallVec<[Fortune; 32]>, may_postgres::Error> {
+        let mut items: SmallVec<[_; 32]> = smallvec::smallvec![Fortune {
             id: 0,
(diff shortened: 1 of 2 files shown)


frameworks/Rust/may-minihttp/src/main.rs
```diff
@@ -195,8 +195,8 @@ impl PgConnection {
         Ok(worlds)
     }
 
-    fn tell_fortune(&self) -> Result<Vec<Fortune>, may_postgres::Error> {
-        let mut items = vec![Fortune {
+    fn tell_fortune(&self) -> Result<SmallVec<[Fortune; 32]>, may_postgres::Error> {
+        let mut items: SmallVec<[_; 32]> = smallvec::smallvec![Fortune {
```
nbrady-techempower, line 199: This should be dynamically sized. If you're doing this because you saw it somewhere else, please let me know. I may have missed it.

Xudong-Huang: oh, I didn't notice that. But you can see it here
https://github.com/TechEmpower/FrameworkBenchmarks/blob/288381e39da7358f8477f66bd023126f3b89355d/frameworks/Rust/ntex/src/db.rs#L175-L185

Xudong-Huang: `SmallVec` is dynamically backend, if more elements comes, it will use heap malloc.

Xudong-Huang:
no need update

PR #7701, opened 2022-11-14
Update FaF llvm version to fix build

errantmind:
Update FaF llvm version to fix build


frameworks/Rust/faf/faf.dockerfile
```diff
@@ -13,4 +13,4 @@ RUN RUSTFLAGS="-Ctarget-cpu=native -Ztune-cpu=native -Zmutable-noalias=yes -Clin
    && strip ./target/x86_64-unknown-linux-gnu/release/faf-ex
 
 EXPOSE 8089
-CMD nice -20 ./target/x86_64-unknown-linux-gnu/release/faf-ex
+CMD nice -19 ./target/x86_64-unknown-linux-gnu/release/faf-ex
```
joanhey: Still not in the rules, but for me this is not fair play.
All the frameworks could do it, but not only 1 fw.
And I'll never use it so in production.

joanhey (changes requested):

frameworks/Rust/faf/src/main.rs
```diff
@@ -6,11 +8,12 @@ use faf::const_config::*;
 use faf::const_http::*;
 use faf::extern_http_date;
 use faf::util::{const_len, memcmp};
+use std::intrinsics::likely;
 
 const ROUTE_PLAINTEXT: &[u8] = b"/p";
```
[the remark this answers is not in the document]
errantmind: Not sure where this is coming from. I changed this many months ago and this code you are quoting is not up to date.

frameworks/Rust/faf/faf.dockerfile
errantmind: It seems a bit off to be commenting on code long unchanged in this pull request

PR #7716, opened 2022-11-19
Updates for CakePHP 4.4.

ADmad:
Consolidated all actions into a single controller.


joanhey (approved):

joanhey:
@ADmad could you check why is it slower with Adapterman ??
Thank you

ADmad:
Sorry, currently I don't have the time to profile  the app using Adapterman to figure out why it's slower with it.

joanhey:
@ADmad thank you for the changes
I try to my best, but it's very difficult to maintain all php fw with all the new changes.
Thank you

Issue #7703, opened 2022-11-15
PHP 8.2 update [info]


joanhey:
## PHP 8.2
We start to test 8.2 RC5, final release date Dec 8 2022.

PHP update versions info:
- PHP 8.1 check #6894
- PHP 8.0 check #6184, that is closed now

PR #7748, opened 2022-11-28
[php] Add Frankenphp platform

joanhey:
It's a new PHP platform that use Caddyserver.

https://frankenphp.dev/

Initial push, later we will add optimizations and the` Worker` mode.

@dunglas  please, review

#7703


joanhey:
@nbrady-techempower ready to be merged

The fail again with nginx-unit. I'll check it again.

PR #7757, opened 2022-12-02
[php] Patch to fix Nginx Unit dockerfile

joanhey:
Thanks to Timo @tippexs from Nginx :+1: 

#7703

### DIFF
--- a/frameworks/PHP/php/deploy/unit-entrypoint.sh
+++ b/frameworks/PHP/php/deploy/unit-entrypoint.sh
@@ -1,0 +1,80 @@
+#!/usr/bin/env bash
+
+set -e
+
+curl_put()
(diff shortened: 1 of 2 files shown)


joanhey:
Ready to be merged !!

tippexs (commented):
Thanks for letting me know. This is an interim fix. We will ship a new version of our `docker-entrypoint.sh` script (https://github.com/nginx/unit/blob/master/pkg/docker/docker-entrypoint.sh) with the next version of NGINX Unit 1.29. Therfore the containers will be updated accordingly. 

There is nothing wrong patching it now BUT we are about to release a new Version within the next 2 weeks probably. So maybe a manual fix is not worth it.

PR #7763, opened 2022-12-05
[Python] Add Flask with socketify.py WSGI

cirospaciari:
Added Flask with socketify.py WSGI


cirospaciari:
weird, local all works fine, i will check the toml and json config

PR #7777, opened 2022-12-08
[Python] Add const as an separated option in Robyn and add version without const

cirospaciari:
Robyn uses "const" with basically caches the string in Rust and avoid calling python at all, i think this is trick,  so i added an non-const with is an more realistic and fair approach.
About it:
https://sansyrox.github.io/robyn/#/architecture?id=const-requests

If const is allowed so i think others frameworks should add an native cache too to be fair, so at least it should be an separated option.


nbrady-techempower:
That is interesting. Thanks for adding the separate test. Interested to see how the two compare. I'll allow `const` for now, but only in plaintext. We'll see how it does.

cirospaciari:
> That is interesting. Thanks for adding the separate test. Interested to see how the two compare. I'll allow `const` for now, but only in plaintext. We'll see how it does.

Basically little less than half of the performance in non-const
robyn const 1,265,155
robyn 521,705

If this will be allowed in plaintext results in-memory cache that never crosses GIL should be allowed in this test too but is not really realistic, because will not reflect true dynamic plaintext results

PR #7779, opened 2022-12-09
[PHP] Update imi PHP 8.2

Yurunsoft:


joanhey:
Add to the PHP 8.2 update [info] #7703

PR #7780, opened 2022-12-09
Add Nginx Njs

joanhey:
To check Njs overhead


frameworks/C/nginx/njs/hello.js
```diff
@@ -0,0 +1,11 @@
+function hello(r) {
```
lolgab, line 1: The file is not formatted consistently

joanhey: I copied from njs-examples:
https://github.com/nginx/njs-examples#hello-world-example-httphello

PR #7782, opened 2022-12-10
[Python] Add Django socketify.py WSGI 

cirospaciari:
Added Django using socketify.py WSGI


cirospaciari:
django now is working great with socketify :rocket:

Issue #7321, opened 2022-05-02
Update Citrine to Ubuntu 22.04


joanhey:
### Why?
For the kernel.

The new kernel in 22.04 fix a lot of performance degradation, from the Spectre / Meltdown Mitigation security fixes.
That we noted from round 18 to 19.
https://www.phoronix.com/scan.php?page=news_item&px=FSGSBASE-For-Linux-5.9

And the [Intel Xeon Gold 5120 CPU](https://ark.intel.com/products/120474/Intel-Xeon-Gold-5120-Processor-19_25M-Cache-2_20-GHz) CPUs in citrine, was one of the most affected to this security patch.
https://www.phoronix.com/scan.php?page=article&item=linux-419-mitigations
https://www.phoronix.com/scan.php?page=article&item=spectre-meltdown-2

nbrady-techempower:
Great point @joanhey -- I'll see if we can get this done this week so we can give plenty of time to review results before we start putting deadlines on R21.

nbrady-techempower:
I'm not going to be able to get to this for this round, but it will be the first thing we do for Round 22. Also, may be some news around new machines at the same time. ;)

joanhey:
Please, update the kernel before change to new machines.

So we can check the performance change!!

sebastienros:
> Why?

Because it's the LTS and that's what everyone is deploying, right? Even if it was slower, we should follow the LTS IMO.

volyrique:
[...]

PR #7783, opened 2022-12-10
Update wrk to version 4.2.0

volyrique:
The `-mtune` compiler option does not have an effect on x86-64 (since we are already using `-march`), but it definitely affects other platforms, AArch64 in particular.

### DIFF
--- a/toolset/wrk/concurrency.sh
+++ b/toolset/wrk/concurrency.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-let max_threads=$(cat /proc/cpuinfo | grep processor | wc -l)
+let max_threads=$(nproc)
 echo ""
 echo "---------------------------------------------------------"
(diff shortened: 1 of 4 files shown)


nbrady-techempower:
Thanks for doing this @volyrique!

joanhey:
With the first frameworks in the actual run, look like they are slower with this change.
We will wait to finish, and check the results.

volyrique:
I would like to wait till a full run has completed, so that I could look at the raw data (`results.zip`). My current suspicion is that the update to Ubuntu 22.04 is actually the culprit.

joanhey:
I don't thing that is for Ubuntu 22.04.
It's a docker container, and they don't use the kernel. They use the kernel from the host.

More Req/s more content switch, that's the problem with #7321.
But with the hosts, not with the Wrk docker.

PR #7784, opened 2022-12-11
Update axum to 0.6.1 and update other packages

romanstingler:


romanstingler:
@jplatte hey I updated axum and the other deps maybe you can have a look at it.
I just started this week with rust.
So it would be beneficial if someone could check it.

frameworks/Rust/axum/Cargo.toml
```diff
@@ -29,27 +29,37 @@ name = "axum-pg"
...
+debug = false
+split-debuginfo = 'off'
+debug-assertions = false
+overflow-checks = false
 lto = true
+panic = 'unwind'
+incremental = true
 codegen-units = 1
+rpath = false
+strip = "symbols"
```
jplatte, line 65: Why all these new settings? Many are set to their default values, and `incremental = true` could actually be based for performance.

romanstingler: ahhh, damn, I forgot to remove them, I just tested different compiler options. I will set them to the previous version with lto and codegen-units.

jplatte:
Did you run a `cargo update` inside the directory with `Cargo.toml`? That might upgrade some indirect dependencies (I'm asking because you also bumped many minimum version requirements in `Cargo.toml` for no apparent reason).

frameworks/Rust/axum/src/database_mongo.rs
```diff
@@ -14,14 +15,15 @@ use mongodb::Database;
 pub struct DatabaseConnection(pub Database);
 
 #[async_trait]
-impl<B> FromRequest<B> for DatabaseConnection
+impl<S, B> FromRequest<S, B> for DatabaseConnection
```
jplatte, line 18: All these `FromRequest` implementations should be `FromRequestParts`, because they don't care about the body of the request. Then you can also revert the change in extractor order for the handler functions.

romanstingler: I have pushed the changes FromRequest ->  FromRequestParts.

Yeah, I have checked all included crates for the latest version and updated them.

In my opinion, the benchmarks should represent a performance figure which is representative at the time when the benchmark is executed. 
Meaning, that the latest packages should be used. 
This also gives the possibility to compare benchmark history and in cases of regression maybe find crates that don't have benchmarks in their CI.

There is theoretically a 1.0.0-RC.1 of hyper available, but the latest update was in October, and it dropped 
`hyper::server::conn::AddrIncoming` so there would be more refactoring necessary. 
But as there are updates on the 0.14.x version this should be fine.

 I tested on my local notebook and all tests pass.
 
 I highly appreciate your time and patience with a Rust noob.

frameworks/Rust/axum/src/main_mongo_raw.rs
```diff
@@ -35,8 +35,8 @@ async fn db(DatabaseConnection(db): DatabaseConnection) -> impl IntoResponse {
 }
 
 async fn queries(
-    DatabaseConnection(db): DatabaseConnection,
     Query(params): Query<Params>,
+    DatabaseConnection(db): DatabaseConnection,
```
jplatte, line 39: This change can now be reverted.

frameworks/Rust/axum/src/main_mongo_raw.rs
```diff
@@ -56,8 +56,8 @@ async fn queries(
 }
 
 async fn updates(
-    DatabaseConnection(db): DatabaseConnection,
     Query(params): Query<Params>,
+    DatabaseConnection(db): DatabaseConnection,
```
jplatte, line 60: Same here.

frameworks/Rust/axum/src/main_pg.rs
```diff
@@ -59,8 +59,8 @@ async fn fortunes(DatabaseConnection(conn): DatabaseConnection) -> impl IntoResp
 }
 
 async fn updates(
-    DatabaseConnection(conn): DatabaseConnection,
     Query(params): Query<Params>,
+    DatabaseConnection(conn): DatabaseConnection,
```
jplatte, line 63: Same here.

frameworks/Rust/axum/Cargo.toml
```diff
@@ -29,26 +29,27 @@ name = "axum-pg"
...
 tokio-pg-mapper-derive = "0.2.0"
-tokio-postgres = "0.7.5"
-tower = { version = "0.4.12", features = ["util"] }
-tower-http = { version = "0.3.4", features = ["set-header"] }
-yarte = "0.15.6"
+tokio-postgres = "0.7.7"
+tower = { version = "0.4.13", features = ["util"] }
+tower-http = { version = "0.3.5", features = ["set-header"] }
+yarte = "0.15.7"
+
```
jplatte, line 52: unnecessary extra newline

```suggestion
```

jplatte (commented):
So it seems like you are misunderstanding dependency requirements in `Cargo.toml`. While they don't hurt, every version increase outside of the one for `axum` doesn't really make a difference, because a requirement like `futures = "0.3.21"` is equivalent to `futures = ">=0.3.21 <0.4"` ([docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html)). Please run `cargo update` to actually get all of the latest crate versions (including of indirect dependencies).

romanstingler:
@jplatte 
I addressed your comments and did some cleanup based on clippy's suggestions.
I tested it locally, and it verified and run the benchmarks just fine.

offtopic: wo kann ich dir das Lehrgeld überweisen? :)

nbrady-techempower:
Is this good to be merged?

jplatte (approved):
> Is this good to be merged?

Yes.

PR #7792, opened 2022-12-13
Updating Node/MongoDB Handlers to Use Default _id Index

winrid:
The current benchmarks for Node/Mongo do not use an index for queries. I'm not sure how much data we're storing in here yet, but the other tests are using default indexes (like default primary key index in PG), so this seems appropriate.

### Explanation

When you use `id` instead of `_id` you are not using an index, since you're creating a new field, without an index. We could explicitly create an index in the code, however it's fine to just use `_id` as a `Number` and use the default btree index that you get along with that.

- Moved from "id" to "_id" to use default btree like postgres/mysql tests.
- Refactored callback based code to use async/await - how most NodeJS code is written today. Performance is about the same.
- Fixed connection related race conditions in the native mongodb driver benchmark.
- Added additional parallelism in some places using promises and `Promise.all`.
- ~~Replaced multiple sequential queries with findOneAndUpdate which also returns the new version.~~

### DIFF
--- a/frameworks/JavaScript/express/helper.js
+++ b/frameworks/JavaScript/express/helper.js
@@ -1,8 +1,7 @@
 module.exports = {
 
     sanititizeTotal : (total) => {
-
-        var totalIterations;
+        let totalIterations;
         if (!total || typeof(total) != 'number') {
             totalIterations = 1;
         } else if(total < 501 && total > 0) {
@@ -16,7 +15,6 @@ module.exports = {
     },
 
     randomizeNum : () => {
-
         return Math.floor(Math.random() * 10000) + 1
     }
 }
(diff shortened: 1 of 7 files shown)


winrid:
I have verified the latest commit compiles and runs without errors.

winrid:
Made some more improvements. Verified functional.

winrid:
It's so weird that it fails in CI but works locally:

```
Verification Summary
| nodejs
|       fortune       : PASS
|       plaintext     : PASS
|       db            : PASS
|       update        : PASS
|       json          : PASS
|       query         : PASS
|       cached-query  : PASS


```

looking into it...

winrid:
`express-mongodb`, `nodejs-mongodb`, and `nodejs-mongodb-raw` fixed, ready for re-run:

```
Verification Summary
| express-mongodb
|       query         : PASS
|       db            : PASS
|       update        : PASS
|       fortune       : PASS
```

winrid:
Nice! Pretty good performance improvement.

PR #7798, opened 2022-12-17
[php] With Workerman without worker mode

joanhey:
For each request include the `.php` file from the plain php benchmark.
Using the [Adapterman lib](https://github.com/joanhey/AdapterMan).

It's experimental.
But if it's fast, we will dedicate more time.

#7703


joanhey:
Bench in my local PC, with the `json.php` test without worker mode.
With 100 concurrent connections.
image

PR #7799, opened 2022-12-17
[ci skip] Fix Robyn benchmark_config.json

joanhey:
Fix 
`Framework robyn does not define a default test in benchmark_config.json`


cirospaciari:
Sorry for this, my bad. Thank's to fixing it

PR #7815, opened 2022-12-21
Update NGINX Unit, SNUnit and change json lib

lolgab:
Previous version of NGINX Unit had a bug that made the benchmarks fail.


lolgab:
Can I have a review on this?
Without this change the benchmark doesn't run at all for this framework.

joanhey:
Still with nginx-unit v1.29 fail.
And with php with v1.28 and the fix with the entrypoint, also fail.

I updated php with nginx-unit v1.29, but I think that will fail too.

PR #7821, opened 2022-12-22
[CSharp] Optimize DataHeader

adamradocz:
Reverse the assignation order of the arrays to allow the JIT to not add bounds checks.


sebastienros:
Does it really matter? It's done in a static constructor, so only once at startup. That should have no impact on the results and the code is already hard enough to read ;)

adamradocz:
> Does it really matter? It's done in a static constructor, so only once at startup. That should have no impact on the results and the code is already hard enough to read ;)

You decide. I just made the PR. :)

nbrady-techempower:
Going to close this for now. If someone from the aspnetcore team wants this merged in, happy to reopen.

PR #7825, opened 2022-12-23
[Rust] viz improves

fundon:
* add diesel case
* improve pg logic
* improve sqlx logic


fundon:
Done. It can be merged now.

PR #7826, opened 2022-12-23
Clojure: Add Kit framework

nikolap:
Support for the Kit framework (Clojure) https://kit-clj.github.io/.

Note: I wasn't clear on what line needs to be added to the github workflows build file. If you could point me to it, can update the PR.


joanhey:
You don't need to touch the github workflows build file.
It's automatic, with new frameworks.

PR #7830, opened 2022-12-28
add luxure benchmark

FreeMasen:
- [x] Run tfb --new
- [x] add lua code
- [x] Add luxure.dockerfile
- [x] Update readme
- [x] Run `./tfb --mode verify --test luxure`
  - This worked, however I did try running `./tfb --mode benchmark --test luxure` and the higher concurrency does encouter a fatal error specifically `16384 for plaintext`, if this is something that should be omitted via configuration, please let me know how to do that
- [ ] Update `.github/workflows/build.yml`
  - I didn't see anything specific about frameworks in this file after reviewing. If I am missing something please let me know and I will update
- [x] Double check no unneeded files included

### DIFF
--- a/frameworks/Lua/luxure/README.md
+++ b/frameworks/Lua/luxure/README.md
@@ -1,0 +1,22 @@
+# Luxure Benchmarking Test
+
+### Test Type Implementation Source Code
+
+* [JSON](./server.lua)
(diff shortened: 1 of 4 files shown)


joanhey:
You don't need to touch the Github actions.
The build check which files changed, and only run the tests in that frameworks.

PR #7831, opened 2022-12-28
added php wolff framework

n0nag0n:
Just added the wolff framework. I'm on a roll :)

It's on php 8.2 as well.


joanhey:
Remember that is NOT only add more php frameworks.
You need to maintain the fw, with new versions of php and the fw !!!

joanhey:
You are Welcome !!!

joanhey:
But thanks to this benchmark, every time I see less fake benchmarks.

PR #7833, opened 2023-01-01
[mORMot] better scaling on CPUs with many cores 

pavelmash:
- using multiple HTTP request queues for better scaling
- using radix tree for routing


synopse:
Notes about this pull request:
1. The framework async HTTP server has been tuned to please high-end HW like Citrine.
2. It can now bind several HTTP server instances via SO_REUSERPORT, in the same process, to leverage resources.
3. The new radix tree routing is the official way of routing in the mORMot framework, so is used as required by the TFB specifications. It is actually as fast as manual routing with a "case" block. And it uses RTTI to register the routes.
4. Numbers are higher, and the server seems more stable for most use cases.

PR #7849, opened 2023-01-07
Improve rocket.rs implementation

elibroftw:
Changes

- Improved random number generation to use `rand::thread_rng().gen_range` instead of a shuffled array which isn't truly random at all and which required a mutex to lock, lazy_static for the array (meaning not concurrent), and is only 4 lines of code instead of the 47.
- Improved JSON example to be more inline with real world where dynamic strings may be used instead of string literals
- Used clamp instead of triple three if statements
- Upgraded dependencies
- Updated README.md to be more accurate/future-proof


elibroftw:
I also think that `workers: available_parallelism().expect("could not get parallelism").get() * 16,` should be removed as there is no comments explaining why the workers was set to this and what this number equals on the server the benchmark is run on. Rocket.rs guidelines states that the default number of 512 should not be changed unless "unless physical or virtual resources are _scarce_"

wezm:
Looks like a good improvement

elibroftw:
Don't know why the update SQL is panicking when I didn't touch that part :(.

nbrady-techempower:
@elibroftw There's been some changes and now conflicting files. Please feel free to reopen if you have a chance to work on this.

elibroftw:
Can you reopen this?

PR #7850, opened 2023-01-08
[may_minihttp] update db client logic

Xudong-Huang:
* use latest rust docker image
* create database clients concurrently
* refine database client logic

### DIFF
--- a/frameworks/Rust/may-minihttp/may-minihttp.dockerfile
+++ b/frameworks/Rust/may-minihttp/may-minihttp.dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.65
+FROM rust:1.66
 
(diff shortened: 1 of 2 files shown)


frameworks/Rust/may-minihttp/may-minihttp.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:1.65
+FROM rust:latest
```
nbrady-techempower, line 1: Can you please set this to a specific version. I think a few have gotten through with `latest` that need to be updated.

Xudong-Huang: changed to rust:1.66

PR #7856, opened 2023-01-12
ki-lang http server

ctxkiwi:
New language with its builtin http server. I opened a pull request yesterday, but closed it soon after because it wasnt handling pipe connections correctly.


ctxkiwi:
very strange that it failed. It doesnt do that in my local docker. Anyway. ill check it out later. Im currently working on a new version of the compiler, i might close this pull request until im ready with that.

PR #7877, opened 2023-01-23
Update MongoDB to version 6.0 and PostgreSQL to version 15

volyrique:
Also, revert the recent experimental changes related to `wrk` that are meant to diagnose the performance regression present since the first continuous benchmarking run that includes PR #7783. Since neither #7832 nor #7852 lead to any significant differences in the results, the `wrk` update in #7783 is apparently a red herring, and the real culprit is elsewhere (most probably OS kernel updates in the host environment).

### DIFF
--- a/toolset/databases/mongodb/create.js
+++ b/toolset/databases/mongodb/create.js
@@ -1,7 +1,8 @@
+disableTelemetry()
 db = db.getSiblingDB('hello_world')
 db.world.drop()
 for (var i = 1; i <= 10000; i++) {
-  db.world.save( { _id: i, id: i, randomNumber: Math.min(Math.floor(Math.random() * 10000) + 1, 10000) })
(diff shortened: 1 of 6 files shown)


volyrique:
@nbrady-techempower Even though most of the checks were expected to fail, I went through the output to verify that there were no surprises. Most of the failures were in implementations that also had issues during the last continuous benchmarking run (i.e. were expected to fail), but there were some that were caused by unrelated problems - for example, many PHP and Rust implementations errored out with a `no space left on device` message.

nbrady-techempower:
Thank you @volyrique!!

sebastienros:
I noticed a huge improvement on Multi Queries and Data Updates benchmarks today on ASP.NET benchmarks. And I was able to confirm it was due to this change. 15 to 20 % RPS increase.

PR #7880, opened 2023-01-28
MySQL: Update the base image to Ubuntu 22.04

volyrique:
Also, do a minor cleanup of the toolset Dockerfiles.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,26 +1,26 @@
 FROM buildpack-deps:bionic
 
+ARG DEBIAN_FRONTEND=noninteractive
+# WARNING: DON'T PUT A SPACE AFTER ANY BACKSLASH OR APT WILL BREAK
 # One -q produces output suitable for logging (mostly hides
 # progress indicators)
-RUN apt-get -yqq update
+RUN apt-get -yqq update && apt-get -yqq install \
+      -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" \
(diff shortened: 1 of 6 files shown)


volyrique:
@nbrady-techempower Similar to my other toolset PRs, the causes of the test failures are unrelated to the changes themselves (e.g. issues that occurred during the last continuous benchmarking run as well). For instance, all implementations in Groovy and Julia use MySQL, and they pass the tests without problems. Anyway, we are not changing the MySQL version here, just the base image, so actually failures should happen much earlier than the moment we start running the frameworks (e.g. while building the Docker images), which is also the case for the cleanups I have done.

I remember that there is a mechanism to tag implementations as broken, so perhaps we should do another pass and update the metadata based on the results from the last continuous benchmarking run (or the last couple of runs). In this way it would probably become a bit easier to land toolset changes (which trigger checks in all test implementations) and we might also reduce the time to complete a benchmarking run a bit too.

nbrady-techempower:
@volyrique 

> I remember that there is a mechanism to tag implementations as broken, so perhaps we should do another pass and update the metadata based on the results from the last continuous benchmarking run (or the last couple of runs). In this way it would probably become a bit easier to land toolset changes (which trigger checks in all test implementations) and we might also reduce the time to complete a benchmarking run a bit too.

This is a good idea. It's been some time. In the benchmark config, you can label a test as broken with `"tags": ["broken"]`

Thanks again for the update!

PR #7896, opened 2023-02-05
http4k- upgrade http4k, java, gradle, simplify jetty loom server, tweak undertow workers

daviddenton:


volyrique:
A minor comment - in a couple of Dockerfiles (basically the ones related to GraalVM, I think) there are instructions such as:
```
FROM frolvlad/alpine-glibc
```
This does not appear to use a fixed version of the base image, so it impedes reproducibility.

daviddenton:
No problem. Will just pin to latest one I can find in Dockerhub. Thanks!

PR #7897, opened 2023-02-05
Update SNUnit benchmark and dependencies

lolgab:


volyrique:
Note that PR #7928 marks SNUnit as a broken implementation, so you could use this opportunity to fix it.

joanhey:
SUnit is using Nginx Unit, and all the frameworks using it pass the Github Actions.
Nginx Unit fail only in Citrine, not when testing in local or Github actions.

lolgab:
> Note that PR #7928 marks SNUnit as a broken implementation, so you could use this opportunity to fix it.

The problem is in NGINX Unit, not specific to the Scala framework. I'm updating to the latest `1.29.1` that got released some days ago.

volyrique:
You need to remove the broken tag.

lolgab:
> You need to remove the broken tag.

Done ✅

PR #7899, opened 2023-02-07
ntex updates

fafhrd91:
* add cached queries
* set core affinity


frameworks/Rust/ntex/src/db.rs
```diff
@@ -29,6 +29,7 @@ pub struct PgConnection {
     rng: WyRand,
     updates: Vec<Statement>,
     buf: RefCell<BytesMut>,
+    cached: Vec<World>,
```
volyrique, line 32: I don't think that this actually complies with [requirement 7b](https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#requirements-6):
> vii. The implementation may use any caching library or approach desired, with the following caveats. [...] b. The implementation should use a bona-fide cache library or server with features such as read through, a replacement algorithm, and so on. Implementations should not create a plain key-value map of objects. If the platform or framework does not ship with a canonical caching capability, use a caching library or separate-process caching server that is realistic for a production environment. A "reference implementation" would use Google's Guava CacheBuilder for in-VM or Redis for separate-process caching.

fafhrd91: makes sense. i will delete cached queries bench. other frameworks should be reviewed as this impl was inspired by other implementation

fafhrd91: @nbrady-techempower

volyrique: Yeah, from a cursory look everything faster than `officefloor-r2dbc` in [the last completed run](https://tfb-status.techempower.com/results/05ddf07c-bbc9-4cea-8799-c1103640136c) is not compliant, but I don't really know Go.

fafhrd91: removed in #7900

PR #7900, opened 2023-02-07
ntex: remove cached bench

fafhrd91:
remove cached queries bench, as it is not complaint with [requirement 7b](https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#requirements-6)


fafhrd91:
I used this [code](https://github.com/TechEmpower/FrameworkBenchmarks/blob/a3dac7455e9dec7a855418fd819193cb9ac5ca3b/frameworks/Rust/viz/src/main_pg.rs#L26) as example

Issue #6064, opened 2020-10-09
Citrine will be DOWN beginning Oct 12, 2020


msmith-techempower:
I am very excited to be writing this issue (finally)!

Over the past several months, I have been rewriting the toolset (the software which actually orchestrates the entire verification/benchmarking/docker-communication/etc process) from scratch in Rust. At present, I would say that the new toolset is nearing MVP status, and as such I need to take down Citrine so that I can test actual results of running benchmarks with the new toolset. I hope that Citrine will not be down long, but I cannot promise a timeline today.

### What's wrong with the current toolset?

The current toolset is the result of organic iteration of software that has taken many twists and turns throughout the years. It originally had no verification, no way to tell whether a test had been successfully benchmarked, no containers of any sort, no way to to for certain that a process that was started (which can, and often does, spawn its own child processes) has been stopped, and many more shortcomings. Slowly over time, many of these niceties were added onto the toolset, but in a learning fashion - I was not sure at the time I started working on it how verification would work; I was not sure that Docker would be suitable and work; etc. With the privilege of hindsight, I can see what did and did not work so well. The new toolset aims to be built from the ground up having the luxury of only making the right decisions from these lessons.

The current toolset is a Python application, though many contributors may not know it since it also packaged as a Docker image and there is a handy script (`tfb`) in the root of the repository. The idea here was simple - we will not force contributors to have Python installed to run the toolset since Docker is a prerequisite to running the toolset. This was nice at the time and served its purpose, but it has some very simple drawbacks: namely that DockerInDocker is required for anything to work, and that causes very complicated Docker configuration issues everywhere *and* not-yet-solved issues on Windows hosts. Another reason we opted for the Docker image of the toolset is because, at the time, the Docker libraries used widely would not run on Windows natively in Python.

Another issue with the current toolset is adding new test types is extremely cumbersome, confusing, and very technical. In fact, to date we have never had a pull request submitted attempting to add a new test type. On the one hand, it would require much discussion and usually we open an RFC for that, but it is also extremely difficult to wrap one's head around.

### What's great about the new toolset?

[...]

PR #7909, opened 2023-02-09
Reduce the size of the toolset Docker images

volyrique:
The toolset Dockerfiles were using variants of `buildpack-deps` as their base images, but this is not really necessary because we do not compile anything (except in the wrk case); switch to the underlying `ubuntu` images instead. Also, try to reduce the number of layers in the images by following best practices.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM buildpack-deps:bionic
+FROM ubuntu:18.04
 
 ARG DEBIAN_FRONTEND=noninteractive
 # WARNING: DON'T PUT A SPACE AFTER ANY BACKSLASH OR APT WILL BREAK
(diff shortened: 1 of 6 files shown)


volyrique:
@nbrady-techempower This is yet another toolset PR that results in many expected failures. I checked several tests that used each of the supported databases manually, and I couldn't find any issues. Obviously a problem in the top-level Dockerfile or the wrk image would mean that pretty much nothing would work, but there are plenty of tests that pass (e.g. the C and the Groovy implementations).

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM buildpack-deps:bionic
+FROM ubuntu:18.04
```
nbrady-techempower, line 1: Does 22.04 fail? I remember there was an issue with upgrading in the past, but the details escape me.

volyrique: Yes, it does, and it seems that it has something to do with the fact that the toolset uses Python 2. Given that the plan (#6064) is to replace the current toolset with one rewritten from scratch, I have decided not to bother with it. While this container is not that critical with respect to performance as the database and the wrk ones, during the first time a developer runs the `tfb` script the contrast between `buildpack-deps` and `ubuntu` is quite stark, so IMHO this change is still worth it while the toolset transition has not been completed, especially for people with slower Internet connections.

PR #7920, opened 2023-02-14
[aspnetcore] List should not assume result size

sebastienros:
It's against the rules to initialize the Fortunes collection with prior knowledge of the results size.


joperezr (approved):
LGTM

PR #7910, opened 2023-02-09
New benchmark: Vert.x-Web Kotlinx

ShreckYe:
See the README.md for more details.

PR #7932, opened 2023-02-19
Bump the dependency versions and upgrade the JDK version to 17 in all Vert.x-related portions except "vertx-web-scala"

ShreckYe:
In the benchmark portions "vertx", "vertx-web", and "vertx-web-kotlin-corotines", bump Vert.x to the latest version 4.3.8 and other dependencies to their latest versions, and upgrade JDK to 17 [as it's supported since 4.2](https://vertx.io/blog/whats-new-in-vert-x-4-2/#java-17). "vertx-web-scala" is unchanged here because the latest version of the Vert.x Scala libraries is only 3.9.

This is part of an effort to ensure fairer comparison among all the benchmark portions using Vert.x, which are written in different languages and depending on different utility libraries. Also see #7910, #7933, and #7934.


franz1981:
Ehi @ShreckYe I'm aware that in techempower everyone can change the benchmark of everyone, but please, next time coordinated with @vietj (or me, as fallback) when you do this; both changing JDK version and bumping up Netty version (I'm saying that because I'm a NEtty committer as well) have implication beyond numbers in a public benchmarks.

ShreckYe:
@franz1981 Hi Franz, sorry about this. I thought the maintainers would assign you guys as reviewers so I didn't let you guys know before hand. I will remember to coordinate with you guys on Discord if I contribute something next time. If these changes have unwanted side effects you can always revert them.

PR #7928, opened 2023-02-17
Update the set of broken implementations

volyrique:
This pull request updates the set of broken implementations based on the following runs:
[adce24e2-9277-45b2-845c-3dbce439d727](https://tfb-status.techempower.com/results/adce24e2-9277-45b2-845c-3dbce439d727)
[05ddf07c-bbc9-4cea-8799-c1103640136c](https://tfb-status.techempower.com/results/05ddf07c-bbc9-4cea-8799-c1103640136c)
[3d6c6cc0-92ff-4861-86dc-42cef0ecc50f](https://tfb-status.techempower.com/results/3d6c6cc0-92ff-4861-86dc-42cef0ecc50f)
Note that the code in the repository has not been deleted - it is just not going to run in CI and continuous benchmarking.

As a small courtesy, I am pinging the contributors to the affected implementations, at least in the cases where I could determine a relevant person - apologies to those who have been mentioned by mistake:
@5hirish
@achlipala
@agrison
@bdw429s
@cacoco
@caturner81
@chunariov
@daviddenton
@doubaokun
@dragosv
@fakeshadow
@gi0baro
@gkbrk
@gotzmann
@greenlaw110
@hamiltont
@itsumura-h
@jcheron
@joanhey
@Kaliumhexacyanoferrat
@kekekeks
@kubo39
@lucipacurar
@luoxiaojun1992
@mkurz
@polachok
@richerarc
@seanmonstar
@steveklabnik
@sumeetchhetri
@theLastOfCats
@tiangolo
@tommilligan
@whitfin

There are also plenty of implementations, I think, that consistently have issues with one or a few of the tests, so the question is what we should do about them. The first option is to leave them alone (which is this PR's approach), but we could also mark them as broken - unfortunately that means that perfectly working functionality would be excluded from continuous benchmarking. The middle ground is to remove the affected test entries from the relevant `benchmark_config.json` and `config.toml` files.

**P.S.** FYI here is the list of implementations that have at least one permutation tagged as broken in the past:
```
C/octane
C/onion
C++/cpoll_cppsp
C++/ffead-cpp
C++/ulib
Erlang/chicagoboss
Go/revel
Groovy/hot
Haskell/servant
Haskell/snap
Haskell/spock
Haskell/yesod
Java/dropwizard
Java/spring-webflux
Lua/octopus
OCaml/morph
Perl/dancer
Perl/kelp
Perl/web-simple
PHP/hamlet
PHP/hhvm
PHP/hyperf
PHP/workerman
[...]

PR #7937, opened 2023-02-20
Refector openswoole

doubaokun:
Fix #7928


frameworks/PHP/openswoole/benchmark_config.json
```diff
@@ -14,34 +14,13 @@
...
       "orm": "Raw",
       "platform": "openswoole",
       "webserver": "none",
       "os": "Linux",
       "database_os": "Linux",
-      "display_name": "Open Swoole",
-      "notes": "",
-      "versus": "php"
-    },
-    "postgres": {
```
joanhey, line 27: Postgres variant is deleted in config.json.
But you created a open swoole postgre server, and never will be executed.

Please add again the postgres variant here.

joanhey:
Add to #7703 
PHP 8.2

frameworks/PHP/openswoole/openswoole-no-async.dockerfile
```diff
@@ -1,15 +1,21 @@
-FROM php:8.1-cli
+FROM php:8.2-cli
+
+RUN apt-get update && apt-get install -y git > /dev/null
+
+RUN docker-php-ext-install opcache pdo_mysql > /dev/null
 
 RUN pecl install openswoole > /dev/null && \
```
volyrique, line 7: I don't really know PHP, but does this pin the version of Openswoole? I noticed that the Dockerfile for the PostgreSQL permutation was using a different approach.

frameworks/PHP/openswoole/openswoole.dockerfile
```diff
@@ -1,18 +1,21 @@
-FROM php:8.1-cli
+FROM php:8.2-cli
+
+RUN apt-get update && apt-get install -y git > /dev/null
+
+RUN docker-php-ext-install opcache pdo_mysql > /dev/null
 
 RUN pecl install openswoole > /dev/null && \
```
volyrique, line 7: Ditto.

frameworks/PHP/openswoole/openswoole-no-async.dockerfile
volyrique: Ah, I see, but the main point still stands - does this pin the version of Openswoole?

joanhey: No, pecl don't use semver or an updating system.
Both Swoole and OpenSwoole, release a new version every 2-3 months from 2021.
So they use a 'fail fast' solution.

I try that all the PHP ecosystem in this bench, is updated and working.
But I was busy, the last weeks. Both are failing for the last ~3 runs only.
And a lot of times, the run fail for a fw, is downloading from the servers. Not that the fw is failing, so I wait some fails.

Also with PHP, we use the major version. All the minor versions are automatic.
The same with composer, that we use semantic versioning, and still sometimes fail.

When fail I try to fix it or create an issue in the fw (so the bench also is used to find bugs) and if necessary I marked them as broken.
Without the fail fast way, we'll have very old versions. And use a lot of my free time, now I have near to 500 PRs merged. If a need to update the pecl versions each 2-3 months, I will need 4000 PRs.

No PHP framework stay more than ~3 months failing in the runs.

Only check the work for major PHP versions from 2021: #6184, #6894, #7703 (still updating)

joanhey: You made the PR to mark as broken a lot of frameworks.
I do it normally for the PHP frameworks.

volyrique: Ultimately it is up to the TechEmpower team to set policies, but my impression is that they prefer when all dependency versions are pinned, which also makes it possible to reproduce results. The reason you have opened so many pull requests is because you have changed one or a few implementations in each one, but there is no requirement to do it that way - if you update versions every quarter, you can just do 4 PRs in a year.

joanhey: I was thinking so, the first time that I tried to update a major PHP version.
Some years ago, with few php frameworks, I created 1 PR to update all the PHP frameworks.
It was a nightmare, in almost all aspects.

nbrady-techempower: It would be ideal if we could pin the version of OpenSwoole. If you used the same opening lines as the postgres dockerfile at least the compilation step could be cached. Would there be any issues using OpenSwoole with other tests if it's using the configure with postgres option?

joanhey: The mysql variants aren't compiled.
The postgres version is pined, but it don't execute till update the bench config. I send a PR to doubaokun:fix_openswoole branch, but still not merged.

The good think is that now the 2 mysql variants dockerfiles are identical, and so 1 variant use 1s.

joanhey: Now 1s
https://github.com/TechEmpower/FrameworkBenchmarks/actions/runs/4228688851/jobs/7344367870#step:9:1355

Before ~3m 40s each
https://tfb-status.techempower.com/unzip/results.2022-11-13-10-29-24-097.zip/results/20221107183025/openswoole/build/openswoole.log

https://tfb-status.techempower.com/unzip/results.2022-11-13-10-29-24-097.zip/results/20221107183025/openswoole-no-async/build/openswoole-no-async.log

joanhey:
Ready to be merged.

Later I'll add the Postgre change.

doubaokun:
@joanhey thanks, will check. The threads reduced to half in this PR, so metrics could be lower and not comparable.

PR #7951, opened 2023-02-23
refactor Rust/hyper to futures 0.3 & update toolchain

polachok:
I don't think it makes a lot of sense to continue using futures 0.1 and tokio-core in 2023, basically no one in real life does that.

@steveklabnik 
@seanmonstar

### DIFF
--- a/frameworks/Rust/hyper/Cargo.lock
+++ b/frameworks/Rust/hyper/Cargo.lock
@@ -3,16 +3,15 @@
 version = 3
 
 [[package]]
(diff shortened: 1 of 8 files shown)


seanmonstar (approved):
Yes, agreed! Looks great to me, thanks!

(I was thinking of streaming updating this since that other PR pinged me it was failing to build, now I need to figure out what else to do XD)

steveklabnik:
Absolutely agreed. Thank you!

PR #7966, opened 2023-02-25
xitca-web-wasm: Fix the AArch64 build

volyrique:
cc @fakeshadow

It is also interesting to try other WebAssembly runtimes such as Node.js, but I don't know if they are going to work.


fakeshadow:
Thanks. I can give it a try but I doubt it would be trivial. It appears there is no spec (yet) for how a (socket) fd should be lend to the wasm binary and the way `wasmtime` doing it is kinda unique/hacky.

PR #7967, opened 2023-02-25
Add LinqToDB benchmarks

MaceWindu:
Adds 4 tests:
- aspcore-mvc-linq2db-my
- aspcore-mvc-linq2db-pg
- aspcore-mw-linq2db-my
- aspcore-mw-linq2db-pg

Additionally updated `DefaultRandom` to use shared random implementation from runtime instead of manual implementation (could revert).

Update benchmark for mysql not implemented, as currently benchmarks [doesn't](https://github.com/TechEmpower/FrameworkBenchmarks/issues/5874#issuecomment-1140254644) properly assert multi-row updates


nbrady-techempower:
There are too many permutations for this framework already. I like to keep it at a max of 20 because these tests take too long in our continuous benchmarking environment. This would put this framework close to 30. Some tests will have to be removed to fit these in. You can ping the maintainers to see if there's any they want to remove to make room.

PR #7981, opened 2023-03-03
Upgrade Microdot tests to Python 3.11

miguelgrinberg:
Update all tests to use Python 3.11, plus a couple minor performance related updates.


miguelgrinberg:
Closing and reopening a second time, as I can't figure out a way to request a build that drops the docker build cache.

PR #7986, opened 2023-03-04
Fix use the physical CPU count from Express

JHyeok:
Modify to use the number of physical CPUs, not the logical number of CPUs with hyperthreading applied, when using Cluster mode.

```
Working with clusters of Node.js processes it is common to see code using os.cpus().length as the number of child workers to fork.
For some workloads this can negatively impact performance on CPUs that use simultaneous multithreading (SMT).
Latency is doubled because two processes share the same physical CPU core to get their work done.
Additionally there is memory spent for each running worker, as well as time to spawn their processes.
It is better to fork no more child processes than there are physical cores.
```

### DIFF
--- a/frameworks/JavaScript/express/app.js
+++ b/frameworks/JavaScript/express/app.js
@@ -4,7 +4,7 @@
  */
 
 const cluster = require('cluster'),
-  numCPUs = require('os').cpus().length,
+  physicalCpuCount = require('physical-cpu-count'),
   express = require('express');
 
 const bodyParser = require('body-parser');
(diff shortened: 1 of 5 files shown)


sebastienros:
Impact of this PR as we measure it in the ASP.NET lab.

image

sebastienros:
Also, latency was the same or a little higher (json and fortunes). Memory is definitely better (since less processes). But also more failed requests or socket errors on plaintext. Not about the TE environment though.

JHyeok:
Hi, @nbrady-techempower @sebastienros 

I have applied this PR and checked the results at https://tfb-status.techempower.com/ because the environment on my computer is different.

Sorry about that.

I'll create a PR to fix it as soon as I can.

JHyeok:
@sebastienros Thanks for the comment. I could see the problem.

Issue #7930, opened 2023-02-18
Can you add some frameworks about io_uring, such as monoio


favouriter:
I hope I can pay attention to the framework test related to io_uring

volyrique:
The current benchmarking environment does not support `io_uring` properly because the host kernel is too old; issue #7321 tracks the OS update.

fakeshadow:
FYI you can look into https://github.com/monoio-rs/monoio-http which can at least be used to implement json and plaintext benchmark.

nbrady-techempower:
The Citrine environment has been upgraded.

PR #7992, opened 2023-03-08
Gh actions update to Ubuntu 22.04

joanhey:
I tested in another branch, without problems.

https://github.com/joanhey/FrameworkBenchmarks/actions/runs/4359524306/jobs/7621411385
https://github.com/joanhey/FrameworkBenchmarks/actions/runs/4359524306/jobs/7621411385#step:1:4

So some frameworks will verify and test `io_uring`, now that Citrine use Ubuntu 22.04.

#7930 https://github.com/TechEmpower/FrameworkBenchmarks/pull/7928#issuecomment-1434857626 #7321

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,7 @@ name: build
 on: [ push, pull_request ]
 jobs:
   setup:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       # Required for workflow triggers like the auto-label for failing PRs
       - name: Save PR number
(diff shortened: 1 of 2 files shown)


joanhey:
@nbrady-techempower Could we update the tfb dockerfile to Ubuntu 22.04 ?

I know that is using python 2.7, but it's possible to update too.

joanhey:
Another question:
Why not use the official docker images for databases ?

Will be faster for github actions, and almost all the apps will use the official images.

nbrady-techempower:
Thanks @joanhey!

To be honest, I simply don't remember why we didn't do that for the databases. If you want to take a shot at it and see if it works, please do. I can't think of a reason why it wouldn't, but I don't remember the specifics.

volyrique:
> Could we update the tfb dockerfile to Ubuntu 22.04 ?
> 
> I know that is using python 2.7, but it's still possible to update.

We discussed that [here](https://github.com/TechEmpower/FrameworkBenchmarks/pull/7909#discussion_r1105063544) - the gist of it is that there are issues, but if you want to give it a try too, go ahead.

PR #7996, opened 2023-03-08
[php] Plain PHP update to v 8.2

joanhey:
#7703


joanhey:
The GH actions change it's working !!

PR #7997, opened 2023-03-08
[gin] Output message when gin-std server is ready

sebastienros:
I could use this to detect when the application has started before sending some request. Currently nothing is displayed at all.

@0uep is there a better way to do this?


0uep:
Sorry @sebastienros 
I do not know (I do not master Gin)

PR #7998, opened 2023-03-08
[toolset] Use Mongodb official image

joanhey:
Use Ubuntu 22.04


joanhey:
Built time 11s https://github.com/TechEmpower/FrameworkBenchmarks/actions/runs/4367670824/jobs/7639294377#step:9:762

And working !!!

PR #8002, opened 2023-03-09
[toolset] Use PostgreSQL official docker image

joanhey:
Use Debian 11 (no Ubuntu image)

Left old `postgresql.conf`, for checking till testing finish (we can left it for historical check).
And add a full v15 `postgresql.conf.sample` as information.

### DIFF
--- a/toolset/databases/postgres/config.sh
+++ b/toolset/databases/postgres/config.sh
@@ -1,0 +1,1 @@
+cat /tmp/postgresql.conf >> $PGDATA/postgresql.conf
--- a/toolset/databases/postgres/postgres.dockerfile
+++ b/toolset/databases/postgres/postgres.dockerfile
@@ -1,58 +1,15 @@
-FROM ubuntu:22.04
+FROM postgres:15-bullseye
 
-ARG PG_VERSION=15
+ENV POSTGRES_USER=benchmarkdbuser
+ENV POSTGRES_PASSWORD=benchmarkdbpass
+ENV POSTGRES_DB=hello_world
 
-COPY 60-postgresql-shm.conf pg_hba.conf postgresql.conf ./
-COPY create-postgres.sql create-postgres-database.sql /tmp/
-
-# Prepare the PostgreSQL APT repository
-ARG DEBIAN_FRONTEND=noninteractive
-ADD "https://www.postgresql.org/media/keys/ACCC4CF8.asc" /etc/apt/keyrings/postgresql.asc
-RUN chmod 644 /etc/apt/keyrings/postgresql.asc && \
-    apt-get -yqq update && \
-    apt-get -yqq install \
-      apt-utils \
-      locales \
-      lsb-release && \
-    echo "deb [ signed-by=/etc/apt/keyrings/postgresql.asc ] http://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > \
-      /etc/apt/sources.list.d/pgdg.list && \
-    locale-gen en_US.UTF-8
-
-ENV LANG=en_US.UTF-8
-ENV LANGUAGE=en_US:en
-ENV LC_ALL=en_US.UTF-8
-
-# Install PostgreSQL on the database machine
-RUN apt-get -yqq update && \
-    apt-get -yqq install \
-      -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" \
-      postgresql-${PG_VERSION} \
-      postgresql-contrib-${PG_VERSION} && \
-    sed -i "s|PG_VERSION|${PG_VERSION}|g" postgresql.conf && \
-    mv pg_hba.conf postgresql.conf /etc/postgresql/${PG_VERSION}/main && \
-    mkdir /ssd && \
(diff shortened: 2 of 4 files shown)


joanhey:
Faster build and working !!!

After pass a run, we can check if we modifiy the conf.

sebastienros:
Seems like it's the cause of slower multi-queries benchmark slowdown. Spent some time trying to reproduce a regression in aspnet, but the results didn't match. Then looked in TE to find a change in the PostgreSQL image and found this PR with a timestamp that matches the regression.

image

joanhey:
We are still waiting to finish the actual run in Citrine, that use the official image.
I tried my best to use the same configuration, than before. The old config is still in the repo to check it.

But checking the partial results in Citrine, the `asp.net core [platform, pg]` is faster with the new image in the multi-queries benchmark.

image

sebastienros:
asp.net seems to be the same here, but every other one I have checked are slower. Do you confirm?

Before: https://www.techempower.com/benchmarks/#section=test&runid=9072dc54-398e-4e79-b4bc-0d0a51c5aa8d&test=query
Current: https://www.techempower.com/benchmarks/#section=test&runid=0f03793d-ac4c-4869-b1fa-51fcef26d48f&test=query

joanhey:
It don't exist any official Postgresql image with Ubuntu 22.04. So I used the Debian 11 image.

And the kernels are different, and will affect the results. 
Ubuntu 22.04 -> Kernel 5.15.x  |  Debian 11 -> Kernel 5.10.x  (I think).

joanhey:
We need to wait, till finish the run or a second run, to have conclusions.

sebastienros:
The docker base file kernels are of no impact. Docker uses the host kernel.

I was just pointing that this PR has an impact the queries benchmarks. The same way updating to 15 improved other benchmarks. If a change in the configuration file is responsible though then maybe we want to see if that was a good thing. I am totally fine with using the default, it's just better to understand why it changed.

joanhey:
Please check the configs, both are there, perhaps I made a mistake.

sebastienros:
I checked and it seems a few arguments are actually changed. Most values set in the new file postgresql-min.conf are overriding the default which I believe were used before.

Examples: `max_connections = 2000`; `wal_level = minimal`; ...

Do you know where this file is coming from and why these settings were chosen in favor of the defaults? Unless I didn't understand how the files are set.
I have no informed opinion on either values, just wondering if there is an actual reason behind the change, like known optimizations, of if we should just keep the defaults.

sebastienros:
I found the old file, so I assume it has the same content as the new "min" file?

joanhey:
I think so, but if anybody  check it perhaps exist an small change.
But like I said, I only removed the items that are configurable via the docker args.

And I din't copy also, the pg_hba.conf, but I think that don't need to affect the bench results.
https://github.com/joanhey/FrameworkBenchmarks/blob/fb26873bd518f5a32427cfd542d705b105213a51/toolset/databases/postgres/pg_hba.conf

Who made the changes in the old config ?
https://github.com/joanhey/FrameworkBenchmarks/blame/fb26873bd518f5a32427cfd542d705b105213a51/toolset/databases/postgres/postgresql.conf

joanhey:
Mysql use an almost default config, not like PostgreSQL.
And PostgreSQL also use a kernel optimization:
https://github.com/joanhey/FrameworkBenchmarks/blob/fb26873bd518f5a32427cfd542d705b105213a51/toolset/databases/postgres/60-postgresql-shm.conf

That I'll add after a run to Mysql too.
https://github.com/TechEmpower/FrameworkBenchmarks/pull/8003#issuecomment-1466867043

joanhey:
Or almost try to use their recommendations:

https://www.percona.com/search?search=mysql%20conf%20optimized
https://www.percona.com/search?search=postgre%20conf%20optimized

volyrique:
Another issue that I have noticed is that the number of frameworks that have failed to start has jumped significantly as well, and many of them (e.g. `ntex-db`) use PostgreSQL, but that might just be a coincidence. Otherwise I have compared [the old](https://github.com/TechEmpower/FrameworkBenchmarks/blob/060ae5cc4e836098efa1f8fb44bd7cf1d18a64f9/toolset/databases/postgres/postgresql.conf) and [the new](https://github.com/TechEmpower/FrameworkBenchmarks/blob/d382c07c517e9b291df84d6bb0f3cfc74eefaa3d/toolset/databases/postgres/postgresql-min.conf) configuration files, and, as far as I can tell, there is virtually no difference; `max_connections` and `wal_level` in particular have the same values in both cases. The old `pg_hba.conf` file just used default settings, I believe, except that there was one entry to allow remote connections. So, it looks the only significant change is the move from Ubuntu 22.04 to Debian 11, which might mean differences in the versions of the system libraries and possibly the way they are built.

volyrique:
Only the multiple queries and database updates tests are affected - the rest just exhibit the normal run-to-run variation/noise. E.g. both `drogon-core` and `may-minihttp` regressed by 5-6%, which is significant, and it is not just them.

joanhey:
Yes, there are frameworks that have patched libpg, only for more performance.
But don't change their frameworks to use the new auth system, instead using the flawed md5 ?

PR #8003, opened 2023-03-09
[toolset] Use Mysql official docker image

joanhey:
Use Debian 11 (no Ubuntu image also)

### DIFF
--- a/toolset/databases/mysql/mysql.dockerfile
+++ b/toolset/databases/mysql/mysql.dockerfile
@@ -1,46 +1,9 @@
-FROM ubuntu:22.04
+FROM mysql:8.0-debian
 
-ARG MYSQL_VERSION=8.0
+ENV MYSQL_ROOT_PASSWORD=root
+ENV MYSQL_USER=benchmarkdbuser
+ENV MYSQL_PASSWORD=benchmarkdbpass
+ENV MYSQL_DATABASE=hello_world
 
-COPY create.sql /tmp/
-COPY my.cnf ./
-
-ARG DEBIAN_FRONTEND=noninteractive
-ADD "https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x859be8d7c586f538430b19c2467b942d3a79bd29" \
-    /etc/apt/keyrings/mysql.asc
-RUN chmod 644 /etc/apt/keyrings/mysql.asc && \
-    apt-get -yqq update && \
-    apt-get -yqq install \
-      apt-utils \
-      locales \
-      lsb-release && \
-    echo "deb [ signed-by=/etc/apt/keyrings/mysql.asc ] http://repo.mysql.com/apt/ubuntu $(lsb_release -cs) mysql-${MYSQL_VERSION}" > \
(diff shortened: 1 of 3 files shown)


toolset/databases/mysql/mysql.dockerfile
```diff
@@ -1,46 +1,9 @@
-FROM ubuntu:22.04
+FROM mysql:8.0-debian
```
volyrique, line 1: Can you switch to `mysql:8-oracle` - it supports AArch64 as well?

joanhey: We don't need AArch64 now.

But more important, is that all the toolset use APT packages, and I don't want to change that.

volyrique: We don't need it for the continuous benchmarking or CI set up by the TechEmpower team, but people might be running the benchmarks in other contexts. For example, the .NET developers at Microsoft are [using](https://devblogs.microsoft.com/dotnet/arm64-performance-improvements-in-dotnet-7/) them to track AArch64 performance.

IMHO since the change is so simple, it is worth making it, so that parts of the community don't need to apply patches downstream. After your changes (which look great BTW) the usage of `apt` is completely transparent, since the Dockerfile does not run a package manager directly.

Anyway, that's my two cents.

joanhey: I think that use Oracle Linux is NOT a good way.

More interesting is start a discussion about using Mariadb, instead or alternative to Mysql.

The Maridb docker image use Ubuntu 22.04, and it supports 4 architectures.

volyrique: > I think that use Oracle Linux is NOT a good way.

From my point of view I don't see a real difference. On the other hand replacing MySQL with MariaDB sounds like a good idea, especially from a platform support perspective.

joanhey:
Another detail, is that the Mysql image don't have a kernel optimization like PostgreSQL.

Firs we will run so, to check the difference between our image and the official.
And later we will try to add the kernel optimization, like in PostgreSQL.

volyrique:
Are you referring to the [60-postgresql-shm.conf](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/databases/postgres/60-postgresql-shm.conf) file? Is it even relevant to MySQL?

joanhey:
Of course:

- https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/managing_monitoring_and_updating_the_kernel/adjusting-kernel-parameters-for-database-servers_managing-monitoring-and-updating-the-kernel
- https://www.linkedin.com/pulse/configuring-huge-pages-mysql-server-red-hat-linux-juan-soto/
- https://github.com/perusio/high-performance-web-kernel-parameters/blob/master/dbserver/05-mysql.conf
- https://www.cyberciti.biz/tips/linux-hugetlbfs-and-mysql-performance.html
- ....

joanhey:
Possibly is for that, than PostgreSQL is a lot of faster in the bench.
We will know it shortly.

PR #8005, opened 2023-03-09
Updates to aspnetcore PG database access

roji:


roji:
@DamianEdwards pushed your changes from https://github.com/aspnet/Benchmarks/pull/1801 into this, can you please take a look?

Will leave this open for some last perf experiments I want to do in the coming days.

sebastienros (approved):

roji:
> Will leave this open for some last perf experiments I want to do in the coming days.

These aren't ready to be integrated here. Let's merge this PR, I'll continue experimenting and possibly submit a separate one later.

benaadams (approved):

roji:
Note: added another commit here to disable SQL parsing/rewriting (see https://github.com/aspnet/Benchmarks/pull/1590#issuecomment-1474907586).

frameworks/CSharp/aspnetcore/PlatformBenchmarks/Program.cs
```diff
@@ -14,6 +14,14 @@ public static async Task Main(string[] args)
     {
         Args = args;
 
+#if NPGSQL
+        // This disables SQL parsing/rewriting, which requires using positional parameters and NpgsqlBatch everywhere.
+        // This helps commands where there are no parameters (Fortunes); when there are parameters, their ParameterName
+        // being null already triggers positional parameters and disables parsing)
+        // Note that Dapper and EF aren't yet compatible with this mode.
+        AppContext.SetSwitch("Npgsql.EnableSqlRewriting", false);
+#endif
```
DamianEdwards, line 23: Is this something we have enabled in our copy of the benchmark, or is it not required there for other reasons?

roji: We can't do it in our copy since we have Dapper and EF, which this is incompatible with... This one is tricky, since it's an AppContext switched cached at startup, but we don't know whether the application will be used for raw, dapper or EF.

This could be an #if (we already have #if DATABASE, #if NPGSQL vs. MYSQLCONNECTOR)...

DamianEdwards: I see. Ideally this would be a property that be set directly on the command. What kind of gains are we talking about by setting it here?

roji: See https://github.com/aspnet/Benchmarks/pull/1590#issuecomment-1475291022.

I did a measurement a long time ago and it wasn't very significant, but we're at the point where every little bit helps (I can measure again at some point).

DamianEdwards: Using `AppContext` switches makes me slightly uneasy as they're usually reserved in .NET for restoring unsupported behaviors, rather than opting-in to legitimate optimizations. I think this is one of those things that's very much in the contentious or at least debatable zone WRT to keeping in the spirit of the benchmark.

DamianEdwards: We discussed this offline and I'm OK with this going in now.

DamianEdwards (approved):

sebastienros:
@nbrady-techempower ready to merge, thanks

PR #8007, opened 2023-03-09
drop permissions of container to match host system

gdams:
When you run Techempower as a non-elevated user, it will create report files that that the user cannot clean up.

This PR ensures that the Group/User IDs match the host systems.


gdams:
Hi @nbrady-techempower - could you please review/merge (or tag an appropriate reviewer)? Thanks so much!

PR #8010, opened 2023-03-11
[toolset] Tfb dockerfile update to Ubuntu22.04

joanhey:
Possibly the last version to use Python2. Debian 11 has removed Python2 .


joanhey:
Looks like it exist a problem with #8007

Dockerfile
```diff
@@ -1,8 +1,10 @@
-FROM ubuntu:18.04
+FROM buildpack-deps:jammy
```
volyrique, line 1: Couldn't we use `ubuntu:22.04` here and add a few packages to the `apt-get install` command? `buildpack-deps` is a significantly slower download; it also adds much more development packages than we need.

Dockerfile
```diff
@@ -13,21 +15,27 @@ RUN apt-get -yqq update && apt-get -yqq install \
       git-core \
       libmysqlclient-dev          `# Needed for MySQL-python` \
       libpq-dev \
-      python-dev \
-      python-pip \
+      python2 \
+      python2.7-dev \
+      curl \
```
volyrique, line 20: Nit - could you, please, keep the packages in alphabetical order?

Dockerfile
joanhey: The MySQL-python need compilation, so I need to use it.
image

volyrique: Yes, I updated my comment - you just need to add `gcc`.

joanhey: Updated tu Ubuntu 22.04.
What we gain in one place, we lose it the other.

joanhey:
Ready !!

volyrique:
IMHO you should remove the commented out code and order the packages as I suggested before.

joanhey:
The commented code will be there till is running correctly.

The docker suggestions, are not useful in the bench, like use less layers.
With that we loose the docker cache, and need more time to build the images.

Example:
Your last PHP-H2O dockerfile need more than 4 minutes to build.
My last dockerfile use less than 2 min.

If only 300 frameworks use an extra 2 min (more use a lot of more):  300 * 2 min = 600 min = 10 hours.
More important than the failing frameworks, a lot fail in the build stage in 20-50 s.

I think that we need to create an Issue, to create a list with the build time of each framework.
So we can check the frameworks than use more time, and check them.

joanhey:
The continuous bench, use docker cache.
But only an small change break it.

Not only that, I have seen how almost all dockerfiles using PHP use the first one (from Cakephp) and if change anything (for example a PHP update, OS lib, ...) break a lot of the rest.
Only an small change in the dockerfile, break the docker cache.

Is for that than I want a list of build times for the frameworks.

volyrique:
> Last dockerfile using 2 min:

It is [actually](https://tfb-status.techempower.com/unzip/results.2023-03-01-21-49-10-529.zip/results/20230223190414/php-h2o/build/php-h2o.log) 2 minutes and 39 seconds, so more like 3 minutes, and yes, it is not the same `h2o` version, so not comparable.

joanhey:
After check my last php-h2o I can change it to built in 1min.

But will be better than your try to cache the H2o build for H2o, H2o-ruby, H2o-php and H2o-cr.

Thank you.

joanhey:
My last word about this stupid conversation.
Bye.

PR #8011, opened 2023-03-11
[php] Update Cakephp to PHP 8.2

joanhey:
#7703


joanhey:
Confirmed the problem with PR #8007 

https://github.com/TechEmpower/FrameworkBenchmarks/actions/runs/4392558706/jobs/7692392335#step:7:1068

joanhey:
We need to revert PR #8007

PR #8013, opened 2023-03-11
[toolset] 🔥 Revert "drop permissions of container to match host system (#8007)"

joanhey:
@nbrady-techempower fail in GH actions and in local

https://github.com/TechEmpower/FrameworkBenchmarks/actions/runs/4392558706/jobs/7692392335#step:7:1068
https://github.com/TechEmpower/FrameworkBenchmarks/actions/runs/4392721211/jobs/7692665726#step:7:1068
https://github.com/TechEmpower/FrameworkBenchmarks/actions/runs/4393389504/jobs/7693834515#step:7:1077

In local fail with permission denied, when building the wrk dockerfile.

Affected PRs: #8010 #8011 #8012 #8014 #8015 #8016 #8017 #8020

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,5 @@
 FROM ubuntu:18.04
 
-ARG USER_ID
-ARG GROUP_ID
 ARG DEBIAN_FRONTEND=noninteractive
 # WARNING: DON'T PUT A SPACE AFTER ANY BACKSLASH OR APT WILL BREAK
 # One -q produces output suitable for logging (mostly hides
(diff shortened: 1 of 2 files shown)


joanhey:
Perhaps we need to add `dockerfile` (tfb) and `tfb` to GH actions, to run the tests.

nbrady-techempower:
@joanhey good point. for some reason, i thought we were doing it that way and that's why i accepted those PRs. good catch

gdams:
What can I do to get these changes back in? More than happy to diagnose the issues

PR #8014, opened 2023-03-11
Improve the axum entry

jplatte:
Mainly: Upgrade dependencies, use state instead of extensions.

### DIFF
--- a/frameworks/Rust/axum/Cargo.lock
+++ b/frameworks/Rust/axum/Cargo.lock
@@ -95,7 +95,7 @@ dependencies = [
  "tokio-pg-mapper-derive",
  "tokio-postgres",
  "tower",
- "tower-http",
+ "tower-http 0.4.0",
  "yarte",
 ]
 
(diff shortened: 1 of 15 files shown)


jplatte:
No idea what's broken in CI, seems to be about Docker which I didn't modify.

joanhey:
Yes, the toolset is failing. Check #8013

volyrique:
@jplatte You can also rebase on top of 110b441205b2b62bb6eeb5ebcbc37ac6c95db09a.

jplatte:
@volyrique That commit is already part of the history on my branch.

volyrique:
The point is that that commit should be the last commit before your changes - right now your PR is on top of a broken commit.

jplatte:
Okay. Just doing `git rebase 110b441` would be a no-op, so I was a bit confused what you meant.

jplatte:
I suspect it will still fail though. The `checkout` action by default checks out a merge commit, not the PR head for running tests.

PR #8018, opened 2023-03-12
[micronaut] Remove repeated tests [ci skip]

joanhey:
Remove repeated `json` and `plaintext`, from 9 database variants.
They use unnecessary  time and resources.

image

PD: possibly more frameworks do the same.

### DIFF
--- a/frameworks/Java/micronaut/benchmark_config.json
+++ b/frameworks/Java/micronaut/benchmark_config.json
@@ -26,8 +26,6 @@
         "versus": "None"
       },
       "vertx-pg-client-graalvm": {
-        "json_url": "/json",
-        "plaintext_url": "/plaintext",
         "db_url": "/db",
         "query_url": "/queries?queries=",
         "fortune_url": "/fortunes",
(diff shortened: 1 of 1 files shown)


joanhey:
@sdelamo perhaps you want to maintain 1 separate for vertx, graalvm,...

sdelamo:
ping @dstepanov

dstepanov:
It makes sense, but keep "vertx-pg-client-graalvm" so we have graalvm results

joanhey:
Ready

dstepanov:
LGTM

dstepanov:
Maybe we can rename "vertx-pg-client-graalvm" to `micronaut graalvm` for consistency

joanhey:
If we change the name, we'll loose the history.
But we can change the display name, that now is `Micronaut Vertx PG Client GraalVM`.

dstepanov:
What history? I don't think there has been a "release" with this version.

joanhey:
image

I can change it, like you want.

dstepanov:
Please do

PR #8021, opened 2023-03-13
drop permissions of container to match host system

gdams:
@joanhey @nbrady-techempower I've fixed the failing GH action


gdams:
@joanhey can I see the error you're seeing?

joanhey:
image

joanhey:
Test any framework in local.

joanhey:
You can also test in GH actions, only add `[ci lang-only Pascal]` in the commit name.

volyrique:
I am also able to reproduce the same issue on Ubuntu 22.04 while running Docker as an unprivileged user (that has been added to the `docker` group).

@gdams, note that it seems that the GitHub actions don't really test anything when changing the top-level Dockerfile and/or the `tfb` script, so we don't see any issue here. I suggest changing `php-h2o` to use `1c8407e3679a4e94dd77c2755b62b1ae48613bd6` as `H2O_VERSION` in [the Dockerfile](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/PHP/php/php-h2o.dockerfile), so that your changes cover at least one implementation.

Dockerfile
```diff
@@ -29,4 +31,9 @@ RUN apt-get -yqq update && apt-get -yqq install \
 
 ENV FWROOT=/FrameworkBenchmarks PYTHONPATH=/FrameworkBenchmarks
 
+# Drop permissions of user to match those of the host system
+RUN addgroup --gid $GROUP_ID user
```
volyrique, line 35: You should add the group and the user only if they do not exist already - for example, on my macOS `13.1 (22C65)` system with Docker Desktop version `4.17.0 (99724)` this fails with the following error:
```
[+] Building 1.5s (6/7)                                                                                                                                                                                                                                                       
 => [internal] load build definition from Dockerfile                                                                                                                                                                                                                     0.0s
 => => transferring dockerfile: 1.40kB                                                                                                                                                                                                                                   0.0s
 => [internal] load .dockerignore                                                                                                                                                                                                                                        0.0s
 => => transferring context: 2B                                                                                                                                                                                                                                          0.0s
 => [internal] load metadata for docker.io/library/ubuntu:18.04                                                                                                                                                                                                          1.2s
 => [1/4] FROM docker.io/library/ubuntu:18.04@sha256:1e32b9c52e8f22769df41e8f61066c77b2b35b0a423c4161c0e48eca2fd24f75                                                                                                                                                    0.0s
 => CACHED [2/4] RUN apt-get -yqq update && apt-get -yqq install       -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold"       cloc       dstat                       `# Collect resource usage statistics`       git-core       libmysqlclient  0.0s
 => ERROR [3/4] RUN addgroup --gid 20 user                                                                                                                                                                                                                               0.2s
------                                                                                                                                                                                                                                                                        
 > [3/4] RUN addgroup --gid 20 user:
#6 0.222 addgroup: The GID `20' is already in use.
------
executor failed running [/bin/sh -c addgroup --gid $GROUP_ID user]: exit code: 1
Unable to find image 'techempower/tfb:latest' locally
docker: Error response from daemon: pull access denied for techempower/tfb, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.
```
FYI `id -u` and `id -g` return `501` and `20` respectively in my environment.

gdams: I've switched this to use a check first

gdams:
okay @joanhey I think I've fixed this now, let me know if it works on your system?

joanhey:
I said use `[ci lang-only Pascal]` in the commit, because is built in 6 min.

Change in PHP/PHP need about 30 min, and possibly will be marked it as failed.

gdams:
> I said use [ci lang-only Pascal] in the commit, because is built in 6 min.
> Change in PHP/PHP need about 30 min, and possibly will be marked it as failed.

okay I've amended the commit name

joanhey:
The PHP CI is still running :smile: 

I will test it in local.

Dockerfile
```diff
@@ -29,4 +31,13 @@ RUN apt-get -yqq update && apt-get -yqq install \
 
 ENV FWROOT=/FrameworkBenchmarks PYTHONPATH=/FrameworkBenchmarks
 
+# Check if Group is already created
+RUN if ! getent group $GROUP_ID; then \
+      addgroup --gid $GROUP_ID user; \
+    fi
+
+# Drop permissions of user to match those of the host system
+RUN adduser --disabled-password --gecos '' --uid $USER_ID --gid $GROUP_ID user
```
volyrique, line 40: IMHO you should check if the user exists already too.

gdams: As long as you're depending on the `ubuntu` images you'll never find a user called `user`.

gdams: Or do you mean check if the UID exists?

volyrique: It is not the user name that would be the problem, but the user ID.

gdams: how about I check it like this?

```bash
RUN if ! getent passwd $USER_ID; then \
      adduser --disabled-password --gecos '' --uid $USER_ID --gid $GROUP_ID user; \
    fi
```

gdams: Hmm actually the next USER step will fail if the user hasn't been created

volyrique: Looks good to me.

gdams: okay I think I have a fix for that edge-case, see what you think @volyrique

volyrique: That is better, but I think that it can actually be simplified - apparently the `USER` instruction [accepts](https://docs.docker.com/engine/reference/builder/#user) UIDs (and GIDs) as well. So, keep the check, but remove the `else` case, and change the subsequent instruction to:
```
USER $USER_ID
```
If I am not mistaken, that should cover all cases - we don't actually care about the actual username.

gdams: Updated PTAL

gdams:
@joanhey @nbrady-techempower this should be good to go now

volyrique:
Unfortunately I still see the `wrk` issue that @joanhey mentioned before. On macOS the toolset no longer produces the error I talked about previously, and fails in the same way instead. @gdams, how exactly are you testing this? Presumably we should see the same problems.

gdams:
> Unfortunately I still see the wrk issue that @joanhey mentioned before. On macOS the toolset no longer produces the error I talked about previously, and fails in the same way instead. @gdams, how exactly are you testing this? Presumably we should see the same problems.

@volyrique ofcourse, the challenge here was that `/var/run/docker.sock` was not owned by the non-root user. I've fixed this by using `chown` and then using [gosu](https://github.com/tianon/gosu) to drop the permissions.

gdams:
hmm actually locally I'm seeing the following error now:

```output
techempower/mysql:latest: error: 'Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)'
techempower/mysql:latest: Check that mysqld is running and that the socket: '/var/run/mysqld/mysqld.sock' exists!
techempower/mysql:latest: mysqladmin: [Warning] Using a password on the command line interface can be insecure.
techempower/mysql:latest: mysqladmin: connect to server at 'localhost' failed
````

gdams:
@volyrique how about now?

volyrique:
It works for me now, both on Linux and macOS.

joanhey:
I don't know if it will affect the new #8036 tfb dockerfile cache.

So I think that will be merged first #8036, so we can test later than the permissions work.

joanhey:
In the GH actions is not needed to touch the user.
Perhaps will be better execute it if we pass the ARGS.

gdams:
> In the GH actions is not needed to touch the user.
Perhaps will be better to execute it if we pass the ARGS.
 
Is it not better to test in GH actions in the same way that your end users would?

gdams:
> So I think that will be merged first https://github.com/TechEmpower/FrameworkBenchmarks/pull/8036, so we can test later than the permissions work.

I've rebased your changes so we should be good to go here

PR #8028, opened 2023-03-14
Update to anansi 0.14.1

saru-tora:


frameworks/Rust/anansi/anansi-raw.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:1.64
+FROM rust:latest
```
nbrady-techempower, line 1: Please keep these tied down to a specific version.

saru-tora: Done.

PR #8036, opened 2023-03-15
GH actions cache tfb dockerfile build

joanhey:
Now use 15-21 s.


joanhey:
We only need to delete old cache builds, as the max cache is 10GB per repository.

joanhey:
Later will be good to try it, with the wrk dockerfile build.

joanhey:
There are 2 PRs open that touch the tfb dockerfile, so we will test that work correctly.

PR #8057, opened 2023-03-21
[mORMot] speed up  pipelining; speed up docker build

pavelmash:
mORMot using a modified libpq that allows to send the PipelineSync command without flushing prev. commands to socket

### DIFF
--- a/frameworks/Pascal/mormot/src/raw.pas
+++ b/frameworks/Pascal/mormot/src/raw.pas
@@ -193,7 +193,7 @@ begin
      hsoIncludeDateHeader  // required by TPW General Test Requirements #5
     ] + flags);
   fHttpServer.HttpQueueLength := 10000; // needed e.g. from wrk/ab benchmarks
-  fHttpServer.ServerName := 'M'; // better plaintext performance
+  fHttpServer.ServerName := 'mmt';
   // use default routing using RTTI on the TRawAsyncServer published methods
   fHttpServer.Route.RunMethods([urmGet], self);
   // wait for the server to be ready and raise exception e.g. on binding issue
--- a/frameworks/Pascal/mormot/src/raw.pas
+++ b/frameworks/Pascal/mormot/src/raw.pas
@@ -193,6 +193,7 @@ begin
      hsoIncludeDateHeader  // required by TPW General Test Requirements #5
     ] + flags);
   fHttpServer.HttpQueueLength := 10000; // needed e.g. from wrk/ab benchmarks
+  fHttpServer.ServerName := 'M'; // better plaintext performance
   // use default routing using RTTI on the TRawAsyncServer published methods
   fHttpServer.Route.RunMethods([urmGet], self);
   // wait for the server to be ready and raise exception e.g. on binding issue
--- a/frameworks/Pascal/mormot/src/raw.pas
+++ b/frameworks/Pascal/mormot/src/raw.pas
@@ -233,6 +233,7 @@ begin
     pStmt.SendPipelinePrepared;
     pConn.PipelineSync;
   end;
+  pConn.Flush; // we use modified libpq what not flush inside PQPipelineSync - flush manually
   for i := 0 to cnt - 1 do
   begin
     pStmt.GetPipelineResult;
--- a/frameworks/Pascal/mormot/src/raw.pas
+++ b/frameworks/Pascal/mormot/src/raw.pas
@@ -193,7 +193,7 @@ begin
      hsoIncludeDateHeader  // required by TPW General Test Requirements #5
     ] + flags);
   fHttpServer.HttpQueueLength := 10000; // needed e.g. from wrk/ab benchmarks
-  fHttpServer.ServerName := 'mmt';
+  fHttpServer.ServerName := 'M';
   // use default routing using RTTI on the TRawAsyncServer published methods
   fHttpServer.Route.RunMethods([urmGet], self);
   // wait for the server to be ready and raise exception e.g. on binding issue
--- a/frameworks/Pascal/mormot/mormot.dockerfile
+++ b/frameworks/Pascal/mormot/mormot.dockerfile
@@ -11,8 +11,7 @@ RUN /bin/bash -c ./setup_and_build.sh
 
 FROM ubuntu:22.04
 COPY --from=builder /build/bin/fpc-x86_64-linux/raw /usr/local/bin/raw
-
-RUN apt-get update && apt-get install -yqq postgresql-client
+COPY libpq.so.5.16 /usr/lib/x86_64-linux-gnu/libpq.so.5
 
 EXPOSE 8080
 CMD ["raw"]
--- a/frameworks/Pascal/mormot/mormot.dockerfile
+++ b/frameworks/Pascal/mormot/mormot.dockerfile
@@ -11,7 +11,7 @@ RUN /bin/bash -c ./setup_and_build.sh
 
 FROM ubuntu:22.04
 COPY --from=builder /build/bin/fpc-x86_64-linux/raw /usr/local/bin/raw
-COPY libpq.so.5.16 /usr/lib/x86_64-linux-gnu/libpq.so.5
+COPY --from=builder /build/libpq.so.5.16 /usr/lib/x86_64-linux-gnu/libpq.so.5
 
 EXPOSE 8080
 CMD ["raw"]
--- a/frameworks/Pascal/mormot/setup_and_build.sh
+++ b/frameworks/Pascal/mormot/setup_and_build.sh
@@ -35,11 +35,13 @@ echo "Download statics from $URL ..."
 wget -qO- "$URL" | tar -xz -C ./libs/mORMot/static
 
 # uncomment for fixed commit URL
-URL=https://github.com/synopse/mORMot2/tarball/c57d48846f050f4f68856758d28cf76450addaf2
+URL=https://github.com/synopse/mORMot2/tarball/edf377a61aa5ac99add92d144480b12095f578c2
 #URL="https://api.github.com/repos/synopse/mORMot2/tarball/$USED_TAG"
 echo "Download and unpacking mORMot sources from $URL ..."
 wget -qO- "$URL" | tar -xz -C ./libs/mORMot  --strip-components=1
 
+# download our modified libpq what do not PQflush inside PQPipelineSync
+wget https://github.com/pavelmash/postgres/releases/download/5.16_PQpipelineSync_noflush/libpq.so.5.16
 
 # uncomment line below to echo commands to console
 set -x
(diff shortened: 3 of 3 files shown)


frameworks/Pascal/mormot/src/raw.pas
```diff
@@ -193,6 +193,7 @@ constructor TRawAsyncServer.Create(
      hsoIncludeDateHeader  // required by TPW General Test Requirements #5
     ] + flags);
   fHttpServer.HttpQueueLength := 10000; // needed e.g. from wrk/ab benchmarks
+  fHttpServer.ServerName := 'M'; // better plaintext performance
```
sebastienros, line 196: This optimization has been disallowed for other frameworks by the past:

> We expect the Server header to be whatever is normal for the platform or framework. If the framework does not normally provide a Server response header, we nevertheless require that you provide one as this roughly normalizes network load across all implementations.

frameworks/Pascal/mormot/src/raw.pas
```diff
@@ -233,6 +234,7 @@ function TRawAsyncServer.GetRawRandomWorlds(cnt: PtrInt; out res: TWorlds): bool
     pStmt.SendPipelinePrepared;
     pConn.PipelineSync;
   end;
+  pConn.Flush; // we use modified libpq what not flush inside PQPipelineSync - flush manually
```
sebastienros, line 237: Sounds like a discussion that concluded by the requirement of SYNC calls. I would suggest you double check before using a modified libpq that prevents them.

frameworks/Pascal/mormot/src/raw.pas
pavelmash: Default Server: header produced by mORMot2 framework is `mORMot2 (Linux)` what is too long. So we override it to short one-letter version, as most other top-rated frameworks did, for example - .net use ["K" instead of "Kestrel"](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/CSharp/aspnetcore/PlatformBenchmarks/BenchmarkApplication.Fortunes.cs#L24)

frameworks/Pascal/mormot/src/raw.pas
pavelmash: I checked. Requirements is to send a SYNC command, but not to flush client socket (do write syscall) after SYNC.
So we modify a lipbq to avoid socket flushing - see [what we did](https://github.com/pavelmash/postgres/commit/41d2dfa21ae39fe591dc0f00e97d2053b492c3cf). 
At last Just-JS do the same (send all pipeline commands in one-two socket write) and it's implementation is valid

frameworks/Pascal/mormot/src/raw.pas
sebastienros: Must have been confused with the `/p` in plaintext routes. Thanks for confirming.

joanhey: It's disallowed to all play with the same rules.
But still not formalized in the rules and verification tests. :(

frameworks/Pascal/mormot/src/raw.pas
joanhey: IMO will be forbidden to use a modified libpq.
We are testing frameworks !!

frameworks/Pascal/mormot/src/raw.pas
pavelmash: @joanhey - so what your suggestion for `Server` header value? 
To be the same as `webserver` in `benchmark_config.json`, or, if it is equal to "None" as for many frameworks - to be  the same as  `framework` ?

frameworks/Pascal/mormot/src/raw.pas
pavelmash: There is a requirement what "each query must be separated by a Sync message", the problem is what in case of `libpq`  PQpipelineSync function, what add  a `S` message into buffer also flush a socket.  This is not what we need for best performance.
Form other hand `libpq` do not export public function for direct message manipulation. So we decide to do a small modification - just  [comment one line](https://github.com/pavelmash/postgres/commit/41d2dfa21ae39fe591dc0f00e97d2053b492c3cf) . 
This do not break any rules.
Many frameworks what implements PG protocol by itself also works in this way.

frameworks/Pascal/mormot/src/raw.pas
joanhey: The problem is not the name, but the minimum number of characters.
4 chars more in the server header, seems not to much. But if we multiply these 4 extra chars for 6M req/s, exist a big difference.

pavelmash: Here is examples of top rated frameworks what use one-letter for `Server:`
 -  may [Server: M](https://github.com/Xudong-Huang/may_minihttp/blob/master/src/response.rs#L108)
 -  .net [Server: K](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/CSharp/aspnetcore/PlatformBenchmarks/BenchmarkApplication.Fortunes.cs#L24)
 - ntex [Server: N](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Rust/ntex/src/utils.rs#L7)
 and so on....

I don't think all of them can quickly change their implementation. 
For *fair play* we also use `Server: M`  for mORMot

frameworks/Pascal/mormot/src/raw.pas
volyrique: I have identified the same performance issue in `h2o`, and I have started working on a proper upstream `libpq` patch, which I hope to post soon. @pavelmash, the commit in your branch is IMHO not good enough to be submitted for review because it changes the semantics of an existing public interface, which has been there since PostgreSQL 14.

The changes in this PR that concern database query pipelining do not break any rule as far as I can tell because the protocol messages are exactly the same before and after the patch. Instead, the changes reduce the number of system calls that are performed for database access, which is particularly important, since the machines used for continuous benchmarking are severely affected by speculative execution vulnerability mitigations, for example.

synopse: @joanhey Some framework have their own PostgreSQL access layer, e.g. for asynchronous efficient process.
It is not forbidden to use another client, and other frameworks are welcome to use the patched libpq.

frameworks/Pascal/mormot/src/raw.pas
joanhey: Yes, I know.
It's for that, than we need to clarify and add rule and verification.

joanhey: For example, with Nginx you can't change the server name.

frameworks/Pascal/mormot/src/raw.pas
joanhey: Why don't send the changes to libpq?
Or why are not accepted ?

volyrique: As I said, I am working on doing that for my own version of the changes.

frameworks/Pascal/mormot/src/raw.pas
pavelmash: `Server` header value is changed to `mmt` by [658165b824] .
@joanhey - hope rule will be changed - I propose to use at last 3 character. For example google respond with `server: gws`, so 4 characters is too much. 
And, for sure, we should add validation for new rule, to ensure all participants are in the same situation

frameworks/Pascal/mormot/src/raw.pas
pavelmash: And while @volyrique  working on contribution to `libpq`  we can accept this PR and verify a performance difference. This will be a good reason for Postgres team to accept such changes to trunc. In any case changing libpq in trunc is a long story...

pavelmash: Not many framework uses libpq at all. But [url=https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/C%2B%2B/lithium/compile_libpq.sh]some of them[/url] patch it.

frameworks/Pascal/mormot/src/raw.pas
cirospaciari: > 

We need to clarify this, today socketify.py uses `Server: uWebSockets_20`, we can remove Server header or add a minimal size like 3 characters, in my case I can change to `Server: uWS`

frameworks/Pascal/mormot/src/raw.pas
cirospaciari: > IMO will be forbidden to use a modified libpq. We are testing frameworks !!

Database benchmarks should be separated, we can rank the fastest databases connections and their respective languages outside HTTP Servers, this would reduce the repetitive benchmarks that basically only benchmark the better connector implementation.

With fewer database tests we should be able to add to HTTP servers benchmarks different payload sizes.

volyrique: FYI my `libpq` patch is available [here](https://www.postgresql.org/message-id/CACV6eE5arHFZEA717%3DiKEa_OewpVFfWJOmsOdGrqqsr8CJVfWQ%40mail.gmail.com).

cirospaciari: > FYI

anyone should use any connectors, after all, optimizations are optimizations. The bad thing is to try to compare WebFrameworks with databases instead of direct testing the database connectors/solutions.

synopse: I don't think DB tests should be separated. Optimizations are indeed optimizations.
The real bad thing is that we compare frameworks which are proofs of concept, not usable in real-purpose code, with production ready code.
For instance, some frameworks use a direct DB connection, but only handle integer fields, so that they could shine in TFB benchmarks. Or others don't allow to change the fortunes template without re-compiling the executable - which IMHO voids the aim of a HTML template, which is to separate view and model.
I would add a requirement with some complex process instead, something with data more complex than the /fortunes test, just to validate the framework is usable for any real usecase.

joanhey: It's better to create a general issue or discussion, than talk in a PR.
We need to try to have a fair play, for all frameworks.
And later create individual issues, for each specific item.

@synopse not only the HTML template need compilation. I see some frameworks than cache the response in Fortunes, so bypassing the template, composition and html scape stage in each request.

PD: @pavelmash you can use only 1 letter server header, till is NOT added as a rule and verification.

pavelmash: @joanhey - reverted by [1a48475f76c033e40fc4ddb3fa7bcbfed0bd0314] to use 1 letter, thanks!
I agree what separate issue should be created for discussion.

Hope this PR will be merged now. At last this can help Postgres team to estimate changes we propose to `libpq` what implemented by @volyrique as [libpq patch](https://www.postgresql.org/message-id/CACV6eE5arHFZEA717%3DiKEa_OewpVFfWJOmsOdGrqqsr8CJVfWQ%40mail.gmail.com). In case Postgres team decline such patch I will rollback usage of modified `libpq`

PR #8062, opened 2023-03-23
Display startup message for Gin (Go)

sebastienros:
My previous attempt didn't work as it was using logging which is disabled in the scenario. This time I used `fmt.Println` and verified it was correctly outputting the text.

image


sebastienros:
`query` doesn't pass the validation even though I haven't touched this part.

PR #8079, opened 2023-03-28
[python] Mark Django as broken

joanhey:
Fail since the end of February.

@cirospaciari @SerhiyRomanov @TonyPythoneer @ZuluPro


cirospaciari:
Pretty old Python and Django versions, if no body fixes it until April 08, I will take a look.

joanhey:
Django test is using the last v4.17

PR #8081, opened 2023-03-28
[php] Mixphp update to PHP 8.2

joanhey:
And update to Ubuntu 22.04.

#7703

@onanying 
Fail to update with:

-  Swoole
   `Uncaught Error: Creation of dynamic property Swoole\Http\Request::$param is deprecated in /mixphp/vendor/mix/vega/src/Router.php on line 128 in /mixphp/src/Error.php:50`
-  Workerman-Mysql  
   `ERROR  Use of "parent" in callables is deprecated in /mixphp/vendor/mix/database/src/Connection.php on line 53 in /mixphp/src/Error.php on line 50`


onanying:
fixed: https://github.com/mix-php/mix/commit/134745869928da5ceedfcc93ff90f2a426aff32d

PR #8095, opened 2023-03-31
Micronaut Framework 3.8.8

sdelamo:


volyrique:
I think that renaming `micronaut-vertx-pg-client-graalvm.dockerfile` inside `frameworks/Java/micronaut` to `micronaut-graalvm.dockerfile` should fix the test failure.

PR #8110, opened 2023-04-01
Salvo 0.38 and use moka for cache

chrislearn:
Moka is a memory cache library, use it for compliance.


frameworks/Rust/salvo/benchmark_config.json
```diff
@@ -47,6 +46,39 @@
...
+        "orm": "Raw",
+        "platform": "Rust",
+        "webserver": "Hyper",
+        "os": "Linux",
+        "database_os": "Linux",
+        "display_name": "Salvo [PG]",
+        "notes": "",
+        "versus": ""
+      },
+      "cache-diesel": {
```
volyrique, line 64: Is there a particular reason to introduce permutations specifically for the cached queries test instead of expanding `diesel` and `pg`? That is increasing the runtime quite a lot.

frameworks/Rust/salvo/salvo-diesel.dockerfile
```diff
@@ -1,12 +1,9 @@
-FROM rust:1.62.1
-
-RUN apt-get update -yqq && apt-get install -yqq cmake g++
+FROM rust:latest
```
volyrique, line 1: You should use a fixed version; similarly for the rest of the Dockerfiles.

PR #8122, opened 2023-04-03
[Python] CherryPy: update to v18.8.0 and disable logs

remittor:
Fix from author: https://github.com/cherrypy/cherrypy/issues/1836#issuecomment-570950575


remittor:
Errors during verification are caused only by congestion of the СI.
```
Accessing URL http://tfb-server:8080/db: 
VERIFYING QUERY COUNT FOR http://tfb-server:8080/db
ERROR:root:Terminating process 2089 by timeout of 20 secs.
   FAIL for http://tfb-server:8080/db
     Only 496 executed queries in the database out of roughly 512 expected.
     See https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#specific-test-requirements
   FAIL for http://tfb-server:8080/db
     Only 487 rows read in the database out of roughly 512 expected.
     See https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#specific-test-requirements
{'Date': 'Mon, 03 Apr 2023 20:01:14 GMT', 'Content-Length': '33', 'Content-Type': 'application/json', 'Server': 'CherryPy/18.8.0'}
{"id": 3657, "randomNumber": 218}
```

Perhaps it is worth reducing the requirements for СI.

remittor:
I restarted project verification when America and Europe are still sleeping.
Verification was successful !!!

**Therefore, it is necessary to reduce the requirements for tests if they are run on СI.**

nbrady-techempower:
> Therefore, it is necessary to reduce the requirements for tests if they are run on СI.

I have not seen this enough times to do that but I'll keep an eye on it. I would rather have to rerun once or twice than build a bespoke verification case just for CI. That would undoubtedly lead to cases where verification passes CI but not on citrine.

PR #8130, opened 2023-04-05
[php] Enable JIT

joanhey:
PHP JIT is not enabled.

By default the `ppa:ondrej/php`, have disabled JIT. Even when you enable it, in your `php.ini`. I found it when creating tests for ngx-php, and later checking the benchmark, are also disabled.
Because some code have segfaults. I'll create an Issue in Odrej repo, because it's OK to be disabled by default, but we need to enable it, in dockerfile without a `sed`, for frameworks/apps that work without problems.

So I tested in benchmark mode first in local now, and I'll enable selectively in some fws.
Using function mode, not tracing.
Now only `Workerman-jit` and `php-ngx-pgsql`, to check the difference.

### DIFF
--- a/frameworks/PHP/php-ngx/php-ngx-pgsql.dockerfile
+++ b/frameworks/PHP/php-ngx/php-ngx-pgsql.dockerfile
@@ -32,7 +32,7 @@ RUN sed -i "s|app.php|app-pg.php|g" /deploy/nginx.conf
 
 RUN export WORKERS=$(( 4 * $(nproc) )) && \
     sed -i "s|worker_processes  auto|worker_processes $WORKERS|g" /deploy/nginx.conf
-
+RUN sed -i "s|opcache.jit=off|opcache.jit=function|g" /etc/php/8.1/embed/conf.d/10-opcache.ini
 EXPOSE 8080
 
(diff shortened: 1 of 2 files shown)


kitrio:
This is a related link.
PHP JIT segfault issue https://github.com/php/php-src/issues/7817

joanhey:
I have read these Issues, after I found that it was disabled.
By default JIT is always disabled, the developer need  to enable it in the php.ini.

But with the Oerdnj change, we can't enable it.
.../conf.d/10-opcache.ini 
```ini
; configuration for php opcache module
; priority=10
zend_extension=opcache.so
opcache.jit=off
```
Is loaded after the php.ini, and always overwrite the value to off.
We can never enable it, from the php.ini. And without any information that is disabled.
IMO will be better to inform better, before or when you enable it and not always disabled.

I tested in benchmark mode with some PHP frameworks, without any segfault.

In this same benchmark, I saw a lot of changes, because create errors in some frameworks.
Examples:  php_fpm keepalive, persistent db connections, ..... that later were reverted.
Because the problem was in the frameworks, not in the optimal configuration.

The important part, is that we as developers can enable it in the php.ini, for testing or because our code don't fail.

PR #8132, opened 2023-04-06
Switch to the wrk package provided by Ubuntu 22.04

volyrique:
Also, make a small change to `h2o_mruby` to test. Note that previously `wrk` was compiled with a lot of optimizations enabled, but since the bottlenecks in the tests are usually elsewhere, I don't expect any significant performance regressions.

### DIFF
--- a/toolset/wrk/wrk.dockerfile
+++ b/toolset/wrk/wrk.dockerfile
@@ -1,33 +1,13 @@
-ARG UBUNTU_VERSION=jammy
-
-FROM "buildpack-deps:${UBUNTU_VERSION}" AS compile
-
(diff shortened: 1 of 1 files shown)


volyrique:
@nbrady-techempower, apparently I have failed to limit the testing just to `h2o_mruby`, so the failures are the pre-existing ones, i.e. the ones that are visible in the continuous benchmarking results.

joanhey:
:+1:  but please, try to not mix in the same PR, changes in the toolset with frameworks.

volyrique:
I moved the `h2o_mruby` change to PR #8134 because it did not serve its purpose here anyway.

PR #8133, opened 2023-04-07
xitca-web-wasm: Add an implementation of the JSON serialization test

volyrique:
@fakeshadow, I would appreciate a look.


fakeshadow:
Thanks. LGTM

PR #8160, opened 2023-04-19
Roda

timuckun:


timuckun:
Upgraded the roda-sequel benchmarks to the latest version of ruby and the various gems in the project.

frameworks/Ruby/roda-sequel/.devcontainer/Dockerfile
nbrady-techempower, line 1: Can you remove the files in this folder. They look like they don't pertain to the actual test implementation.

timuckun:
I could do that but they are helpful when developing. They allow me to develop the project without having to install ruby and associated gems on my own machine. I thought they might be useful for others who want to work on the project.

I can delete them if you want though...

timuckun:
Removed the devcontainer as requested.

nbrady-techempower:
@timuckun Thanks. I appreciate you looking to do that, but we don't want to store that here. It's possible it doesn't work well for others and then they're opening up issues about it here, etc, etc. The only things we want in this repo are the items necessary for the test implementations to work.

timuckun:
No worries.

PR #8161, opened 2023-04-19
Micronaut R2DBC: Set `initialSize` of the pool to 0

dstepanov:
It looks like there is a bug in R2DBC Pool that uses one connection when `initialSize` and `maxSize` are equal.

https://github.com/micronaut-projects/micronaut-data/issues/2136


graemerocher (approved):

PiotrDuz:
Just want to point to my answer which shows the cause of the problem, and also possible solution: 
https://github.com/micronaut-projects/micronaut-data/issues/2136#issuecomment-1517531162

PR #8177, opened 2023-04-25
Lithium delete repeated json and plaintext

joanhey:
in db variants

@matt-42


joanhey:
After check it, a lot of frameworks are **repeating json and plaintext** tests.

Created issue #8178

Issue #8178, opened 2023-04-25
A lot of repeated json and plaintext tests


joanhey:
After check it, a lot of frameworks are **repeating json and plaintext** tests.
Some are for different platforms or libs, and it's OK.
**But many only because have them in database variants.**

**Plaintext repeated:**

- [ ]  Sinatra **14**
- [ ]  Officefloor  **11** @sagenschneider 
- [ ]  Falcon  **11** @remittor 
- [ ]  Fastapi **11** @micaelmalta 
- [ ]  Chi  **10** @0uep 
- [ ]  Flask  **7** @remittor 
- [ ]  Roda  **6**  @timuckun  #8160
- [ ]  Nestjs **6** @JHyeok 
- [ ]  ...

Work in progress.

remittor:
![image](https://user-images.githubusercontent.com/9350115/234288762-f3b2d446-6219-419e-96ac-f262550d45c4.png)

@joanhey , do I need to remove only test `falcon [meinheld-orjson]` ?
All other tests have different http servers.

remittor:
![image](https://user-images.githubusercontent.com/9350115/234290188-8ae140a9-1e76-4d9d-a6aa-f8a761f922a8.png)

@joanhey , `PyPy` variants (`flask-raw` & `flask`) already marked as `broken` : https://github.com/TechEmpower/FrameworkBenchmarks/pull/8169/commits/f8e8239a242fa57a40c727ab25bcfcac199afee0#diff-bf7a33c9cd47aa475c2802732158ab4d7009ed11dafefea8a6114204558925d3R112

joanhey:
@remittor orjson looks like a different json lib, but the plaintext and fortunes don't use json.
 
About flask:
![image](https://user-images.githubusercontent.com/249085/234306188-d042b8a7-61f1-4b3f-81f3-b905a75fe88a.png)

![image](https://user-images.githubusercontent.com/249085/234306562-50ba56d5-a254-4757-a3da-f8b07c508f41.png)

joanhey:
I don't know the difference between Sinatra and Sinatra Sequel.

But in the plaintext test, the results are identical.

![image](https://user-images.githubusercontent.com/249085/234402305-70252d76-b451-46dd-9f71-47a6a65931e4.png)
![image](https://user-images.githubusercontent.com/249085/234402414-20ef4112-6c29-4e90-80b1-0f5e57fb3d3f.png)

[...]

Issue #8137, opened 2023-04-09
[Python] Falcon unification issues


cirospaciari:
@remittor Thanks for unifying the Falcon code, I really appreciate that and it's the right move for sure. Just opened this issue so we can chat about it.

https://github.com/TechEmpower/FrameworkBenchmarks/pull/8111

This PR has some issues:
Falcon does not need to add Server and Date when using socketify and it adds unnecessary overhead, checking global scope is slow on CPython (but acceptable on PyPy).
The servers should emit these headers or a simple wrapper only on apps that need that to emit this would be a better solution, and fairer with servers that already include this natively.

Performance diffs for socketify WSGI on CPython:
Before: 433,316
After: 356,027

remittor:
> checking global scope is slow on CPython (but acceptable on PyPy).
> socketify WSGI on CPython:  Before: 433,316  After: 356,027

I think that the drop in speed is not caused by the addition of the common function `add_ext_headers`.
It is very likely that the drop in speed caused the downgrade of `CPython` to version 3.9 (previously used version 3.10).

Speeds for Falcon+FastWSGI: before: 497,777 after: 496,506

cirospaciari:
For socketify CPython can run on 3.11 the latest version (with includes a lot of performance improvements) I think all servers can run on 3.11

remittor:
> I think all servers can run on 3.11. ( and soon 3.12)

I don't mind at all. I just thought that there would be compatibility issues with some modules.
So I suggest you do the following:
1) Convert all `falcon` tests to python 3.10...311.
2) Fix `django` (since you yourself promised to do it)
3) Help to understand the reason for not started `flask` + `uwsgi` (everything works on CI)
4) Help to understand the reason for not working `flask` test `update` (everything works on CI)

cirospaciari:
Agreed, unfortunately, I'm working a lot in my current job, and spending my little free time with my family, soon I should be able to have more time  to work on this

PR #8180, opened 2023-04-25
[Python] Falcon: use 3.11 python and delete repeated tests

remittor:
Issues: #8137, #8178


remittor:
One test did not get the required number of responses because `CI` service was overloaded.

PR #8184, opened 2023-04-27
[Python] FastWSGI: add ASGI support and async DB tests

remittor:
I decided to make a competitor for this project: https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Python/uvicorn


joanhey:
Sorry, I only will ask !!

I manage almost all PHP frameworks, but I don't create all the variants with each framework.

With the plain PHP, of course than I try all variants, but not with each PHP framework.
With a framework I test with the fastest 1-3 different platforms variants, but not with 10 variants for each one.

joanhey:
When we test it with plain PHP, normally the result is the same for all frameworks.

joanhey:
There are more PHP frameworks than with Python or any other language.
If we do the same with the PHP frameworks, we will need 2 weeks for finish each run.

remittor:
> ...but I don't create all the variants with each framework.

I don't add `FastWSGI` server to every `Python` framework. I added `FastWSGI` server only to the most interesting `Python` frameworks.

Total tests in section `PHP.db`: 85
Total time for 'PHP' tests: 21h 51min

Total tests in section `Python.db`: 59
Total time for 'Python' tests: 13h 49min

Total time for 'Java' tests: 23h 46min

joanhey:
Yes, it's what I am saying.

Not for python, but for all.
And like I said PHP have a lot of frameworks, but we don't test it with all servers and platforms.

Any new framework need to be added to the bench. But not each framework with 12 servers and frameworks.
The plain language, can use all variants, and the frameworks chose between 1-3.

joanhey:
And it isn't only about time, but with noise with the results.

joanhey:
Every day this bench is bigger, and together we need to find better ways to be faster and useful.

PR #8201, opened 2023-05-08
go-fiber added scratch docker image, bumped golang version

lospejos:
go-fiber added scratch docker image, bumped Golang version to 1.20


joanhey:
Hi @lospejos 
I understand the advantages of using an scratch docker image.
But in the benchmark, please could you explain the benefit ?
Because we are testing exactly the same binary.

I think that we need to create an issue, about that.

lospejos:
_But in the benchmark, please could you explain the benefit?_

Well, my main concern was to test exact the same things we do use in general / production. Besides, since there are _exact the same binaries_, and `scratch` variant consumes less resources (disk space), why not to use it?

I didn't made thorough tests, but what chances are that when using scratch image we can get less overhead (getting rid of OS context switches, resources consumption of other background processes in OS that require resources etc)?

Anyway, this is just my proposal, and you are free to reject it.

joanhey:
Thank you for the answer.

I didn't say to no use scratch docker, only decide with or without. Because the others frameworks have the same results.
Please check  #8209

Now I'll show you some results, from the actual run:
image

image
image

From last run:
image
image

I understand to try it with one fw, that is well maintained, to know if exist big variations.
But all the results are the same.

Do you think that we need to use the two variants in all frameworks, to have the same results?

nbrady-techempower:
Closing as this seems out of date. Feel free to re-open if you have time to work on it.

PR #8206, opened 2023-05-11
[Python] Bump Granian to 0.4

gi0baro:
SSIA


gi0baro:
@nbrady-techempower this now pass tests

PR #8208, opened 2023-05-12
Update to Vert.x 4.4.2 with IO uring  available on linux + improvements

vietj:
This PR updates the vertx benchmark with Vert.x 4.4.2 and uses IO_Uring when available.

In addition application start can be racy with server start and sometimes the server is up and running before the cached prepared statement are visible to the application leading to runtime errors (NullPointerException). We should start the server after the prepared statement have been created and made visible to the HTTP server threads.

### DIFF
--- a/frameworks/Java/vertx/pom.xml
+++ b/frameworks/Java/vertx/pom.xml
@@ -10,9 +10,10 @@
 		<maven.compiler.target>17</maven.compiler.target>
 		<!-- the main class -->
 		<main.class>vertx.App</main.class>
-		<stack.version>4.3.8</stack.version>
-		<jackson.version>2.14.2</jackson.version>
-		<netty.version>4.1.89.Final</netty.version>
+		<stack.version>4.4.2</stack.version>
+		<jackson.version>2.15.0</jackson.version>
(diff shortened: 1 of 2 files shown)


vietj:
the CI failure seems to be related to the fact that plaintext and json are not executed by the tests ?

joanhey:
The CI tests are making the requests. But for any reason the server don't respond.

vietj:
ok I think I know the issue @joanhey , does the plaintext/json tests run a database ?

vietj:
Issue solves, all green now :-) thanks @joanhey for the tip

joanhey:
Why not use only 1 run, and will be faster to run the tests ?
The bench need to build the docker 2 times, when with 1 will be ready.

vietj:
@joanhey can you elaborate ?

franz1981:
@vietj Probably @joanhey means that instead of using 2 docker configurations for DB and non-DB endpoints, just stick with a single one

vietj:
ah ok, I need to remember why we did two endpoints at some points, perhaps we considered using another database

PR #8219, opened 2023-05-18
[xitca-web] update wasm bench.

fakeshadow:
possible perf improvement by using aot compiled cwasm binary.

other change:
upgrade dep.
re-introduce zero copy parsing for fortune.
reduce use of indirect pointers of query params.


fakeshadow:
@nbrady-techempower Can we get this PR merged? It's ready.

PR #8220, opened 2023-05-19
Fixed docker images for Lisp frameworks Woo, Ningle and NingleX.

svetlyak40wt:
I've upgraded these frameworks to a newer base image `debian:bullseye-slim` and also modified dockerfiles to make future builds more stable. Now versions of Quicklisp, SBCL, ASDF are fixed.

Interesting, it it possible to share some env variables between different docker files on a single platfrom? For example, to move common code from all Lisp dockerfiles into a single place.


joanhey:
You need to delete "broken" from tags: in the benchmark_config.json, in all frameworks.

https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Lisp/ningle/benchmark_config.json#L27

svetlyak40wt:
@joanhey I've removed tags.

Issue #8199, opened 2023-05-07
Frameworks fail because use Debian Stretch EOL


joanhey:
In the last run, some frameworks failed in Docker build, because use **Debian Stretch** that is End Of Live.

### Frameworks failing

- Ningle, Ninglex : @gwbrown @GertThiel
  https://tfb-status.techempower.com/unzip/results.2023-05-07-07-21-52-737.zip/results/20230501013557/ningle/build/ningle.log
- Racket :  @Bogdanp
  https://tfb-status.techempower.com/unzip/results.2023-05-07-07-21-52-737.zip/results/20230501013557/racket/build/racket.log
- Servant : @naushadh @phadej
  https://tfb-status.techempower.com/unzip/results.2023-05-07-07-21-52-737.zip/results/20230501013557/servant/build/servant.log
- Warp : @phadej
  https://tfb-status.techempower.com/unzip/results.2023-05-07-07-21-52-737.zip/results/20230501013557/warp/build/warp-shared.log
- Wizardo-inline : @facundominguez
  https://tfb-status.techempower.com/unzip/results.2023-05-07-07-21-52-737.zip/results/20230501013557/wizzardo-inline/build/wizzardo-inline.log
- Woo : @gwbrown @GertThiel
  https://tfb-status.techempower.com/unzip/results.2023-05-07-07-21-52-737.zip/results/20230501013557/woo/build/woo.log

Please update the dockerfiles.

Before the next run, they will be market as broken. 

### Debian EOL
![image](https://user-images.githubusercontent.com/249085/236694550-d15dc5af-fb8a-4377-b84a-1d6e087b7f4a.png)

svetlyak40wt:
I'm going to fix Woo, Ningle and NingleX soon.

svetlyak40wt:
Fixed Lisp frameworks: https://github.com/TechEmpower/FrameworkBenchmarks/pull/8220

svetlyak40wt:
Byt the way, why do Lisp framework benchmarks are marked as having "Stripped implementation"?

What should be done to make them non-stripped?

gwbrown:
Honestly I don’t know why they’re marked as stripped - they were like that when I made my changes and I just didn’t do anything about it. They’re certainly more realistic than some of the other framework implementations marked as realistic. 

(Thanks for fixing these BTW!)

facundominguez:
Fixed wizzardo-inline in #8223.

[...]

PR #8223, opened 2023-05-23
Update wizzardo-inline after Debian Stretch EOL

facundominguez:
Addresses #8199

Changed the docker container to rely on nixpkgs which should make it harder to break. Also updated the benchmark code to the latest from the inline-java repo.


facundominguez:
I don't know what the "PR: Please Update" label is asking to update and update it in which way.

nbrady-techempower:
@facundominguez Not sure what happened there. That's usually only applied on a failed job. Never seen that happen before. Anyway, this LGTM. Thanks!

PR #8224, opened 2023-05-24
Upgrade Quarkus to new major version 3

Sanne:
Also polished our pom.xml files and docker image definitions, as many things have evolved in Quarkus since the last PR.
Also:
 - upgraded to using JDK17: JDK11 is getting old and not really representative of expected usage
 - upgraded to use Red Hat UBI images to better reflect production environments


joanhey:
Please could you remove Plaintext and Json tests from Hibernate Reactive.
I think that only change how the ORM work, but these tests don't use db.
image
 Related to #8178

Sanne:
Hi @joanhey , in practice in Quarkus depending on the use of the blocking vs reactive database stack, there are other subtle changes as the whole request can be handled on the reactive pipeline (or not).

So I'd prefer keeping it, if possible - the two stacks can sometimes perform very differently depending on this aspect; they don't seem to be doing this right now but that's possibly a coincidence.

Is the goal to reduce the overall tests being run? We can help with that by removing one of our three stacks, as one is being phased out and not very interesting anymore - but I'd prefer to handle such things in a follow up PR if  that's ok with you so that we can have at least some nightlies to compare Quarkus 3 with Quarkus 2 before changing the tests.

franz1981 (commented):
LGTM! I would remove the `quarkus` one as well :P

Sanne:
> > Is the goal to reduce the overall tests being run? We can help with that by removing one of our three stacks
> 
> @Sanne This would be great, thank you!

Done that in #8241

PR #8234, opened 2023-05-30
Update Helidon Nima to 4.0.0-ALPHA6

credmond-git:
Update Helidon Nima to 4.0.0-ALPHA6
Java to JDK 20.
Postgresql jdbc driver to: 42.6.0 (with loom fixes) Use hikaria db connection pool.

### DIFF
--- a/frameworks/Java/helidon/README.md
+++ b/frameworks/Java/helidon/README.md
@@ -2,7 +2,7 @@
 
 This is the Helidon portion of a [benchmarking test suite](../) comparing a variety of web development platforms.
 Two Helidon APIs are implemented: Reactive and Nima. The Reactive API has been around since the first
-version of Helidon while Nima is a new, blocking API based on JDK 19 virtual threads. 
+version of Helidon while Nima is a new, blocking API based on JDK 20 virtual threads. 
 
 The code is organized into two Maven modules, namely, *reactive* and *nima*. Both modules implement the
 same set of tests outlined below. 
--- a/frameworks/Java/helidon/helidon-nima.dockerfile
+++ b/frameworks/Java/helidon/helidon-nima.dockerfile
@@ -14,6 +14,5 @@ EXPOSE 8080
 CMD java --enable-preview \
     -XX:+UseNUMA \
     -XX:+UseParallelGC \
-    -XX:+DisableExplicitGC \
     -XX:+UseStringDeduplication \
     -jar app.jar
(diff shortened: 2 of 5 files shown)


credmond-git:
If you @sprsquish @danielkec  could please double check the changes that would be great. 
I didn't change much, only upgrading versions and switching to the postgresql driver (now that it has the loom fixes)

spericas:
@credmond-git Have you tested those -XX switches? 

For reference, release notes https://www.postgresql.org/about/news/postgresql-jdbc-4260-released-2613/

credmond-git:
Sorry i am not sure what @spericas mean about the -XX switches? I did test with the new loom friendly re-entrant locks in JDBC 42.6.0. That is why i made this change to see how loom could work with better locking support in a thread per request system using the existing JDBC method. 
The results were very close to the reactive postgres driver, but i left the setting to `pgclient` so you the project author could make the decision. I personally would perfer to set it to `hikari` as that is more representative of the market you are building for with nima.

spericas:
@credmond-git I mean these:

```
    -XX:+UseNUMA \
    -XX:+UseParallelGC \
    -XX:+DisableExplicitGC \
    -XX:+UseStringDeduplication \
```

credmond-git:
Ah, i tested with and without them using the quick benchmark, and they have the roughly the same performance. So they could be removed. 
But generally:
UseParallelGC is better for throughput, where as G1 is better for lower latency
UseNUMA should only help if applicable, it should not cause any performance impact if the machines are not NUMA . 
DisableExplicitGC shouldn't matter, i don't see anywhere that is explicitly calling GC, so i could remove this. 
UseStringDeduplication this one is harder to say, it seems like there are a lot of duplicate strings, as we are always returning the same subset of results. So it could improve memory usage and performance. But i haven't thoroughly run test around this. 

Reviewing several other projects, I removed DisableExplicitGC  as these seem to be the most common set of flags. 
    -XX:+UseNUMA \
    -XX:+UseParallelGC \
    -XX:+UseStringDeduplication \

spericas:
@credmond-git OK, thanks. We'll do some testing and then figure out the review process.

credmond-git:
I removed -XX:+UseStringDeduplication, the remaining two should be considered safe. 
``` 
    -XX:+UseNUMA \
    -XX:+UseParallelGC
```

spericas:
@nbrady-techempower We've done some additional testing over the weekend. PR is ready to merge! Thanks.

PR #8255, opened 2023-06-13
Add broken tags to FW's that haven't started/stopped correctly in last 3 full runs + script for detecting failing frameworks

ecruz-te:
Python web scraper to detect any frameworks that have failed the past 3 completed runs with the same error

USAGE: `python tfb-fail-detector.py`


joanhey (changes requested):
Prologue is fixed in #8254

frameworks/Nim/prologue/benchmark_config.json
```diff
@@ -18,7 +18,8 @@
       "database_os": "Linux",
       "display_name": "Prologue",
       "notes": "",
-      "versus": "httpbeast"
+      "versus": "httpbeast",
+      "tags": ["broken"]
```
joanhey, line 22: Fixed in #8254

joanhey (changes requested):

joanhey:
Please don't close frameworks, without inform their developers.

nbrady-techempower:
> The script is a very good start, like a new useful helper.
> 
> Please add another method to show also the URL to see the error. And perhaps also show the devs, of the fw, in the Techempower repo.

These are some good ideas. Let's iterate on that when we have time.

PR #8262, opened 2023-06-20
[Quarkus] Replace lambdas with method references and remove unnecessary casts

hatahet:


franz1981:
Please @hatahet what's the motivation of these changes?
Have you tried in a lab?
@Sanne too

Sanne (approved):

PR #8265, opened 2023-06-20
Few Quarkus Tweaks

franz1981:


Sanne (approved):

franz1981:
@Sanne I've added a change on Rocker to enable it to be as fast as the one used on Vertx benchmark.
Plaintext, similarly, is making uses of vertx heap buffers which `Buffer::getBytes` is faster then copying from direct Netty buffers to heap buffers

PR #8268, opened 2023-06-21
Python3 update

ecruz-te:
Upgrades toolset to python3.10

- Changes mostly made using the stage 1 and stage 2 suggested by `future` package (https://python-future.org/automatic_conversion.html)
- Changes MySQL-python to mysqlclient for python3 support


Dockerfile
```diff
@@ -16,26 +16,26 @@ RUN apt-get -yqq update && apt-get -yqq install \
       gcc \
       git-core \
       gosu \
-      libmysqlclient-dev          `# Needed for MySQL-python` \
+      default-libmysqlclient-dev  `# Needed for mysqlclient` \
+      build-essential             `# Needed for mysqlclient` \
       libpq-dev \
-      python2 \
-      python2.7-dev \
+      python3.10 \
```
volyrique, line 22: Have you thought about simplifying this to `python3`? It would minimize the changes if in the future we switch to Ubuntu 24.04, for example.

entrypoint.sh
```diff
@@ -3,4 +3,4 @@ set -euox pipefail
 
 chown -R $USER_ID /var/run/
 
-gosu $USER_ID python2 /FrameworkBenchmarks/toolset/run-tests.py "$@"
\ No newline at end of file
+gosu $USER_ID python3.10 /FrameworkBenchmarks/toolset/run-tests.py "$@"
```
volyrique, line 6: Same here - how about simplifying to `python3`?

Dockerfile
```diff
@@ -16,26 +16,26 @@ RUN apt-get -yqq update && apt-get -yqq install \
...
       docker==4.0.2 \
-      MySQL-python \
+      mysqlclient \
       psutil \
       psycopg2-binary \
       pymongo \
-      requests
+      requests==2.28.1 `# urllib3 incompatibility https://github.com/docker/docker-py/issues/3113#issuecomment-1525500104`
+
     # Fix for docker-py trying to import one package from the wrong location
```
volyrique, line 39: You can remove these obsolete lines as well.

PR #8273, opened 2023-06-23
Further Python 3 fixes and clean-up

volyrique:
@nbrady-techempower, one of my comments in #8268 was misunderstood, hence this PR. Also, I noticed that the toolset had an issue with its usage of `pymongo` (it was using some functionality that had been deprecated and subsequently removed in the latest version), so here is a quick fix, along with some clean-up.


nbrady-techempower:
Thank you @volyrique!

PR #8312, opened 2023-07-09
Optimize Common Lisp test based on Woo webframework

svetlyak40wt:
I've optimized Common Lisp tests based on Woo framework.

- replaced CL-Markup with Spinneret because Spinneret 30% faster and.
- replaced Jonathan with ST-JSON, because latter is 10-15% faster.
- rewrote all database queries to use prepared query and binary protocol
  to work with Postgresql.
- added proper type declarations.
- and set compiler optimization policies.

Also, I've added a few minor changes to the test runner:

- when you pass `--debug` option, `DEBUG=true` environment variable is passed into the docker image.
- also, now you can specify `debug_port` option in the `benchmark_config.json` file. If you do so, then this port will be exposed from docker container if test was started with `--debug` option. I use this feature to connect to the process inside docker from Emacs and to play with the code in interactive mode.

On my virtual machines this optimized version shows x2 performance boost comparing to the version of the test in the `master` branch.


svetlyak40wt:
@nbrady-techempower please, look at this PR, is it possible to merge it if errors in non Lisp tests are not caused by my changes?

toolset/utils/docker_helper.py
```diff
@@ -218,6 +227,10 @@ def watch_container(docker_container, docker_file):
                 remove=True,
                 log_config={'type': None})
 
+#            print('WAITING while you connect to 4005')
+#            import time
+#            time.sleep(15)
+
```
joanhey, line 233: Please try to don't mix toolset changes, with frameworks changes in the same PR.

joanhey: The toolset change need to be approved first.
If it isn't approved, your changes to the frameworks will be not merged also.

svetlyak40wt: Ok, I'll separate this pull into two.

svetlyak40wt:
I've created two separate pull requests:

- https://github.com/TechEmpower/FrameworkBenchmarks/pull/8313
- https://github.com/TechEmpower/FrameworkBenchmarks/pull/8314

Closing this one.

PR #8314, opened 2023-07-10
A few changes to the test runner when used with --debug option

svetlyak40wt:
- when you pass `--debug` option, `DEBUG=true` environment variable is passed into the docker image. This way benchmark author can enable logging or open a debugger port when running the test in the debug mode.
- also, now you can specify `debug_port` option in the `benchmark_config.json` file. If you do so, then this port will be exposed from docker container if test was started with `--debug` option. I use this feature to connect to the Lisp process inside docker from Emacs and to play with the code in interactive mode.

### DIFF
--- a/frameworks/Lisp/woo/benchmark_config.json
+++ b/frameworks/Lisp/woo/benchmark_config.json
@@ -10,6 +10,7 @@
                 "query_url": "/queries?queries=",
                 "update_url": "/updates?queries=",
                 "port": 8080,
+                "debug_port": 4005,
                 "approach": "Stripped",
                 "classification": "Micro",
                 "database": "Postgres",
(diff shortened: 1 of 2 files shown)


svetlyak40wt:
@nbrady-techempower probably some documentation should be updated to reflect these changes?

nbrady-techempower:
@svetlyak40wt https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Codebase-Framework-Files here now, thanks again!

svetlyak40wt:
@nbrady-techempower I see your updated to documentation. In line:

```
debug_port (optional): The port the server will be listening on if started with the --debug option. Will default to port if not supplied.
```

 The part "Will default to port if not supplied." is not correct, it is better to rewrite it as "Will be opened in addition to the `port` is specified. Because otherwise one might think that if `debug_port` is specified, then `port` will not be exposed, but really both `port` and `debug_port` will be exposed.

PR #8324, opened 2023-07-17
[php:symfony] fix and add postgresql

fullbl:
Should fix all symfony environments and add postgresql to them!


frameworks/PHP/symfony/benchmark_config.json
joanhey, line 1: Remember to change the database,  in the permutations.

joanhey (changes requested):

frameworks/PHP/symfony/benchmark_config.json
fullbl: thanks for the hint! I changed them and also the dockerfiles involved!

joanhey (approved):

joanhey:
Fixes https://github.com/symfony/symfony/issues/49447

frameworks/PHP/symfony/symfony-raw.dockerfile
```diff
@@ -7,30 +7,31 @@ RUN LC_ALL=C.UTF-8 add-apt-repository ppa:ondrej/php > /dev/null && \
...
+ENV COMPOSER_ALLOW_SUPERUSER=1
+RUN composer install --optimize-autoloader --classmap-authoritative --no-dev --quiet --no-scripts
+RUN cp deploy/postgresql/.env . && composer dump-env prod && bin/console cache:clear
 
-RUN php bin/console cache:clear 
 RUN echo "opcache.preload=/symfony/var/cache/prod/App_KernelProdContainer.preload.php" >> /etc/php/8.2/fpm/php.ini
 
 EXPOSE 8080
 
-CMD service php8.2-fpm start && \
```
joanhey, line 35: Why don't use service ??

fullbl: service should be already started at this point!

joanhey: Well we can use restart.

fullbl: I really don't get it, here we are building the image, so the service will start when container will run with the correct configuration, or am I wrong? I never saw the need to start the service manually!

joanhey: It's only simplicity:
`service php8.2-fpm start`
vs
`/usr/sbin/php-fpm8.2 --fpm-config /etc/php/8.2/fpm/php-fpm.conf`

joanhey:
@nbrady-techempower  that's ready to be merged
It's only small details discussion.

joanhey:
This change has created a great regression performance in Swoole and Workerman.
For example, Workerman in Fortunes had ~70K req/s , now ~47K req/s.

I'll investigate it, but I think that the first try will be to use again MySQL, perhaps Doctrine have a performance problem with PostreSQL.

fullbl:
why not doubling it and testing for both?

joanhey:
Normally I try that 1 PR only change 1 think, so it's easier to check performance changes.

The problem also happen without db connection.
Example, Workerman in Json had ~199K req/s, now ~137K req/s.

joanhey:
**Json**
Faster with PHP-FPM. But a lot slower in all workers mode: Workerman, Swoole and FrankenPHP.

Before:
image


Now:
image

fullbl:
that's weird, though I have never used swoole or frankenphp, the edits I made shouldn't impact that much!

joanhey:
I think that is the new PHP annotations, they work well for a shared nothing like PHP-FPM.
But not with persistent apps like the workers.
I'll test a little more, and if necessary I'll create an issue in Symfony repo.

joanhey:
The edits or changes that make a developer, always impact.
But the interesting thing is understand why is slower now !!

joanhey:
We'll wait till it's fixed the performance problems in the actual runs.

Issue #8397, a lot of frameworks have performance problems.

frameworks/PHP/symfony/symfony-franken.dockerfile
```diff
@@ -19,13 +19,13 @@ ADD . /symfony
 WORKDIR /symfony
 
 RUN mkdir -m 777 -p /symfony/var/cache/{dev,prod} /symfony/var/log
-RUN composer install --no-dev --no-scripts --quiet
+
+ENV COMPOSER_ALLOW_SUPERUSER=1
+RUN composer install --optimize-autoloader --classmap-authoritative --no-dev --quiet --no-scripts
```
stof, line 24: Running that before running `composer require runtime/frankenphp-symfony` is breaking it because `composer require` will perform a partial update and re-dump the autoload (and in that command, you don't enable the autoload optimizations).
There was a reason to dump the optimized autoloader after that before.

PR #8326, opened 2023-07-18
Micronaut 4

dstepanov:
FYI @graemerocher @sdelamo


sdelamo (changes requested):
Could we use properties, the default in Micronaut Framework 4,  instead of yml?

dstepanov:
> Could we use properties, the default in Micronaut Framework 4, instead of yml?

That would improve the startup which is not benchmarked by the framework.

graemerocher (approved):

sdelamo (approved):

yawkat:
we need feedback from the vertx team before merging this

@franz1981

franz1981:
@dstepanov Hi Denis, just curious, which bug was? is something related Vertx we have any responsability (and can fix)?

franz1981:
@dstepanov thanks for sharing, hope will be fixed soon!!!

dstepanov:
@nbrady-techempower Can you merge, please?

PR #7977, opened 2023-03-01
Add option for extra docker args to tfb script

calderast:
Add the option for extra docker arguments to be added to the docker run command for added user configurability. The user can specify the desired arguments, if any, in an environment variable EXTRA_DOCKER_ARGS before running a test, e.g. `export EXTRA_DOCKER_ARGS="--cpus=2.0 --memory=3G"`.
If no value is specified for EXTRA_DOCKER_ARGS, the existing functionality will not be impacted.

Issue #8029, opened 2023-03-14
Add configurability to Java benchmarks


calderast:
The Java Engineering team at Microsoft is using some of the TechEmpower framework benchmarks to test the performance of our JDK.
We've run into a few places where it would be helpful to add configurability to the Java benchmarks:

- It would be helpful to be able specify the JDK under test
- It would be helpful to be able to append JVM arguments/override existing JVM arguments (see example [here](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/spring/spring.dockerfile#L34) for the spring benchmark)
- It would be helpful for the user to be able to specify extra docker arguments (I submitted PR #7977 for this, but after a second look I'm not sure if those args actually go to to the docker container under test, or just an intermediate container)

Please let me know if you have any thoughts on how to tackle these issues in a way that would benefit everyone using these benchmarks!

nbrady-techempower:
Ah, yes. Now that I think about it, those args are just getting tossed to the toolset container.

I think what would need to happen is a new toolset arg [here](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/toolset/run-tests.py) to pass docker arguments to the test docker container. 

We don't have any bandwidth to add new features to the toolset right now, but certainly if somebody want to tackle this, I'll give guidance where I can.

NateBrady23:
Closing as part of bulk cleanup.

PR #8329, opened 2023-07-19
Add ability to configure test docker container

calderast:
Adds the arguments --test-container-memory and --extra-docker-runtime-args to allow the user to configure the test docker container.
e.g. `./tfb --test spring --test-container-memory 1G --extra-docker-runtime-args cpu_period:10000 cpu_quota:50000`

Reverts the pull request https://github.com/TechEmpower/FrameworkBenchmarks/pull/7977, as the extra docker arguments were being passed to an intermediate container instead of the test container.

### DIFF
--- a/toolset/utils/docker_helper.py
+++ b/toolset/utils/docker_helper.py
@@ -13,8 +13,6 @@ from toolset.databases import databases
 
 from psutil import virtual_memory
 
-# total memory limit allocated for the test container
-mem_limit = int(round(virtual_memory().total * .95))
 
 class DockerHelper:
     def __init__(self, benchmarker=None):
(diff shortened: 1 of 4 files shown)


calderast:
cc @nbrady-techempower, this PR addresses some of the issues mentioned in https://github.com/TechEmpower/FrameworkBenchmarks/issues/8029

nbrady-techempower:
Fantastic! I've added this to the wiki as well. Thank you!

toolset/utils/docker_helper.py
```diff
@@ -207,6 +205,17 @@ def watch_container(docker_container, docker_file):
...
+            # Total memory limit allocated for the test container
+            if self.benchmarker.config.test_container_memory is not None:
+                mem_limit = self.benchmarker.config.test_container_memory
+            else:
+                mem_limit = int(round(virtual_memory().total * .95))
+
+            # Convert extra docker runtime args to a dictionary
+            extra_docker_args = {}
+            if self.benchmarker.config.extra_docker_runtime_args is not None:
+                extra_docker_args = {key: int(value) if value.isdigit() else value for key, value in (pair.split(":") for pair in self.benchmarker.config.extra_docker_runtime_args)}
```
mkurz, line 218: @calderast @nbrady-techempower You should not split all colons, just on the first occurence, see #8363

mkurz:
@calderast @nbrady-techempower Is it possible to pass volume(s) to mount that should be mounted to `--extra-docker-runtime-args` ? Like `docker run ... -v /path/on/host:/path/in/container ...`
It does work for me when modifing `docker_helper.py` by adding
```python
     ...
     volumes = {'/path/on/host': {'bind': '/path/in/container', 'mode': 'rw'}},
     ...
```

https://github.com/TechEmpower/FrameworkBenchmarks/blob/e01316ba339cd89e9430b9f88dfac2459722b176/toolset/utils/docker_helper.py#L220-L239

I don't work with python, maybe this can be passed already as json? I couldn't get it work? Or does the string need to be converted to a python dictionary? If so, I guess this has to be done as special case in the code for volume?

Thanks!

mkurz:
@calderast Any chance you take a look at my [previous comment](https://github.com/TechEmpower/FrameworkBenchmarks/pull/8329#issuecomment-1680559171)? Thanks!

PR #8349, opened 2023-08-01
cppcms fixes

ecruz-te:


joanhey:
Good luck !!
But what work for one, don't work for others.
And so all learn from this tool.

PR #8358, opened 2023-08-09
Can I merge the changes by just modifying the CPU quantity parameter?

shaovie:
Can I make further modifications to the parameters? It's just a single parameter change, but it has a significant impact on performance. I'm begging you, please.
@nbrady-techempower


nbrady-techempower:
You don't have to beg. Citrine is currently down right now and we're looking at getting it back online as soon as we can.

PR #8365, opened 2023-08-18
[xitca-web] update dep and db api.

fakeshadow:
update deps.
use dedicated api for pipelined pg query.
change xitca-web [diesel] to full stack classification.


frameworks/Rust/xitca-web/xitca-web-wasm.dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM rust:1.67 AS compile
+FROM rust:latest AS compile
```
nbrady-techempower, line 1: Please keep this locked down to a specific version

fakeshadow: fixed.

PR #8375, opened 2023-08-23
[Scala] Add Apache Pekko

otrosien:
Adds Apache Pekko (https://pekko.apache.org/) as framework implementation. The test mimics the code for testing Akka.


otrosien:
@nbrady-techempower do you think this can make it still into the upcoming round?

PR #8396, opened 2023-08-29
[c++]a lightweight, concise i/o event demultiplexer implementation in c++11

shaovie:
I hope we can merge in before it opens up


shaovie:
@nbrady-techempower ^_^

PR #8398, opened 2023-08-29
[Helidon] Updates benchmark config file

spericas:
Updates benchmark config file. Now using Helidon 3 and Helidon 4 as display names.


spericas:
@nbrady-techempower

PR #8406, opened 2023-09-03
[viz] deps updates

fundon:
* diesel-async v0.4
* sqlx v0.7
* sailfish v0.8


frameworks/Rust/viz/viz-diesel.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:1.66.0
+FROM rust:latest
```
nbrady-techempower, line 1: Please keep these locked down to a version number

PR #8407, opened 2023-09-05
Add Bun.sh with good defaults.

jtwebman:
This adds Bun.sh 1.0.0 to the mix for TypeScript and JavaScript


frameworks/TypeScript/bun/benchmark_config.json
```diff
@@ -0,0 +1,30 @@
...
+        "database": "Postgres",
+        "framework": "bun",
+        "language": "Typescript",
+        "flavor": "None",
+        "orm": "Micro",
+        "platform": "Bun",
+        "webserver": "Elysia",
+        "os": "Linux",
+        "database_os": "Linux",
+        "display_name": "Bun Elysia Posgres",
```
muuvmuuv, line 24: Postgres* :)

frameworks/TypeScript/bun/bun.dockerfile
```diff
@@ -0,0 +1,38 @@
+FROM debian:bookworm as build
```
muuvmuuv, line 1: There is an official docker image which might fit better? https://hub.docker.com/r/oven/bun

jtwebman: Switch

marcbarbosa:
@jtwebman Since you are using a specific framework (Elysia), I believe we should move your implementation to `frameworks/TypeScript/Elysia/` directory.

nbrady-techempower:
Yes, please move this to `frameworks/TypeScript/elysia` and update the `package.json` to reflect. `frameworks/TypeScript/bun` would be reserved for a "vanilla" bun implementation without using another web framework.

nbrady-techempower:
closing in favor of #8480

PR #8410, opened 2023-09-06
Can you add this SPRAPP under Golang

hiqsociety:
first time doing this. sorry. it's based on hertz but we will be having a direction on our own. this is "prefork mode" but we will not have non-prefork mode.


nbrady-techempower:
@hiqsociety Please check out the wiki for adding tests. You're missing a lot here.  https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Development-How-To-Add-Frameworks-or-Tests

Issue #7646, opened 2022-10-19
Create benchmarks using bun


Meir017:
the deno version - https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/TypeScript/deno/src/main.ts

FYI @Jarred-Sumner

muuvmuuv:
Yes, would love other JavaScript and TypeScript engines compared here: deno and bun. Also zig (which bun is written in) as a language is also missing.

jtwebman:
I built one real quick this weekend here: https://github.com/TechEmpower/FrameworkBenchmarks/pull/8407

masfahru:
I think we should wait until [the implementation of `node:cluster` module](https://github.com/oven-sh/bun/issues/2428) is done

ThatOneCalculator:
Bun does have mutlithreading with the `Bun.workers` API.

chlorophant:
Does this example also load balance to each process somehow? If so, does anyone know how that mechanism works?

jtwebman:
It doesn't. Node wrote their own into their stuff https://nodejs.org/api/cluster.html Bun.js doesn't support that yet but you probably could add a simple one https://github.com/oven-sh/bun/issues/2428 as at least then you are comparing apples to apples.

NateBrady23:
Closing as part of bulk cleanup.

PR #8416, opened 2023-09-12
[Bun] Add Bun runtime

marcbarbosa:
Initial (vanilla) implementation for [Bun](https://bun.sh): `/json` and `/plaintext`

Reference: https://github.com/TechEmpower/FrameworkBenchmarks/issues/7646

### DIFF
--- a/frameworks/TypeScript/bun/bun.dockerfile
+++ b/frameworks/TypeScript/bun/bun.dockerfile
@@ -1,0 +1,13 @@
+FROM oven/bun:1.0
+
+EXPOSE 8080
+
+WORKDIR /app
+
+USER bun
+
+COPY ./src .
+
+ENV NODE_ENV=production
+
+CMD ["bun", "index.ts"]
--- a/frameworks/TypeScript/bun/src/index.ts
+++ b/frameworks/TypeScript/bun/src/index.ts
@@ -1,0 +1,21 @@
+const HELLO_WORLD_STR = "Hello, World!";
+const options: ResponseInit = { headers: { "Server": "Bun" } };
+
+const server = Bun.serve({
+  port: 8080,
+  fetch(req: Request) {
+    const { pathname } = new URL(req.url);
+
+    if (pathname === "/json") {
+      return Response.json({ message: HELLO_WORLD_STR }, options);
+    }
+
+    if (pathname === "/plaintext") {
+      return new Response(HELLO_WORLD_STR, options);
+    }
+
+    return new Response("", { status: 404 })
+  },
+});
+
+console.log(`Listening on localhost:${server.port}`);
--- a/frameworks/TypeScript/bun/README.md
+++ b/frameworks/TypeScript/bun/README.md
@@ -1,0 +1,22 @@
+# [Bun](https://bun.sh/) - A fast all-in-one JavaScript runtime
+
+## Description
+
+Bun is an all-in-one toolkit for JavaScript and TypeScript apps. It ships as a single executable
+called `bun`​.
+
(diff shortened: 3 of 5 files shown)


frameworks/TypeScript/bun/bun.dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM oven/bun
```
nbrady-techempower, line 1: Could you lock this down to a specific version please

marcbarbosa: Done

frameworks/TypeScript/bun/src/index.ts
```diff
@@ -0,0 +1,16 @@
+type HandlerFn = (req?: Request) => Promise<Response> | Response;
+
+const options: ResponseInit = { headers: { "Server": "Bun" } };
+
+const handlers = new Map<string, HandlerFn>();
+handlers.set("/json", () => Response.json({ message: "Hello, World!" }, options));
+handlers.set("/plaintext", () => new Response("Hello, World!", options));
```
WebReflection, line 7: out of curiosity, shouldn't this be `Response.text("Hello, World!", options)` instead?

frameworks/TypeScript/bun/src/index.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const server = Bun.serve({
+  port: 8080,
+  fetch(req: Request) {
+    return handlers.get(new URL(req.url).pathname)?.call(req);
```
WebReflection, line 12: as a benchmark, I am not sure this it idiomatic or well performing + the `call(req)` when functions are arrows look off ..

```js
fetch: (req: Request) => handlers.get(new URL(req.url).pathname)?.()
```

looks already better, but I am not sure that's how people really write routes in Bun neither .. if it's idiomatic benchrmaks we're after, this PR doesn't look like the best we can do ... thoughts?

frameworks/TypeScript/bun/src/index.ts
marcbarbosa: > out of curiosity, shouldn't this be Response.text("Hello, World!", options) instead?

Not really, `new Response()` returns `text/plain` as content type already, required by the `/plaintext` scenario.

> or better, what is Response.json in there?

`Response.json` returns the object serialized and `application/json` as content type, required by the `/json` scenario.

frameworks/TypeScript/bun/src/index.ts
marcbarbosa: I agree, actually I'm refactoring my PR to simplify this part and not implementing anything other than what is strictly necessary. Just plain `if` statements.

frameworks/TypeScript/bun/src/index.ts
WebReflection: not in the docs and not standard though, maybe a comment would do

marcbarbosa: This is the documentation for `Response.json`: https://developer.mozilla.org/en-US/docs/Web/API/Response/json_static

And I believe `text/plain` is the default content type.

Do you think an extra comment is really needed?

WebReflection: I’ve missed that … my bad. No comment needed

nbrady-techempower:
@marcbarbosa Let me know when you're done refactoring. This looks ready.

marcbarbosa:
@nbrady-techempower Done

Jarred-Sumner:
This needs parallelism to be potentially competitive with native languages — you can use Worker and enable reusePort. request.url already runs the URL parser so you’re running it twice here

PR #8422, opened 2023-09-15
[TypeScript/Ditsmod]: added tests implementation for Ditsmod.

KostyaTretyak:
Hello. Please accept the new framework called [Ditsmod](https://ditsmod.github.io/en/) for testing.


KostyaTretyak:
I wrote `platform: Node.js` in the `benchmark_config.json` file, but now I noticed that other projects have `nodejs` or `NodeJS` there. Won't this be a problem? Is it better to write `nodejs`?

PR #8431, opened 2023-09-21
Improve plain text response + bump version

dstepanov:


dstepanov:
Please review @graemerocher @sdelamo

graemerocher (approved):

dstepanov:
@nbrady-techempower Please review

frameworks/Java/micronaut/benchmark_config.json
```diff
@@ -25,6 +25,29 @@
         "notes": "",
         "versus": "None"
       },
+      "graalvm": {
+        "json_url": "/json",
+        "plaintext_url": "/plaintext",
```
KostyaTretyak, line 30: It doesn't make sense to mix tests for "plaintext and json" with tests intended for database. See https://github.com/TechEmpower/FrameworkBenchmarks/issues/8178

dstepanov: We have them repeated here to see how does Micronaut perform on Graalvm native build

KostyaTretyak: So you need to add two tests. In one test there will be a reference for the database, in the other (plaintext and json) - not.

dstepanov: We had the same config before and had to remove it because of some bug in Graalvm. Now we returning it back and it will test the default configuration of database access and the plain text/json using the native compilation I don’t see a need to duplicate it.

KostyaTretyak: Don't you get confused by the results table when you look at the results on the "plaintext" tab, but see endings like `frameworkname [postgres]`? To me, in such cases, the reference to the database is confusing, because it really has nothing to do with the test. In your case, it is also mentioned for the database (PG).

dstepanov: Most of the frameworks I saw do it the same, the testing framework should emit the database tag for the tests that aren't using it.

PR #8452, opened 2023-09-28
[rails] Upgrade to Rails 7.1.0.rc1

p8:
Also update the framework defaults to 7.1.


p8:
@nbrady-techempower Yes, I'd like to see this in the next run.
I won't expect any performance degradation only improvements and a more accurate result.

PR #8454, opened 2023-09-28
[Bun] use Bun.spawn to increase performance

masfahru:


masfahru:
Current (preview) CITRINE results:
image

I've tested this pr on 8 vcpus vm, the results:
image

PR #8455, opened 2023-09-28
refactor(ditsmod): added test with simplified use of DI.

KostyaTretyak:
Please add these changes for TechEmpower's Round 22.


KostyaTretyak:
Tests failed due to Docker Auth timeout.

PR #8466, opened 2023-10-02
[PHP] Optimize Swoole test

NathanFreeman:


joanhey:
Hi, you are removing the no-async mode.
I added it to test vs normal async, and the performance is better.

Any reason to eliminate it?
I'm at your service for any information.

frameworks/PHP/swoole/database.php
```diff
@@ -0,0 +1,185 @@
...
+                PDO::ATTR_ERRMODE            => PDO::ERRMODE_EXCEPTION,
+                PDO::ATTR_EMULATE_PREPARES   => false
+            ]
+        );
+
+        self::$db      = $pdo->prepare(Operation::WORLD_SELECT_SQL);
+        self::$fortune = $pdo->prepare(Operation::FORTUNE_SQL);
+        self::$random  = $pdo->prepare(Operation::WORLD_SELECT_SQL);
+        self::$update  = $pdo->prepare(Operation::WORLD_UPDATE_SQL);
+        self::$query   = $pdo->prepare(Operation::WORLD_SELECT_SQL);
```
joanhey, line 99: I think that here you can also simplify it, without repeating the same SQL to create the prepared statement.

NathanFreeman:
> Hi, you are removing the no-async mode. I added it to test vs normal async, and the performance is better.
> 
> Any reason to eliminate it? I'm at your service for any information. PD: after check the `db-no-async.php` need a little refactor (it's from 2019)

I did not remove the no-async mode.
`default` and `postgres` is no-async mode.
`no-async` and `async-postgres` is async mode.
`ENABLE_COROUTINE` is the switch to enable or disable no-async mode.

frameworks/PHP/swoole/swoole-server.php
```diff
@@ -10,7 +10,7 @@
 
 $server  = new Server('0.0.0.0', 8080);
 $setting = [
-    'worker_num'        => swoole_cpu_num() * 4,
+    'worker_num'        => swoole_cpu_num(),
```
joanhey, line 13: This is correct for the async, or all coroutines will fight for the CPU.
But not for the sync, that use the same OS context switch and need more processes for db connections.

joanhey (changes requested):

frameworks/PHP/swoole/swoole-server.php
NathanFreeman: I have already considered this situation now.

joanhey (approved):

NathanFreeman:
@joanhey Hi.
Could you merge it into the master branch when you have free time?
Thanks.

joanhey:
They are not accepting more PRs til finish the actual run, that is the Round 22.
When finish the remaining PRs will be merged.

NathanFreeman:
@joanhey Thank you

PR #8480, opened 2023-10-09
[TypeScript] Add ElysiaJS

masfahru:


masfahru:
Tested in 3 VMs (each with 1 core and 2 vCPUs), with HyperExpress as a comparison.
image
https://www.techempower.com/benchmarks/#section=test&shareid=f7d589d9-1733-4cc9-9f79-990d2e952d47&test=composite



Hopefully this test can be included in the next official round.

PR #8485, opened 2023-10-11
[xitca-web] enable xitca-web-iou bench.

fakeshadow:
re-enable `xitca-web-iou` since round 22 has already started and unrealistic bench would not interfere with official score anymore.
reduce loc and feature gate.
update deps.


fakeshadow:
@nbrady-techempower Just in case Please merge this PR after official round 22 started. Would really hate unrealistic score polluting the leader board. Thank you.

Issue #8420, opened 2023-09-14
No longer accepting plaintext only frameworks / Limited number of tests mutations


nbrady-techempower:
Hi everyone!

As the number of new frameworks submitted to the benchmarks grows, the amount of time it takes to complete a full run does as well. Because of this, we will be implementing the following rules:

- New frameworks that only implement plaintext will no longer be accepted. Of course, we'd like all frameworks to implement all tests to get a better idea of performance in various areas of the framework, but we expect at least 2 different tests to be implemented. Ideally plaintext or json and one db test.

- The number of test mutations will be limited to 10. We do not mind if you open up pull requests between runs to try out various mutations for your framework so long as the total number at any given time does not exceed 10.

After the next round, we will ping framework maintainers to make these changes. We will also look to remove tests that are older and no longer maintained. 

Thank you!

fakeshadow:
Rules like these show how popular the project is and I agree with both.
On top of it I suggest composite score being calculated per mutation which would offer a quick view of per mutation detail.

gi0baro:
@nbrady-techempower on the number of mutations I proposed https://github.com/TechEmpower/FrameworkBenchmarks/pull/8055 some time ago but then left it given the community feedbacks. It might be worth a while re-check it

joanhey:
I like it a lot, but it exist a problem long ago.
At the moment the framework is removed, all the history of the framework will disappear from the Rounds.

Like I said before the Rounds need to be immutable.
For example, in PHP we need to change the name because was php5, after the change plain PHP don't appear in the old Rounds. 
We have the numbers, and the work done, but don't show in the Rounds.

otrosien:
One framework to remove: Baratine. The domain baratine.io is not registered to the project anymore, and the github project has last changes 7 years ago (https://github.com/baratine/baratine)

joanhey:
In reality Baratine is marked as Stripped.

Why not remove all the stripped framework from the runs ??

fakeshadow:
> In reality Baratine is marked as Stripped.
> 
> Why not bypass all the stripped frameworks from the runs ??
> 
> https://github.com/search?q=repo%3ATechEmpower%2FFrameworkBenchmarks+%5C%22Stripped%5C%22+OR+%5C%22stripped%5C%22+path%3A%2F%5Eframeworks%5C%2F%2F&type=code

[...]

PR #8498, opened 2023-10-14
Update aspnet core benchmarks

sebastienros:
Minimal APIs implementation added, using Dapper only.
MVC updated, using Entity Framework only.
Platform updated, using ADO.NET only.
Middleware implementation is being removed.

Kept for future PRs for simplicity:
- AOT
- Cached Worlds in MVC and Minimal APIs


sebastienros:
@DamianEdwards @roji @davidfowl

joanhey:
As your removing tests, perhaps it'll be good to remove also the aspnetcore with only plaintext and json.

image

Per #8420 "No longer accepting plaintext only frameworks / Limited number of tests mutations"

sebastienros:
@joanhey maybe I made a mistake in the config files, but the other scenarios are implemented for it, in a distinct docker file. Will check why it doesn't run.

sebastienros:
@joanhey it's run with the aspnetcore-pg name

image

Did I miss something to group them, or is this only a behavior from the tests?

joanhey:
Also while don't maintain almost one mysql permutation?
So the future users will know the difference about using Mysql vs Postgresql.

joanhey:
Why use the time to create 2 docker images to run the same benchmark.
Exist a issue about that, if we create a bench for every endpoint, we can have better numbers thanks to the JIT.
We need to clarify this situation, that help the synthetic bench but it isn't realistic in a normal app.

image

joanhey:
Some time ago, I separated the name of aspnetcore to identify it in the composite score.
Because before exist only one, and that it wasn't realistic to identify that mvc is not as fast as the hard coded platform.
To help the users choose the right permutation.

image

I think that still need to be separated this classification.

sebastienros:
@joanhey Thanks for the feedback.

I have grouped the the default runs in the same docker file, and added MySQL as a separate one, with only the database benchmarks.

> I think that still need to be separated this classification

I am not sure if you are asking me to change anything in the config files. Right now the "default" profile should be used for this and it contains all the scenarios. I just know that how it looks like today on the public tables is great, thanks for whatever you did here.

joanhey:
I only changed the framework name. And this PR now have it OK.
image

sebastienros:
@joanhey I am ready to get it mergeg then if you are ok with it

joanhey (approved):

PR #8516, opened 2023-11-03
Fix aspnetcore database benchmarks

sebastienros:
WRK would report socket errors on reads that didn't show up in verification tests.

image


sebastienros:
@nbrady-techempower ready to merge, unless you have a schedule for these. Cheers!

PR #8521, opened 2023-11-05
Created Node.js tests with postgres.js module

srisaiswaroop:
1. Added the postgres.js module for postgres queries in addition to sequelize currently used.
2. Removed chakra as it is not running at all.
3. Added proper display names for tests so that its clear from the results.
4. Added separate tests for each of the postgres.js tests.
5. Used slow-json-stringify to improve json parsing.
6. Uses different sort method to slightly improve perfomance.
7. Updated the nodejs version to pickup latest stable version instead of manually updating every year.
8. Added Environment variable node_env as production to improve performance.
9. Implemented all database routes using postgresjs including cached using node-cache.
10. Improved fortunes by avoiding handlebars based rendering.


joanhey (changes requested):
It isn't permitted to use latest version. Use always Major.Minus version.

Please don't create extra permutations for free, each permutation have it's cost in time to build the dockerfile, ....

So create a default with all the tests, and don't add extra permutations only for plaintext, cached queries, ...

joanhey:
Exist a discussion about the templates in the Fortunes test, to mark the frameworks than don't use template engine.

You are removing the handlebars templating. I understand that it will be faster, but please maintain a permutation with handlebars. Almost all developers will use a template engine to create html. Still I think that it's good to check the performance loose when using the template engine.

A benchmark is to help the core developers, to improve the code for performance.
Adding tricks or removing modules that are normally used, don't help the core and final developers. Still some frameworks are doing that, but we are trying to limit that artificial tricks to have a fair play.

Please don't see a benchmark, only like a number competition.

srisaiswaroopwenergy:
@joanhey

I've implemented the changes you suggested.

Regarding the use of handlebars, I haven't removed any handlebars code from the existing tests that run under Node.js. I simply avoided using handlebars in the new test variant I added. This decision is based on the fact that JavaScript itself introduced a template feature. To clarify, this isn't some unusual technique I've employed; it's standard JavaScript knowledge.

You can learn more about template literals in JavaScript on the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals). This core language feature was introduced in JavaScript in June 2015, and its purpose is to create string templates, as the name suggests. You can find an example of HTML template rendering at the bottom of the MDN documentation page linked above.

srisaiswaroop:
@joanhey

I've implemented the changes you suggested.

Regarding the use of handlebars, I haven't removed any handlebars code from the existing tests that run under Node.js. I simply avoided using handlebars in the new test variant I added. This decision is based on the fact that JavaScript itself introduced a template feature. To clarify, this isn't some unusual technique I've employed; it's standard JavaScript knowledge.

You can learn more about template literals in JavaScript on the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals). This core language feature was introduced in JavaScript in June 2015, and its purpose is to create string templates, as the name suggests. You can find an example of HTML template rendering at the bottom of the MDN documentation page linked above.

srisaiswaroop:
@joanhey

I've implemented the changes you suggested.

Regarding the use of handlebars, I haven't removed any handlebars code from the existing tests that run under Node.js. I simply avoided using handlebars in the new test variant I added. This decision is based on the fact that JavaScript itself introduced a template feature. To clarify, this isn't some unusual technique I've employed; it's standard JavaScript knowledge.

You can learn more about template literals in JavaScript on the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals). This core language feature was introduced in JavaScript in June 2015, and its purpose is to create string templates, as the name suggests. You can find an example of HTML template rendering at the bottom of the MDN documentation page linked above.

joanhey (approved):

PR #8525, opened 2023-11-07
[PHP] Swoole enable opcache

NathanFreeman:


joanhey:
The OPCache is enabled for Swoole, as it's using the cli.

And the `opcache.enable_file_override` enabled, it won't give any gain in these tests.

But for me it's OK this change, as don't break, only add a little bit noise.

PR #8528, opened 2023-11-08
SPRAPP, New Golang Framework

hiqsociety:


frameworks/Go/sprapp/README.md
```diff
@@ -0,0 +1,5 @@
+SPRAPP https://github.com/cloudxaas/sprapp/
```
li-jin-gou, line 1: Why isn't this called hertz-prefork? @hiqsociety  Should we add a new hertz-prefork ?

li-jin-gou: cc @nbrady-techempower 
image

kolinfluence: the idea is like fiber-prefork.
will be using like fiber framework

PR #8547, opened 2023-11-16
Upgrade hserver Framework version and optimize configuration

heixiaoma:
Upgrade dependencies, optimize performance.


heixiaoma (commented):
Upgrade dependencies, optimize performance.

PR #8550, opened 2023-11-17
Remove blocking I/O from ReactPHP implementation

lcobucci:
The existing ReactPHP implementation doesn't leverage the async capabilities, creating wrong results when benchmarking things.

This optimises the ReactPHP app, allowing things to be a bit more comparable.

### DIFF
--- a/frameworks/PHP/reactphp/app.php
+++ b/frameworks/PHP/reactphp/app.php
@@ -1,134 +1,147 @@
 <?php
+
+use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface as Request;
+use React\EventLoop\Loop;
+use React\MySQL\ConnectionInterface as DbConnection;
+use React\MySQL\Factory as DbFactory;
 use React\Http\Message\Response;
+use React\MySQL\QueryResult;
+use React\Promise\PromiseInterface;
 
-function init()
-{
-    global $world, $fortune, $update;
-    $pdo = new PDO(
-        'mysql:host=tfb-database;dbname=hello_world',
-        'benchmarkdbuser',
-        'benchmarkdbpass',
-        [
-            PDO::ATTR_DEFAULT_FETCH_MODE => PDO::FETCH_ASSOC,
-            PDO::ATTR_EMULATE_PREPARES   => false
-        ]
-    );
-    $world   = $pdo->prepare('SELECT id,randomNumber FROM World WHERE id=?');
-    $update  = $pdo->prepare('UPDATE World SET randomNumber=? WHERE id=?');
-    $fortune = $pdo->prepare('SELECT id,message FROM Fortune');
-    $fortune->setFetchMode(PDO::FETCH_KEY_PAIR);
-}
+use function React\Promise\all;
 
-function router(Request $request)
+/** @return Closure(Request):ResponseInterface */
+function requestHandler(): Closure
 {
-    return match($request->getUri()->getPath()) {
-        '/plaintext' => text(),
-        '/json'      => json(), 
-        '/db'        => db(),
-        '/fortunes'  => fortune(),
-        '/query'     => query($request),
-        '/update'    => updateraw($request),
-        // '/info'      => info(),
-        default      => new Response(404, [], 'Error 404'),
+    $connection = establishDbConnection('benchmarkdbuser:benchmarkdbpass@tfb-database/hello_world?idle=0.5');
+
+    $world = static function (int $id) use ($connection): PromiseInterface {
+        return $connection->query('SELECT id,randomNumber FROM World WHERE id=?', [$id]);
+    };
+
+    $fortune = static function () use ($connection): PromiseInterface {
+        return $connection->query('SELECT id,message FROM Fortune');
+    };
+
+    $update = static function (int $id, int $randomNumber) use ($connection): PromiseInterface {
+        return $connection->query('UPDATE World SET randomNumber=? WHERE id=?', [$randomNumber, $id]);
+    };
+
+    return static function (Request $request) use ($world, $fortune, $update): ResponseInterface | PromiseInterface {
+        return match($request->getUri()->getPath()) {
+            '/plaintext' => Response::plaintext('Hello, World!'),
+            '/json'      => Response::json(['message' => 'Hello, World!']),
+            '/db'        => db($world),
+            '/fortunes'  => fortune($fortune),
+            '/query'     => query(queryCount($request), $world),
+            '/update'    => updateraw(queryCount($request), $world, $update),
+            // '/info'      => info(),
+            default      => new Response(404, [], 'Error 404'),
+        };
     };
 }
 
-function text()
-{
-    return new Response(200, [
-        'Content-Type' => 'text/plain'
-    ], 'Hello, World!');
+function establishDbConnection(
+    #[SensitiveParameter]
+    string $uri,
+): DbConnection {
+    $connection = (new DbFactory())->createLazyConnection($uri);
+
+    $interrupt = $connection->quit(...);
+
+    $connection->on('close', static function () use (&$interrupt) {
+        Loop::removeSignal(SIGINT, $interrupt);
+        Loop::removeSignal(SIGTERM, $interrupt);
+    });
+
+    Loop::addSignal(SIGINT, $interrupt);
+    Loop::addSignal(SIGTERM, $interrupt);
+
+    return $connection;
 }
 
-function json()
+/** @param Closure(int):PromiseInterface $world */
+function db(Closure $world): PromiseInterface
 {
-    return new Response(200, [
-        'Content-Type' => 'application/json'
-    ], json_encode(['message' => 'Hello, World!']));
+    $id = mt_rand(1, 10000);
+
+    return $world($id)->then(
+        static fn (QueryResult $result): ResponseInterface => Response::json($result->resultRows[0]),
+    );
 }
 
-function db()
+function queryCount(Request $request): int
 {
-    global $world;
+    $count = (int) ($request->getQueryParams()['q'] ?? 1);
 
-    $world->execute([mt_rand(1, 10000)]);
-
-    return new Response(200, [
-        'Content-Type' => 'application/json'
-    ], json_encode($world->fetch()));
-}
-
-function query($request)
-{
-    global $world;
-
-    $query_count = 1;
-    $q = (int) $request->getQueryParams()['q'];
-    if ($q > 1) {
-        $query_count = min($q, 500);
-    }                                                          
-
-    while ($query_count--) {
-        $world->execute([mt_rand(1, 10000)]);
-        $arr[] = $world->fetch();
+    if ($count > 1) {
+        return min($count, 500);
     }
 
-    return new Response(200, [
-        'Content-Type' => 'application/json'
-    ], json_encode($arr));
+    return 1;
 }
 
-function updateraw($request)
+/** @param Closure(int):PromiseInterface $world */
+function query(int $queryCount, Closure $world): PromiseInterface
 {
-    global $world, $update;
+    $processQueries = static function (int $count) use ($world): iterable {
+        while ($count--) {
+            $id = mt_rand(1, 10000);
 
-    $query_count = 1;
-    $q = (int) $request->getQueryParams()['q'];
-    if ($q > 1) {
-        $query_count = min($q, 500);
-    }
+            yield $world($id)->then(static fn (QueryResult $result): array => $result->resultRows[0]);
+        }
+    };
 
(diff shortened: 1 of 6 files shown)


lcobucci:
@WyriHaximus please feel free to correct things even further :+1:

frameworks/PHP/reactphp/server.php
```diff
@@ -1,20 +1,28 @@
...
-$server = new React\Http\Server($loop, function (Psr\Http\Message\ServerRequestInterface $request) {
-    return router($request);
-});
+$interrupt = static function () use ($server, $socket, &$interrupt): void {
+    echo 'Interrupting server', PHP_EOL;
 
-$socket = new React\Socket\Server('0.0.0.0:8080', $loop);
-$server->listen($socket);
+    $socket->close();
+    $server->removeAllListeners();
```
WyriHaximus, line 21: No need to remove all the listeners, that will be done the moment the socket is closed: 
```suggestion
```

frameworks/PHP/reactphp/app.php
```diff
@@ -1,134 +1,149 @@
...
 
-function text()
-{
-    return new Response(200, [
-        'Content-Type' => 'text/plain'
-    ], 'Hello, World!');
-}
+    $fortune = static function () use ($connection): PromiseInterface {
+        return $connection->query('SELECT id,message FROM Fortune');
+    };
```
WyriHaximus, line 25: Seems like a waste creating these every request, maybe a `global` or a singleton *shivers* is in order here?

frameworks/PHP/reactphp/reactphp.dockerfile
```diff
@@ -6,25 +6,26 @@ RUN apt-get update -yqq && apt-get install -yqq software-properties-common > /de
...
+COPY composer.json .
 RUN composer install --prefer-dist --optimize-autoloader --no-dev --quiet
 
+COPY . .
+
 EXPOSE 8080
 
-CMD php server.php
+ENTRYPOINT ["/usr/bin/php"]
+CMD ["server.php"]
```
WyriHaximus, line 31: Shameless plug maybe, but you have tried my ReactPHP focussed docker images? Also comes with JIT turned on out of the box for a nice boost. Here is a multi stage example so you don't have your dev tools in your final image: https://github.com/wyrihaximusnet/docker-php#example-usage

Just a suggestion FYI.

WyriHaximus (changes requested):

frameworks/PHP/reactphp/app.php
lcobucci: This isn't being created on each request, though.

The function scope and returned closure behaves as a singleton.

frameworks/PHP/reactphp/app.php
```diff
@@ -1,134 +1,149 @@
...
-    return match($request->getUri()->getPath()) {
-        '/plaintext' => text(),
-        '/json'      => json(), 
-        '/db'        => db(),
-        '/fortunes'  => fortune(),
-        '/query'     => query($request),
-        '/update'    => updateraw($request),
-        // '/info'      => info(),
-        default      => new Response(404, [], 'Error 404'),
+    $connection = establishDbConnection('benchmarkdbuser:benchmarkdbpass@tfb-database/hello_world?idle=0.5');
```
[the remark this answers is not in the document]
lcobucci: Not really, the lazy connection is created only once and reused everywhere

frameworks/PHP/reactphp/reactphp.dockerfile
lcobucci: I didn't want to change everything too drastically. If you think it would have huge impact on the benchmark, then maybe it'd make sense. However, I think we could/should defer that to another PR.

WyriHaximus: Yeah I get that, hence putting it in as suggestion :)

frameworks/PHP/reactphp/app.php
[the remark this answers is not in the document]
WyriHaximus: Pretty sure reading the code all the way up https://github.com/friends-of-reactphp/mysql/blob/0.7.x/src/Io/LazyConnection.php there is no way it gets reused. Unless I'm missing something.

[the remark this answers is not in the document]
lcobucci: I'll run tests to verify it but I'm pretty sure the reuse of `$connection` here combined with an 500ms idle limit makes `LazyConnection` to not open new internal connections in `LazyConnection#connecting()`

frameworks/PHP/reactphp/app.php
WyriHaximus: :facepalm: Just noticed it's returning a closure, you're right about here and `$connection`. Ignore this you're right :+1: . Going to feel stupid for a couple of minutes :joy: .

lcobucci: HAHAHAHA I imagined that the closure was indeed being missed from your review 😂

There's no need for feeling stupid, I can't think of anyone who never overlooked things ❤️

frameworks/PHP/reactphp/app.php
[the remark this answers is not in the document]
lcobucci: Just confirmed that we're indeed keeping the connection open (due to the closure thingy and the idle limit).
Sent several requests to the `/db` endpoint and `show processlist` in MySQL displayed the same PID for all requests.

frameworks/PHP/reactphp/app.php
joanhey: For good results, here we need to only execute the prepared statement.

WyriHaximus: > For good results, here we need to only execute the prepared statement.

Even better. 0.7 will come with a connection pool: https://github.com/friends-of-reactphp/mysql/issues/175

joanhey:
Please update to PHP 8.3
- #8579

lcobucci:
> Please update to PHP 8.3
> - #8579

Sure, will do around next week

lcobucci:
@joanhey a6657b758d397c73ccb60201403de52af0d9daca handles the PHP 8.3 upgrade :+1:

joanhey:
@lcobucci 
Could you check ReactPHP, as it's failing for a long time.
https://tfb-status.techempower.com/unzip/results.2025-01-04-12-20-28-252.zip/reactphp

WyriHaximus:
@joanhey Just checked those logs but don't see any obvious failures there. I will have a look at the code and docker file locally and see if I can spot what's wrong there.

WyriHaximus:
@joanhey Do you remember when it started breaking? Because it's a MySQL related issue, so one of these 3 broke it; https://github.com/TechEmpower/FrameworkBenchmarks/pull/8715, https://github.com/TechEmpower/FrameworkBenchmarks/pull/9009, or https://github.com/TechEmpower/FrameworkBenchmarks/pull/9183.

joanhey:
I can check when started to fail.
But it's more important, if anything fail than show the problem to the devs.

joanhey:
We update Mysql from 5.7 to 8.0 and now to 9.0,
Anybody can check in local this.
But for me, the worst think is than don't show where is the problem the framework.

joanhey:
If it fail for the Mysql, it's OK.
But why fail the tests without DB ?

joanhey:
Like I said, problems always exist, but the more important  think is that the language or framework, show you where is the problem.

joanhey:
Here started to fail:
May 7, 2024

image

I'll investigate what changed in the bench, but please also search what changed in ReactPHP !!

joanhey:
Changes in the toolset :
https://github.com/TechEmpower/FrameworkBenchmarks/commits/master/toolset

Changes in the benchmark code for ReactPHP:
https://github.com/TechEmpower/FrameworkBenchmarks/commits/master/frameworks/PHP/reactphp

joanhey:
The last run that work ReactPHP:
Here we show the last commits 
https://github.com/TechEmpower/FrameworkBenchmarks/commits/39e29cb04e7d78074ee363aea1e8e2e3d738cbc9/

The next run that failed:
https://github.com/TechEmpower/FrameworkBenchmarks/commits/c3e4826cf93907ed0175c7d8fc09488ab1fb4167/

I don't see anything bad in the commits.
Please check it !!

joanhey:
A good benchmark it isn't only for performance checks.
I use it to fix problems too.

Check: #9408

WyriHaximus:
> I can check when started to fail. But it's more important, if anything fail than show the problem to the devs.

Yes, which is why I just filed: https://github.com/TechEmpower/FrameworkBenchmarks/pull/9527

WyriHaximus:
> Changes in the toolset : https://github.com/TechEmpower/FrameworkBenchmarks/commits/master/toolset
> 
> Changes in the benchmark code for ReactPHP: https://github.com/TechEmpower/FrameworkBenchmarks/commits/master/frameworks/PHP/reactphp

Yeah checked those, had to put in a crude version of https://github.com/TechEmpower/FrameworkBenchmarks/pull/9527 to find the issue.
 
> PD: if you check the commits, you will see than I try my best, but it is not enough sometimes

It's a lot, and we're thankful for that.

> I'll investigate what changed in the bench, but please also search what changed in ReactPHP !!

Nothing actually, it was the MySQL upgrades to a version with the auth method ReactPHP currently supports enabled. (We need to add support for the never ones 😅 .)

WyriHaximus:
> If it fail for the Mysql, it's OK. But why fail the tests without DB ?

Because the connectivity test is done against the `/db` endpoint instead of `/plaintext`, not sure how to control that so filed  https://github.com/TechEmpower/FrameworkBenchmarks/pull/9528 as a step in the right direction. Because having this as out coming makes it a lot clearer where the problem might be:
image

joanhey:
You can test it locally
`./tfb --mode verify --test reactphp`

And if you want to test it directly calling the urls:
`./tfb --mode debug --test reactphp`

joanhey:
After add #9527

image

joanhey:
#9527 and #9528 never will be merged, as don't pass the tests.

joanhey:
Plaintext and json work without problems.
If you want, we can disable temporally the db tests till fixed.

WyriHaximus:
> Plaintext and json work without problems. If you want, we can disable temporally the db tests till fixed.

Lets start there, will file a PR for that 👍

WyriHaximus:
Given #8303 I've opened #9533

WyriHaximus:
> #9527 and #9528 never will be merged, as don't pass the tests. You need to test it locally till fix the problem.

FWIW: https://github.com/TechEmpower/FrameworkBenchmarks/pull/8715, https://github.com/TechEmpower/FrameworkBenchmarks/pull/9009, and  https://github.com/TechEmpower/FrameworkBenchmarks/pull/9183 all got merged with failing tests. 

> Plaintext and json work without problems. If you want, we can disable temporally the db tests till fixed.

Yup just filed https://github.com/TechEmpower/FrameworkBenchmarks/pull/9534 for that, will rebase the others once that is passing and merged

joanhey:
I created #9529 yesterday, now it's duplicated.

About the Databases update, normally fail, because always exist frameworks that fail.
With the Databases we test than run with the frameworks than normally work.

WyriHaximus:
> I created #9529 yesterday, now it's duplicated.

Missed that as it wasn't mentioned in the PR title, sorry.

> About the Databases update, normally fail, because always exist frameworks that fail. Even frameworks than don't use the database. With the Databases we test than run with the frameworks than normally work.

As a maintainer I get it, especially something this big I get it. And I get that #9528 will never be merged, that thing was a discussion starter. However, #9527 is perfectly fine to merge as it doesn't impact the benchmark negatively and will actually positively aid in finding issues with it.

> PD: the frameworks failing in the last run https://tfb-status.techempower.com/results/924216cc-3d66-43a2-ad8b-4d423e968e3e

Ow sweet thanks, it's kinda good to see we're not the only ones 😅 .

joanhey:
#9257 only show the error, but don't fix the problem.

The tests will fail, and no framework is merged if don't pass the tests in the PR.

joanhey:
I don't mentioned in the title, because the important thing is the update to PHP8.4.
But I comment it that it's necessary disable the db tests and fix the broken run.

WyriHaximus:
> #9257 only show the error, but don't fix the problem.
> 
> The tests will fail, and no framework is merged if don't pass the tests in the PR.

#9538 will not fix the database issue, but at least respond with an error

WyriHaximus:
> I don't mentioned in the title, because the important thing is the update to PHP8.4. But I comment it that it's necessary to disable the db tests to fix the broken run.

Fair, I was just expecting it in it's own dedicated PR. But that's up to you as maintainer how you prefer to handle that :)

PR #8640, opened 2023-12-14
[Viz]: v0.6.0

fundon:
* viz v0.6
* hyper v1.0


frameworks/Rust/viz/viz.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:1.72
+FROM rust:latest
```
NateBrady23, line 1: Please keep this locked down to a specific version

fundon: Moved to #8658

PR #8660, opened 2023-12-21
Upgrade axum benchmark dependencies

jplatte:
I actually wanted to upgrade axum itself to 0.7, but found that a few hyper server settings were used here that hyper 1.0 seems to not support as nicely. I'll post another PR or update this one once I know what to do about them.


jplatte:
@NateBrady23 you merged previous axum-related PRs, mind taking a look?

PR #8668, opened 2023-12-28
[ruby/rails] Use 3 threads per worker

p8:
Using 3 threads instead of 5 (as used in master) seems optimal.
Running this locally I got the following results:

```
+---------------------+---------+------+-----+-----+-----+-------+--------------+
|                     |plaintext|update| json|   db|query|fortune|weighted_score|
+---------------------+---------+------+-----+-----+-----+-------+--------------+
| 2 threads per worker|    20870|  9478|66764|16220|11634|  10543|          1042|
| 1 thread  per worker|    20799|  9819|83123|17434|10334|   8653|          1048|
| 4 threads per worker|    29257| 10051|69532|18108|11752|  11132|          1093|
| 5 threads per worker|    33152| 10203|77062|18459|11721|  11711|          1114|
| 3 threads per worker|    25077| 10382|84429|16497|12338|  11246|          1141|
+---------------------+---------+------+-----+-----+-----+-------+--------------+
```

### DIFF
--- a/frameworks/Ruby/rails/config/database.yml
+++ b/frameworks/Ruby/rails/config/database.yml
@@ -5,7 +5,7 @@ default: &default
   password: benchmarkdbpass
   host: tfb-database
   timeout: 5000
-  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 3 } %>
 
 development:
   <<: *default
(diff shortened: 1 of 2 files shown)


joanhey:
Remember than your local server, it's different from the Citrine servers.
It'll be better to test in Citrine, as we have continuous runs, try to add separate PRs after each run to test with number of threads.

Also from these numbers I see an strange behavior.
The `json` test is ~3x faster than the `plaintext`, that is not normal.

Happy new year !!!

p8:
Happy new year @joanhey !
I'll see if things can be improved in separate PRs.
The difference between `json` and `plaintext` deserves an investigation.

PR #8678, opened 2024-01-09
[ruby/rack] Enable Rubys M:N thread scheduler for Puma

p8:
With the M:N thread scheduler thread creation and management cost are reduced.


NateBrady23:
@p8 You have a few PRs marked as drafts that have been open for a bit. Anything to do here?

p8:
@NateBrady23 These are just waiting for a new run with the existing version, so we can compare if this improves things.

PR #8679, opened 2024-01-09
[ruby/rails] Enable Rubys MN threads

p8:
With the M:N thread scheduler thread creation and management cost are reduced.

Waiting for the initial run on the new servers so we can compare if this improves things.


p8:
This doesn't seem that stable yet. Closing for now.

PR #8681, opened 2024-01-10
[php] React fix deprecation

joanhey:
@lcobucci


lcobucci:
Nice catch! I've totally missed that 🤦

PR #8714, opened 2024-01-20
PostgreSQL update to v16

joanhey:
Released September 14, 2023.
https://www.postgresql.org/about/news/postgresql-16-released-2715/

After update the Citrine servers, it'll be good to first run with the last commit with full results, to compare the real differences with the new servers.
Or wait to update the databases, after 1-2 full runs in the new servers.

And after check the new specs, we need also to check the database configurations.

### DIFF
--- a/toolset/databases/postgres/postgres.dockerfile
+++ b/toolset/databases/postgres/postgres.dockerfile
@@ -1,4 +1,4 @@
-FROM postgres:15-bullseye
+FROM postgres:16-bullseye
 
 ENV POSTGRES_USER=benchmarkdbuser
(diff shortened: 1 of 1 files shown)


toolset/databases/postgres/postgres.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM postgres:15-bullseye
+FROM postgres:16-bullseye
```
volyrique, line 1: I suggest switching to Debian Bookworm as well, to benefit from the work that has gone into C library, i.e. `FROM postgres:16-bookworm`.

PR #8742, opened 2024-02-20
[Rust] Fix the Rocket framework bench

Nautilus-1021:
There was 2 benchmarks for the [Rocket](https://rocket.rs/) framework, one with the Diesel database driver and one with the SQLX database driver. Both were tagged as broken, And don't exactly know why.

# The SQLX test

But I fixed the SQLX one, and I updated the dependencies. ~~But I removed the database update benchmark because when I tested it locally, it crashed because of a deadlock in the database. It could be because of the large amount of requests running simultaneously, but I thought that can't be the problem and that there was securities, but apparently no, so because I can't fix it with my capacities, I removed the update test.~~

# The Diesel test

The original code of the Diesel test is still in the repository, so if someone wants to fix it, he can.


Nautilus-1021:
Sorry for all the force pushs, I did a mistake while trying to rebase my branch. But now everything is good.

And I fixed the Rust version in the Dockerfile

volyrique:
> Both were tagged as broken, And don't exactly know why.

I think that PR #8255, which was the one that marked both implementations as broken, was quite clear - the implementations failed to start consistently in the continuous benchmarking environment, where consistently means in at least 3 consecutive runs. Just check all runs that completed before 13.06.2023, and you will notice the issue.

Nautilus-1021:
> > Both were tagged as broken, And don't exactly know why.
> 
> I think that PR #8255, which was the one that marked both implementations as broken, was quite clear - the implementations failed to start consistently in the continuous benchmarking environment, where consistently means in at least 3 consecutive runs. Just check all runs that completed before 13.06.2023, and you will notice the issue.

Thank you for the explanation

PR #8789, opened 2024-03-05
[New Framework]: Oxygen.jl

ndortega:
Hi there, I'm the author of the Oxygen.jl framework and this is my first attempt at registering the framework, please let me know if anything needs to be changed before this can be merged


joanhey:
- #8420

frameworks/Julia/oxygen/oxygen.dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM julia:latest
```
NateBrady23, line 1: Can you tie this down to a specific version please

ndortega: @NateBrady23 
How specific are we talking? Would setting it to version 1 be specific enough or are you looking for an exact 1.x.x version?

NateBrady23: If their version convention is such that breaking changes can be introduced in 1.x, then 1.x.x would be best.

ndortega: Julia is super good about backward compatibility, so setting to the latest stable version 1 should be fine.

pankgeorg: It's not a compatibility issue. But 1.10 does different scheduling which brings different performance. It's not good enough yet, but it's significantly better (mainly because you can avoid being blocked from precompilation)

PR #8819, opened 2024-03-22
Add mrhttp

MarkReedZ:
Adding mrhttp the fastest python web framework

    https://github.com/MarkReedZ/mrhttp


joanhey:
- #8420

MarkReedZ:
Plaintext and json are supported

PR #8851, opened 2024-04-10
[Rust] update may_minihttp dependency

Xudong-Huang:
* update may_minihttp dependency to 0.1 from crates.io
* update to use latest rust toolchain

### DIFF
--- a/frameworks/Rust/may-minihttp/may-minihttp.dockerfile
+++ b/frameworks/Rust/may-minihttp/may-minihttp.dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.72
+FROM rust:latest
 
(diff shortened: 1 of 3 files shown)


frameworks/Rust/may-minihttp/may-minihttp.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:1.72
+FROM rust:latest
```
joanhey, line 1: Please, use a a specific version.

Xudong-Huang: done

joanhey (approved):

PR #8852, opened 2024-04-10
[C++] [userver] Upscale to the new environment

itrofimow:
Userver is configured to use fixed amount of threads for different things, and that amount was bound to the vCPU-s available.
Since the environment update doubled available vCPU-s, I'd like to upscale userver accordingly.

### DIFF
--- a/frameworks/C++/userver/userver_configs/static_config.yaml
+++ b/frameworks/C++/userver/userver_configs/static_config.yaml
@@ -1,7 +1,7 @@
 # yaml
 components_manager:
     event_thread_pool:
-        threads: 5
+        threads: 9
         dedicated_timer_threads: 1
     coro_pool:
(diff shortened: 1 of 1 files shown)


joanhey:
Why not calculate it based in the number vCPUs.
We also run this benchmark in other environments: cloud, local, ...

itrofimow:
I think we already had this conversation, didn't we?

My users (myself included) have all the means required to configure userver as they see fit, and for this particular load I prefer it configured this way.

joanhey:
You can do it in the dockerfile with sed !!
Give to your users a default !!!
And later they can change it like they want !!

But this framework will have very bad results with the cloud bench :(

joanhey:
No more comments !!
Do it like you want.

PR #8860, opened 2024-04-11
Add --reverse-order option and switch order after each run on citrine

NateBrady23:


joanhey:
It is a good time for try it !!!
Each time that fail the continuous run start at the beginning again, the last frameworks never know the results !!!

PR #8908, opened 2024-04-17
[Ruby] Add rage-rb

rsamoilov:
Rage is a fast web framework compatible with Rails. It uses an event-driven architecture and implements a lightweight, cooperative concurrency model based on Ruby Fibers.

https://github.com/rage-rb/rage

===

I'm not sure I've worked out the size of the DB pool correctly. It's possible I will need to reduce it in case there are errors under high concurrency. But before that, I'd like to see how it performs on your setup with these settings.


rsamoilov:
hey @NateBrady23 , is there something I could fix or help clarify to get this merged?

PR #8951, opened 2024-04-28
Various toolset fixes and updates

volyrique:
* Fix the MongoDB Docker image
* Update the toolset Docker images to Ubuntu 24.04
* Update Dool to version 1.3.1
* Fix any warnings that are printed by the toolset scripts


NateBrady23:
@volyrique Thank you so much for taking care of this!

PR #8993, opened 2024-05-07
[ruby/padrino] Use `update_columns` instead of `update`

p8:
Unlike `update_columns`, `update` runs in a transaction and runs callbacks. Transactions aren't a requirement, so we can just call `update_columns` for improved performance.


p8:
DataMapper doesn't support `update_columns`

PR #8997, opened 2024-05-08
[ruby] Combine *-sequel frameworks with their main frameworks

p8:
`rack-sequel` and `sinatra-sequel` only use a different ORM (sequel) than `rack` and `sinatra`. There is no need to test `/json` and `/plaintext` for `rack-sequel` and `sinatra-sequel`, as these calls don't use the ORM.

For `sinatra-sequel` the `/json` and `/plaintext` calls were already removed from the benchmark config in 1f94f389dcc30f1795b23953ad152560d5433f9f. This change does the same for `rack-sequel`.

To still see these results in the combined table, the framework is set to `rack` and `sinatra`.


p8:
Hmm, ok I'll probably need to merge the directories instead.

PR #9011, opened 2024-05-11
Update ZIO HTTP Project

khajavi:
Updated the ZIO HTTP project to the latest RC version.

### DIFF
--- a/frameworks/Scala/zio-http/src/main/scala/Main.scala
+++ b/frameworks/Scala/zio-http/src/main/scala/Main.scala
@@ -1,42 +1,48 @@
-import zhttp.http._
-import zhttp.service.Server
-import zio.{App, ExitCode, URIO}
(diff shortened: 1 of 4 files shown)


frameworks/Scala/zio-http/src/main/scala/Main.scala
```diff
@@ -1,42 +1,48 @@
...
+      ),
+  )
+
+  private val config = Server.Config.default
+    .port(8080)
+    .enableRequestStreaming
+
+  private val nettyConfig = NettyConfig.default
+    .leakDetection(LeakDetectionLevel.DISABLED)
+    .maxThreads(8)
```
Srujithchaithanya1, line 40: may be you can set the number of threads dynamically based on the available CPU cores to make better use of system resources instead of 8

Srujithchaithanya1 (commented):
Everything seems fine

PR #9016, opened 2024-05-13
[ruby] Reduce `random_id` calls in queries and update

p8:
Calling `Array#sample` with a size x is faster than calling `rand` x times.
Depends on: #9027

### DIFF
--- a/frameworks/Ruby/grape/config.ru
+++ b/frameworks/Ruby/grape/config.ru
@@ -4,6 +4,8 @@ require 'yaml'
 require_relative 'config/auto_tune'
 
 MAX_PK = 10_000
+ID_RANGE = (1..MAX_PK).freeze
+ALL_IDS = ID_RANGE.to_a
 QUERIES_MIN = 1
 QUERIES_MAX = 500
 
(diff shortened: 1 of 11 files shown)


NateBrady23:
That's really interesting. My kneejerk reaction to this is, doesn't `.sample` just call `rand(x)` or the equivalent `rb_random_int` behind the scenes `x` times?

p8:
I don't think I understand the [implementation.](https://github.com/ruby/ruby/blob/master/array.c#L6540)
It is a bit faster though:
```ruby
require 'benchmark/ips'

Benchmark.ips do |x|
  # Configure the number of seconds used during
  # the warmup phase (default 2) and calculation phase (default 5)
  x.config(:time => 5, :warmup => 2)

  QUERY_RANGE = (1..10_000).freeze # range of IDs in the Fortune DB
  x.report("rand") do
    Array.new(500) { rand(QUERY_RANGE) }
  end

  ALL_IDS = QUERY_RANGE.to_a # enumeration of all the IDs in fortune DB
  x.report("sample") do
    ALL_IDS.sample(500)
  end

  # Compare the iterations per second of the various reports!
  x.compare!
end
```

```
ruby 3.2.4 (2024-04-23 revision af471c0e01) [x86_64-darwin22]
Warming up --------------------------------------
                rand     1.666k i/100ms
              sample     6.395k i/100ms
Calculating -------------------------------------
                rand     16.709k (± 1.4%) i/s -     84.966k in   5.085923s
              sample     65.047k (± 2.7%) i/s -    326.145k in   5.017962s

Comparison:
              sample:    65047.4 i/s
                rand:    16709.4 i/s - 3.89x  slower
```

PR #9019, opened 2024-05-14
[ruby/rails] Update trilogy to 2.8.1

p8:
This fixes connection errors and memory leaks.


p8:
The trilogy adapter doesn't yet support caching_sha2_password authentication.

PR #9024, opened 2024-05-15
[php] Symfony add RoadRunner

joanhey:
Using RoadRunner v2023.3.
The new version 2024, still don't work with symfony php-runtime.

Clean the dockerfiles.
And use image `composer/composer` (2.9Mb) vs `composer` (199MB).


frameworks/PHP/symfony/.rr.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+server:
+  command: "php /symfony/public/runtime.php"
+  env:
+    - APP_RUNTIME: Runtime\RoadRunnerSymfonyNyholm\Runtime
+
+http:
+  address: "0.0.0.0:8080"
+
+  middleware: ["headers"]
```
rustatian, line 12: @joanhey Not needed as well here.

frameworks/PHP/symfony/symfony-roadrunner.dockerfile
```diff
@@ -0,0 +1,25 @@
...
+COPY --link . .
+
+ENV APP_RUNTIME="Runtime\RoadRunnerSymfonyNyholm\Runtime"
+RUN composer require runtime/roadrunner-symfony-nyholm --update-no-dev --no-scripts --quiet
+RUN cp deploy/postgresql/.env . && composer dump-env prod && bin/console cache:clear
+
+EXPOSE 8080
+
+RUN rr -v
+ENTRYPOINT ["rr", "serve"]
```
rustatian, line 25: @joanhey RR by default (at least on spiral tests) started with `./rr serve -o "http.pool.num_workers = 64"`
-o option means overwrite the default value. If you use default `rr server` here, you may set num_workers directly in the configuration (.rr.yaml)

frameworks/PHP/symfony/.rr.yaml
```diff
@@ -0,0 +1,25 @@
...
+  headers:
+    response:
+      Server: "RoadRunner"
+  #pool:
+    #num_workers: 0
```
rustatian, line 17: @joanhey Here, as I mentioned below, you may use `pool.num_workers: 64`

joanhey: By default use 1 worker by CPU.
The new servers have 28 cores and 56 threads. The old servers had only 14 cores and 28 threads.

rustatian: Generally, RR respects `cgroup` limitation in the Dockerfile automatically (if you have these limits).
Hard to say what is more correct, at some point you won't benefit from increasing workers number, but where is this point.. At x2 or x3.. For Hyper-Threading threads, you may start from x1.5 worker's number and then try x2.5. I expect, somewhere in-between would be the perfect number 😄

joanhey: How work RR, with threads or with forks ?

rustatian: RR is a Go application which uses `fork_exec` system call to start PHP CLI processes. It is not a SAPI, so, RR works with both, with TSRM (ZTS) PHP and non-ZTS PHP.

rustatian: So basically, workers are child processes with PHP CLI inside. RR then uses IPC to communicate with them, using our own protocol called Goridge.

joanhey: The PHP is running like forks so.
We can use till x 4 workers I think.

I'm the 2º contributor to ngx-php (php embeded in Nginx) and Workerman. I also created Adapterman.
But I never touched RoadRunner, before last week, to add Symfony RR.

rustatian: Good to know 👍 Yeah, you may try x4 at first.

RoadRunner is not only a server, it is an ecosystem of plugins. So, you may create you own plugin, connect and extend RR. You may take a look at the RR repo, there are a lot of them. Plugin then can send payloads and receive responses from the pools of workers (aka PHP CLI processes).

joanhey: I think than the opcodes (opcache) are not reused, because RR is using the `exec()`, with `fork()` alone need to work.
Each php cli process executed is independent and don't share the opcodes.

rustatian: Yes, I mentioned about that before, that Go uses fork_exec system call (fork followed by exec). So, every CLI PHP process had its own SHM pointer, not shared with other CLI processes. But, this doesn’t matter, since everything stays in memory. Unlike FPM which unloads the PHP script after execution. FPM approach would benefit from OPCache.

PR #9025, opened 2024-05-15
[php] Update Laravel with RoadRunner v2023.3

joanhey:


frameworks/PHP/laravel/deploy/roadrunner/.rr.yaml
```diff
@@ -7,16 +10,14 @@ logs:
   level: error
 http:
   address: 0.0.0.0:8080
-  middleware: ["headers", "static", "gzip"]
+  middleware: ["headers"]
```
rustatian, line 13: @joanhey 👋 
You don't need to use middleware at all here. You may safely remove the whole `middleware` line + headers section.

joanhey: If I remove it, don't pass the tests.
The test need a server: header

PR #9021, opened 2024-05-14
[PHP, Spiral/RoadRunner]: update `Dockerfile` to address RR `v2024.x.x` requirements

rustatian:
Starting from the RR `v2024.1.0` it requires `protobuf` and `grpc` extension installed. They used to improve the performance of parsing proto-requests.

PR #9032, opened 2024-05-16
[php] Fix #9021 fail to build grpc

joanhey:


joanhey:
The update test fail, like before, but it'll be fixed in another PR.
Actually with the #9021 fail building the container, and it's very slow building til fail.

PR #9034, opened 2024-05-17
Spiral: Add Cycle ORM config

roxblnfk:
## What was changed:

Added Cycle ORM config with `cache=true`

## Why?

https://github.com/TechEmpower/FrameworkBenchmarks/pull/9021#issuecomment-2116174844

I suppose that unnecessary requests arise because each of the RoadRunner workers synchronizes the ORM schema with the database.
Enabling cache might help avoid the issue.

### DIFF
--- a/frameworks/PHP/spiral/app/config/cycle.php
+++ b/frameworks/PHP/spiral/app/config/cycle.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+use Cycle\ORM\SchemaInterface;
+
+return [
+    'schema' => [
+        /**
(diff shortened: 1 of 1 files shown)


joanhey:
The problem with the update test is necessary to fix it to pass the tests.
The ORM for the number of queries need to update that number of rows in the database.
Well forcing to update all the worlds and/or forcing to have different random number for ID.

roxblnfk:
Can we just hop off for Postgres instead of MySQL?

msmith-techempower:
Closing for hygiene - reopen or make a new PR when ready.

PR #9038, opened 2024-05-18
toolset: Update pymongo to version 4.7.2

volyrique:


volyrique:
Since this is a toolset change, it triggered checks for all implementations. A lot of the test failures concern frameworks that do not use MongoDB; other cases had failed during the last complete continuous benchmarking run, so they do not signify a regression either. Implementations that actually use MongoDB and that work fine normally such as `tornado` and `tornado-py3` finish the tests successfully.

PR #9049, opened 2024-05-22
Micronaut performance improvements

dstepanov:
- More database pool threads
- Shared Netty event loop and Vertx PgClient event loop
- Let context switching
- Latest version of Gradle / JVM


frameworks/Java/micronaut/micronaut-vertx-pg-client/src/main/java/benchmark/ConnectionHolder.java
```diff
@@ -0,0 +1,140 @@
...
+
+    @Property(name = "datasources.default.url") String url;
+    @Property(name = "datasources.default.username") String user;
+    @Property(name = "datasources.default.password") String password;
+    @Property(name = "datasources.default.maximum-pool-size") int maxPoolSize;
+
+    public Future<PgConnection> get() {
+        Future<PgConnection> c = conn.get();
+        if (c == null) {
+            NettyHttpRequest<?> req = (NettyHttpRequest<?>) ServerRequestContext.currentRequest().get();
```
yawkat, line 43: i think i prefer ThreadExecutorMap, unless you have a reason to use the ServerRequestContext

yawkat (approved):

graemerocher (approved):

frameworks/Java/micronaut/micronaut-vertx-pg-client/src/main/java/benchmark/ConnectionHolder.java
yawkat: but the request is only used to get the event loop.

yawkat: just cast it. i think it should work

dstepanov: Ok

yawkat: thanks. does it work? :sweat_smile:

dstepanov:
@NateBrady23 Please review :)

PR #9078, opened 2024-06-01
Another PostgreSQL configuration file clean-up

volyrique:
This change makes it easier to check how exactly we deviate from the default settings.


NateBrady23:
I like that. Thanks!

PR #9084, opened 2024-06-03
[micronaut] Disable testresources and downgrade docker images

yawkat:
This is an attempt to fix the JIT-only regression introduced by f2b0b607acb50c71890f00693bc2f467d217ba4f .


yawkat:
@dstepanov @graemerocher

dstepanov (approved):

PR #9115, opened 2024-06-13
Perl Kelp: Fix and modernize

bbrtj:
I am the current maintainer of the Perl Kelp framework. This PR fixes and rewrites its benchmark code. I took the approach to only fix what has already been done by the author, so I neither added the new cache test or more databases (like Postgre) nor removed stuff that were already there (like `t/main.t` test).

Status:
- `kelp` (MariaDB) passes tests with `tfb`
- `kelp-mongodb` passes tests with `tfb`
- bundled test file passes with `prove -l`

I haven't contributed any code here up until now, so let me know if anything else needs to be changed.


bbrtj:
I've chosen that the server should replace processes by new ones after handling `10000` requests. While it would surely perform better with an even higher number, I think this is a sensible production value. On top of that, it's the same as the default number of requests handled by Hypnotoad (mojolicious), so it will be more accurate overview of the performance differences between the two.

Side note: dancer benchmark uses Starman server with a default of `1000` requests before process shutdown, so that value should likely be bumped. Would be nice if all Perl benchmarks used the same value (possibly ten thousand or higher to better compete against other languages).

bbrtj:
- Added tests on multiple Perl webservers, 8 configurations in total now
- removed the .toml file, as it seems to not be required anymore (tell me if I'm wrong)
- increased the number of requests per process to 100,000

bbrtj:
I'm finished, ready for review and merge. After it's merged, I'll make another PR which will increase requests per process of all perl framework benchmarks to the same value (one hundred thousand).

PR #9151, opened 2024-07-09
Perl: Fix non-functional benchmarks

bbrtj:
This PR attempts to unify the value of maximum requests per child to one hundred thousand across all Perl frameworks. This value can have drastic effects on performance and it was varying greatly across the frameworks. `100000` is not an unrealistic value and should allow the frameworks to still compete with other, faster languages. On the other hand, it will make comparisons between different Perl frameworks more accurate.

As a result of this PR, Mojolicious and Plack should become a bit slower, while Dancer and Web::Simple should become quite a lot faster. Kelp was already set to a proper value in recently merged PR.

### DIFF
--- a/frameworks/Perl/dancer/README.md
+++ b/frameworks/Perl/dancer/README.md
@@ -8,7 +8,7 @@
 
 * Dancer
 * Dancer::Plugin::Database
-* DBD::mysql
+* DBD::MariaDB
 * Starman (if using Starman as web server)
 * Plack (for plackup)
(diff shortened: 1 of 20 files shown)


bbrtj:
This PR is sure to reignite some issues with old frameworks no longer passing tests. May need more work to make them all functional.

bbrtj:
I am finished, ready for review and merge. All existing Perl framework tests should now be functional.

PR #9159, opened 2024-07-11
Gnet is unrealistic

uNetworkingAB:
gnet is definitely not realistic.

* Its entire "HTTP parser" is a 1-line seek for "\r\n\r\n" and doesn't even handle a TCP segment being... segmented. It does no header parsing whatsoever.
* Its response is fully pre-formatted except for the Date.

gnet is baically a TCP benchmark, not an HTTP benchmark. Any TCP server could be made as complex as gnet with less than 5 lines of code added.

Therefore, it definitely is not "realistic". Especially not since it literally placed 1st in Round 22. How this slipped past quality control is beyond me, but also doesn't shock me.

### DIFF
--- a/frameworks/Go/gnet/benchmark_config.json
+++ b/frameworks/Go/gnet/benchmark_config.json
@@ -4,7 +4,7 @@
     "default": {
       "plaintext_url": "/plaintext",
       "port": 8080,
-      "approach": "Realistic",
+      "approach": "Unrealistic",
       "classification": "Platform",
       "database": "None",
       "framework": "gnet",


uNetworkingAB:
Quoting their own page:

**Note that the HTTP implementation of gnet on TechEmpower is half-baked and fine-tuned for benchmark purposes only and far from production-ready.**

NateBrady23:
> How this slipped past quality control is beyond me, but also doesn't shock me.

We rely on community members like yourself to look at this. These kinds of toxic comments don't belong here. You can keep them to your own open source repo with comments like, "You have the formatting of a psychopath." Keep that out of here.

uNetworkingAB:
Then how come a simple PR took 2 months to review, if there obviously is nothing being reviewed?

Taking a look for more than 5 seconds into gnet would make it obvious that it's not a valid subject.

NateBrady23:
You couldn't even get this right. "Stripped" needs to be capitalized. And here we are trusting the king of quality control. Happy to remove your framework. Seeing your tweets now as well. You're not interested in reviewing code and making things better; you're interested in being right and you're too toxic about it. Most PR's take a while to review. We work 60 hours a week for paying clients. This isn't the focus. You have unanswered PR's for longer than 2 months, yeah? You want to be miserable and yell at someone, do it somewhere else.

PR #9160, opened 2024-07-11
Remove uwebsockets

NateBrady23:
Unfortunately the maintainer is no longer able to make pull requests to this repo.

PR #9162, opened 2024-07-11
Update readme according to recent changes in the project

angrymouse:
According to #9160, readme might not be up to date and thus not reflect current status of the project.
This PR clears out any irrelevant points from readme that might cause confusion to people and possibly get their project removed from benchmarks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Build Status](https://github.com/TechEmpower/FrameworkBenchmarks/workflows/build/badge.svg?branch=master&event=push)](https://github.com/TechEmpower/FrameworkBenchmarks/actions?query=workflow%3Abuild+branch%3Amaster)
 
-If you're new to the project, welcome! Please feel free to ask questions [here](https://github.com/TechEmpower/FrameworkBenchmarks/issues/2978). We encourage new frameworks and contributors to ask questions. We're here to help!
+If you're new to the project, welcome! Please feel almost-free to ask questions [here](https://github.com/TechEmpower/FrameworkBenchmarks/issues/2978) (but only positive ones). We encourage new frameworks and contributors to ask questions, but not ones that criticize us. 
 
 This project provides representative performance measures across a wide field of web application frameworks. With much help from the community, coverage is quite broad and we are happy to broaden it further with contributions. The project presently includes frameworks on many languages including `Go`, `Python`, `Java`, `Ruby`, `PHP`, `C#`, `F#`,`Clojure`, `Groovy`, `Dart`, `JavaScript`, `Erlang`, `Haskell`, `Scala`, `Perl`, `Lua`, `C`, and others.  The current tests exercise plaintext responses, JSON serialization, database reads and writes via the object-relational mapper (ORM), collections, sorting, server-side templates, and XSS counter-measures. Future tests will exercise other components and greater computation.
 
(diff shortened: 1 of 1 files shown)


NateBrady23:
The open source community is full of amazing people. Not once in ten years doing this have we have had to block anyone for their conduct. I've had to delete extremely toxic comments, not "questions from new contributors."

I looked at your profile and you don't seem like one of those toxic people, so not sure I understand this.

angrymouse:
> I looked at your profile and you don't seem like one of those toxic people, so not sure I understand this.
Thanks @NateBrady23, not trying to spread hate but certainly to make a statement and actually start discussion about is it the right approach or not.
I am a big fan of uWebSockets(.js) for the performance reasons of it, and it seems very unfair to me that comment of library of maintainer about benchmarks on third party platform (not even spamming in issues/PRs, anything people would encounter when visiting the project), gets his *library* deleted from the benchmark list. He could have made anonymous account and said exactly the same thing.
Alex Hultman isn't very kind person - but getting the library removed for even not-the-most-constructive feedback about the benchmark isn't exactly a good look to me. 
I'm all for kind and nice conversations, although I would also love to be able to say criticism without risk of my library getting removed for that.
Thanks for maintaining and hope the best future for this benchmark, let's keep it as comprehensive as possible, without author bias.

NateBrady23:
Absolutely. I don't mind criticism. Look through the issues, we have tons of that! I also understand that this benchmark project isn't for everyone. Some people think it's about winning first; others think it's about comparing and finding optimization solutions you hadn't thought about, etc. 

The problem with this person is they took it too far. They accused us of favoring other PRs over theirs, made strange comments we had to delete, and blasted us on Twitter saying they disliked the project and it was a joke (before getting banned.) 

So certainly any reasonable person with (even harsh) criticisms is welcome to bring those to us. We just always hope to have a respectful conversation around it. 

As far as removing the framework, I'm not sure there's another way. It's unfair keeping the framework if the core maintainer isn't allowed to make updates to the tests. We also remove frameworks if core maintainers ask us to. We don't want a stale implementation or something they don't promote here if they don't want it here. 

In this particular case, the maintainer made it pretty clear they don't support or endorse the project, which is absolutely their right.

e3dio:
> The problem with this person is they took it too far. They accused us of favoring other PRs over theirs

This is so horrible that you had to ban a person who has large amounts of knowledge in this field and remove probably the fastest project of its type ? Favoring a PR ?

> made strange comments we had to delete

Strange comments? What strange comments? All his posts I see are highly knowledgeable and contribute greatly to the conversation about library performance, valuable to anyone. Maybe there was an insensitive comment somewhere ?
 
> and blasted us on Twitter saying they disliked the project and it was a joke

I have seen far, far worse on Twitter / X, not exactly a place to go if you want a safe space. At least none of that happened in your GitHub repo? Are we not allowed to have independent thoughts anymore anywhere ? And it appears there was a legitimate point being made about invalid HTTP servers being at the top of the list ?

> In this particular case, the maintainer made it pretty clear they don't support or endorse the project, which is absolutely their right.

He's got your benchmarks on the front page of the [ReadMe](https://github.com/uNetworking/uWebSockets.js), maybe he just wants the results to be accurate ? 

I personally use uWS.js and I think it's great so I don't like to see it shut down anywhere for frivolous reasons, maybe if he promise to be nice you can let him update the project or something

PR #9169, opened 2024-07-17
[php] Fix Swoole and faster docker builds

joanhey:
The last version require `libbrotlienc` and break the build in last runs. Also break the semantic versioning.
We can add `libbrotli-dev `before install Swoole, but this will break the docker cache from our last PR. And Swoole need some time to compile.

Now we use `phpswoole/swoole:5.1.3-php8.3` docker image, to have faster docker builds and fix this and future problems.


joanhey:
Ping @NathanFreeman

joanhey:
Ready to be merged !!

PR #9182, opened 2024-07-30
Bumped version for Java, Jooby, Maven and other libraries/dependencies

lospejos:


jknack:
@lospejos This line: [app.mvc(new Resource(app.require(DataSource.class)));](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/jooby/src/main/java/com/techempower/MvcApp.java#L20) need to be:

```java
      app.mvc(new Resource_(app.require(DataSource.class)));
```

That will fix the build error.

PR #9198, opened 2024-08-08
Patch/elixir updates

atavistock:
Update the version of Erlang and Elixir for most current.  There have been significant performance changes over the last year which should have a noticeable impact on performance.  This didn't require and changes to the web/test code itself.

Update the version of Phoenix and some other libraries to most current.  Again there were some optimizations which should have an impact on the overall performance.  This did require a small confirmation changes to how libraries are loaded.

Overall there are no changes to the actual controller or html delivery here.


frameworks/Elixir/phoenix/lib/hello_web/controllers/page_controller.ex
```diff
@@ -41,11 +42,11 @@ defmodule HelloWeb.PageController do
       message: "Additional fortune added at request time."
     }
 
-    fortunes = [additional_fortune | Repo.all(Fortune)]
+    fortunes =
+      [additional_fortune | Repo.all(Fortune)]
+      |> Enum.sort_by(& &1.message)
```
josevalim, line 47: Have you benchmarked this one? I actually expect `sort` to be more efficient, because `sort_by` needs to do additional passes on the data.

frameworks/Elixir/phoenix/lib/hello_web/controllers/page_controller.ex
```diff
@@ -29,7 +30,7 @@ defmodule HelloWeb.PageController do
     worlds =
       Stream.repeatedly(&random_id/0)
       |> Stream.uniq()
-      |> Stream.map(fn idx -> Repo.get(World, idx) end)
+      |> Stream.map(&Repo.get(World, &1))
       |> Enum.take(size(params["queries"]))
```
josevalim, line 34: We can optimize this in two ways:

1. Wrap the whole operation on a `Repo.checkout`, so we avoid checking out the connection multiple times
2. Use Task.async/await or Task.async_stream so we run the queries in parallel

frameworks/Elixir/phoenix/lib/hello_web/controllers/page_controller.ex
```diff
@@ -54,9 +55,13 @@ defmodule HelloWeb.PageController do
...
       Stream.repeatedly(&random_id/0)
       |> Stream.uniq()
-      |> Stream.map(fn idx -> Repo.get(World, idx) end)
+      |> Stream.map(&Repo.get(World, &1))
       |> Stream.map(fn world -> %{id: world.id, randomnumber: :rand.uniform(@random_max)} end)
       |> Enum.take(size(params["queries"]))
+      # If this is not sorted it sometimes generates
+      #  FAIL for http://tfb-server:8080/updates/20
+      #  Only 20470 executed queries in the database out of roughly 20480 expected.
+      |> Enum.sort_by(& &1.id)
```
josevalim, line 64: Do you know why we need to sort it? I can't see anything that would explain it. :D

atavistock: I'm not sure on this.  I was digging into why this intermittently failed, found the Rails version did a sort at the end,  and found comments in other languages about this exact issue.

https://github.com/TechEmpower/FrameworkBenchmarks/blob/db303286a70ce0bd90f2ca02c95b01d7c5456548/frameworks/Rust/tide/src/handlers.rs#L124-L137

josevalim: Thanks for sharing. I don't understand why but at least it is "consistent". :) I will try to dig deeper later.

And, if possible, please let me know if the parallelism or Repo.checkout changes I suggest above are allowed. I think they could improve the timing of a few endpoints. :)

atavistock: I'm not affiliated with this project other than enjoying Elixir wanting to make it show well.  ;) 

Your feedback is **greatly** appreciated.  IMHO using `Repo.checkout` and `Tasks` are within both the rules of many of the tests and within spirit of the benchmarks.   I can make a PR for those but won't be able to until ~7 hours from now.

josevalim: I see, thank you! :heart:

After thinking a bit about this, I would probably write this as:

```elixir
      Stream.repeatedly(&random_id/0)
      |> Stream.uniq()
      |> Enum.take(size(params["queries"]))
      |> Enum.sort()
      |> Enum.map(fn id ->
        world = Repo.get(World, id)
        %{id: world.id, randomnumber: :rand.uniform(@random_max)}
      end)
```

The first three parts of the pipeline can be abstracted into a function which we can reuse across actions:

```
    def sample(params_count) do
      Stream.repeatedly(&random_id/0)
      |> Stream.uniq()
      |> Enum.take(size(params_count))
    end
```

atavistock: I've done some iterations of running benchmarks and these are my observations.

- The greatest impact seems to be `Repo.checkout` where there are a lot of queries.  On examples with a very large number of queries and concurrency reduced the time by **200%**
- There was a small but consistent reduction simply do to restructuring to build the ids first.
- Parallelization using `Task` seems to have had a negative effect.  I'm not confident on this because the difference was in the margin of error kind of range and inconsistent.   I'm guessing because the the db is so small and on the same docker container that running the queries is nearly no cost?

Putting in a new PR with the changes that have a clear benefit.

PR #9212, opened 2024-08-14
Update Elysia

zoto-ff:
* update `bun` and `elysia` versions
* make all tests faster
* refactor and fixes
* use `bun build --compile`


frameworks/TypeScript/elysia/src/postgres.ts
```diff
@@ -9,16 +9,16 @@ const sql = postgres({
   max: 1,
 });
```
zoto-ff, line 10: why `max: 1`? is it rule or some kind of performance hack?

NateBrady23: I can't speak to a performance hack, but the number of max connections is not a rule.

PR #9227, opened 2024-08-22
Add CodeBehind Framework

elanatframework:
**CodeBehind** is a modern back-end framework under ASP.NET Core. CodeBehind was developed by [Elanat](https://elanat.net) in 2023 and competes with Microsoft's default web frameworks (**ASP.NET Core MVC** and **Razor Pages** and **Blazor**). CodeBehind is an engineering masterpiece that simultaneously provides the possibility of development based on MVC, Model-View, Controller-View, only View and Web-Forms. The type of structure and naming in CodeBehind is a nostalgia that reminds of former Microsoft Web-Forms.

ASP.NET Core web frameworks list

This pull request is the source code of the FrameworkBenchmarks test, which is made with the CodeBehind framework and is ready to be compiled in .NET Core version 8.
It is necessary to explain that the CodeBehind framework creates the final View class on the first request and then compiles it. Before starting test, you need to make a request to the server once.
More information:
[How is the list of views finally made?](https://github.com/elanatframework/Code_behind/blob/elanat_framework/doc/how_is_the_list_of_views_finally_made.md)

Also, after completing the project and before performing the test, it is necessary to call the `Initialization` method in the `program.cs` class with the value true as shown below.
`SetCodeBehind.CodeBehindCompiler.Initialization(true);`

By setting the above code, the final View class is created and compiled only once, then the compiled assembly is also stored in the drive.; A new request after the application sleeps on the web server causes the assembly to be placed in memory from the drive.

> Note: If you make a request to the server once before testing and the test is done before the application goes to sleep on the web server, there is no need to change the Initialization method.

GitHub link:
https://github.com/elanatframework/Code_behind


elanatframework:
Can anyone from .NET experts solve this problem?

NateBrady23:
@elanatframework `Cannot locate specified Dockerfile: codebehind.dockerfile`

Your Dockerfile needs to be named `codebehind.dockerfile`

elanatframework:
In the CodeBehind framework, the controller is specified in the attributes section of the View page.
Requests first go to the View, and the View creates a new instance of the Controller class.

Views are collected in a class and then compiled.
[More information](https://github.com/elanatframework/Code_behind/blob/elanat_framework/doc/how_is_the_list_of_views_finally_made.md)

Does the `wwwroot` directory also get added?
Is there access to execute for compilation in your test?

PR #9232, opened 2024-08-24
Disable bulk update

oxTriggerMan:
This PR is to address the issue that some frameworks do not modify database records one by one according to the Data Update requirements, but instead combine these modified records into a single SQL to achieve batch updates, which does not meet the requirements of the Data Update stress test.

Below are the requirements for Data Update:

```shell
Requirements summary

This test exercises database writes. Each request is processed by fetching multiple rows from a simple database table, converting 
the rows to in-memory objects, modifying one attribute of each object in memory, updating each associated row in the database 
individually, and then serializing the list of objects as a JSON response. The test is run multiple times: testing 1, 5, 10, 15, and 20 
updates per request. Note that the number of statements per request is twice the number of updates since each update is paired 
with one query to fetch the object. All tests are run at 512 concurrency.
```

see https://github.com/TechEmpower/FrameworkBenchmarks/issues/2204

### DIFF
--- a/frameworks/PHP/imi/ApiServer/Controller/PgController.php
+++ b/frameworks/PHP/imi/ApiServer/Controller/PgController.php
@@ -179,22 +179,18 @@ class PgController extends HttpController
         }
         $db = Db::getInstance(self::POOL_NAME);
         $stmtSelect = $db->prepare('SELECT id, randomnumber FROM World WHERE id = ? LIMIT 1');
-        $stmtUpdate = $db->prepare('UPDATE World SET randomNumber = CASE id' . \str_repeat(' WHEN ?::INTEGER THEN ?::INTEGER ', $queryCount) . 'END WHERE id IN (' . \str_repeat('?::INTEGER,', $queryCount - 1) . '?::INTEGER)');
+        $stmtUpdate = $db->prepare('UPDATE World SET randomNumber=? WHERE id=?');
         $list = [];
-        $keys = $values = [];
         while ($queryCount--)
         {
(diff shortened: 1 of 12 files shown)


joanhey:
Sorry, but you are only changing the PHP frameworks.
And they only do the same that do other languages, if we change the PHP frameworks, we need to change also all the other languages implementations.

Java:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/vertx/src/main/java/vertx/App.java#L191-L205

Rust:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Rust/actix/src/db.rs#L64-L81

and the same in almost all languages and frameworks.

Another observation is that a lot of ORMs also do it so.

What say the requirements is:
https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#database-updates

> Each row must be selected randomly using one query in the same fashion as the single database query test (Test #2 above). This includes that it is not acceptable to read just the randomNumber field. As with the read-only multiple-query test type (#3 above), use of IN clauses or similar means to consolidate multiple queries into one operation is not permitted. Similarly, use of a batch or multiple SELECTs within a single statement are not permitted.

About the update statements:

That is exactly what you are changing, the bulk updates !!

Please close this PR.

joanhey (changes requested):
This change is not accepted.
As the requirements permit bulk updates.

oxTriggerMan:
Oh, I'm sorry, I misunderstood the meaning of this paragraph. I close this pr now.

PR #9238, opened 2024-08-27
feat: log resource usage statistics from docker stats

akupiec:
I have noticed that the stats.txt file contains extensive hardware logs for each tested framework, but for some reason, the reported memory usage is way off. For example, it shows ~2GB instead of ~5MB.

After some investigation, I discovered that there is a thini called `dool` that collects stats from the host machine. Unfortunately, it is not possible to distinguish which part of the resources are used by the system itself, benchmark runner, or testing container.

Armed with the knowledge that there are tools capable of collecting resource usage specifically for a Docker containers, I found that Docker provides a lot of useful and compact informations through [docker stats](https://docs.docker.com/reference/cli/docker/container/stats/).

### TL;DR:
**Problem:**
`dool` logs info relevant for host toolset **and** tested container without clear separation.

**Solution:**
Use resource logs from [docker stats](https://docs.docker.com/reference/cli/docker/container/stats/).

PS. 
I'm not a python developer so feel free to roast me 🔥

### DIFF
--- a/toolset/benchmark/benchmarker.py
+++ b/toolset/benchmark/benchmarker.py
@@ -1,3 +1,6 @@
+import threading
+
+from docker.models.containers import Container
 from toolset.utils.output_helper import log, FNULL
 from toolset.utils.docker_helper import DockerHelper
 from toolset.utils.time_logger import TimeLogger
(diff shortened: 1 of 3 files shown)


NateBrady23:
Let's give it a shot, thanks for this!

PR #9258, opened 2024-09-07
[php] Enable jit for workerman and webman

walkor:
Enable jit for workerman and webman and remove unnecessary files.


frameworks/PHP/workerman/benchmark_config.json
```diff
@@ -41,35 +39,12 @@
...
-      "os": "Linux",
-      "database_os": "Linux",
-      "display_name": "workerman-async-db",
-      "notes": "",
-      "versus": "php",
-      "tags": [
-        "broken"
-      ]
-    },
-    "php8-jit": {
```
joanhey, line 70: Please, maintain normal and JIT versions.
That's for test PHP and Workerman it's a really good candidate. And PHP 8.4 come with a complete new JIT.

But the worst thing is that it'll disappear from all historic visualizations.

joanhey (changes requested):

frameworks/PHP/workerman/workerman-pgsql.dockerfile
```diff
@@ -11,17 +13,13 @@ RUN apt-get install -yqq php8.3-cli php8.3-pgsql php8.3-xml > /dev/null
 COPY --from=composer/composer:latest-bin --link /composer /usr/local/bin/composer
 
 RUN apt-get install -y php-pear php8.3-dev libevent-dev git > /dev/null && \
-    pecl install event-3.1.3 > /dev/null && echo "extension=event.so" > /etc/php/8.3/cli/conf.d/event.ini
-
-COPY --link php.ini /etc/php/8.3/cli/php.ini
+    pecl install event-3.1.3 > /dev/null && echo "extension=event.so" > /etc/php/8.3/cli/conf.d/30-event.ini
```
joanhey, line 16: The last event extension is 3.1.4

walkor: Perhaps we can directly use the command `pecl install event`, which will install the latest version of the event.

frameworks/PHP/webman/benchmark_config.json
```diff
@@ -2,8 +2,27 @@
   "framework": "webman",
   "tests": [{
```
joanhey, line 3: New feature:
If you add your github user to the maintainers line, you will be pinged with any new change.

```
{
  "framework": "webman",
  "maintainers": ["walkor"],
  "tests": [{
  ...

```

You cand do the same with Workerman.

frameworks/PHP/workerman/workerman-pgsql.dockerfile
joanhey: Before I used event without version, but sometimes had problems.

joanhey: But you can use it if you want.

joanhey: Another advantage, it's that so we can check the performance of the event extension when change.
And so we know if a performance change in Workerman or in event.

walkor: Yes, you are right.

frameworks/PHP/webman/benchmark_config.json
```diff
@@ -20,7 +39,7 @@
       "webserver": "None",
       "os": "Linux",
       "database_os": "Linux",
-      "display_name": "webman",
+      "display_name": "webman-pgsql",
```
joanhey, line 43: We are also changing the display_name, for easier visibility.

From "webman-pgsql" to "webman [pgsql]"

Or if we have more options like in Workerman :  "workerman [JIT, pgsql]"

frameworks/PHP/workerman/benchmark_config.json
```diff
@@ -62,7 +104,7 @@
       "webserver": "None",
       "os": "Linux",
       "database_os": "Linux",
-      "display_name": "workerman-postgres",
+      "display_name": "workerman[pgsql]",
```
joanhey, line 107: Please use an space or we loss the visibility gain.

workerman[pgsql] vs
workerman [pgsql]

frameworks/PHP/workerman/benchmark_config.json
```diff
@@ -41,35 +62,32 @@
...
       "database_os": "Linux",
-      "display_name": "workerman-postgres",
+      "display_name": "workerman [jit, pgsql]",
       "notes": "",
       "versus": "php"
     },
-    "async": {
-      "db_url": "/db",
-      "fortune_url": "/fortunes",
+    "without-jit": {
```
joanhey, line 69: Please don't change the name.
Use the same names, or all the history will be lost.

joanhey (changes requested):

frameworks/PHP/workerman/benchmark_config.json
joanhey: Now I'm busy.
Give sometime and I'll check it.

walkor: Ok thanks.

frameworks/PHP/workerman/benchmark_config.json
walkor: @joanhey 
The without-jit testing has been added. All changes have been locally verified and are working fine.

PR #9263, opened 2024-09-09
[Phoenix] Use transaction for multiple updates, multiple queries

srcrip:
# Changes

- Uses one connection from the db pool (by opening a transaction) for multiple updates and multiple queries bench for phoenix. This seems to speed up these two a lot.

### DIFF
--- a/frameworks/Elixir/phoenix/lib/hello_web/controllers/page_controller.ex
+++ b/frameworks/Elixir/phoenix/lib/hello_web/controllers/page_controller.ex
@@ -1,5 +1,4 @@
 defmodule HelloWeb.PageController do
-
   use HelloWeb, :controller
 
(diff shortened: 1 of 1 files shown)


Schultzer (approved):

srcrip:
@NateBrady23 this one should speed up the Phoenix benchmark dramatically, it would be very nice to get this in before the next round. Can you tell me when that is btw?

NateBrady23:
You can follow https://tfb-status.techempower.com to get a general idea.

srcrip:
@NateBrady23 do you know if bandit is now the default on the latest runs for Phoenix? The test results don't seem consistent based on what I got in my testing. I'm wondering if somehow it's using cowboy by default now? there used to be two separate categories and they seem to be gone now.

PR #5317, opened 2019-12-17
Add a new framework - gnet

panjf2000:

PR #9273, opened 2024-09-13
gnet: fix a bug of Content-Length header when using wildcat

andy2pan:
I've just discovered that `gnet` was taken off the leaderboard in #9159. I feel the need to clarify a few things.

> gnet is definitely not realistic.
>
> - Its entire "HTTP parser" is a 1-line seek for "\r\n\r\n" and doesn't even handle a TCP segment being... segmented. It does no header parsing whatsoever.
> - Its response is fully pre-formatted except for the Date.

***For 1***, `gnet` used a handmade HTTP parser at first (#5317) and then switched to another legit one - [wildcat](https://github.com/evanphx/wildcat) to parse HTTP requests (#7160) before I found that `gnet` had been failing the benchmark test of techempower.

However, that failure didn't reproduce on my computer and the code worked flawlessly when I ran the benchmark locally, which made it extremely hard for me to diagnose the issue because I wasn't able to debug the code on techempower's benchmark machines. Back then, I tried to fix that issue for a long time and created a few PRs intending to resolve it, see #7177 and #7214, but they didn't work. 

Finally, I had to give up because I had bigger stuff on my plate at the time, so I decided to get rid of `wildcat` for the moment and get back to fixing it when I got some time. Instead of reverting to the handmade HTTP parser, I took another action: drop `wildcat` and use a simple approach of seeking "\r\n\r\n" as a temporary solution in #7235. ***To be fair, I didn't intend to use this trick to get some performance boost because I ran the benchmarks (`wrk` with HTTP pipeline, same as techempower's benchmark) locally with both the simple parser and the standard parser before I created #7235 and there is almost no difference between the two results***:

```bash
# Standard parser

[gnet] 2024-09-13T16:23:21.58014+08:00	INFO	logging/logger.go:256	Launching gnet with 10 event-loops, listening on: tcp://:8085
2024/09/13 16:23:21 echo server with multi-core=true is listening on tcp://:8085
Running 15s test @ http://127.0.0.1:8085/plaintext
  4 threads and 256 connections
    Latency     1.49ms    4.36ms 110.24ms   96.23%
    Req/Sec   660.28k    91.25k    0.87M    76.00%
  Latency Distribution
     50%  586.00us
     75%    1.24ms
     90%    2.86ms
     99%   15.44ms
  39515552 requests in 15.05s, 4.75GB read
  Socket errors: connect 7, read 115, write 0, timeout 0
Requests/sec: 2625812.44
Transfer/sec:    323.04MB

# Simple parser

 [gnet] 2024-09-13T16:25:39.754574+08:00	INFO	logging/logger.go:256	Launching gnet with 10 event-loops, listening on: tcp://127.0.0.1:8085
2024/09/13 16:25:39 echo server with multi-core=true is listening on tcp://127.0.0.1:8085
Running 15s test @ http://127.0.0.1:8085/plaintext
  4 threads and 256 connections
    Latency     1.58ms    2.92ms  66.21ms   92.04%
    Req/Sec   660.23k    96.44k    0.92M    68.11%
  Latency Distribution
     50%  634.00us
     75%    1.54ms
     90%    3.72ms
     99%   13.95ms
  39560512 requests in 15.09s, 4.72GB read
  Socket errors: connect 7, read 117, write 0, timeout 0
Requests/sec: 2621492.58
Transfer/sec:    320.01MB
```

Therefore, the main bottleneck of performance in the techempower's benchmark of plaintext is ***not*** the HTTP parsing due to the simplicity of requests. I only used that simple parser for 1) it has way less code compared to the previous handmade parser that is kind of messy, 2) it doesn't affect the performance result. Moreover, I was under the impression that I'd be fixing that problem soon and switched back to `wildcat`, but I was overwhelmed by other stuff and have since forgotten about that..., I'm sorry for that.

***For 2***, I don't understand neither the word "pre-formatted" or what he was complaining about. To me, it's a normal action and I don't believe that it would affect plaintext performance at all, just like the HTTP parsing.

Today, a GitHub issue pointed out a bug in my HTTP example of `gnet`, which made me realize this could be the root cause that led the test code of `gnet` of techempower to fail the benchmark test constantly, check out https://github.com/gnet-io/gnet-examples/commit/3f7a3df16cf0e142074113e03c9985bd53a196e7. That's when I found out that `gnet` had been taken off the list. To be frank, I'm fine with taking off `gnet` for its simple HTTP parser cuz it indeed can hardly be identified as a legit HTTP parser (even though I've shown that the parser doesn't seem to have a noticeable impact on the performance of plaintext), but since the bug of using `wildcat` in `gnet` may have been fixed and `wildcat` is a legit HTTP parser, I'd like to update `gnet` to use `wildcat` and propose to add it back to the ranking list.


andy2pan:
@NateBrady23 Please take a review at this when you have some time.

PR #9287, opened 2024-09-21
[xitca-web] add bench for async orm

fakeshadow:
add `xitca-postgres-diesel` as async ORM where it integrates into `diesel` and `diesel-async` ecosystem.

### DIFF
--- a/frameworks/Rust/xitca-web/benchmark_config.json
+++ b/frameworks/Rust/xitca-web/benchmark_config.json
@@ -86,6 +86,28 @@
         "notes": "",
         "versus": ""
       },
+      "orm": {
+        "json_url": "/json",
+        "plaintext_url": "/plaintext",
+        "db_url": "/db",
+        "fortune_url": "/fortunes",
+        "query_url": "/queries?q=",
+        "update_url": "/updates?q=",
+        "port": 8080,
+        "approach": "realistic",
+        "classification": "fullstack",
+        "database": "postgres",
+        "framework": "xitca-web [orm]",
+        "language": "rust",
+        "orm": "full",
+        "platform": "none",
+        "webserver": "xitca-server",
+        "os": "linux",
+        "database_os": "linux",
+        "display_name": "xitca-web [orm]",
+        "notes": "",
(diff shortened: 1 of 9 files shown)


frameworks/Rust/xitca-web/benchmark_config.json
```diff
@@ -86,6 +86,28 @@
...
+        "plaintext_url": "/plaintext",
+        "db_url": "/db",
+        "fortune_url": "/fortunes",
+        "query_url": "/queries?q=",
+        "update_url": "/updates?q=",
+        "port": 8080,
+        "approach": "realistic",
+        "classification": "fullstack",
+        "database": "postgres",
+        "framework": "xitca-web [orm]",
```
joanhey, line 100: Please use the same name for the framework.
All use the same framework.

image

fakeshadow: Using different names is for the purpose of being honest with accurate representing. Currently there are two issues with the composite score:
1. It counts in "stripped" variant regardless your filter setting. In `xitca-web`'s case it means the `[unrealistic]` score will be shown as realistic.
2. There are too many frameworks implementing partial categories to boost their composite score. Strictly speaking this practice is not an issue as long as the partial bench code share the same abstraction level and/or framework's public API sets. But in reality I see lots of them just throw everything their frameworks have to offer out of the window and drop down to lower level.

That said I don't mind making requested change. Just want to speak out the motivation of making multiple framework names: To make honest and transparent representation of one framework.

joanhey: Surprise !! But this seems an issue in the visualize results of the benchmark.
But you can create only a separate `[unrealistic]`

@NateBrady23

fakeshadow: I have fixed the config and all variants should be collapsing into a single name now. Had to remove one variant called `axum [xitca]` in the process due to naming conflict as related framework does not maintain by me. 

Please let me know if further changes are needed.

PR #9297, opened 2024-09-27
[new] Perl/Feersum benchmark

vividsnow:
Perl [Feersum](https://metacpan.org/dist/Feersum) native interface benchmark


vividsnow:
Hi

I'm confused regarding the rule for updating multiple rows in one query. According to the [wiki](https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#database-updates):

At the same time, if I understand correctly, there is a test on the number of executed update queries, which checks for ~ 1 update per statement: https://github.com/TechEmpower/FrameworkBenchmarks/actions/runs/11078743989/job/30786616269?pr=9297#step:9:1149.

So, is it ok to update multiple rows in one statement?

msmith-techempower:
Correct, but only for the Updates test specifically.

Is this ready to merge or does this answer change your understanding to the point where you want to update?

vividsnow:
Hi, @msmith-techempower

Does this form of prepared statement count as a bulk update?
```sql
with t(v,i) as (values (?,?), (?,?), ...)
update World w
set randomNumber = t.v
from t
where t.i = w.id
```

msmith-techempower:
Yes, that query does write multiple updates in a single statement (a "bulk update"), and it does not do a single "bulk read" that returns rows; therefore, I believe it should be allowed.

vividsnow:
As far as I understand the verification process steps: [1](https://github.com/TechEmpower/FrameworkBenchmarks/blob/ddd09520c10926c0e2c73a005b1f79f5a68142a8/toolset/test_types/verifications.py#L425-L442) and [2](https://github.com/TechEmpower/FrameworkBenchmarks/blob/ddd09520c10926c0e2c73a005b1f79f5a68142a8/toolset/test_types/verifications.py#L468-L472), it will not tolerate executed queries which are significantly less than the expected value:

https://github.com/TechEmpower/FrameworkBenchmarks/blob/ddd09520c10926c0e2c73a005b1f79f5a68142a8/toolset/test_types/verifications.py#L348

Therefore, until the verification process is changed, I consider this pull request ready to merge.

PR #9300, opened 2024-09-29
[Laravel/ripple] Update some configurations to be compatible with the v0.6

cclilshy:
Update some configurations to be compatible with the v0.6 core, this version addresses some of the exceptions that occurred earlier.

### DIFF
--- a/frameworks/PHP/laravel/laravel-ripple.dockerfile
+++ b/frameworks/PHP/laravel/laravel-ripple.dockerfile
@@ -33,16 +33,16 @@ RUN mkdir -p bootstrap/cache \
             storage/framework/views \
             storage/framework/cache
 
-# Configure
 RUN echo "PRP_HTTP_LISTEN=http://0.0.0.0:8080" >> .env
-RUN echo "PRP_HTTP_COUNT=64" >> .env
+RUN echo "PRP_HTTP_WORKERS=64" >> .env
+RUN echo "PRP_HTTP_RELOAD=0" >> .env
+RUN echo "PRP_HTTP_SANDBOX=1" >> .env
 
+# Configure
 RUN composer install --quiet
+RUN composer require cclilshy/p-ripple-drive --quiet
 RUN php artisan optimize
 
-RUN composer require cclilshy/p-ripple-drive --quiet
-RUN composer update --quiet
-
+# Start
 EXPOSE 8080
-
-ENTRYPOINT ["php","artisan","p:server","start"]
+ENTRYPOINT ["php","artisan","ripple:server","start"]
--- a/frameworks/PHP/laravel/benchmark_config.json
+++ b/frameworks/PHP/laravel/benchmark_config.json
@@ -139,7 +139,7 @@
 			"notes": "",
 			"versus": "php"
 		},
-		"pripple": {
+		"ripple": {
 			"json_url": "/json",
 			"db_url": "/db",
 			"query_url": "/queries/",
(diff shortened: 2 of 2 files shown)


cclilshy:
In one test, I got the following results. The difference was huge, but no unexpected exceptions occurred, just extra parameters

`-s pipeline.lua -- 16`

According to the conclusion: Currently Ripple does not support the same long connection to submit another request before the end of the response

If you focus only on this feature, is there a clear correlation between the following large differences in test results? Thank you for your guidance.

```bash
wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 512 --timeout 8 -t 16 http://tfb-server:8080/plaintext
Running 15s test @ http://tfb-server:8080/plaintext
  16 threads and 512 connections
    Latency    76.46ms   23.91ms 824.06ms   78.31%
    Req/Sec   409.19     73.58   633.00     78.35%
  Latency Distribution
     50%   75.06ms
     75%   87.18ms
     90%  100.04ms
     99%  132.38ms
  97821 requests in 15.09s, 19.03MB read
Requests/sec:   6482.30
Transfer/sec:      1.26MB
 Concurrency: 256 for plaintext
 wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 256 --timeout 8 -t 16 http://tfb-server:8080/plaintext -s pipeline.lua -- 16
Running 15s test @ http://tfb-server:8080/plaintext
  16 threads and 256 connections
    Latency     0.00us    0.00us   0.00us    -nan%
    Req/Sec   157.38      3.05   160.00     87.50%
  Latency Distribution
     50%    0.00us
     75%    0.00us
     90%    0.00us
     99%    0.00us
  256 requests in 15.06s, 51.00KB read
Requests/sec:     17.00
Transfer/sec:      3.39KB
STARTTIME 1727616870
ENDTIME 1727616885
```

joanhey:
In the last test you don't only added `-s pipeline.lua -- 16`, also use 256 concurrency `-c 256`.
Half the concurrency from the firs test `- c 512 `.

So it's normal to have less req/s in the last test, less concurrency = less req/s.

cclilshy:
Thank you for your concern, the test is consistent with my inference, but the focus in the short term is not to solve the pipeline.lua support

Do you have any other suggestions for this commit?

frameworks/PHP/laravel/laravel-ripple.dockerfile
```diff
@@ -34,15 +34,14 @@ RUN mkdir -p bootstrap/cache \
             storage/framework/cache
 
 # Configure
-RUN echo "PRP_HTTP_LISTEN=http://0.0.0.0:8080" >> .env
-RUN echo "PRP_HTTP_COUNT=64" >> .env
-
 RUN composer install --quiet
-RUN php artisan optimize
```
joanhey, line 41: Add again this line, before the # Start

frameworks/PHP/laravel/laravel-ripple.dockerfile
```diff
@@ -34,15 +34,14 @@ RUN mkdir -p bootstrap/cache \
...
-
 RUN composer require cclilshy/p-ripple-drive --quiet
```
joanhey, line 38: You can run the require and update in 1 line.
Check : 
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/PHP/laravel/laravel-workerman.dockerfile#L26-L27

cclilshy:
I did that, currently it follows a uniform code. maybe the `ripple:publish` of config will need to be submitted next time.

cclilshy:
No, lose the '.env 'configuration and its listening address will be 127.0.0.1 instead of 0.0.0.0, so I'll do better next time, excuse me, it urgently needs to use the latest way

PR #9313, opened 2024-10-07
[spring] Improve database tests

sdeleuze:
Hi, I am Sébastien, Spring Framework committer, here is a PR for `spring` and `spring-webflux` that improves the scalability and reliability of database access under high concurrency.


sdeleuze:
@NateBrady23 We managed to fix with this PR an issue that prevented the `/update` test to run correctly with `spring` in the last CI run (+ other `spring-webflux` refinements) and the performance gains of removing the postgres deadlock are massive, so please consider merging it before running the 2024 official run.

Issue #8061, opened 2023-03-23
Should we drop md5 auth in PostgreSQL ?  We are in 2023 !!!


joanhey:
With the new PostgreSQL official docker image, we have again the problem with md5 auth.
I can fix it, but **is it realistic to use md5 auth in 2023** ?

Some frameworks that fail, they had the time and energy to patch libpq for performance. So they can use some time to add the new scram (sha-256) auth in their frameworks.

More info: #7557 #8002

Give your feedback.

nbrady-techempower:
@joanhey How many frameworks are failing because of this? Do you know offhand and can ping the contributors here?

fakeshadow:
It should be default to sasl auth at this point. But there are some points I consider worth to mention: 

1. It is still realistic to use md5. Do note that sasl auth can still happen on plain TCP connection which make it not any more safer realistically speaking. I believe most people use plain TCP in trusted local network anyway.
2. Some if not most frameworks fail to do sasl auth are not using libpq. They are likely to use their language native db client which cause the problem(lack of feature) and it's considerable more work to write a native db driver than fork and patch libpq.

fakeshadow:
@fafhrd91 Maybe it worth to implement sasl auth in ntex postgres.

joanhey:
I think that fail for the auth ~30 frameworks permutations.

fakeshadow:
md5 auth is already deprecated from postgres and will be removed.
It's fair to switch to sasl as default and people had years of time to catch up.

Issue #7557, opened 2022-08-19
High Number of frameworks having problems starting


billywhizz:
in the last run and the current one just-js and a number of other frameworks seem to have failed to start. it has never failed before and nothing has changed since round 21.

it seems the number of frameworks failing in same way has increased significantly in last round and the one currently running. is there some environmental issue causing problems maybe?

https://tfb-status.techempower.com/results/15bfed85-1093-485d-b161-5f61b7abd0be

nbrady-techempower:
A few spot checks have me believing it's the postgres upgrade. @joanhey does that seem right to you?

joanhey:
After some research, it seems correct that it affects these fails.
I have had to look at other fws, because Just doesn't give any errors.
The others return `... authentication type: 10`

It's again the problem with the auth encryption. That come from postgresql v10.

New in Postgresql 14:

> Change the default of the [password_encryption](https://www.postgresql.org/docs/14/runtime-config-connection.html#GUC-PASSWORD-ENCRYPTION) server parameter to scram-sha-256 (Peter Eisentraut)
> 
> Previously it was md5. All new passwords will be stored as SHA256 unless this server setting is changed or the password is specified in MD5 format. Also, the legacy (and undocumented) Boolean-like values which were previously synonyms for md5 are no longer accepted.

Now I'll send a PR.

billywhizz:
@joanhey thanks for investigating - so this means all frameworks must support/use scram-sha-256 authentication in their PG driver?

joanhey:
`Scram-sha-256` was introduced in Postgesql v10. (October 5, 2017)
And it's the recommended method for a most secure connection. And any new PG driver need to support it.

I changed the default to md5 again in `postgresql.conf` in the PR.
And Just-js is working again, and the rest.

Historically this bench use md5, because at the time of this change, most pg drivers didn't suppor scram-sha-256.
Something similar happens with Mysql 8.

billywhizz:
thanks @joanhey - i will look at adding scram-sha-256 support to the postgres driver for just-js.

joanhey:
A new think in Postgresql 14 is **pipeline** mode in `libpg`, and also work with older servers versions.

Some fws in this bench use hand made versions.

https://www.postgresql.org/docs/current/libpq-pipeline-mode.html
https://www.percona.com/blog/how-postgresql-pipeline-mode-works/

billywhizz:
[...]

PR #9348, opened 2024-10-22
Upgrade aiohttp

Dreamsorcerer:


Dreamsorcerer:
Database credentials don't work anymore? I haven't touch those in the PR..

Dreamsorcerer:
@NateBrady23 or anyone know what to change for the DB credentials? Would be good to get this upgrade in as there's been a lot of performance optimisations in recent releases (though they may be a magnitude more noticeable in real world scenarios than these simple benchmarks).

volyrique:
Could it be an issue with the authentication method (issue #8061 might be related)?

Dreamsorcerer:
> Could it be an issue with the authentication method (issue #8061 might be related)?

I'm unclear. We just create a DSN with user/password, so I don't see anything we'd need to change for that:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/1beae77c8272ec937ce3ad1a6959582198925552/frameworks/Python/aiohttp/app/main.py#L32

Appears that asyncpg implemented the auth method 6 years ago, so I doubt it's an issue there either..

volyrique:
Yes, but the point is that the password is not necessarily sent in cleartext to the server - see [here](https://www.postgresql.org/docs/17/auth-password.html). `asyncpg` might not be using the correct method (the defaults have changed much more recently than 6 years ago, as the issue I linked to demonstrates, and some methods might have been removed - other frameworks have been broken by database updates), hence failing with an authentication error. The error message doesn't seem to go into that kind of detail, so I would probably run a packet capture to see what is happening on the wire, unless `asyncpg` provides that level of detail somehow (I am not familiar with it).

Just an idea, since you were asking for suggestions.

msmith-techempower:
The issues here don't SEEM to be related to the md5 issue... closing for now, but feel free to open a new PR when you get the issues resolved.

msmith-techempower:
It doesn't run successfully locally or in CI with your changes https://github.com/TechEmpower/FrameworkBenchmarks/actions/runs/13909801752/job/38921326418?pr=9348

Currently, on main aiohttp verifies successfully. 

This may be related to #7557 since the only changes that would result in the password authentication failing given your changes appear to be the db driver versions being updated. Maybe these new versions disabled md5 auth?

Dreamsorcerer:
Ahhh, I didn't realise it was still working on master. Apparently sqlalchemy changed rendering of URL objects.

This should be good to merge now.

msmith-techempower (approved):

PR #9380, opened 2024-11-03
[crystal/lucky] Fix tests and update dependencies

p8:
Reduce connection pool size to avoid connection errors like:

     psycopg2.OperationalError: connection to server at "10.0.0.2", port
     5432 failed: FATAL:  sorry, too many clients already

### DIFF
--- a/frameworks/Crystal/lucky/lucky.dockerfile
+++ b/frameworks/Crystal/lucky/lucky.dockerfile
@@ -1,4 +1,4 @@
-FROM crystallang/crystal:1.12.1
+FROM crystallang/crystal:1.14.0
 
 WORKDIR /lucky
 COPY shard.lock shard.lock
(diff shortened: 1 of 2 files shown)


frameworks/Crystal/lucky/lucky.dockerfile
```diff
@@ -14,7 +14,7 @@ ENV LUCKY_ENV production
 
 RUN shards build bench --release --no-debug
 
-ENV DATABASE_URL postgres://benchmarkdbuser:benchmarkdbpass@tfb-database:5432/hello_world?initial_pool_size=56&max_idle_pool_size=56
+ENV DATABASE_URL postgres://benchmarkdbuser:benchmarkdbpass@tfb-database:5432/hello_world?initial_pool_size=10&max_idle_pool_size=10
```
joanhey, line 17: Sorry, only an small question !!
All are playing with different pool sizes for the databases.

But we are learning (all languages) witch one is better ???

joanhey: That's only a question for reflexion !!
So we can learn all together.

Nothing about this PR.

p8: No problem @joanhey  😄 
I think the pool size is already per CPU core.
This was causing the runs to fail as it was making too many connections.

PR #9381, opened 2024-11-04
Revert "[php] Fix Leaf v3.6 dockerfiles"

Mmagkooo:
Reverts TechEmpower/FrameworkBenchmarks#9250


joanhey:
Sorry, Why do you want to revert this commit ??

joanhey (changes requested):
Please explain why do you want to revert this commit??

Actually it's working OK.

Issue #9408, opened 2024-11-21
PHP 8.4 update [info]


joanhey:
## PHP 8.4

We start to test v8.4 , released 21 Nov 2024.
https://www.php.net/releases/8.4/en.php

PHP update versions info:

- https://github.com/TechEmpower/FrameworkBenchmarks/issues/8579
- https://github.com/TechEmpower/FrameworkBenchmarks/issues/7703
- https://github.com/TechEmpower/FrameworkBenchmarks/issues/6894
- https://github.com/TechEmpower/FrameworkBenchmarks/issues/6184

## Fail
- [ ] Amp work but with deprecations. We need to update it to v3 in another PR
- [x] CodeIgniter4 Waiting for new version https://github.com/codeigniter4/CodeIgniter4/issues/9116
- [ ] Comet Work but show deprecated warnings. https://github.com/gotzmann/comet/issues/41
- [x] Fat-free  ~waiting for new version fixed~ https://github.com/f3-factory/fatfree-core/commit/a8adbd8fe0439ae652ff5faf0469782bc8dc4d84 and more https://github.com/f3-factory/fatfree-core/commits/master/
- [ ] Fuel
- [x] Laravel-s work with Laravel and Lumen but with a lot of deprecation warnnings 
- [x] Leaf work but show deprecated warnings https://github.com/leafsphp/leaf/pull/271
- [x] Ngx-php ~not ready~ https://github.com/rryqszq4/ngx-php/pull/197
- [ ] Openswoole ~Not ready~ work with v25.2.0 but the postgres variant fail (so this variant still use php/8.2)
- [x] Phalcon work with v5.9.0
- [x] Roadrunner is ready but fail. We need to test it more
- [x] Slim work ~but show deprecated warnings~ Fixed using php-di v7 https://github.com/slimphp/Slim/pull/3353
- [x] Unit server  is ready but fail 
- [x] Webman work ~but show deprecated warnings.~ https://github.com/walkor/webman/issues/423

We need to check that the JIT is ON in all frameworks !!
As in PHP8.4 changed and the Odrej ppa config is special.

NateBrady23:
Closing as part of bulk cleanup.

PR #9416, opened 2024-11-22
[php] Webman update to PHP 8.4

joanhey:
Work but show deprecated warnings

#9408

EDIT: fail query and update tests


walkor:
Ok, I will check that later

PR #9431, opened 2024-11-28
Update Scala SNUnit

lolgab:
Update framework to latest version and Scala runtime to latest version as well.


lolgab:
Thank you @NateBrady23

PR #9444, opened 2024-12-06
Ktor optimize and update

inemtsev:
- Updated Ktor version
- Updated /update method to use batch write inline with other top performing frameworks
- Cleanup


inemtsev:
Ready to Review/Merge

PR #9446, opened 2024-12-07
[rails] Upgrade to Rails 8.

p8:


skipkayhil:
👋 @p8 should this have bumped `load_defaults` in `config/application.rb`?

p8:
👋 @skipkayhil Yes, although I don't think any of the new defaults affect the results.

PR #9453, opened 2024-12-09
[toolset] Mongodb update to v8

joanhey:
Released 11 Sep 2024.
https://www.mongodb.com/resources/products/mongodb-version-history#mdbeightzero

Testing with CI


joanhey:
It's very difficult to check the results !!!
It'll be good than if we change a database version, only check the permutations that use this database.

Perl with mongo pass.

PR #9463, opened 2024-12-15
Update Ubuntu version for Ur/Web

achlipala:
Apparently the previous Ubuntu version is no longer accessible where the Docker configuration was looking for it, causing build failures.


joanhey (commented):
If you choose an LTS version (24.04), you don't need to change it every 9 months.

The v24.10 will end in july 2025 (6 months).

achlipala-biz:
Ah, great suggestion, @joanhey!  I had originally used a non-LTS version because it included a new package I wanted, but now that package has made its way to LTS.

PR #9467, opened 2024-12-18
Use go1.23 with fasthttp framework.

gaby:
- Update go from 1.19 to 1.23
- Update all dependencies and Dockerfiles.


gaby:
@NateBrady23 Why was this closed? The other frameworks are using 1.23, seems unfair.

PR #9499, opened 2025-01-07
[rails] Enable config.api_only

p8:
`config.api_only` disables a lot of middleware like Cookies and Session so we no longer have to delete them specifically.


p8:
@msmith-techempower I'd like to see this in round 23, if possible.

msmith-techempower (approved):

PR #9501, opened 2025-01-08
[crystal/spider-gazelle] Fix tests and update dependencies

p8:
Reduce connection pool size to avoid connection errors like:

     psycopg2.OperationalError: connection to server at "10.0.0.2", port
     5432 failed: FATAL:  sorry, too many clients already


p8:
@msmith-techempower I'd like to see this in round 23, if possible. Thanks 🙏

msmith-techempower (approved):

PR #9505, opened 2025-01-08
Relax HTML validation for Fortunes test

jonathanhefner:
Some frameworks, such as Next.js, automatically inject `<meta>`, `<link>`, and `<script>` tags into pages rendered by their templating system.  Since the Fortunes test is meant to exercise the templating system (as opposed to raw string concatenation), it should allow these tags.

Furthermore, React warns against nesting a `<tr>` directly inside a `<table>` ("\<tr> cannot be a child of \<table>") because browsers will automatically wrap the `<tr>` elements in a `<tbody>`, causing a mismatch with the virtual DOM.  Therefore, the Fortunes test should allow optional `<tbody>` (and `<thead>`) tags.

This commit relaxes the HTML validation for the Fortunes test to allow these tags by simply ignoring them when building the comparison string.

### DIFF
--- a/toolset/test_types/fortune/fortune_html_parser.py
+++ b/toolset/test_types/fortune/fortune_html_parser.py
@@ -8,8 +8,17 @@ from toolset.utils.output_helper import log
 
 
 class FortuneHTMLParser(HTMLParser):
+    IGNORED_TAGS = (
+        "<meta>", "</meta>",
+        "<link>", "</link>",
+        "<script>", "</script>",
+        "<thead>", "</thead>",
+        "<tbody>", "</tbody>",
+    )
+
     def __init__(self):
         HTMLParser.__init__(self, convert_charrefs=False)
+        self.ignore_content = False
         self.body = []
 
     valid_fortune = '''<!doctype html><html>
@@ -41,7 +50,7 @@ class FortuneHTMLParser(HTMLParser):
         # and since we did not specify xml compliance (where
         # incorrect casing would throw a syntax error), we must
         # allow all casings. We will lower for our normalization.
-        self.body.append("<!{d}>".format(d=decl.lower()))
+        self.append("<!{d}>".format(d=decl.lower()))
 
     def handle_charref(self, name):
         '''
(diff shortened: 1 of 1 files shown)


joanhey:
I understand you.
But who need to change?  the framework, than create extra magic, or the test ?

If any dev need to be so, why the framework add extra magic ?

jonathanhefner:
@joanhey As I hinted at in my commit message, it is possible to send exact HTML in these frameworks, but it boils down to raw string concatenation, and isn't a realistic way of writing an app.

If TechEmpower benchmarks are intended to reflect realistic usage of their frameworks, then this test should not demand exact output.  In fact, the test already [accommodates alternative HTML](https://github.com/TechEmpower/FrameworkBenchmarks/blob/ddd09520c10926c0e2c73a005b1f79f5a68142a8/toolset/test_types/fortune/fortune_html_parser.py#L48-L58) for character escaping (also [here](https://github.com/TechEmpower/FrameworkBenchmarks/blob/ddd09520c10926c0e2c73a005b1f79f5a68142a8/toolset/test_types/fortune/fortune_html_parser.py#L100-L101) and [here](https://github.com/TechEmpower/FrameworkBenchmarks/blob/ddd09520c10926c0e2c73a005b1f79f5a68142a8/toolset/test_types/fortune/fortune_html_parser.py#L127-L144)).

If a framework sends unnecessary bytes, then it will be penalized in terms of performance, but there is no reason to exclude those (realistic!) results from the benchmark.

msmith-techempower:
I think I regret the rigid way we built the verifier for Fortunes back in the day.

If I were to start over (and I did... there's a Rust verifier sitting collecting dust while I'm focused on other projects), I would make it only _enforce_ that the response include the required elements - matching `<tr>` and `</tr>` tags and each required-and-balanced `<td>` and `</td>` tags with their correct bodies and orders (since that is part of the requirements). We would make any additional bytes beyond what the verification rules enforce _currently_ be a warning with a link to the minimal example.

That would alleviate the strictness you, and others, feel when trying to make their framework pass.

I'm going to think on *this* particular relaxing of the Fortunes rules as a stopgap solution, but I doubt that it will be included for the forthcoming Round-to-be-published.

msmith-techempower:
Okay, talked it over (post round 23) and I think we're good with this as the stopgap solution. LGTM!

PR #9506, opened 2025-01-08
Benchmark Next.js

jonathanhefner:
_This PR is based on top of #9505.  I can rebase it if #9505 is merged first._

This adds benchmark tests for the Next.js framework.


msmith-techempower:
I wrote a response in that other PR about relaxing the Fortunes test - for now, I think the simplest solution would be to _not_ include the Fortunes test in this implementation (as it would fail for now - you can just comment it out with a note that links to the other PR) so as to get NextJS into this round.

jonathanhefner:
@msmith-techempower Alright, understood.  I [disabled the Fortunes test](https://github.com/jonathanhefner/FrameworkBenchmarks/blob/8c60545eaadb3a56fa51186c2089987358419fc4/frameworks/TypeScript/nextjs/benchmark_config.json#L23) and added a [corresponding note in the README](https://github.com/jonathanhefner/FrameworkBenchmarks/blob/8c60545eaadb3a56fa51186c2089987358419fc4/frameworks/TypeScript/nextjs/README.md?plain=1#L31-L47).

jonathanhefner:
@msmith-techempower The Docker image was failing to build because it copies an empty directory that wasn't tracked by git.  I've added a `.keep` file to the directory, so the build should work now.

PR #9509, opened 2025-01-08
Quote `randomNumber` column in PostgreSQL schema

jonathanhefner:
_I've opened this PR because it is a point of frustration when implementing tests for new frameworks, but I've marked it as draft because the change may be too breaking._

The [test specs][] say that the `World` table has an `id` column and `randomNumber` column (note the camelCasing).  Indeed, the [PostgreSQL schema][] uses `randomNumber` in its `CREATE TABLE` statement.  However, the `CREATE TABLE` statement does not quote the column name, so PostgreSQL treats the name as if it were lowercase.

Also note that the PostgreSQL schema has `CREATE TABLE World` _and_ `CREATE TABLE "World"` statements.  Presumably, this is because the former actually creates a `world` table because it too is not quoted. (Whereas the latter actually creates a `World` table.)

This commit modifies the `CREATE TABLE "World"` statement to quote the `randomNumber` column so that its camel case is preserved.

[test specs]: https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview
[PostgreSQL schema]: https://github.com/TechEmpower/FrameworkBenchmarks/blob/ddd09520c10926c0e2c73a005b1f79f5a68142a8/toolset/databases/postgres/create-postgres.sql

### DIFF
--- a/toolset/databases/postgres/create-postgres.sql
+++ b/toolset/databases/postgres/create-postgres.sql
@@ -34,12 +34,12 @@ INSERT INTO Fortune (id, message) VALUES (12, 'フレームワークのベンチ
 
 CREATE TABLE  "World" (
   id integer NOT NULL,
-  randomNumber integer NOT NULL default 0,
+  "randomNumber" integer NOT NULL default 0,
   PRIMARY KEY  (id)
 );
 GRANT ALL PRIVILEGES ON "World" to benchmarkdbuser;
 
-INSERT INTO "World" (id, randomnumber)
+INSERT INTO "World" (id, "randomNumber")
 SELECT x.id, least(floor(random() * 10000 + 1), 10000) FROM generate_series(1,10000) as x(id);
 
 CREATE TABLE "Fortune" (


jonathanhefner:
> So, the impact here is that anyone making a query with the non-quoted version will be in error?

Correct.

Before this PR:

  ```console
  > SELECT "randomNumber" FROM "World" LIMIT 1;
  ERROR:  column "randomNumber" does not exist
  LINE 1: SELECT "randomNumber" FROM "World" LIMIT 1;
                 ^
  HINT:  Perhaps you meant to reference the column "World.randomnumber".


  > SELECT randomNumber FROM "World" LIMIT 1;
   randomnumber
  --------------
           2338
  (1 row)
  ```

After this PR:

  ```console
  > SELECT "randomNumber" FROM "World" LIMIT 1;
   randomNumber
  --------------
           5197
  (1 row)


  > SELECT randomNumber FROM "World" LIMIT 1;
  ERROR:  column "randomnumber" does not exist
  LINE 1: SELECT randomNumber FROM "World" LIMIT 1;
                 ^
  HINT:  Perhaps you meant to reference the column "World.randomNumber".
  ```

msmith-techempower:
I'm doing kind of a "spring cleaning" of these legacy PRs. Here's my thinking on this one:

I think this has potential, but I don't want to start failing tests for switching over from requiring no quotes to suddenly requiring them - there are a lot of codebases and ORMs that would probably be impacted. A better solution, imho, is to support both. I thought of just adding both columns to your PR, but it occurred to me that perhaps there are implementations using `SELECT * ...` which would now have an additional column that might bleed into outputs.

So, I'm closing for now, but feel free to reopen if you want to take up the mantle on this issue.

PR #9510, opened 2025-01-08
[php] Lumen update to PHP/8.4

joanhey:
#9408


joanhey:
The last update for Lumen.
The next round will be eliminated.
image

msmith-techempower:
If it has already been archived, I think it makes sense to remove now; why wait for the round to run?

joanhey:
It'll be last round, and still Lumen is faster than Laravel.
So the devs need to know it !!

PR #9512, opened 2025-01-09
[rails] Reduce headers to required headers only

p8:
|           branch_name|fortune|
|----------------------|-------|
|                master|  14884|
|reduce-default-headers|  15730|


p8:
@msmith-techempower I'd like to see this in round 23, if possible. Thanks 🙏

msmith-techempower (approved):

PR #9520, opened 2025-01-11
H2O: Track connection load across the worker threads

volyrique:


joanhey:
@volyrique h2o is failing in the runs

`h2o: h2o_app: /tmp/h2o-build/lib/common/socket/evloop.c.h:817: h2o_socket_notify_write: Assertion 'sock->super._cb.write == NULL' failed.`

volyrique:
Thanks, I will have a look on the weekend.

PR #9521, opened 2025-01-12
[ruby/sinatra] Don't use `JSON.fast_generate`

p8:
`fast_generate` is not faster for small payloads.

|         branch_name|update|   db|query|fortune|weighted_score|
|--------------------|------|-----|-----|-------|--------------|
|              master|  8301|36433|15826|  28025|          1089|
|remove-fast-generate|  8897|38895|16317|  27660|          1143|


p8:
@msmith-techempower I'd like to see this in round 23, if possible. Thanks!

PR #9543, opened 2025-01-16
Add rage-sequel

rsamoilov:
The PR adds the rage-sequel test, which is a variation of the existing [rage](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Ruby/rage) test using the Sequel ORM. It follows the example of the existing [sinatra](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Ruby/sinatra) and [sinatra-sequel](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Ruby/sinatra-sequel) tests.

### DIFF
--- a/frameworks/Ruby/rage-sequel/Gemfile
+++ b/frameworks/Ruby/rage-sequel/Gemfile
@@ -1,0 +1,7 @@
+source "https://rubygems.org"
+
+gem "rage-rb", "~> 1.10"
+
+gem "pg", "~> 1.0"
+gem 'sequel', '~> 5.0'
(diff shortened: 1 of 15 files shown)


p8:
@rsamoilov Nice!
The non database `json` and `plaintext` calls are probably similar to the regular `rage` tests?
For `sinatra` we don't make those calls for `sequel` either to reduce the total test time:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Ruby/sinatra-sequel/benchmark_config.json

rsamoilov:
Thanks, fixed!

PR #9549, opened 2025-01-18
[FrankenPHP] Remove unnecessary file-exists checks.

Alliballibaba2:
Optimizes FrankenPHP/Symfony/Laravel by deactivating the `file server`/`try files` checks and forwarding directly to the PHP runtime (which is all we need in this case).

### DIFF
--- a/frameworks/PHP/symfony/deploy/Caddyfile
+++ b/frameworks/PHP/symfony/deploy/Caddyfile
@@ -11,18 +11,13 @@
 :8080
 
 route {
-    root * /symfony/public
 
-    # If the requested file does not exist, try index files
-    @indexFiles file {
-        try_files {path} {path}/runtime.php runtime.php
-        split_path .php
+	# FrankenPHP!
+	# disable static files for this benchmark
+	# by using php instead of php_server
+    rewrite runtime.php
+    php {
+        root /symfony/public
     }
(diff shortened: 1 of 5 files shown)


Alliballibaba2:
Issue explaining why this prevents unnecessary requests to the filesystem https://github.com/dunglas/frankenphp/pull/1311

frameworks/PHP/symfony/deploy/Caddyfile
```diff
@@ -11,18 +11,13 @@
...
-    # If the requested file does not exist, try index files
-    @indexFiles file {
-        try_files {path} {path}/runtime.php runtime.php
-        split_path .php
+	# FrankenPHP!
+	# disable static files for this benchmark
+	# by using php instead of php_server
+    rewrite runtime.php
+    php {
+        root /symfony/public
```
msmith-techempower, line 20: Did you mean to remove the `@phpFiles` stuff from here? You added it to the franken Caddyfile - seems weird to remove it from symfony.

Alliballibaba2: Yes, the previous configuration will check for files, directory index files and general php files, which is unnecessary in this case (there is only 1 entry file).

Rewriting directly to the entryfile will remove all 'file exists' checks..

Issue #9440, opened 2024-12-04
Check frameworks for HTTP compliance


Kaliumhexacyanoferrat:
Currently there are a lot of frameworks marked as `realistic` that do not provide a proper HTTP webserver implementation but just respond to the test tools in the expected manner. As they skip a lot of scenarios a real-world web server has to handle, they are of course faster.

After the discussion in #9391 I've created a project intended to check a HTTP web server for compliance with the HTTP specification. This is a very rough draft because this is a functionality not very much projects would need - so I am willing to spend more time only, if you are interested in integrating such tool into your test framework. The executable will take an HTTP endpoint and return a status code whether the server is compliant or not.

You can find the project here: https://github.com/Kaliumhexacyanoferrat/http-conformity

The information provided by the tool could be used in various ways:

- Add another filter to the results dashboard
- Do not allow frameworks that are not compliant to be marked as `realistic`
- Fail the build if a non-compliant framework is added
- Exclude non-compliant frameworks

Please let me know what you think :)

joanhey:
We created our own functional tests for Adapterman and Workerman.
https://github.com/joanhey/AdapterMan/actions/runs/11104804508

All the tests:
https://github.com/joanhey/AdapterMan/actions/runs/11104804508/job/30849613841#step:5:1

As only 1 test is specific to PHP, recently we have move it to it's own project to use with any server or app server.
And we are working in selecting which tests to run, it's easy  (WIP).
https://github.com/joanhey/TestMan

Actually we have 103 tests and 511 assertions.
![image](https://github.com/user-attachments/assets/30fd56a4-5970-4edb-bbf7-6b4cb9f49dd1)

But for run the tests we need a file in the server to return the values.
That's not possible to add in all the frameworks from this bench.

So with the code that we have from this bench, the number of tests that we can run it's very small and never will be realistic. Only will detect the more radical implementations.

Kaliumhexacyanoferrat:
Nice, thanks - yes, without the server fulfilling some contract we can test only a few assertions. Nevertheless detecting and disabling radical implementations on official rounds already would be a great improvement as they tend to be on top - so I would encourage such an usage of your suite with just a few of the tests.

joanhey:
We can start a list of the tests, that we can use in this bench:

- 404 Not Found
- 400 Bad Request
- POST method and return the same that with the GET
- ...

Kaliumhexacyanoferrat:
[...]

PR #9560, opened 2025-01-22
[ruby/rack] Mark Rack as Realistic instead of Stripped

p8:
The[ requirements](https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#general-test-requirements) state that:

    Implementations that are not based on a realistic HTTP implementation
    will be marked as Stripped.

The Rack tests use webservers (Puma, Unicorn, Falcon) based on realisitic HTTP implementation, and the [Rack specification](https://github.com/rack/rack/blob/main/SPEC.rdoc) does as well.

So if we are going to [mark some frameworks as non realistic](https://github.com/TechEmpower/FrameworkBenchmarks/issues/9440) because they don't conform to a realistic HTTP implementation, why should Rack be marked as Stripped when it does conform?


joanhey:
LGTM this change.
It was changed a lo of time ago, without any explication in:
https://github.com/TechEmpower/FrameworkBenchmarks/commit/e684b548bc3eb9bc60fc9ee6ede2af1861e57612#diff-33dc9963b3f86ae43192c37d5134b3560ec3f56259b23ce9bf2afdca68458f27R12

p8:
Also frameworks hardcoding the Content-Length should probably be marked as Stripped?
https://github.com/TechEmpower/FrameworkBenchmarks/blob/0c959df4a17c13d76abff47c4b048bfb4465d914/frameworks/Go/goev/src/main.go#L23
https://github.com/TechEmpower/FrameworkBenchmarks/blob/0c959df4a17c13d76abff47c4b048bfb4465d914/frameworks/Rust/actix/src/main_server.rs#L25
https://github.com/TechEmpower/FrameworkBenchmarks/blob/0c959df4a17c13d76abff47c4b048bfb4465d914/frameworks/Rust/ntex/src/main_plt.rs#L13

PR #9565, opened 2025-01-24
[clace] Added localhost path

akclace:
Added app at localhost path also. The previous test run for Clace failed https://tfb-status.techempower.com/unzip/results.2025-01-23-22-49-18-014.zip/results/20250117182557/clace possibly because the app was at tfb-server path only.


akclace:
Thanks

PR #9571, opened 2025-01-27
[ruby/roda-sequel] Only set headers not created by servers

p8:
Servers like Iodine set the `Date` header itself, so no need to duplicate that effort.

|                         branch_name|  json|   db|query|update|fortune|plaintext|weighted_score|
|------------------------------------|------|-----|-----|------|-------|---------|--------------|
|                              master|213942|52774|37141| 22971|  49660|   227545|          2923|
|remove-redundant-headers|280527|89348|61538| 24688|  57966|   335448|          3752|

### DIFF
--- a/frameworks/Ruby/roda-sequel/hello_world.rb
+++ b/frameworks/Ruby/roda-sequel/hello_world.rb
@@ -15,9 +15,19 @@ class HelloWorld < Roda
     rand(MAX_PK) + 1
   end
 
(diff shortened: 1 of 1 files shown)


p8:
@NateBrady23 Can we get some of these PR merged and running in a new run?

p8:
Thanks @NateBrady23 !

PR #9594, opened 2025-02-18
ihp: update to v1.3.1

mpscholten:
This updates the IHP framework from v.1.0.1 to v1.3.1

This fixes the test failure here https://tfb-status.techempower.com/results/91a66052-9d86-446c-b31a-eadbd669ed08


msmith-techempower:
This took 23 minutes to build and run verification for me locally... what in the world is going on here?

For context, this passed verification but I have a hard time wanting to continue implementations that take half an hour to run.

msmith-techempower:
Okay, ate lunch and talked this over with Nate - it's already in and other frameworks already take this much or longer to run. LGTM

mpscholten:
Thanks for your feedback 👍 you're right that it takes far to long. I'll look into how we can further optimize this to get the build time down. IMO it should be possible in 2 - 3 min instead.

PR #9598, opened 2025-02-21
[php] Update Flight and add Workerman variant

joanhey:

### DIFF
--- a/frameworks/PHP/flight/benchmark_config.json
+++ b/frameworks/PHP/flight/benchmark_config.json
@@ -23,6 +23,29 @@
       "display_name": "flight",
       "notes": "",
       "versus": "php"
-    } 
+    },
+    "workerman": {
+      "json_url": "/json",
+      "plaintext_url": "/plaintext",
+      "db_url": "/db",
+      "fortune_url": "/fortunes",
+      "query_url": "/db-multiple?queries=",
+      "update_url": "/updates?queries=",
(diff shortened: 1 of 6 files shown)


n0nag0n:
Thanks @joanhey

n0nag0n:
@joanhey Out of curiosity, what performance boost did you notice with Adapterman/workerman and Flight?

joanhey:
I don't bench locally.

The first performance boost will be with normal Flight, after the last changes.
With Adapterman I think that will be +200%. 
Even Flight now work like a persistent app, Flight need to optimize the code for better performance in that situations.

joanhey:
When finish the actual run, I'll update it.

Before:
| Framework | JSON |1-query |20-query | Fortunes | Updates | Plaintext |
| ---- | ----| ---| ---| ---| --- | --- |
|flight |	186,482 |118,685 | 33,523 | 106,888 | 4,473 | 193,698	


After:
| Framework | JSON |1-query |20-query | Fortunes | Updates | Plaintext |
| ---- | ----| ---| ---| ---| --- | --- |

n0nag0n:
I just checked and it looks like the implementation is broken somehow :-/

joanhey:
Still failing when benchmarking returning 0 req always, but pass the tests.

When I have free time, I'll check it to find the problem.
But it'll be not easy, as don't show any error.

PR #9609, opened 2025-02-27
[sisk] Fix warning

p8:
`benchmark_config.json` wasn't correct resulting in the following warning:

    Framework sisk does not define a default test in benchmark_config.json


p8:
Thanks for merging all these PR's @NateBrady23.
Would it be possible to restart the current run, so we don't have wait a week to start seeing these results? 🙏

PR #9639, opened 2025-03-10
Update ASP.NET Core metadata

sebastienros:
Would you be able to update the <T> moniker to `ASP.NET Core [Platform]` instead of the currently `ASP.NET Core` value?

/cc @DamianEdwards


frameworks/CSharp/aspnetcore/benchmark_config.json
```diff
@@ -67,7 +67,7 @@
         "webserver": "Kestrel",
         "os": "Linux",
         "database_os": "Linux",
-        "display_name": "ASP.NET Core [Minimal APIs, Pg, Dapper]",
+        "display_name": "ASP.NET Core [Minimal APIs]",
         "notes": "",
         "versus": "aspnetcore"
```
DamianEdwards, line 72: did we want to change the `"versus"` value to, so that it's compared to the platform test? What is this actually referring to now?

sebastienros: >  versus: The name of another test

In this case it's the `framework` value concatenated with the name of the test: `[default], minimal, mvc, aot`. So to compare to Platform (which is `default`) we just use the framework name -> `aspnetcore`.

DamianEdwards: I thought the framework name was `ASP.NET Core` now?

sebastienros: Yes, but there are two `framework` fields in the json ... I was referring to the top property (assuming that's what the doc describes) https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Codebase-Framework-Files#benchmark-config-file

sebastienros: The one in each "test":

> framework (metadata): name of the framework (only used to display information on the results site)

this one is display-only

sebastienros: > ok so the actual framework name is "aspnetcore" and the display name is "ASP.NET Core"?

Nothing after this change will display only "ASP.NET Core". There will always be `[{test name}]` at least.

Going back to your original question, putting `aspnetcore` in the `versus` properties means that we are comparing to the Platform test, since it's the `default` test. If we wanted to compare to MVC for instance we would have to use the value `aspentcore-mvc`.

DamianEdwards: > Nothing after this change will display only "ASP.NET Core". 

Right, but the filters panel will show the frameworks as "ASP.NET Core [Platform]", etc. yes? What is the "aspnetcore" value used for then I wonder?

sebastienros: Yes, the filters will show "ASP.NET Core [Platform]" now.

> What is the "aspnetcore" value used for then I wonder?

I would assume only technical things the versus property, or the tooling. But I am wondering how `aspnetcore` is even showing today then. Maybe I need to update the `config.toml` file too. Not sure what is the source of truth between these two.

msmith-techempower: `config.toml` is in limbo atm - it's safe to ignore.

msmith-techempower:
Is this ready to merge?

sebastienros:
Yes, please merge

PR #9659, opened 2025-03-12
[ntex]: add io-uring based runtime

fafhrd91:
* add io-uring based runtime


p8:
This increases the number of test mutation to 12:
> The number of test mutations will be limited to 10. We do not mind if you open up pull requests between runs to try out various mutations for your framework so long as the total number at any given time does not exceed 10.
https://github.com/TechEmpower/FrameworkBenchmarks/issues/8420

fafhrd91:
Oh, I missed that. Removed some permutations.

PR #9667, opened 2025-03-13
Enforce limiting frameworks to a maximum number of test mutations

p8:
> The number of test mutations will be limited to 10.
https://github.com/TechEmpower/FrameworkBenchmarks/issues/8420


msmith-techempower:
I think what I would prefer is that users would get a warning when working on their implementation locally - specifically, it should call out that the max has already been reached, and if they are still set on getting their framework into the tests they need to contact maintainers and remove deprecated or discarded ones to make room.

This puts a lot of unfair effort on the part of a potential contributor to do more than just add their framework... I just don't have a better solution off hand.

p8:
@msmith-techempower I'm showing a warning now, similar to the warning if the "default" test variant is missing.

p8:
Made sure broken tests aren't taken into account.

joanhey:
Why not run only 10 mutations as maximum per run.

Using the firs 7-8 mutations always, and the rest will be run randomly in each run.

If any framework want a mutation than always run, they need to place in the first mutations in the `benchmark_config.json`.

Also we can add a new tag in the json, to define mutations than will run randomly selected.
Showing the warning always, of course.

p8:
Nice! Thanks @msmith-techempower !

PR #9682, opened 2025-03-17
php/spiral: cleanup

roxblnfk:
- enabled ORM Schema cache
- disabled unused modules


roxblnfk:
Is it possible to get output from this?

```
spiral: Exception: The command '/bin/sh -c php app.php configure > /dev/null 2>&1' returned a non-zero code: 255
```

joanhey:
Change
`php app.php configure > /dev/null 2>&1`

to
`php app.php configure`

to see the output.

PR #9687, opened 2025-03-18
[ruby/hanami] Return json and plaintext from the routes.

p8:


p8:
@msmith-techempower I didn't get a notification.

PR #9722, opened 2025-03-24
Runtime improvements

inemtsev:
- Performance improvements
- Upgrade Kotlin compiler to latest version


inemtsev:
Looks good. Let's review and merge please :)

msmith-techempower (approved):

frameworks/Kotlin/ktor/ktor/src/main/kotlin/org/jetbrains/ktor/benchmarks/Hello.kt
```diff
@@ -30,23 +31,34 @@ data class World(val id: Int, var randomNumber: Int)
...
+    val jsonResponse = json.encodeToString(Message("Hello, world!"))
 
     routing {
         get("/plaintext") {
             call.respond(helloWorldContent)
         }
 
         get("/json") {
-            call.respondText(Json.encodeToString(Message("Hello, world!")), ContentType.Application.Json)
+            call.respondText(jsonResponse, ContentType.Application.Json)
```
bjhham, line 61: Hey, I was just making some updates for Ktor and I noticed this might be in violation of one of the test rules:

> For each request, an object mapping the key message to Hello, World! must be instantiated.

Is this not applicable here, or should I change it?

inemtsev: @bjhham I just checked, you are correct, I missed this rule. We should fix this.

msmith-techempower: Bah, I'm going too quickly this morning. You are correct - you need to instantiate the object on every request with the `json` test.

Your `plaintext` is fine.

Unfortunately, Github won't let me un-merge - can you please open a second PR and resolve?

bjhham: Yep, I'll open one soon with a few tweaks to response streaming and an update to 3.1.2. as well.  Thanks for updating it to 3.0 btw, I had intended to in 2024 but it slipped my mind.

inemtsev: No worries, I really enjoy using Ktor.

I struggled to optimize "/update" endpoint though, maybe you know a better way 😅

PR #9725, opened 2025-03-24
Update CICD logic

msmith-techempower:
Attempts to address #9682


msmith-techempower:
Well, I've clearly broken everything and forgot to disable CICD `>_<`.

Issue #9718, opened 2025-03-23
Don't remove docker layers (set `forcerm` = false)


joshka:
Currently the test runner intentionally removes each layer unconditionally in a docker file.
https://github.com/TechEmpower/FrameworkBenchmarks/blob/b9f7af61769d62058acc9fe493b0f549ad3293ad/toolset/utils/docker_helper.py#L42

This prevents using the layers for caching build output of deps. E.g. in a Rust docker container you can do something like:

```dockerfile
COPY Cargo.toml Cargo.lock ./
RUN mkdir src && echo "fn main() {}" > src/main.rs && cargo build --release && rm -rf src

COPY . .
RUN RUSTFLAGS="-C target-cpu=native" cargo install --path . --locked
```

This will only build the dependencies once and then cache that for all future builds. This allows faster experimentation (cutting out 30s-minutes of each build test cycle depending on the tree / CPU speed). Cache busting is fairly simple (update the toml / lock file, or more generally just `touch` them so they're detected as newer), and the same technique applies to other languages not just Rust.

There was a unanswered question about this in 2018 about whether to remove it. https://github.com/TechEmpower/FrameworkBenchmarks/pull/3574#discussion_r181799480 by @msmith-techempower. I'm curious if there's any known blockers to doing so?

c-cesar:
Maybe the best solution is simply tell deve to tweak this option, or add a flag to tfb command to diable it. 
This way we can have fast local testing and don't mess with citrine setup.

PR #9727, opened 2025-03-24
Add --force-rm flag

joshka:
[feat: add --force-rm flag to configure removing intermediate docker layers](https://github.com/TechEmpower/FrameworkBenchmarks/commit/15c254722862df4089ff8210d5f41d7874ec6c7a)

This commit makes it possible to turn off the removal of intermediate
docker layers when building the tfb containers by adding a `--force-rm`
flag to the tfb script. This is useful in situations where you want to
inspect the intermediate layers for debugging purposes, or for caching
builds of dependencies as a docker layer to speed up the build process.

Note that the default behavior is to remove the intermediate layers to
avoid filling up the disk with unused layers on the citrine server.

Fixes: https://github.com/TechEmpower/FrameworkBenchmarks/issues/9718

[hyper: cache dependencies to reduce build time](https://github.com/TechEmpower/FrameworkBenchmarks/commit/e8711902ccbbe05032e3e76902672f735d70a20f)

This change reduces the time it takes to build the hyper Docker image by
caching the dependency builds. This is done by building a dummy binary
before copying the source code into the image.

### DIFF
--- a/toolset/run-tests.py
+++ b/toolset/run-tests.py
@@ -208,6 +208,10 @@ def main(argv=None):
         nargs='*',
         default=None,
         help='Extra docker arguments to be passed to the test container')
+    parser.add_argument(
+        '--force-rm',
+        default=True,
+        help='Remove intermediate docker containers after running.')
 
     # Network options
     parser.add_argument(
--- a/frameworks/Rust/hyper/hyper.dockerfile
+++ b/frameworks/Rust/hyper/hyper.dockerfile
@@ -1,8 +1,17 @@
 FROM rust:1.85 AS hyper
 
 WORKDIR /src
-COPY . .
-RUN RUSTFLAGS="-C target-cpu=native" cargo install --path . --locked
+ENV RUSTFLAGS="-C target-cpu=native"
+
+# Cache dependency builds (requires passing --force-rm False to tfb command)
+COPY Cargo.toml Cargo.lock /src/
+RUN mkdir src \
+    && echo "fn main() {println!(\"if you see this, the build broke\")}" > src/main.rs \
+    && cargo build --release \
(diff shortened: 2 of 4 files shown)


toolset/run-tests.py
```diff
@@ -208,6 +208,10 @@ def main(argv=None):
         nargs='*',
         default=None,
         help='Extra docker arguments to be passed to the test container')
+    parser.add_argument(
+        '--force-rm',
+        default=True,
+        help='Remove intermediate docker containers after running.')
```
msmith-techempower, line 214: I think that if we're going to do this in the aim of improving ergonomics for contributors who need to quickly iterate implementations again and again and want those layers to persist between attempts, then this should be `False` by default, and we need to update our CI scripts to forcibly set to `True` (and our continuous scripts).

msmith-techempower (changes requested):
Can we also prepend `[CI fw-only Rust/hyper]` to the next commits?

steveklabnik (commented):
I don't really know docker and so cannot review this PR.

joshka:
> I don't really know docker and so cannot review this PR.

No problem - you were auto tagged because I previously added you to the maintainers config on the hyper test, and because I included the hyper change to show how / why this makes sense as a change.

---

If you're interested a bit in how this works, each command run from a docker file creates a saved layer which can be used by future calls to docker to jump straight to that point rather than running that command again when building an image. Those intermediate layers take up space though, and tend to exhaust the benchmarking storage, so the benchmark turned that saving off.

It's useful to turn it on however for speeding up development activities. The trick to doing so in Rust projects is to compile an empty main.rs against all the dependencies listed in your manifest and lock file and then delete binary that you just created. This saves a docker layer that has a cache of all the compiled dependencies, which speeds up future compilation times.

PR #9753, opened 2025-03-30
[csharp/tetsuweb] Remove Tetsuweb

p8:
TetsuWeb hasn't seen an update in 6 years:
https://github.com/gkbrk/TetsuWeb/tree/master

Also most downloads are probably for the TechEmpower tests:
https://www.nuget.org/packages/Tetsu.Web/0.1.5#versions-body-tab


msmith-techempower:
I love all red commits

PR #9767, opened 2025-04-02
Aiohttp switch to orjson

Reskov:
Switch to `orjson` since it outperforms `ujson` https://dev.to/dollardhingra/benchmarking-python-json-serializers-json-vs-ujson-vs-orjson-1o16

```python
pip install aiohttp orjson ujson

from aiohttp.web import Response
from aiohttp import web
import orjson
import ujson

def json_response(payload):
    return Response(
        body=orjson.dumps(payload),
        content_type="application/json",
    )


async def orjson_handler(request):
    return json_response(
        {'message': 'Hello, World!'},
    )

async def ujson_handler(request):
    return web.json_response(
        text=ujson.dumps({'message': 'Hello, World!'}),
        content_type="application/json",
    )

app = web.Application()
app.router.add_get('/orjson', orjson_handler)
app.router.add_get('/ujson', ujson_handler)

if __name__ == '__main__':
    web.run_app(app, port=8080)
```

### Local benchmark
```
 wrk -t8 -c20 -d20s http://localhost:8080/orjson 
Running 20s test @ http://localhost:8080/orjson
  8 threads and 20 connections
    Latency   622.93us    1.14ms  38.03ms   99.60%
    Req/Sec     3.56k   165.63     3.94k    82.90%
  569116 requests in 20.10s, 93.35MB read
Requests/sec:  28310.04
Transfer/sec:      4.64MB
```

```
 wrk -t8 -c20 -d20s http://localhost:8080/ujson
Running 20s test @ http://localhost:8080/ujson
  8 threads and 20 connections
    Latency   805.26us    3.19ms  77.12ms   99.31%
    Req/Sec     3.48k   218.00     3.76k    96.52%
  556324 requests in 20.10s, 99.21MB read
Requests/sec:  27675.67
Transfer/sec:      4.94MB
```

### DIFF
--- a/frameworks/Python/aiohttp/app/views.py
+++ b/frameworks/Python/aiohttp/app/views.py
@@ -1,11 +1,10 @@
-from functools import partial
 from operator import attrgetter, itemgetter
 from pathlib import Path
 from random import randint
 
 import jinja2
-import ujson
-from aiohttp.web import Response, json_response
+import orjson
+from aiohttp.web import Response
 from sqlalchemy import select
 
 from .models import sa_fortunes, sa_worlds, Fortune, World
@@ -16,7 +15,6 @@ READ_ROW_SQL = 'SELECT "randomnumber", "id" FROM "world" WHERE id = $1'
 READ_SELECT_ORM = select(World.randomnumber)
 WRITE_ROW_SQL = 'UPDATE "world" SET "randomnumber"=$2 WHERE id=$1'
 
-json_response = partial(json_response, dumps=ujson.dumps)
 template_path = Path(__file__).parent / 'templates' / 'fortune.jinja'
 template = jinja2.Template(template_path.read_text())
 sort_fortunes_orm = attrgetter('message')
@@ -34,6 +32,11 @@ def get_num_queries(request):
         return 500
     return num_queries
 
+def json_response(payload):
+    return Response(
+        body=orjson.dumps(payload),
+        content_type="application/json",
+    )
 
 async def json(request):
     """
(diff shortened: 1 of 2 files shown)


Reskov:
Builds fail on `/updates`, but I'm not sure if switching the json response library can cause this issue. Maybe it is because of the update isn't triggered when randomNumber matching previous, and we need to have something similar.
 https://github.com/TechEmpower/FrameworkBenchmarks/blob/40b97c5db222bab3c7c7950ff43b5ee573dae090/frameworks/Python/fastapi/app_orm.py#L152

Reskov:
@Dreamsorcerer please take a look. 

1) The benchmark performance are unstable locally `./tfb --mode verify --test aiohttp` So I can not measure the actual impact of switching from `ujson` to `orjson`. But orjson.dumps outperforms ujson on a separate benchmark.

2) The changes of the db is necessary to prevent framework tests flaky

3) psycopg2 library is not used.

Dreamsorcerer:
Changes are now overlapping my PR, but yeah, looks good to me. Either can be merged and then conflicts fixed.

Dreamsorcerer:
Note that the places your last commit deviates from my changes though, I already micro-benchmarked.
e.g.
```
> python3 -m timeit 'from random import randint, sample; a = sample(range(1, 10001), 1000); a.sort()' 'for x in a: y = randint(1, 10000)'
100 loops, best of 5: 2.14 msec per loop
> python3 -m timeit 'from random import randint, sample; a = sample(range(1, 10001), 1000); a.sort()' 'x = tuple(zip(a, sample(range(1, 10001), 1000)))
              for y,z in x: pass'
200 loops, best of 5: 1.79 msec per loop
```

msmith-techempower (approved):

PR #9796, opened 2025-04-09
aiohttp base nginx setup

Reskov:


Reskov:
@Dreamsorcerer please review

frameworks/Python/aiohttp/nginx-entrypoint.sh
```diff
@@ -0,0 +1,53 @@
...
+done
+cat >> /etc/nginx/conf.d/default.conf <<EOF
+    keepalive 32;
+}
+
+server {
+    listen 8080;
+
+    access_log off;
+    error_log /var/log/nginx/error.log crit;
```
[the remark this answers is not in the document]
Reskov: Agreed. Updated PR to `use error_log stderr;`

frameworks/Python/aiohttp/nginx-entrypoint.sh
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    server {
+        listen 8080 reuseport;
+
+        access_log off;
+        error_log stderr error;
+
+        location / {
+            proxy_pass http://aiohttp;
```
Dreamsorcerer, line 50: We can use a Unix socket here.

Dreamsorcerer: Actually, not here, above.

frameworks/Python/aiohttp/nginx-entrypoint.sh
```diff
@@ -0,0 +1,61 @@
...
+
+http {
+    keepalive_requests 10000000;
+
+    upstream aiohttp {
+        least_conn;
+EOF
+
+for i in $(seq 0 $((CORES-1))); do
+  echo "        server 127.0.0.1:$((8000 + i));" >> /aiohttp/nginx.conf
```
Dreamsorcerer, line 36: It'd be a little more work, but we could create unix sockets for all of these.
Config would look more like:
```
server unix:/run/aiohttp/aiohttp-{i}.sock fail_timeout=0;
```

frameworks/Python/aiohttp/app/app.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+if __name__ == '__main__':
+    uvloop.install()
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--port', type=int, required=True)
+    args = parser.parse_args()
+
+    app = create_app()
+    web.run_app(app, port=args.port)
```
Dreamsorcerer, line 14: Here we would do:
```suggestion
    web.run_app(app, path=args.sock_path)
```

frameworks/Python/aiohttp/aiohttp-nginx.dockerfile
Dreamsorcerer, line 1: I originally assumed you were going to replace the gunicorn versions. If we're keeping them, I'd suggest maybe making this one the default and renaming the other one to `aiohttp-gunicorn`, given that our default recommendation is to use nginx.

frameworks/Python/aiohttp/nginx-entrypoint.sh
joanhey: Always it's better sockets, for performance !!

Reskov: yes, thanks for suggestion i've updated pr

frameworks/Python/aiohttp/aiohttp-nginx.dockerfile
Reskov: I agree with this approach. Consolidating the default `nginx` since it more performant (for both standard and ORM templates) while keeping a separate `aiohttp-gunicorn` makes sense.

Reskov:
@joanhey @Dreamsorcerer Thanks for the review. I've applied the suggested fixes.

Dreamsorcerer (approved):

msmith-techempower (approved):

Dreamsorcerer:
> Surprisingly, gunicorn outperforms nginx on my local machine.

Something looks wrong there, it appears to be exactly double the performance. I wonder if there's something wrong with the load balancing?

Reskov:
I'll try to double check the configuration, but any advice would be extremely useful to me. Locally I also tried to deploy using haproxy, results was better but still worse than gunicorn has.

frameworks/Python/aiohttp/nginx-entrypoint.sh
```diff
@@ -0,0 +1,64 @@
...
+
+events {
+    worker_connections 65535;
+}
+
+http {
+    keepalive_requests 10000000;
+
+    upstream aiohttp {
+        least_conn;
```
Dreamsorcerer, line 35: Is there any difference in performance if we drop this and use the default round-robin? Guessing this isn't really needed given that requests should be around 1ms.

frameworks/Python/aiohttp/nginx-entrypoint.sh
```diff
@@ -0,0 +1,64 @@
...
+EOF
+
+for i in $(seq 0 $((CORES-1))); do
+  echo "        server unix:/run/aiohttp-$i.sock fail_timeout=0;" >> /aiohttp/nginx.conf
+done
+
+cat >> /aiohttp/nginx.conf <<EOF
+        keepalive 32;
```
Dreamsorcerer, line 43: I'm also wondering if this should be much higher? i.e. The aiohttp servers can handle many simultaneous connections.

With the above config, it might be that the workers are creating upto 65535, of which 32 of them are allowed to be kept alive. So, we probably want this to be the same value.

Dreamsorcerer: 65535 connections could also start slowing down the event loop, so also worth testing with lower numbers too, once these are aligned.

Reskov: I've tried different options, someone helps but all of them insignificantly. So i decided to revert nginx as default proxy and keep it as separate docker file https://github.com/TechEmpower/FrameworkBenchmarks/pull/9807

PR #9814, opened 2025-04-12
Optimize http headers creation

franz1981:
This is optimizing http header creation


franz1981:
This is using a timer per event loop and a scratch buffer/header for each connection opened.
It could be further improved (through a central storage (using 4 volatile fields) reducing by far the memory pressure and the number of required timer(s).

franz1981:
I think that the original code for this benchmark is broken because it doesn't stop the timer after a connection is closed. I will put this in draft and fix it first

franz1981:
This is ready to go as well since the problems were already present before and won't likely get any worse with this 🙏

franz1981:
@msmith-techempower ready to go :pray:

msmith-techempower (approved):

PR #9828, opened 2025-04-16
[Python] Upgrade Blacksheep framework

bymoye:
1. Upgrade Blacksheep
2. use msgspec
3. add nginx-unit


bymoye:
@msmith-techempower

bymoye:
Use custom unitd startup method to fix nginx-unit and test event sequence.
I have tested it locally and did not encounter any problems.
Can you review it? Thx!
@msmith-techempower

bymoye:
@msmith-techempower Please test and merge

PR #9835, opened 2025-04-18
Make logger less spammy

inemtsev:
- Logger less spammy
- Some tests didn't even use loggers, this will make things more consistent and probably improve performance


inemtsev:
@bjhham @msmith-techempower Can you guys please review? I saw in some tests that excessive logging can bring down performance, so I made it less spammy.

msmith-techempower (approved):

PR #9874, opened 2025-04-30
Add extensions used in Laravel, Symfony.

kitrio:
1. Added PHP extension (Intl ,bcmath) to Docker setup.

Symfony and Laravel use bcmath and intl, The current setup is using polyfills.
These extensions (bcmath, intl) are commonly used in PHP projects and are recommended to be enabled in default setups.

2. Changed the Laravel opcache configuration to match Symfony.

Laravel makes heavy use of Symphony components, and the recommended settings for these full-stack PHP frameworks are similar.


msmith-techempower (approved):

joanhey:
Good change, but I think that it is NOT important for this bench.

Thank you !!

PR #9881, opened 2025-05-06
[php] Update for Workerman/5

joanhey:
The namespace for the Timer change in Workerman/5.
We fix it with this PR, or the next run will fail almost all.

We add also reuseport.
Actually all AdapterMan implementations are not using JIT.


joanhey:
Only fail PHP with Caddy server, that this PR don't touch, and also fail in the last runs.

I'll try to fix it in another PR.

joanhey:
This PR is ready to be merged !!

PR #9903, opened 2025-05-18
Aiohttp: Add pypy3 and minor optimisation

Dreamsorcerer:
I've also switched docker images to use the latest tag, so these don't need to be updated in future. Plus some other cleanup to reduce maintenance.


frameworks/Python/aiohttp/aiohttp-gunicorn.dockerfile
```diff
@@ -1,14 +1,14 @@
-FROM python:3.13
+FROM python:latest
```
joanhey, line 1: `FROM python:3` it's better.
If python use SEM versioning is not problem.
But `latest` is not so !!!
If you will be here for some years, and fix the problems it's OK, but now this line create a problem.
Thank you !!

joanhey (changes requested):

frameworks/Python/aiohttp/aiohttp-gunicorn.dockerfile
Dreamsorcerer: Suppose that makes sense.

Dreamsorcerer: Actually, I'm not sure it'll make any difference. Any changes for a v4 are no more likely to break the small amount of code here than a minor release would. Ideally we'd be able to have Dependabot bump the pinned version automatically...

msmith-techempower: I think I'm okay with going with `latest` and if it breaks something we'll see it pretty plainly.

msmith-techempower (approved):

frameworks/Python/aiohttp/aiohttp-gunicorn.dockerfile
msmith-techempower: Actually, thinking on this for more than one second, it would make it harder for people to understand what version was run on any given continuous benchmark run from day to day. Imagine some breaking change from 3.13 to 3.14 (as a contrived example) but this dockerfile says `latest` - might be hard without some guesswork what's going on. Agreeing with @joanhey that this should be returned to sem. ver.

msmith-techempower (changes requested):

Dreamsorcerer:
> Actually, thinking on this for more than one second, it would make it harder for people to understand what version was run on any given continuous benchmark run from day to day.

I'd assume that would be visible in the logs somewhere...

> Agreeing with @joanhey that this should be returned to sem. ver.

Well, that suggestion was actually to use `3` which wouldn't make any difference to the situation.

I don't mind too much either way though.

PR #9907, opened 2025-05-21
Swift 6.1

adam-fowler:
Also compile fixes for hummingbird2 postgres

@0xTim I couldn't get the vapor-postgres, vapor-sqlkit or vapor-swiftql (don't know where this came from) builds to run, so I've left them to you


0xTim (approved):

adam-fowler:
@0xTim sorry. I had to remove all the Vapor changes as it wants to build all the targets regardless of which ones were actually changed. And it appears to run out of memory, if we build all of them.

msmith-techempower (approved):

PR #9914, opened 2025-05-25
Upgrade versions 05 2025

inemtsev:
- Upgraded versions of Kotlin, Ktor and some minor dependencies
- Made Java config more consistent, should be the same across all Ktor types now


inemtsev:
LGTM. Please review and merge :)

msmith-techempower (approved):

ShreckYe:
@inemtsev Hi I am recently adding some new features to the Ktor Exposed portion. I see you updated the GC to G1GC across all Ktor portions but missed `ktor-exposed-dao` so I am about to update this. However, from what I learned online, ParallelGC brings better throughput compared to G1GC but has higher latency, and I also contributed to the Vert.x portions and they all use ParallelGC, so I am not sure which one to go for. What's your opinion and experience on this from tweaking these benchmarks? Do you recall significant performance differences between them?

inemtsev:
> @inemtsev Hi I am recently adding some new features to the Ktor Exposed portion. I see you updated the GC to G1GC across all Ktor portions but missed `ktor-exposed-dao` so I am about to update this. However, from what I learned online, ParallelGC brings better throughput compared to G1GC but has higher latency, and I also contributed to the Vert.x portions and they all use ParallelGC, so I am not sure which one to go for. What's your opinion and experience on this from tweaking these benchmarks? Do you recall significant performance differences between them?

Hi,

I originally thought of only using G1GC because that is the most common real-life use-case. But, I have heard that ParallelGC can give better results in benchmarks. If you want to try changing all Ktor variants to use the ParallelGC, I'm good with that. Maybe it will give bit better results. 👍

PR #9915, opened 2025-05-27
(Laravel) - force updating to a new number

cosmastech:
In Laravel, if you attempt to update a Model but have made no changes to its data, it will not write to the database. See [Eloquent@save()](https://github.com/laravel/framework/blob/f880ec20247a962fd8b553f5049afe4125e9dac4/src/Illuminate/Database/Eloquent/Model.php#L1203-L1206).

While the chances are vanishingly small that this would happen enough times to significantly affect benchmarks, I wanted to remove noise from the benchmarks.

### DIFF
--- a/frameworks/PHP/laravel/app/Http/Controllers/Controller.php
+++ b/frameworks/PHP/laravel/app/Http/Controllers/Controller.php
@@ -52,7 +52,7 @@ class Controller extends BaseController
 
         while ($queries--) {
             $row = World::query()->find(self::randomInt());
-            $row->randomNumber = self::randomInt();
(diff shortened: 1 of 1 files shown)


frameworks/PHP/laravel/app/Http/Controllers/Controller.php
```diff
@@ -52,7 +52,7 @@ public function updates($queries = 1)
 
         while ($queries--) {
             $row = World::query()->find(self::randomInt());
-            $row->randomNumber = self::randomInt();
+            $row->randomNumber = ($randomInt = self::randomInt()) !== $row->randomNumber ? $randomInt : $randomInt + 1;
```
p8, line 55: If `$randomInt` is `10000` this would fail, as `10001` is invalid.
Other frameworks generate a new random number instead.

PR #9923, opened 2025-05-31
Upgrade hyperlane

eastspire:
Upgrade hyperlane


joanhey:
Please could you add a default test in benchmark_config.json.

Now create a noisy error line in each run (also in local tests)
image

msmith-techempower (approved):

PR #9981, opened 2025-07-03
D improvements

denizzzka:


cyrusmsk:
Having multithreading version of vibe.d - would be nice

denizzzka:
@cyrusmsk There is no much sense: new ones are better than the old (as for me)

cyrusmsk:
> @cyrusmsk There is no much sense: new ones are better than the old (as for me)

But running on all cores should show better performance.
Maybe run several copies of vibe.d instances with one load-balancer .. because otherwise it is single-core vs multi-core comparison

denizzzka:
@cyrusmsk oh, I accidentally read that you want to have multiple versions of vibe.d :-)

Maybe you're right, but for now we will work with what we have

denizzzka:
https://github.com/TechEmpower/FrameworkBenchmarks/pull/9982

PR #10000, opened 2025-07-20
Add Tachyon — a very slow and extremely unsafe Rust server

TachyonConcepts:
Hello TFB team,

If it's not too much trouble, would you consider adding this humble submission to the benchmark suite?

**Tachyon** is a web server written entirely in `unsafe` Rust (yes, every single line).  
It does only one thing — serve `/plaintext` — and it does it... well, slowly. Very slowly.  
In fact, it’s arguably one of the safest entries to benchmark because you’ll never get overwhelmed by throughput.

This PR includes only the minimal files required to make it work inside the TFB framework.  
The server uses `io_uring`, AVX2, and very questionable life choices — all of which are thoroughly documented in the included `README.md`.

Thanks for your time and for maintaining such a fantastic benchmark project 🙏

### DIFF
--- a/frameworks/Rust/tachyon-concept/README.md
+++ b/frameworks/Rust/tachyon-concept/README.md
@@ -1,5 +1,7 @@
 # Tachyon
 
+[Tachyon Repo](https://github.com/TachyonConcepts/TachyonConcept)
+
 **Tachyon** is a web server written in 100% `unsafe` Rust.  
 Yes — **100%**. There isn’t a single `safe` block in the entire codebase. Safety is a suggestion, not a rule.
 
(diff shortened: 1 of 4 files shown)


volyrique:
> ... would you consider adding this humble submission to the benchmark suite?

They are unlikely to consider it - refer to issue #8420.

TachyonConcepts:
> > ... would you consider adding this humble submission to the benchmark suite?
> 
> They are unlikely to consider it - refer to issue #8420.

Good day.
I’ve added support for an additional endpoint — /json, as you requested, so it’s no longer limited to /plaintext.
I hope this update meets the requirements and that TechEmpower will now consider including the framework in the test suite.

PR #9751, opened 2025-03-29
add OCaml httpun using eio

rbjorklin:
Not performing as well as expected. Not ready for merge just yet.

PR #10009, opened 2025-07-27
Add `httpcats`

dinosaure:
This PR is on top of #9751 and, as the previous PR, it is a draft. I can spend a time to improve both of PRs if @rbjorklin wants & tells me what is missing.


rbjorklin:
I did some digging but ran out of steam. Would probably just commit a fix to adjust the number of domains created and then call it a day!

msmith-techempower:
Closing along with #9751 for grooming.

PR #10010, opened 2025-07-27
add new test : CSharp/akazawayun.pro

AkazawaYun:


AkazawaYun:
I find the wiki about how to response the test,  and changed some code, is it ok to merge now ?

PR #10012, opened 2025-07-29
Fix bad `benchmark_config.json` in frameworks

joanhey:
Problem when run `./tfb`:

```
 => => naming to docker.io/techempower/tfb                                                                                                                0.0s
Framework oak does not define a default test in benchmark_config.json
Framework ring-http-exchange does not define a default test in benchmark_config.json
Framework ring-http-exchange does not define a default test in benchmark_config.json
Framework ring-http-exchange does not define a default test in benchmark_config.json
Framework http4k does not define a default test in benchmark_config.json
Framework hyperlane does not define a default test in benchmark_config.json
Framework httpserver does not define a default test in benchmark_config.json
Framework oak does not define a default test in benchmark_config.json
Framework ring-http-exchange does not define a default test in benchmark_config.json
Framework ring-http-exchange does not define a default test in benchmark_config.json
Framework ring-http-exchange does not define a default test in benchmark_config.json
Framework http4k does not define a default test in benchmark_config.json
Framework hyperlane does not define a default test in benchmark_config.json
Framework httpserver does not define a default test in benchmark_config.json
Running Tests...

```

They have bad json or not default.
And create noisy lines when running any test locally.
Perhaps we need to add a test for this file.

PD: as this PR don't touch the code, any fail was there before.

### Worst and need to fix

Hyperlane need to be redone.

Actually each test alone have a dockerfile and is compiling the rust, very bad for the running time and not realistic.
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Rust/hyperlane/benchmark_config.json

@尤雨东  @尤雨东

I can't contact the authors by mention, and the last PRs for this problem don't exist (#9987, #9976, #9959, ...)
https://github.com/TechEmpower/FrameworkBenchmarks/commits/master/frameworks/Rust/hyperlane

This change in Hyperlane can't pass a review.

@msmith-techempower @NateBrady23  What we do with this situation ?

Mark as bad, revert PR, ...   ?

### DIFF
--- a/frameworks/Clojure/ring-http-exchange/benchmark_config.json
+++ b/frameworks/Clojure/ring-http-exchange/benchmark_config.json
@@ -20,9 +20,7 @@
         "display_name": "ring-http-exchange",
         "notes": "",
         "versus": "httpserver"
-      }
-    },
-    {
+      },
       "robaho": {
         "json_url": "/json",
         "plaintext_url": "/plaintext",
@@ -41,9 +39,7 @@
         "display_name": "ring-http-exchange-robaho",
         "notes": "",
         "versus": "httpserver-robaho"
-      }
-    },
-    {
+      },
       "graalvm": {
         "json_url": "/json",
         "plaintext_url": "/plaintext",
@@ -62,9 +58,7 @@
         "display_name": "ring-http-exchange-graalvm",
         "notes": "",
         "versus": "ring-http-exchange"
-      }
-    },
-    {
+      },
       "robaho-graalvm": {
         "json_url": "/json",
         "plaintext_url": "/plaintext",
--- a/frameworks/Java/httpserver/benchmark_config.json
+++ b/frameworks/Java/httpserver/benchmark_config.json
@@ -57,9 +57,7 @@
         "display_name": "httpserver-graalvm",
         "notes": "",
         "versus": "httpserver"
-      }
-    },
-    {
+      },
       "robaho": {
         "json_url": "/json",
         "plaintext_url": "/plaintext",
--- a/frameworks/Kotlin/http4k/benchmark_config.json
+++ b/frameworks/Kotlin/http4k/benchmark_config.json
@@ -2,7 +2,7 @@
   "framework": "http4k",
   "tests": [
     {
-      "apache": {
+      "default": {
         "orm": "Raw",
         "database_os": "Linux",
         "cached_query_url": "/cached?queries=",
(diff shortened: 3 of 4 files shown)


msmith-techempower:
> and the last PRs for this problem don't exist 

Whoa... I have never seen that. How does that even happen?

> Actually each test alone have a dockerfile and is compiling the rust, very bad for the running time and not realistic.

Agreed. Bad oversight on my part. I can fix - none of the Dockerfiles differ in any meaningful way; they just enable features (for routing) at compile time... sort of an odd design choice. Assuming the routing is sane, the worst case is basically zero-cost lookup... so I'll just create one Dockerfile that enables all the routes.

> They have bad json or not default.

I can also fix - just need set a default for each.

joanhey:
I fixed all the bad `benchmark_config.json`.
Only left the Hyperlane problem in this PR.

msmith-techempower (approved):

eastspire:
My account was suspended and only resumed the day before yesterday. I just submitted an action and found that the execution failed. I also saw an error message. My code has not been changed during this period. Is there any change in the processing logic for this？

eastspire:
> > and the last PRs for this problem don't exist
> 
> Whoa... I have never seen that. How does that even happen?
> 
> > Actually each test alone have a dockerfile and is compiling the rust, very bad for the running time and not realistic.
> 
> Agreed. Bad oversight on my part. I can fix - none of the Dockerfiles differ in any meaningful way; they just enable features (for routing) at compile time... sort of an odd design choice. Assuming the routing is sane, the worst case is basically zero-cost lookup... so I'll just create one Dockerfile that enables all the routes.
> 
> > They have bad json or not default.
> 
> I can also fix - just need set a default for each.

Using separate configurations for each route is to minimize the impact of routing on performance as much as possible. Although this impact is minimal or even non-existent, I still hope to optimize the framework performance through code or configuration

joanhey:
The problem is that for each test (route) the run need to create a new image from dockerfile, very bad for the time to run it.

Any framework can do the same trick, but this is NOT the intention of this benchmark, we want realistic results.
Do you'll create a new binary for each route in your app ??

Even exist a issue about that, and another about create a mixed bench, with all the routes (but that all are for the same fw and platform) and from the same dockerfile.

eastspire:
I have completed the modifications in this PR: https://github.com/TechEmpower/FrameworkBenchmarks/pull/10023

PR #10024, opened 2025-08-03
[php] Wolff add Adapterman variant

joanhey:


n0nag0n:
Hey this is cool! Your adapterman stuff is pretty neat. Also FYI, I (finally) found a memory leak in Flight a few weeks back that probably is the reason why Flight adapterman fails in the techempower benchmarks. Hopefully it'll pass now!

msmith-techempower (approved):

joanhey:
Results:

Captura de pantalla_área-de-selección_20250819094700
Captura de pantalla_área-de-selección_20250819094756
Captura de pantalla_área-de-selección_20250819094812
Captura de pantalla_área-de-selección_20250819094827


https://x.com/adaptermanphp/status/1957711631821000884

n0nag0n:
Wow, congrats on the power boost!

joanhey:
And still can be faster.
Now we need a `new Wolff\Kernel($config)` for each request.

Ideally we only need to set one `Kernel` with the `$config`, and only call the `start()` for each request.
But It isn't working that way. :/

PD: also I updated the code in #10061 and it'll be faster the next run.

Issue #7493, opened 2022-07-24
add shelf to the list of benchmarked frameworks


iapicca:
I'm aware that [shelf](https://pub.dev/packages/shelf) is not a framework by definition and more of a middleware
but I believe that adding it to the list of frameworks 
will provide developers with a significant reference about the performance impacts compared to a "vanilla" implementation

iapicca:
sorry I don't have time to follow up, I closed the PR

iapicca:
closing as now shelf has been [merged](https://github.com/TechEmpower/FrameworkBenchmarks/commit/4af4805eba5028dc1e3863dfeab268b849a60e5b)

PR #10025, opened 2025-08-03
Add Shelf Framework

felangel:
https://github.com/dart-lang/shelf/issues/175

cc @kevmoo @iapicca

### DIFF
--- a/frameworks/Dart/shelf/bin/server.dart
+++ b/frameworks/Dart/shelf/bin/server.dart
@@ -1,0 +1,84 @@
+import 'dart:convert';
+import 'dart:io';
+import 'dart:isolate';
+
+import 'package:shelf/shelf.dart';
+import 'package:shelf/shelf_io.dart' as shelf_io;
+
+void main(List<String> args) async {
+  /// Create an [Isolate] containing an [HttpServer]
+  /// for each processor after the first
+  for (var i = 1; i < Platform.numberOfProcessors; i++) {
+    await Isolate.spawn(_startServer, args);
+  }
+
+  /// Create a [HttpServer] for the first processor
+  await _startServer(args);
+}
+
+/// Create a request handler
+Response _handler(Request request) {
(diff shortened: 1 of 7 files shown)


frameworks/Dart/shelf/bin/server.dart
```diff
@@ -0,0 +1,84 @@
...
+
+  /// Create a [HttpServer] for the first processor
+  await _startServer(args);
+}
+
+/// Create a request handler
+Response _handler(Request request) {
+  switch (request.url.path) {
+    case 'json':
+      return _jsonTest();
```
alestiago, line 23: nit: Do we prefer to inline the logic over private functions? From some benchmarks I did ages ago, calling a function over inlining introduces a very minimal delay (when done a lot of times). It was quite minuscule though.

felangel: I tried to maintain the same structure as in the pure Dart examples. I wouldn’t expect it to matter since the compiler would likely decide to inline (we can always annotate with `@pragma(vm:prefer-inline)` if needed).

iapicca:
fixes https://github.com/TechEmpower/FrameworkBenchmarks/issues/7493

@felangel @alestiago 
you may want to take a look at this [workaround](https://github.com/dart-lang/sdk/issues/60815#issuecomment-2917552393) to the performance regression

felangel:
> fixes #7493
> 
> @felangel @alestiago you may want to take a look at this [workaround](https://github.com/dart-lang/sdk/issues/60815#issuecomment-2917552393) to the performance regression

I think we should write idiomatic dart since this aims to be a real world benchmark. Ideally those performance optimizations should be made by the language team 🤷

alestiago:
@msmith-techempower could you help with a review? 👀

msmith-techempower:
Sorry, was out on holiday last week... catching up today >_<

PR #10036, opened 2025-08-06
[D, Juptune] Fix: Don't recreate reader & writer each loop - supports pipelining

BradleyChatha:
I didn't catch this initially as it turns out I only ever really looked at the warmup results, which doesn't use pipelining. I only noticed since I checked the rolling testing page and saw this solution was failing.

Basically after a request finished any buffered data would be discarded, causing subsequent reads to start from a half-deleted request, causing errors.

This PR fixes the issue by correctly moving some initialisation code outside of the request reading loop.


joanhey:
Please could you add another test ?

No longer accepting plaintext only frameworks / Limited number of tests mutations #8420

joanhey:
Also the plaintext have a lot of problems:

image

BradleyChatha:
> Please could you add another test ?
> 
> No longer accepting plaintext only frameworks / Limited number of tests mutations #8420

Ah yeah I can probably get a JSON one done this weekend - it shouldn't take too long to get a decent JSON parser written up (D's standard library has one, but it goes against the point of Juptune's goals to use it).

Anything involving a database will take a while though, since it'll be a while until I get an implementation written for the benchmark to use.

> Also the plaintext have a lot of problems.

Sorry if I was unclear - this PR should fix that. Right now it passes verification since the verification stuff doesn't seem to use pipelining, but once the actual benchmarks run (using the `pipeline.lua` wrk script) then pipelining kicks in and triggers this bug.

I've confirmed locally that via `./tfb --mode benchmark --test juptune --type plaintext` that all calls to `wrk` are actually working this time with this fix.

BradleyChatha:
I've added in a JSON benchmark now - neither Juptune nor D's standard library currently provides a serialiser, so I've put together a really simple one in an effort to properly reflect the intentions of the benchmark.

msmith-techempower (approved):

PR #9950, opened 2025-06-13
[php] Laravel apply opcache preloading

kitrio:
Using the opcache.preloading option to reduce PHP opcode loading time.

PR #10041, opened 2025-08-08
[php] Laravel with RoadRunner

joanhey:
Added protobuf extension for faster performance. Thanks @roxblnfk .

Tried with `octane:start`  but it was easier with `rr serve` directly, to add the Server Header without a middleware.

Update to Laravel 12 and PHP8.4.

Added `config/octane.php` to set  `max-request=10000` for all platforms using Octane, that is better for a benchmark.

https://github.com/TechEmpower/FrameworkBenchmarks/issues/9408

PD: ready to be merged.
Laravel-s is fixed in #9950

### DIFF
--- a/frameworks/PHP/laravel/laravel-roadrunner.dockerfile
+++ b/frameworks/PHP/laravel/laravel-roadrunner.dockerfile
@@ -31,3 +31,4 @@ EXPOSE 8080
 # https://artisan.page/12.x/
 #ENTRYPOINT ["php", "artisan", "octane:roadrunner", "--host=0.0.0.0", "--port=8080", "--workers=auto", "--max-requests=10000", "--rr-config=/laravel/deploy/roadrunner/.rr.yaml"]
 ENTRYPOINT ["/laravel/rr", "serve", "-c", "/laravel/deploy/roadrunner/.rr.yaml"]
+


joanhey:
@roxblnfk could you help to add the Server Header globally and any optimization ?

It's working OK, but need the Server Header.
And I think it's better than RoadRunner add it globally, without an extra Laravel middelware.

roxblnfk:
Try to add this declaration with `http` middleware
```yaml
http:
 headers:
    response:
      Server: "RoadRunner"
```

Oh, I see it's here yet

joanhey:
It's there:
https://github.com/TechEmpower/FrameworkBenchmarks/pull/10041/files#diff-ded17ef494703a5082d5b4388ca0133e8271495276405936e1b9ccad939ad1a8R13

image

image


And perhaps we can simplify the .rr.yaml with Octane.

roxblnfk:
I think Octane just overrides RR config. We can use https://github.com/roadrunner-php/laravel-bridge that more predictable and respects rr.yaml

joanhey:
Added the yaml file to Laravel Octane:

`ENTRYPOINT ["php", "artisan", "octane:roadrunner", "--host=0.0.0.0", "--port=8080", "--rr-config=/laravel/deploy/roadrunner/.rr.yaml"]`

And It's loading the file.

Why rr don't return a Server Header by default ??
And we don't need extra work to add it.

PD: by default all servers add the Server Header, and they only add an option to disable it.

roxblnfk:
Octane overrides middleware set there:
https://github.com/laravel/octane/blob/4a46a6c4886ea42e4fff4aa781497121083d5e46/src/Commands/StartRoadRunnerCommand.php#L92

Now need to understand how to set config `octane.roadrunner.http_middleware`

> Why rr don't return a Server Header by default ??

Security policy

roxblnfk:
Also note [there is](https://github.com/laravel/octane/blob/4a46a6c4886ea42e4fff4aa781497121083d5e46/src/Commands/StartRoadRunnerCommand.php#L88) `max-requests` option. It's bad option BTW, reloading workers like this may decrease performance

Also it doesn't use pipes.
Also [`debug` log-level](https://github.com/laravel/octane/blob/4a46a6c4886ea42e4fff4aa781497121083d5e46/src/Commands/StartRoadRunnerCommand.php#L94) is very slow, make sure it set to warn or error
`logs.output` should be stderr. I have no idea how to configure it

joanhey:
Yes I need to change `max-requests` as I did with Swoole.
`"--max-requests=10000"`

joanhey:
I added the log level in the .rr.yaml file, but perhaps Octane override it.

https://github.com/TechEmpower/FrameworkBenchmarks/pull/10041/files#diff-ded17ef494703a5082d5b4388ca0133e8271495276405936e1b9ccad939ad1a8R10-R12

image


PD: and yes it's better to show the errors in stderr, so we can see it in the runs.

roxblnfk:
Have you reconsidered using [roadrunner-php/laravel-bridge](https://github.com/roadrunner-php/laravel-bridge) 😄? This bridge has a redesigned worker startup process, but under the hood, it uses Octane to reset state.

joanhey:
Bypass `octane:start` and use `rr serve` directly. 
So It's working OK and pass all the tests.

image


@roxblnfk feel free to add or suggest any change in .rr.yaml

https://github.com/TechEmpower/FrameworkBenchmarks/blob/95546ab43b892a7c1ae28a2fedeb5e51f1a40186/frameworks/PHP/laravel/deploy/roadrunner/.rr.yaml

roxblnfk:
> feel free to add or suggest any change in .rr.yaml

the same recommendations as [before](https://github.com/TechEmpower/FrameworkBenchmarks/pull/10041#issuecomment-3167039556).

Additionally, depending on the nature of the load (CPU/IO), you could try adding gzip middleware: less load on IO, more on CPU.

joanhey:
Debug level is now:
```
logs:
  mode: production
  level: error
  output: stdout
  #encoding: json
```

I don't know how to change the `max-requests` in .rr.yaml.

This bench requirements:
https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#requirements

gzip compression is not permitted.

frameworks/PHP/laravel/deploy/roadrunner/.rr.yaml
```diff
@@ -1,23 +1,35 @@
...
-logs:
-  mode: production
-  level: error
 http:
   address: 0.0.0.0:8080
-  middleware: ["headers"]
   pool:
-    #max_jobs: 64 # feel free to change this
+    num_workers: 0
+    max_jobs: 500 # feel free to change this
```
roxblnfk: `max-requests` is here

joanhey: Now I'll change to `max_jobs: 0` if later fail in the runs we'll change it again.

joanhey:
Update in #10053

PR #10058, opened 2025-08-17
Adds support for Hypervel

silnex:
Adds support for [Hypervel](https://hypervel.org/)


silnex:
I deleted about half of the files!

PR #10064, opened 2025-08-21
Fix postgrest

cptwunderlich:
Updating postgrest to fix broken plaintext and json routes.


cptwunderlich:
I also looked at the "updates" benchmark and it should work in theory. However, your benchmark calls that endpoint as a GET, even though it modifies the resource. 
PostgREST is creating a REST API based on your DB schema. And for GET requests, it sets a READ ONLY transaction. So updates fail. Using e.g. POST would work.

cptwunderlich:
Oops, I realized the fortunes benchmark is also broken. Will add that tomorrow.

cptwunderlich:
Locally, I get a failure for "fortunes": 

> Only 506 rows read in the database out of roughly 6144 expected.

I'm not sure what's happening there though.

frameworks/Haskell/postgrest/src/fortunes.sql
```diff
@@ -1,35 +1,18 @@
...
-DECLARE
-   fortunes fortune_t[];
-BEGIN
-   SET LOCAL "response.headers" = '[{"Content-Type": "text/html"}]';
-   SELECT array_agg(CAST((id,message) AS fortune_t)) FROM "Fortunes" INTO fortunes;
-   RETURN convert_to(fortunes_template(fortunes), 'UTF8');
-END
-$$ language plpgsql volatile;
+create or replace function fortunes() returns "text/html" as $$
+   select set_config('response.headers', '[{"Content-Type": "text/html; charset=utf-8"}]', true);
```
steve-chavez, line 14: Hey @cptwunderlich, 

This line shouldn't be necessary. PostgREST will automatically set the content type to the `"text/html"` domain.

cptwunderlich: Well, this adds the type "text/html", but the benchmark harness expects an explicit value of "text/html; charset=UTF-8".
If you remove it, the test fails with:

```
   FAIL for http://tfb-server:3000/rpc/fortunes
     Invalid Content-Type header, found "text/html", did not match "^text/html; ?charset=(UTF|utf)-8$".
```

I can add a comment to clarify.

There are some oddities in this benchmark. E.g., the "updates" benchmark will never work for postgREST, bc. they use a "GET" request for that and you just get a read-only transaction for that in postgrest....

steve-chavez: > Well, this adds the type "text/html", but the benchmark harness expects an explicit value of "text/html; charset=UTF-8".

Got it, thanks for clarifying.

> the "updates" benchmark will never work for postgREST, bc. they use a "GET" request for that and you just get a read-only transaction for that in postgrest....

Yeah, that is weird on how they landed on GET for this. There are no escape hatches on PostgREST for doing a write in a GET.

steve-chavez: Maybe the `update` benchmark should just be omitted

cptwunderlich: It is, it's not in the config.

PR #10090, opened 2025-09-03
Remove ENV RUBY_MN_THREADS=1

schneems:
Reported in https://github.com/puma/puma/issues/3715#issuecomment-3245560096. Setting this value tanks the performance of Puma 7.0.0.pre1. Independently, there are other reports that there are some bugs with this feature

- https://bugs.ruby-lang.org/issues/20638
- https://bugs.ruby-lang.org/issues/19996
- https://github.com/puma/puma/issues/3715#issuecomment-3247114128

I also work at Heroku where I see many many (cannot say a number, but it's a lot) of production Ruby apps, and I don't think I've ever seen one use this in production (not saying it doesn't happen, but that it's not the default).


p8:
This seems like a good idea to me.

msmith-techempower (approved):

PR #10094, opened 2025-09-07
[ruby/sinatra] Use CGI.escape_html

p8:
CGI.escape_html seems more performant.


p8:
@msmith-techempower could some of these PR's get merged? In a couple of hours the next run will start and we'll have to wait 8 days to see if the performance improved.

p8:
Thanks @msmith-techempower !

PR #10114, opened 2025-09-15
Photon - a D programming language fiber based framework

DmitryOlshansky:
This introduces a new bare-bones framework photon-http that is used in combination with photon fiber scheduler for high scalability. I'm not sure which classification to use - stripped or realistic?
Only plaintext and json endpoints are implemented so far.

### DIFF
--- a/frameworks/D/photon-http/benchmark_config.json
+++ b/frameworks/D/photon-http/benchmark_config.json
@@ -6,7 +6,7 @@
         "json_url": "/json",
         "plaintext_url": "/plaintext",
         "port": 8080,
-        "approach": "Stripped",
+        "approach": "Realistic",
         "classification": "Platform",
         "database": "None",
         "framework": "None",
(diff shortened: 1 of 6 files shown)


frameworks/D/photon-http/benchmark_config.json
```diff
@@ -0,0 +1,26 @@
+{
+  "framework": "photon-http",
+  "tests": [
+    {
+      "default": {
+        "json_url": "/json",
+        "plaintext_url": "/plaintext",
+        "port": 8080,
+        "approach": "Stripped",
```
msmith-techempower: From the [rules (#2)](https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#general-test-requirements):

> Implementations that are not based on a realistic HTTP implementation will be marked as Stripped. Rather than stipulate a rigid specification of HTTP requirements, we defer to common sense: the implementation should provide a viable subset of HTTP to reasonably unbiased outsiders familiar with HTTP.

After reviewing the code, I would say this appears to be 'Realistic'.

DmitryOlshansky: Okay! Switched to Realistic

Issue #10047, opened 2025-08-11
The tfb-status.techempower.com is slower every time


joanhey:
Some days ago had problems with a 502 errors.

But if we check the actual runs:

<img width="1721" height="899" alt="Image" src="https://github.com/user-attachments/assets/d4dde0de-af0a-4b72-97e6-9517ff4fe5e9" />

With only 4 frameworks tests more, the time is +40h more. I checked the new changes, and I didn't see anything incorrect or than they need that extra time.

And each run is slower.
So I think that is a problem in the tfb-server.

Please anybody from Techempower could check it ??

Thank you.

p8:
I'm seeing big increases in build_times using the script from https://github.com/TechEmpower/FrameworkBenchmarks/issues/8131

From 17 hours:
```
python3 tfb.py -f results.2025-07-25-13-52-41-279.zip 
Number of frameworks: 325
Number of tests: 849
Total time: 5 days, 19:08:18 (included 0.92 hours of sleep)
Total time: 138.22 + 0.92 = 139.14 hours
Total build  time: 17.47 hours
Total verify time: 118.11 hours
...
0:21:39 (0:12:27 + 0:09:03) drogon-core
...
0:06:51 (0:02:47 + 0:03:59) cutelyst
```

To 70 hours:
```
python3 tfb.py -f results.2025-08-10-03-35-20-578.zip
Number of frameworks: 326
Number of tests: 852
Total time: 7 days, 22:38:24 (included 0.93 hours of sleep)
Total time: 189.71 + 0.93 = 190.64 hours
Total build  time: 70.30 hours
Total verify time: 112.11 hours
....
0:54:05 (0:44:53 + 0:09:03) drogon-core
...
0:41:34 (0:37:30 + 0:03:59) cutelyst
```

p8:
Maybe some networking issue?
```
...
drogon-core: Fetched 155 MB in 15min 2s (172 kB/s)
...
```

Probably unrelated but I'm not sure why Drogon needs to install 8 versions of PostgreSQL?
```
drogon-core: Step 21/42 : RUN sudo apt -y install postgresql-server-dev-all
drogon-core:  ---> Running in 2e77df1ce458
drogon-core: 
drogon-core: WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
drogon-core: Reading package lists...
drogon-core: Building dependency tree...
drogon-core: Reading state information...
drogon-core: The following additional packages will be installed:
[...]

PR #10125, opened 2025-09-24
H2O: Track image build timings

volyrique:
After run [b4d86ad8-764f-47a1-a9b1-83c71bd8b6e9](https://tfb-status.techempower.com/results/b4d86ad8-764f-47a1-a9b1-83c71bd8b6e9) and starting from [58533c19-522b-4202-84b8-416f105fc8ee](https://tfb-status.techempower.com/results/58533c19-522b-4202-84b8-416f105fc8ee) the build time jumped from 5:13 to 45:05 (and stayed that way ever since) with [changes](https://github.com/TechEmpower/FrameworkBenchmarks/compare/3877be38816130223677a2b421e102215356456c...b749f1f8cf72fbc22a74a91e09170fa703c8a8dd) in neither the `h2o` test implementation nor the toolset; the machines in the Citrine environment haven't been restarted as well. So, let's try to find out where exactly the majority of the build time is spent.

### DIFF
--- a/frameworks/C/h2o/h2o.dockerfile
+++ b/frameworks/C/h2o/h2o.dockerfile
@@ -4,6 +4,7 @@ ARG H2O_APP_PREFIX=/opt/h2o-app
 
 FROM "ubuntu:${UBUNTU_VERSION}" AS compile
 
+RUN echo "[timing] Installing system packages: $(date)"
 ARG DEBIAN_FRONTEND=noninteractive
 RUN apt-get -yqq update && \
     apt-get -yqq install \
@@ -31,6 +32,7 @@ RUN apt-get -yqq update && \
       ruby \
       systemtap-sdt-dev
 
+RUN echo "[timing] Building H2O: $(date)"
 ARG H2O_VERSION=3b9b6a53cac8bcc6a25fb28df81ad295fc5f9402
 
 WORKDIR /tmp/h2o-build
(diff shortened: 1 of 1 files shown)


p8:
There are a lot frameworks that have increased build times. This mostly seems a networking issue.
See: https://github.com/TechEmpower/FrameworkBenchmarks/issues/10047

msmith-techempower:
Agreed - not clear what caused that 585-run to go off the rails, but it DOES appear to have been resolved after I brought everything down for some maintenance and kicked off anew.

Happy to include your build times logging, though.

volyrique:
Let me preface this by stating that I am talking about the `h2o` test implementation only. With that in mind, I respectfully disagree, @msmith-techempower. The build time during the last successful run, [278bd157-09fc-42de-b767-7c060ffeb15a](https://tfb-status.techempower.com/results/278bd157-09fc-42de-b767-7c060ffeb15a), was still roughly the same, i.e. 44:43, and while I have made some changes to the code in the meantime, IMHO they don't justify that. Furthermore, there are still no updates to the toolset, neither does it make sense that the environment changes (e.g. the kernel upgrade) have such an effect. Also as a counterpoint, the build time in GitHub Actions has stayed below 5 minutes, which is in an environment with only 4 processors!

I agree that the cause is most probably a network issue, but I suspect that Dockerhub is throttling download speeds, given that the Citrine environment is hammering it every 15 minutes or so. Let's see what the logs tell us.

PR #10139, opened 2025-09-28
Update Django

adamghill:
- Update Django version and other dependencies to their latest versions
- Use Python 3.13
- Use `orjson` instead of `ujson`
- Format all code with `ruff`
- Fix incorrect webserver classification from `Meinheld` to `Gunicorn`
- Use `StreamingHttpResponse` for plaintext endpoint


msmith-techempower:
Closing for grooming - feel free to reopen if you continue working on it.

PR #10148, opened 2025-10-06
Update photon-http to implement http pipelining

DmitryOlshansky:
This is mostly just a version bump. The real work is being done by photon-http v0.6.0 that now supports request pipelining.


dejlek:
I am excited about this one! :)

PR #10180, opened 2025-10-15
Bump Wired.IO to 9.5.3, supports pipelined http. Add platform benchmark

MDA2AV:
**[Wired.IO - Fullstack]**
 - Upgrading to Wired.IO 9.5.3 - Added HTTP/1.1 pipeline support and multi-segmented request buffers for extra large requests.

**[Wired.IO - Platform]**
 - Adding Wired.IO platform benchmark for performance comparison between .NET NetworkStream and raw epoll syscalls. 
 - This test does not use Wired.IO engine for request parsing, instead uses a very naive parser from https://github.com/MDA2AV/Unhinged.
 - In .NET, all frameworks are built on top of .NET Socket class, this benchmark is an attempt to measure a lower level implementation peformance by interoping directly with epoll. The http parsing engine will be improved in the future, regardless it has little to zero impact on these benchmarks' results.

Both benchmarks are Native.AOT

### DIFF
--- a/frameworks/CSharp/wiredio/src/Fullstack/Program.cs
+++ b/frameworks/CSharp/wiredio/src/Fullstack/Program.cs
@@ -14,7 +14,7 @@ internal class Program
         await expressBuilder
             .Port(8080)
             .NoScopedEndpoints()
-            .MapGet("/json", scope => async ctx =>
+            .MapGet("/json", _ => async ctx =>
             {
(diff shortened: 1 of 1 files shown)


Kaliumhexacyanoferrat:
Ah, will remove this from #10184 then.

msmith-techempower (approved):

PR #10194, opened 2025-10-20
Update PostgreSQL to version 18

volyrique:
Also, disable password-based authentication using the MD5 hash algorithm, since that functionality has been [officially deprecated](https://www.postgresql.org/docs/18/auth-password.html).

Fix: #8061

### DIFF
--- a/toolset/databases/postgres/postgres.dockerfile
+++ b/toolset/databases/postgres/postgres.dockerfile
@@ -1,9 +1,7 @@
-FROM postgres:17-bookworm
+FROM postgres:18-trixie
 
 ENV PGDATA=/ssd/postgresql \
     POSTGRES_DB=hello_world \
-    POSTGRES_HOST_AUTH_METHOD=md5 \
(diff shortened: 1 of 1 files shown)


volyrique:
As all toolset changes, this PR is going to fail the CI checks because it triggers tests with all implementations, especially the ones that are known to be broken based on the latest continuous benchmarking run and/or those that don't use PostgreSQL at all. We also have the issue with the test runners running out of disk space - refer to the PHP checks, for example. On the other hand, all C and Ruby tests succeed.

With all that said, some tests are expected to start failing due to the authentication changes. My reading of issue #8061 and especially what the TechEmpower team has said is that the breakage would be acceptable, given that framework authors have had 2 years and a half (since the issue has been opened) to update their code and that password-based authentication using the MD5 hash algorithm is now officially deprecated. Let me know if I am wrong.

volyrique:
FYI @msmith-techempower, in case you haven't looked at this PR because of the CI failures.

PR #10217, opened 2025-10-28
[ruby/rails] Update to Rails 8.1

p8:


p8:
@msmith-techempower I've rebased on master.

PR #10241, opened 2025-11-04
Update .NET version to 10.0 for CSharp/aspnetcore

sebastienros:


sebastienros:
@DamianEdwards @LoopedBard3

sebastienros:
@DamianEdwards 

https://mcr.microsoft.com/en-us/artifact/mar/dotnet/sdk/tag/10.0
https://mcr.microsoft.com/en-us/artifact/mar/dotnet/aspnet/tag/10.0

LoopedBard3 (approved):

DamianEdwards:
@sebastienros those 404 for me. What was your point?

sebastienros:
The tag:10 images are available, and I copy pasted the urls ... they use some weird routing framework and the url they generate is not accessible directly ;)

DamianEdwards:
@sebastienros I see, so they point at the rc.2 version right now?

msmith-techempower:
I don't really see why the UPDATES test would be suddenly failing due to these changes, unless there's a built-in ORM that got updated with 10.0.

sebastienros:
@msmith-techempower I actually ran a fake change in master and the same issues are showing up. We will investigate, the two reasons I can think of is that the verification logic changed, or something happened in a service change of DotNet that the docker tag would have pulled automatically.

I checked our implementation, which is not using an advanced orm (no caching or optimizations) and it looks correct. So a mystery so far.

Kaliumhexacyanoferrat:
@sebastienros maybe you can try updating 

```xml
 <PackageReference Include="Npgsql" Version="9.0.2" />
```

to version `10.0.0-rc.1`. I had very similar issues with my framework, even though the EF adapter of postgres is used there.

sebastienros:
@Kaliumhexacyanoferrat thanks for the suggestion. I updated the package version but it didn't help.

amarant:
[Npgsql 10 is out now](https://www.nuget.org/packages/Npgsql#versions-body-tab)

sebastienros:
After a lot of work locally it finally works. Not touching anything else, the updates verification is super sensitive.

frameworks/CSharp/aspnetcore/src/Platform/Data/RawDbMySqlConnector.cs
```diff
@@ -144,23 +136,34 @@ public async Task<World[]> LoadMultipleUpdatesRows(int count)
     {
         var results = new World[count];
 
+        var ids = new int[count];
+        for (var i = 0; i < count; i++)
+        {
+            ids[i] = Random.Shared.Next(1, 10001);
+        }
+        Array.Sort(ids);
```
eerhardt, line 144: Why doesn't this one have the

`// Ensure unique ids by incrementing duplicates`

logic?

sebastienros: I assume the validation logic doesn't apply for mysql, only postgresql. The code is really to prevent the validate from blocking the PR by preventing any possibility of a lower number of update statements (in case there is an id collision).

sebastienros:
@msmith-techempower I thought I had cc'ed you but maybe I didn't send the message. Not time sensitive.

msmith-techempower (approved):

PR #10251, opened 2025-11-06
update feat version

smthing:


ruroru:
I am not a maintainer, but can you change jdk to amazoncorretto:25, I am pretty sure that openjdk ea (early access) images will be removed at some point as well.

smthing:
Sure, I've completed the modification of replacing the JDK with amazoncorretto:25 as you suggested.

PR #10253, opened 2025-11-06
toolset: Reduce the packages installed by APT

volyrique:
Make sure that APT does not install anything unnecessary.


volyrique:
@msmith-techempower, as usual, a toolset PR like this one is going to fail the checks. However, all Ruby tests pass, which covers both MySQL and PostgreSQL. Furthermore, the JavaScript `nodejs-mongodb-raw` test (among others) passes as well, which covers MongoDB, so overall I think that it is safe to merge.

toolset/wrk/wrk.dockerfile
```diff
@@ -4,11 +4,7 @@ FROM ubuntu:24.04
 COPY concurrency.sh pipeline.lua pipeline.sh query.sh ./
 
 ARG DEBIAN_FRONTEND=noninteractive
-RUN apt-get -yqq update >/dev/null && \
-    apt-get -yqq install >/dev/null \
-      curl \
-      wrk && \
-    chmod 777 concurrency.sh pipeline.sh query.sh
+RUN apt-get install --no-install-recommends -qqUy curl wrk > /dev/null
```
msmith-techempower, line 7: Why are we dropping the `update` call? Does the `-qqUy` shorthand it?

volyrique: Yes, it does, the `-U` parameter in particular.

PR #10268, opened 2025-11-11
Update the set of broken implementations part 2

volyrique:
The list is based on the output of the [tfb-fail-detector.py](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/scripts/tfb-fail-detector.py) script, however I [modified](https://github.com/TechEmpower/FrameworkBenchmarks/blob/65fc874dad60351b8634bba3d8442da12b0522db/scripts/tfb-fail-detector.py#L17) it to consider 11 continuous benchmarking runs (roughly 3 months back) instead of the default 3. Also, I excluded implementations that had been affected by recent commit activity - many haven't seen any changes in years.

### DIFF
--- a/frameworks/C++/ffead-cpp/benchmark_config.json
+++ b/frameworks/C++/ffead-cpp/benchmark_config.json
@@ -198,7 +198,7 @@
 			"display_name": "ffead-cpp [pg-raw-async-prof-pool-m]",
 			"notes": "async memory profiled",
 			"versus": "",
-			"tags": []
+			"tags": ["broken"]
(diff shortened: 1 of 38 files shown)


volyrique:
@msmith-techempower, all test failures are in framework permutations that I didn't touch, but the CI scripts are not clever enough to skip them. Also, note that all variants of `gemini` have been failing since mid-July.

msmith-techempower (approved):

p8:
This is great! Thanks @volyrique !

PR #10282, opened 2025-11-18
[PHP] Add curl, unzip to reduce package installation time.

kitrio:
Warning message from composer in docker container:
```
Composer is operating significantly slower than normal because you do not have the PHP curl extension enabled.
```
so, added curl extension, unzip

~~mark: No composer.lock file present. Updating dependencies to latest instead of installing from lock file. See https://getcomposer.org/install for more information.~~

* Affected framework kumbia, mark, phalcon, laravel-eloquent,reactphp,yii2

### DIFF
--- a/frameworks/PHP/ubiquity/ubiquity-ngx-micro.dockerfile
+++ b/frameworks/PHP/ubiquity/ubiquity-ngx-micro.dockerfile
@@ -9,7 +9,7 @@ RUN LC_ALL=C.UTF-8 add-apt-repository ppa:ondrej/php > /dev/null
 RUN apt-get update -yqq > /dev/null && \
     apt-get install -yqq wget git unzip libxml2-dev cmake make systemtap-sdt-dev \
                     zlib1g-dev libpcre3-dev libargon2-dev libsodium-dev \
-                    php8.3-cli php8.3-curl php8.3-dev php8.3-mbstring libphp8.3-embed nginx > /dev/null
+                    php8.3-cli php8.3-dev php8.3-mbstring libphp8.3-embed nginx > /dev/null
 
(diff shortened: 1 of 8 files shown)


WyriHaximus (approved):
👍

joanhey (commented):
Yes, It's OK this change.
But what we gain with composer, we loose it with docker build downloading and installing 2 more packages.

PD: if you have installed PHP, please add the 2 packages.

joanhey:
About the no `composer.lock` file:
The bench is running continuously, each ~190 hours a new run, so we can check any issues with minor version update of dependencies and performance changes.

The bench don't permit use `latest` for versions (that I think is correct).
For example, with the PHP version we chose the MAJOR.MINOR never the PATCH, PHP 8.3 8.4 8.5 ...  so in each run if a new patch version is ready, the bench use it without any work for us.

The same happens with the `composer.json`:
`"laravel/framework": "^12"`, `"cakephp/cakephp": "^5.0.1"`, `"codeigniter4/framework": "^4.0"`, ...

All use SEM versioning, so they need to work with .minor versions. If the bench fail is a bug in the framework (some times happened) and we can create a issue.

So all is working in a secure way, only permitting minor versions.
Imagine the work if we need to update any minor or patch version of PHP and each framework :disappointed_relieved:, when we can have it for free and check performance changes. So the bench never use outdated versions.

PD: in your production apps, please use always `composer.lock` to avoid problems.

joanhey:
I think that nobody understand what I said before in this PR !!

kitrio:
> I think that nobody understand what I said before in this PR !!

Sorry  comment about "composer.lock " is wrong

curl, unzip  add  is related composer message
```
Composer is operating significantly slower than normal because you do not have the PHP curl extension enabled.
```

PR #10286, opened 2025-11-19
toolset: Fix the --force-rm option

volyrique:
Also, clean up the `wrk` Dockerfile.

@msmith-techempower, I believe that the intention was to have these changes in PR #9727, but they were missed. Furthermore, they are probably the reason why the current continuous benchmarking run seems stuck.

### DIFF
--- a/toolset/continuous/tfb-startup.sh
+++ b/toolset/continuous/tfb-startup.sh
@@ -47,7 +47,6 @@ docker run \
   --results-environment "$TFB_ENVIRONMENT" \
   --results-upload-uri "$TFB_UPLOAD_URI" \
   $(if [ "$TFB_RUN_ORDER" = "reverse" ]; then echo "--reverse-order"; fi) \
-  --force-rm \
   --quiet
 
 echo "zipping the results"


msmith-techempower:
> ... current continuous benchmarking run seems stuck.

Actually, that's very likely due to a recent power outage. I'll get it going again.

toolset/run-tests.py
```diff
@@ -210,7 +210,7 @@ def main(argv=None):
         help='Extra docker arguments to be passed to the test container')
     parser.add_argument(
         '--force-rm',
-        default=False,
+        action='store_true',
```
msmith-techempower, line 213: I don't understand this one - should be a boolean flag, no?

toolset/wrk/wrk.dockerfile
```diff
@@ -1,17 +1,7 @@
...
-ENV accept=accept \
-    duration=duration \
-    levels=levels \
-    max_concurrency=max_concurrency \
-    max_threads=max_threads \
-    name=name \
-    pipeline=pipeline \
-    server_host=server_host
+# Required scripts for benchmarking
+COPY concurrency.sh pipeline.lua pipeline.sh query.sh ./
```
msmith-techempower, line 7: Still need the env vars to pass from the TFB client to the wrk client.

toolset/continuous/tfb-startup.sh
```diff
@@ -47,6 +47,7 @@ docker run \
   --results-environment "$TFB_ENVIRONMENT" \
   --results-upload-uri "$TFB_UPLOAD_URI" \
   $(if [ "$TFB_RUN_ORDER" = "reverse" ]; then echo "--reverse-order"; fi) \
+  --force-rm \
```
msmith-techempower: I don't think we want to do this - some frameworks <coughing loudly at Rust> have, frankly, ridiculous build times and we rely heavily on caching that work.

.github/workflows/build.yml
```diff
@@ -155,7 +155,7 @@ jobs:
           # run-ci.py runs the diffing to see if github actions needs to test this framework. Ideally/eventually,
           # we'd like to try and do the diffing before github_actions_clean & setup.
           # This will run the tests exactly as you would in your own vm:
-          docker network create tfb > /dev/null 2>&1 && docker run --network=tfb -e USER_ID=$(id -u) -e CI=true -v /var/run/docker.sock:/var/run/docker.sock --mount type=bind,source=`pwd`,target=/FrameworkBenchmarks techempower/tfb --mode verify --test-dir $RUN_TESTS --results-environment Github-Actions;
+          docker network create tfb > /dev/null 2>&1 && docker run --network=tfb -e USER_ID=$(id -u) -e CI=true -v /var/run/docker.sock:/var/run/docker.sock --mount type=bind,source=`pwd`,target=/FrameworkBenchmarks techempower/tfb --mode verify --force-rm --test-dir $RUN_TESTS --results-environment Github-Actions;
```
msmith-techempower, line 158: I'm not convinced this will have an effect, in practice. Each Github Action runner should be its own isolated system, so every invocation of this script should start with nothing to remove.

toolset/run-tests.py
volyrique: Yes, it should be and my change enforces it. Previously, if the option was specified on the command line, `args.force_rm` was set to the string parameter of this argument, which sort of worked due to Python's weak typing, I suppose (i.e. a non-empty string evaluates to `True`). `action='store_true'` [implies](https://docs.python.org/3/library/argparse.html#action) a `False` default as well.

Compare with line 85 of this script, that is the `--quiet` argument.

toolset/wrk/wrk.dockerfile
volyrique: That happens [here](https://github.com/TechEmpower/FrameworkBenchmarks/blob/fe1810bf707d54d4616001d879f38c013a4f59a6/toolset/utils/docker_helper.py#L439). These nonsensical defaults are unnecessary.

toolset/continuous/tfb-startup.sh
volyrique: In the previous PR you [wrote](https://github.com/TechEmpower/FrameworkBenchmarks/pull/9727#discussion_r2012483036):
> ... and we need to update our CI scripts to forcibly set to True (and our continuous scripts).

Also, lack of caching was the status quo until the last run. If you have changed your mind, I will be happy to update my patch.

.github/workflows/build.yml
volyrique: See [my other comment](https://github.com/TechEmpower/FrameworkBenchmarks/pull/10286#discussion_r2543657675) - I assumed that this was what you meant with `our CI scripts`.

toolset/run-tests.py
msmith-techempower: Sorry, I meant that I had intended the default to be `False` - if the way to enable it is with the string 'store_true' and you are changing the default to that, I am asking for the rationale.

toolset/wrk/wrk.dockerfile
msmith-techempower: AH, I see. These were being included at build time rather than injected at runtime.

I probably did this, but there's at least a slim chance there's a legacy reason why they were there. If it works, it works.

toolset/continuous/tfb-startup.sh
msmith-techempower: I'm not sure if my memory is completely correct, but previously we ended up _wanting_ to leverage the caches between runs (so much so, that our script specifically checks if we're nearing the disk space limit before doing the `system prune`) due to huge run times.

To confirm - this flag will only remove unused intermediate layers (read: not impact benchmarking run times)?

.github/workflows/build.yml
msmith-techempower: AH. Yeah, I probably didn't fully appreciate how these changes would impact everything.

This is... _at worst_ innocuous. I guess I'm okay with it.

toolset/run-tests.py
volyrique: The default remains `False` after my changes. `action='store_true'` ensures that we store a boolean into `args.force_rm` instead of a string. Also, it makes the usage of the parameter more ergonomic because one can just specify `--force-rm` instead of `--force-rm=something`.

toolset/continuous/tfb-startup.sh
volyrique: Correct, but it sounds like you have changed your mind, so I reverted this specific modification. However, I left the equivalent change in the CI scripts because we run out of disk space when testing all the Rust implementations, for instance, so it might help a bit.

Issue #10303, opened 2025-11-23
PHP 8.5 update [info]


joanhey:
PHP 8.5 released 20 Nov 2025.
https://www.php.net/releases/8.5/es.php

PHP update versions info:

- https://github.com/TechEmpower/FrameworkBenchmarks/issues/9408
- https://github.com/TechEmpower/FrameworkBenchmarks/issues/8579
- https://github.com/TechEmpower/FrameworkBenchmarks/issues/7703
- https://github.com/TechEmpower/FrameworkBenchmarks/issues/6894
- https://github.com/TechEmpower/FrameworkBenchmarks/issues/6184

## Fail

- [ ] Nginx Unit still not available PHP 8.5
  EDIT: Unit is deprecated and unsupported from October 2025. So no official PHP 8.5 version.
  #10764
- [x] Swoole ~still not available PHP 8.5~
- [ ] Phalcon still not available PHP 8.5
- [ ] Ngx-php still not avilable PHP 8.5
- [x] Fat-free  `Non-canonical cast (binary) is deprecated` `Constant PDO::MYSQL_ATTR_INIT_COMMAND is deprecated`  https://github.com/bcosca/fatfree/issues/1296
- [x] Mako ~`upstream sent duplicate header line: "Content-Length: 236", previous value: "Content-Length: 31" while reading response header from upstream`~ Update to Mako 11
- [ ] Ubiquity work but with a lot of deprecated warnings
- [ ] Ripple work but with deprecated warnings
- [x] CakePHP work but with deprecated warnings with PDO constants

## Changes
- OPCache is installed by default
- Constant `DATE_RFC7231` deprecated
- Constant `PDO::MYSQL_ATTR_INIT_COMMAND` deprecated
- Constant `PDO::MYSQL_ATTR_SSL_CA` deprecated (all PDO constants)
- `ReflectionProperty::setAccessible()` deprecated

NateBrady23:
Closing as part of bulk cleanup.

PR #10304, opened 2025-11-23
[php] Update plain PHP to 8.5

joanhey:
PHP 8.5 released 20 Nov 2025.
https://www.php.net/releases/8.5/es.php

Nginx Unit still not avaiable PHP8.5

#10303


joanhey:
@NateBrady23 nice to see you again !!
All OK ?

PR #10335, opened 2025-11-26
[ruby/grape] Use Iodine instead of Unicorn

p8:
Unicorn is no longer maintained.
Replace with Iodine, as it performs better in these benchmarks.


p8:
@msmith-techempower Updated!

PR #10336, opened 2025-11-26
[ruby/hanami] Update Hanami to 2.3

p8:


p8:
@msmith-techempower Updated!

PR #10378, opened 2025-12-08
toolset: Tune the AIO parameters for PostgreSQL 18

volyrique:


fakeshadow:
I believe async IO in pg 18 is for certain disk IO only. It's hard to imagine it would have any impact to the benchmark as almost all of the test tables can fit into PG's shared buffer memory. And even if disk IO happens the benefit would be tiny due to high perf of local SSD

volyrique:
Well, first of all, if you look at the `h2o` and the `may-minihttp` results, you will notice that there was a small performance regression after the upgrade to PostgreSQL 18 (starting with run `40103b8c-8670-4734-8d39-7081cf9441d8`). Now, the blog post that is linked to from the configuration file mentions that the default number of I/O workers (i.e. 3) is too conservative, especially on larger machines, and that having an excess of workers is not much of an issue, so I don't think that it would hurt to try. Hard data is always better than just conjectures.

@msmith-techempower, all C, Go, and Ruby tests pass, so I think that this PR is good to go.

PR #10384, opened 2025-12-10
Increase `io.netty.iouring.ringSize` for some Vert.x benchmarks that use io_uring

ShreckYe:
Found this while working on #10383. I see the following warnings when running these benchmarks on my device:

```
io.netty.channel.uring.IoUringIoHandler processCompletionsAndHandleOverflow
vertx: WARNING: CompletionQueue overflow detected, consider increasing size: 4096
```

```
[vert.x-acceptor-thread-0] WARN  io.netty.channel.uring.IoUringIoHandler - CompletionQueue overflow detected, consider increasing size: 8192
```

After these changes, the warnings are gone and the plaintext result with 32 threads and 16384 connections of the `vertx` portion has improved about 50% as run on my device.

While the results may vary from machine to machine, increasing the size can't reduce performance as it seems to me, so it might be good to set a large enough value to ensure best results during debugging on difference devices and prevent such warnings from showing.


ShreckYe:
@vietj @awmcc90 Please see if these changes are OK to you.

PR #10400, opened 2025-12-14
Fix SQL capitalization and format in all the Vert.x benchmarks other than `vertx-web-kotlinx`

ShreckYe:
As in #10399 but for all the other Vert.x benchmarks.


ShreckYe:
@vietj ping

ShreckYe:
`./tfb --test vertx-postgres --mode verify` passes on my device. Don't know why it fails in the CI.

volyrique:
It doesn't; `vertx-web-postgres` fails.

ShreckYe:
Sorry I misread. So it looks like the failed ones are those which were already broken before this PR.

From my experience working on the `vertx-web-kotlinx` portion they are likely caused by the OnGres SCARM client dependency missing. If needed I can try fixing them by either adding the dependency or updating to Vert.x 5.

volyrique:
My suggestion is either to fix them or tag them as broken. Anything else would just be wasting time and resources in the continuous benchmarking environment, so don't simply leave them as they are.

ShreckYe:
Fixed with #10407.

PR #10401, opened 2025-12-14
[ntex] optimization and fix

fakeshadow:
Optimize fortune test by reducing heap allocation. `Vec<&str>` allocates once per request while `&mut Vec<String>` allocates 12 times. `SmallVec` could be used to reduce it to zero too if desired.

Fix update test by reading the `randomNumber` field from the query result. According to [Requirements-4-vi](https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#requirements-4) "At least the randomNumber field must be read from the database result set."

### DIFF
--- a/frameworks/Rust/ntex/src/db.rs
+++ b/frameworks/Rust/ntex/src/db.rs
@@ -1,4 +1,4 @@
-use std::{borrow::Cow, cell::RefCell};
+use std::cell::RefCell;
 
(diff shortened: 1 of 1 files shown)


fakeshadow:
@fafhrd91 Please let me know if you are Ok with these changes. The update change could have a slight negative impact on score.

msmith-techempower (approved):

PR #10406, opened 2025-12-15
[php] Update to CodeIgniter 4.6.4 with optimizations and fixes

sk757a:


frameworks/PHP/codeigniter/public/index.php
```diff
@@ -1,16 +1,40 @@
...
+        'Your PHP version must be %s or higher to run CodeIgniter. Current version: %s',
+        $minPhpVersion,
+        PHP_VERSION,
+    );
+
+    header('HTTP/1.1 503 Service Unavailable.', true, 503);
+    echo $message;
+
+    exit(1);
+}
```
joanhey: This code is unnecessary in a bench. In reality the performance is worst.

joanhey (changes requested):

frameworks/PHP/codeigniter/public/index.php
sk757a: Yes, you're right. This file, like many others, is taken as is from the actual framework distribution.

joanhey:
Please, Can you explain the optimizations and fixes ??

PR #10407, opened 2025-12-16
Fix the broken Vert.x Web portions by updating to Vert.x 5, and update to Scala 3 and enable io_uring BTW

ShreckYe:
Update all dependencies and toolchains to the latest, migrate to Vert.x 5 and Scala 3, and use io_uring in the Vert.x Web portions of Java, Kotlin, and Scala.

Based on my experience, these portions were broken because the OnGres SCARM client dependency was missing and it seems required by newer versions of PostgresSQL. Vert.x 5 doesn't require adding this dependency explicitly anymore.

Important dependency and toolchain versions:
* Vert.x 5.0.5
* Java/JDK 25 in all 3 projects
* Kotlin 2.3.0-RC3 (to support Java 25)
* Scala 3.7.4
* latest stable build tool versions
  * Maven 3.9.11
  * Gradle 9.2.1
  * sbt 1.11.7

### DIFF
--- a/frameworks/Kotlin/vertx-web-kotlin-coroutines/build.gradle.kts
+++ b/frameworks/Kotlin/vertx-web-kotlin-coroutines/build.gradle.kts
@@ -10,7 +10,7 @@ tasks.wrapper {
 }
 
 plugins {
-    kotlin("jvm") version "2.3.0-RC3"
+    kotlin("jvm") version "2.3.0"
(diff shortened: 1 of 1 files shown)


ShreckYe:
@vietj @franz1981 @oliver-brm Would you mind taking a look when you have a chance? Thanks.

PR #10451, opened 2025-12-25
[js/ditsmod] update to node 24, bun 1.3

kitrio:
update node 20 -> 24 lts, bun 1.1 -> bun 1.3


KostyaTretyak:
Hi @kitrio. I believe you did this update so that the Ditsmod application could benefit from the new versions of Node.js and Bun.js, but it seems that there has been a pretty noticeable slowdown. Node.js by 30%, Bun.js by 50%.

But don't worry, I'll update Ditsmod soon, so maybe I'll update Node + Bun too.

PR #10456, opened 2025-12-26
Netty optimizations + fastjson2

doxlik:
Hello dear TechEmpower team.

Here I want to add alternative test based on Netty Java framework, but utilizing more aggressive optimizations similar to plaintext top tier players and also utilize fastjson2 lib for json serilization. 

I passed verification using "sudo ./tfb --test netty-fast --mode verify"

P.S. This test has no purpose to show that original netty test is bad, on the contrary, it based on it, but adds some maximum possible optimizations + alternative json library.


doxlik:
@franz1981 fyi, maybe you would like it, feel free for suggestions)

franz1981 (commented):
I won't do this.
Two reasons:
- this is very very unrealistic and not prod ready (it's not a secure http 1.1 parser)
- check how we do it right now using SWAR (and indexOf(byte)) in Netty itself

Netty is already low level and I see little benefits to show that if you reduce the parsing cost (increasing it's security risk exposure), it would cost less, since is an expected tradeoff of any framework (unless slow for unrelated reasons).

doxlik:
@franz1981 thank you for your point. Actually I was taking as baseline comparision with aspnetcore Platform which has many of similar optimizations and have 27 mill in latest bench while Netty having realistic varian 14 mil. And on my local machine also original netty test was much slower than kestrel test, then after this changes it became better both by throughput and latency. Also I had variant that was using original netty http codec but still using this aggregation and it had same througput but two times worse latency. 


And making a summary I fully agree that it is not prod ready solution, as original netty test, but at the same time it is just afford to show that netty framework also can be at top of plaintext if having similar to other solutions optimizations. So I agree with you that it is not somthing anyone would run in prod, but at the same time I don't think it worse than other top 10 in plaintext and thus it is interesting to have it in the board just as lab experiment as it yet match TFB rules as much as I checked.

doxlik:
I will think about utilizing the real parser to increase reality if I succeed I will push it here, if not will delete PR.

franz1981:
The real parser would work only if you have a bespoke beader map (likely)? Which won't hold any header string based obj

doxlik:
@franz1981 is it better now? Utilized official decoder but preserved custom response writing.

franz1981 (commented):
I haven't tried to debug it but tbh I don't such difference with the existing netty code..
I will take a better look while back from holidays next year 🙏

doxlik:
You know, I think you are right. Not so much changes, so better will close PR.

PR #10476, opened 2025-12-31
[csharp] Update GenHTTP to 10.4 and rework benchmark structure

Kaliumhexacyanoferrat:
- Switch from Entity Framework to raw connections (which now matches the db classification)
- Move and add unhinged and wired to the benchmark configuration
- Streamline the configuration for all tested engines


Kaliumhexacyanoferrat:
@MDA2AV thoughts?

PR #10488, opened 2026-01-02
[ruby] Set puma workers to 1.25 * nproc

p8:
WEB_CONCURRENCY=auto sets the number of workers to the number of processors. Setting it to 1.25 the number of processors seems to perform better.

### DIFF
--- a/frameworks/Ruby/hanami/hanami.dockerfile
+++ b/frameworks/Ruby/hanami/hanami.dockerfile
@@ -1,7 +1,6 @@
 FROM ruby:4.0
 
 ENV RUBY_YJIT_ENABLE=1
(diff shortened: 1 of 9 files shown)


joanhey:
Puma that is a multitasking (coroutines) need to work optimal with concurrency auto per CPU, if not, the problem is with Puma not with force the workers.

Sorry, Happy new year !!

p8:
@joanhey Happy new year!
Yes, I think there might be an issue with Puma here as well.
I'm trying too see if increasing the threads instead has the same improvement as increasing the workers:
https://github.com/TechEmpower/FrameworkBenchmarks/pull/10487

msmith-techempower (approved):

PR #10520, opened 2026-01-06
[java/baratine] Remove unmaintained Baratine

p8:
Baratine hasn't seen an update in 10 years: https://github.com/baratine/baratine

The benchmark also tests json and plaintext only, which is no longer allowed for new frameworks.


p8:
> @p8 is on his New Year's cleaning arc and I love it!

Thanks @msmith-techempower 😊

PR #10558, opened 2026-01-10
[php] Set jemalloc memory allocator

kitrio:
[https://psalm.dev/articles/psalm-6-docker](https://psalm.dev/articles/psalm-6-docker)
[https://blog.daniil.it/2025/07/10/official-psalm-docker-image/](https://blog.daniil.it/2025/07/10/official-psalm-docker-image/)
According to the psalm blog, specifying "jemalloc" in PHP 8.5  improves PHP performance
First test Laravel, Symfony

### DIFF
--- a/frameworks/PHP/laravel/laravel-workerman.dockerfile
+++ b/frameworks/PHP/laravel/laravel-workerman.dockerfile
@@ -3,8 +3,8 @@ FROM ubuntu:24.04
 ARG DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update -yqq && apt-get install -yqq software-properties-common > /dev/null
-RUN LC_ALL=C.UTF-8 add-apt-repository ppa:ondrej/php > /dev/null && \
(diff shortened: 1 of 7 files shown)


joanhey:
I hope so!!
But always is better check the results, than reading any article in any blog.

I have seen a lot of benchmarks, that are completely wrong.
A benchmark need to be correct, and with fair play with all the frameworks and languages.

joanhey:
I remember some years ago, about the symfony or phalcon benchs, they were completely bad.

With Phalcon, that I know the creator, I send 2 issues that said that the bench was completely wrong. And after my issues they closed the issues !!
I said that they send only the "Hello world!!" but the rest send also all the template 😤 and I finish 🤕 and the issues closed !!
Never believe any post in any blog, make your bench and check the results.

EDIT: https://github.com/phalcon-orphanage/framework-bench still the issues are closed for a benchmark 🫣
They are selling like the faster with a very bad benchmark, for that I started to create commits in this bench !!

msmith-techempower (approved):

PR #10590, opened 2026-01-20
[toolset] Update MySQL to v 9.4

joanhey:
The frameworks don't find the IP for  tfb-database !!
Issues: getaddrinfo for tfb-database failed,  getaddrinfo ENOTFOUND tfb-database, ...

I updated before MySQL, without this problem.

EDIT: update to v9.4, with v9.5 and v9.6 don't find the _tfb-database_ ip, investigating a solution for that.

### DIFF
--- a/toolset/databases/mysql/mysql.dockerfile
+++ b/toolset/databases/mysql/mysql.dockerfile
@@ -1,4 +1,4 @@
-FROM mysql:9.5
+FROM mysql:9.4
 
 ENV MYSQL_ROOT_PASSWORD=root
 ENV MYSQL_USER=benchmarkdbuser


joanhey:
Always will be slow MySQL with any big server !!
Anybody that want good performance with MySQL with a big server, like this one, need to pay for the Enterprise version.

But  anybody using a cloud or small vps server, the difference are minimal.

It's for that than before we have a cloud version of the frameworks bench !!
We can still have a cloud version of the frameworks bench, we only need to run the docker with limited CPUs, than it's very easy with docker !!

EDIT: will be good to create a Github action to only run the tests from the database changed in the toolset !!

joanhey:
@msmith-techempower @NateBrady23 
Why is it happen this problem now ??

msmith-techempower:
I don't know the answer off-hand... I'll have to dig in on a couple and see what's happening.

joanhey:
Ready to be merged with MySQL 9.4.

We need to investigate the problem with versions 9.5 and 9.6.

joanhey:
@msmith-techempower ready to be merged.

msmith-techempower (approved):

Issue #9578, opened 2025-01-31
No more fakes


joanhey:
Please, no more http: length hard encoded.

No more template response cache !!!
The template is OK to cache, but the response for the fortunes is NOT OK.

joanhey:
Now the devs of the framework decide if it's realistic or not.

We need to implement more tests before start the benchmark !!

https://github.com/TechEmpower/FrameworkBenchmarks/issues/9440#issuecomment-2518472173

p8:
> The template is OK to cache, but the response for the fortunes is NOT OK.

I wonder if we should make fortunes response dynamic to prevent this.
Maybe store more fortunes in the database and pick random fortunes, similar to the query test?

joanhey:
Validate response status codes #9605

joanhey:
Shouldn't pre-computed HTTP headers trigger "Stripped" classification? #8205

joanhey:
@p8 
https://github.com/TechEmpower/FrameworkBenchmarks/issues/6529
https://github.com/TechEmpower/FrameworkBenchmarks/issues/6788
https://github.com/TechEmpower/FrameworkBenchmarks/issues/6883

p8:
The `aspnetcore` framework precomputes the header length as well:

https://github.com/TechEmpower/FrameworkBenchmarks/blob/c82a834c5e8aeae14c8203cc05beaaa18dcc7cfc/frameworks/CSharp/aspnetcore-mono/PlatformBenchmarks/BenchmarkApplication.Json.cs#L14-L18

https://github.com/TechEmpower/FrameworkBenchmarks/blob/c82a834c5e8aeae14c8203cc05beaaa18dcc7cfc/frameworks/CSharp/aspnetcore-mono/PlatformBenchmarks/BenchmarkApplication.Json.cs#L12

https://github.com/TechEmpower/FrameworkBenchmarks/blob/c82a834c5e8aeae14c8203cc05beaaa18dcc7cfc/frameworks/CSharp/aspnetcore/src/Platform/BenchmarkApplication.Plaintext.cs#L13-L16

remittor:
`aspcorenet`
<details><summary>image</summary>

![Image](https://github.com/user-attachments/assets/38b6abe4-853b-4216-80b3-db1ee13e87ef)

</details>

[...]

PR #10606, opened 2026-01-21
[CSharp/beetlex] Mark core tests as stripped

p8:
Beetlex hardcodes HTTP headers for the core tests, so it should be marked as Stripped:

https://github.com/TechEmpower/FrameworkBenchmarks/blob/d36a2ede0db5c69d70da421e2f48aea53973c0c5/frameworks/CSharp/beetlex/PlatformBenchmarks/HttpHandler.plaintext.cs#L14-L18

Closes: https://github.com/TechEmpower/FrameworkBenchmarks/issues/9578


Dmole:
That's quite the impactful categorization, can we get the year old website updated?


https://www.techempower.com/benchmarks/

Issue #9916, opened 2025-05-28
To avoid misleading: frameworks that do not implement basic HTTP functions should be removed


lesismal:
Thanks to TechEmpower for providing the community with so many language/framework performance test results, which bring great reference value to everyone.

Some frameworks are just 4th-layer frameworks(for TCP/UDP/UnixSocket...), which implement simple HTTP "\r\n" judgment to perform a request parsing, and then write back the HTTP response buffer. They do not have the basic conditions for being a production environment HTTP service, but they participated in this test and got a high/top performance results.

Many users in the community do not really read the test code of these frameworks, so they do not know the truth, which causes a lot of miscommunication, making users mistakenly believe that some frameworks or even languages ​​can achieve unrealistic performance.

As a performance testing project with great influence, it should be fair and just, so I suggest removing these frameworks with incomplete functions.

If you want to test the 4th-layer performance of these not full-feature HTTP frameworks, you should open a separate test item/combination instead of comparing the 4th-layer framework(with fake HTTP implementation) with the HTTP framework.

volyrique:
My impression is that the approach so far has been to classify any implementations that suffer from the issues that you mention as `Stripped` instead of removing them. Stripped implementations are hidden in the result visualization by default.

lesismal:
> My impression is that the approach so far has been to classify any implementations that suffer from the issues that you mention as `Stripped` instead of removing them. Stripped implementations are hidden in the result visualization by default.

Thanks for your action; this is an improvement.
However, those frameworks are just TCP tests, not HTTP, so I still recommend removing them.

lesismal:
> My impression is that the approach so far has been to classify any implementations that suffer from the issues that you mention as `Stripped` instead of removing them. Stripped implementations are hidden in the result visualization by default.

Sorry for that I didn't notice you are not official, but fine.

lesismal:
@NateBrady23 @msmith-techempower @panjf2000 @andy2pan 

Since I haven't received an official response for a long time, I think it's time to drop the pleasantries and politeness, and make further clarifications, even if this might escalate the matter into a bigger argument. Let's start fighting!

Since all tests are based on HTTP, those frameworks that haven't truly implemented the HTTP protocol don't meet the basic testing requirements in the first place.
These repositories that only implement simple judgments like HTTP line breaks to cope with the tests cannot handle production HTTP business at all, and can't even be considered toys.
[...]

Issue #8205, opened 2023-05-10
Shouldn't pre-computed HTTP headers trigger "Stripped" classification?


synopse:
Reading general requirement number 2 :

> This project intends to measure performance of production-grade deployments of real-world web application frameworks and platforms. All implementations are expected (but not required) to be based on robust implementations of the HTTP protocol. Implementations that are not based on a realistic HTTP implementation will be marked as Stripped. Rather than stipulate a rigid specification of HTTP requirements, we defer to common sense: the implementation should provide a viable subset of HTTP to reasonably unbiased outsiders familiar with HTTP.

and for instance requirement iv of the [plain text requirement](https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#plaintext) :

> However, the rest of the response must be fully composed on the spot.

Some frameworks do not compute the HTTP headers at all, but return a fixed value, including `HTTP/1.1 200 OK` and the `Content-Length:` integer value. This is not a realistic HTTP implementation to my (common) sense. There is no point of allowing to pre-compute the headers but disallowing to pre-compute the response body - the former clearly depending on the later content.

Some sample are for instance [aspnetcore plaintext](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/CSharp/aspnetcore/PlatformBenchmarks/BenchmarkApplication.Plaintext.cs) or [aspnetcore](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/CSharp/aspnetcore/PlatformBenchmarks/BenchmarkApplication.Json.cs). Pre-computing the whole headers is not a realistic approach, my guess is that such implementations should be marked as Stripped, or fixed to use a HTTP headers generation library, as any realistic framework do.

Perhaps there are other implementations with the same dubious pattern.

synopse:
After a quick search, other similar findings - some clearly breaking the requirements:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Go/gnet/src/main.go#L29
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Rust/actix/src/main_server.rs#L23
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Rust/faf/src/main.rs
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Rust/ntex/src/main_plt.rs

[...]

PR #10611, opened 2026-01-21
[CSharp/aspnetcore] Mark platform tests as stripped

p8:
ASP netcore hardcodes HTTP headers for the platform tests, so these tests should be marked as stripped:

https://github.com/TechEmpower/FrameworkBenchmarks/blob/d36a2ede0db5c69d70da421e2f48aea53973c0c5/frameworks/CSharp/aspnetcore/src/Platform/BenchmarkApplication.Plaintext.cs#L12-L16

From the [Test Requirements](https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview):
> All implementations are expected (but not required) to be based on robust implementations of the HTTP protocol. Implementations that are not based on a realistic HTTP implementation will be marked as Stripped.

Closes: https://github.com/TechEmpower/FrameworkBenchmarks/issues/9578


joanhey:
I said that some times, exist more frameworks that handcore the headers results !!!

EDIT: not framework can hand made the `Content-Lenght`, but also another headers !!

p8:
@joanhey I think I have open PR's for all frameworks with hardcoded headers now.

joanhey:
Yes, it's better mark them as striped, than create issues !!!

DamianEdwards:
I'm supportive of this assuming we update the test requirements to explicitly call out hard-coding request/response headers or other HTTP protocol implementations as an example of the "stripped" litmus test and update other benchmarks doing this to be classified in the same way.

joanhey:
Exits more with that trick !!!
And also others with cache the fortunes response, not the fortunes template !!!
As the final response is always the same !!!

And others than are marked as Full Framework, and bypass the template engine !!!

joanhey:
@DamianEdwards  of course
First is not permitted by the bench rules.
Second is not fair play.

This PR for me is completely correct. 
If you find more fw making the same, you know it, please mark them as striped.
Thank you.

DamianEdwards:
>First is not permitted by the bench rules.
>Second is not fair play.

@joanhey can you explain this comment a bit more please, not quite following which thing is "first" and which is "second" that you're referring to? My understanding of the issue this PR is highlighting with the ASP.NET Core platform benchmarks is explicitly that they hard-code the HTTP response protocol elements (headers, etc.) and are essentially just sending the HTTP response directly as bytes over the connection rather than going through an HTTP protocol implementation (the incoming request processing does go through the HTTP parsing stack to be clear, and the platform Fortunes implementation does implement a streaming chunked response but is still hard-coding the HTTP header block).

joanhey:
First not fw can hard code the `Content-Lenght`  
But after check the requirements is not explicitly commented.
But for the fortune test is really clear: 
`The list data structure must be a dynamic-size or equivalent and should not be dimensioned using foreknowledge of the row-count of the database table.
`

And also yes, use chunked, need to be not permitted, but the problem here is that the response is larger than 13, and not beneficial for the bench. What the fw gain for not calculate the response size (very fast for this small string) is for send a larger response, than break any good results in a bench.

p8:
Related discussions:
https://github.com/TechEmpower/FrameworkBenchmarks/issues/8205
https://github.com/TechEmpower/FrameworkBenchmarks/issues/9440
https://github.com/TechEmpower/FrameworkBenchmarks/issues/9578

joanhey:
Any realistic server, can't hardcode the `Content-Lenght`, they need to calculate it for any response.
But we are open to any change.

I explained some time ago, that we need to change the static tests, so it is NOT possible  to hand code the headers.
With the fortunes is worst, when they cache the response directly, so no sort and not UTF8 encode for each request. And some fw only cache the response for this bench.
We have the db cache for that, and still questionable.

joanhey:
@DamianEdwards if you want than the headers are all harcoded, not realistic for me, please create a new issue.
If it's approved is perfect, but at that moment any fw can use hand coded headers !!  and so we'll have a fair play.

DamianEdwards:
@joanhey I think we're in agreement. The requirements should be changed to explicitly state that hard-coded header blocks and values for headers like `Content-Length` should be disallowed for the "Realistic" classification.

In the case of ASP.NET Core and the platform Fortunes benchmark specifically, just wanted to callout that it *doesn't* hardcode `Content-Length` header and instead has a [proper chunked response body implementation](https://github.com/TechEmpower/FrameworkBenchmarks/blob/d36a2ede0db5c69d70da421e2f48aea53973c0c5/frameworks/CSharp/aspnetcore/src/Platform/ChunkedPipeWriter.cs), and it does not used cached template responses or a statically sized list data-structure. The data structure it uses is .NET's `List<T>` [without an initial size hint and the querying, dynamic insertion, and sorting is done on each request](https://github.com/TechEmpower/FrameworkBenchmarks/blob/d36a2ede0db5c69d70da421e2f48aea53973c0c5/frameworks/CSharp/aspnetcore/src/Platform/Data/RawDbNpgsql.cs#L187-L204), followed by the [template being executed](https://github.com/TechEmpower/FrameworkBenchmarks/blob/d36a2ede0db5c69d70da421e2f48aea53973c0c5/frameworks/CSharp/aspnetcore/src/Platform/BenchmarkApplication.Fortunes.cs#L37-L51) (no cached template responses).

msmith-techempower:
@DamianEdwards The key issue here is that we have historically treated `plaintext` as "the one you can cheat on" with an implicit "as long as you're acting in good faith on the other test implementations."

@p8 has been going through and marking frameworks as 'Stripped' when they have both (as far as I have seen) `json` and `plaintext` with hard-coded preambles and aren't doing the work.

I think I'm in favor of leaving this one open for now, but I would expect the `json` test to be updated to execute its work as a traditional developer would treat the data in good faith while building out some service. It should create the object and allow the framework to dynamically determine the content length, set headers, etc. without the user gaming it.

See https://github.com/TechEmpower/FrameworkBenchmarks/issues/9916#issuecomment-3411564238 for more context.

tl;dr - `plaintext` is the "fun" one where frameworks can do whatever they want to try and go brrrrr, but the rest need to follow the rules or the framework will be marked "Stripped".

DamianEdwards:
@msmith-techempower sounds good! On that basis, the current ASP.NET Core platform implementations for all the benchmarks other than plaintext should be marked stripped in their current state, as they all hardcode some or all of the response preamble. We can follow up separately on whether we want to react to that by adjusting them, etc.

PR #10613, opened 2026-01-23
[perl] Update to Perl v5.42

joanhey:
Also changed the image in Plack, as they use `perl:latest`.

And clean fersum dockerfile.


joanhey:
Mark Feersum MySQL as broken, I can't fix it.
And without any change fail in the last runs, because the signing key expired on 2025-10-22T17:26:50Z
https://tfb-status.techempower.com/unzip/results.2026-01-02-05-57-57-267.zip/results/20251225053036/feersum-mysql/build/feersum.log

https://bugs.mysql.com/bug.php?id=119212

PR #10632, opened 2026-01-30
Fix dart3_native regression

iapicca:
### Description

This PR partially rolls back changes introduced in https://github.com/TechEmpower/FrameworkBenchmarks/pull/10493.

### Motivation

Inter-process port sharing is not supported (see https://github.com/dart-lang/sdk/issues/62545). 
As a result, the horizontal “hybrid” scaling approach introduced in the referenced PR does not function correctly. 
This causes both performance and image size regressions.

### Changes

- Reimplemented the multi-stage Dockerfile using `scratch` instead of `busybox`
- Updated the Dart version
- Simplified request handling
- General cleanup (e.g., removing unnecessary `Future`s and `async` usage)

### DIFF
--- a/frameworks/Dart/dart3/dart3.dockerfile
+++ b/frameworks/Dart/dart3/dart3.dockerfile
@@ -12,4 +12,4 @@ COPY --from=build /runtime/ /
 COPY --from=build /app/server /bin/server
 
 EXPOSE 8080
-CMD ["server"]
+ENTRYPOINT ["server"]


p8:
@iapicca Unrelated to this PR, but I'm seeing the following warning on master:
```
Framework dart3 does not define a default test in benchmark_config.json
```

iapicca:
> @iapicca Unrelated to this PR, but I'm seeing the following warning on master:
> 
> ```
> Framework dart3 does not define a default test in benchmark_config.json
> ```
Hi @p8 
I'm happy to include the fix in the current PR, but it's not clear to me what causes the warning.
The `default` seems to be already defined in `dart3` [benchmark_config.json ](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Dart/dart3/benchmark_config.json), am I missing something?

Thank you

p8:
@iapicca I think you need to delete these:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/a6502505ad67783dbcb56e9584482e4973807a69/frameworks/Dart/dart3/benchmark_config.json#L17-L18

iapicca:
@p8 done, thank you

p8:
Nice!

PR #10642, opened 2026-02-01
[rust/ntex] Mark all plt tests as stripped

p8:
plt-neon and plt-neon-uring also use precalculated response headers.


joanhey:
Also it have separate builds for plaintext/json and db tests.
For that have 12 permutations for only 3 platforms.

~Send a PR in the @p8 fork to use rust:1.93. Use rust:latest is NOT permitted.~

msmith-techempower (approved):

PR #10643, opened 2026-02-01
[c#] Mark Unhinged tests as Stripped

joanhey:
Use precalculated response headers.

image

image


MDA2AV:
Is the Content-Length the issue? I can change it to calculate the length (27)

joanhey:
Yes, each fw need to calculate the length for any request.
Thank you !!

MDA2AV:
Ok, just mark it as stripped for now, I'm working on a bigger update for this framework for higher performance and by then  I'll fix this content length issue and mark it back to realistic.

MDA2AV:
is something like this allowed then?
It calculates content length every request, it costs almost zero so same performance. It leaves some byes space on the header for the content length which is only known later on by calculating the buffer tail difference from before and after the message is serialized.
This is a basic example, it can be generalized to any content length easily

```csharp
var headers = "HTTP/1.1 200 OK\r\n"u8 +
                          "Content-Length:   \r\n"u8 +
                          "Server: W\r\n"u8 +
                          "Content-Type: application/json\r\n"u8;
                
ctx.Connection.Write(headers);
var date = DateHelper.HeaderBytes;
ctx.Connection.Write(date);

var tailSnapshot = ctx.Connection.WriteTail;

var utf8JsonWriter = t_writer ??= new Utf8JsonWriter(ctx.Connection, new JsonWriterOptions { SkipValidation = true });
utf8JsonWriter.Reset(ctx.Connection);
JsonSerializer.Serialize(utf8JsonWriter, new JsonMessage { Message = JsonBody }, SerializerContext.JsonMessage);

var contentLength = ctx.Connection.WriteTail - tailSnapshot;
unsafe
{
    byte* dst = ctx.Connection.WriteBuffer + 33;
    int tens = contentLength / 10;
    int ones = contentLength - tens * 10;

    dst[0] = (byte)('0' + tens);
    dst[1] = (byte)('0' + ones);
}
```

PR #10648, opened 2026-02-01
[python/fastapi] Remove duplicate json and plaintext tests

p8:
The plaintext and json test for gunicorn-orm are the same as the default.

image


p8:
The gunicorn-nginx tests have the same results as the default for all the tests.
Maybe those can be removed as well?

PR #10650, opened 2026-02-01
[rust/ntex] Use rust:1.93 as rust:latest is not permitted

joanhey:
PD: almost all rust frameworks use rust:latest  !!
When update the version all will fail.

### DIFF
--- a/frameworks/Rust/ntex/ntex-compio.dockerfile
+++ b/frameworks/Rust/ntex/ntex-compio.dockerfile
@@ -1,4 +1,4 @@
-FROM rust:latest
+FROM rust:1.93
 
 # Disable simd at jsonescape
 # ENV CARGO_CFG_JSONESCAPE_DISABLE_AUTO_SIMD=
--- a/frameworks/Rust/ntex/ntex-db-compio.dockerfile
+++ b/frameworks/Rust/ntex/ntex-db-compio.dockerfile
@@ -1,4 +1,4 @@
-FROM rust:latest
+FROM rust:1.93
 
 # Disable simd at jsonescape
 # ENV CARGO_CFG_JSONESCAPE_DISABLE_AUTO_SIMD=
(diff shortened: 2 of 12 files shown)


fafhrd91:
wtf, where is the rule that latest is not permitted?

joanhey:
This undermines reproducibility, a core requirement for valid benchmarking.
It introduces inconsistency and unreliability in results, are your app faster or is the underlying system ?

But more important is avoid fails, when new versions appear, the bench will not work.
I know than Rust don't introduce breaking changes, but what will happen when appear Rust v2, no framework using latest will work. As minimum need to use rust:1 and need to be discussed.

No framework or language use latest in the bench. Not even the toolset.
Imagine that we use the database images with latest, with an small change will break hundreds of framwork benchs or change the results drastically.

fafhrd91:
seems you dont have much experience with rust if suggest specific rust version. rust has tool that allow reproducible builds but compiler version is not it. also by replacing you increase load on tfb team.

joanhey:
I don't talk about builds, is about results numbers.

How we can compare a bench from this week, with a bench of one year ago using latest in the build ?

fafhrd91:
please dont touch what is not maintained by you

joanhey:
https://github.com/TechEmpower/FrameworkBenchmarks/pull/8789#discussion_r1537793110

https://github.com/TechEmpower/FrameworkBenchmarks/pull/6110#issuecomment-720131286

https://github.com/TechEmpower/FrameworkBenchmarks/pull/5174#issuecomment-546019552

https://github.com/TechEmpower/FrameworkBenchmarks/pull/4930#discussion_r303956137

https://github.com/TechEmpower/FrameworkBenchmarks/pull/4268#issuecomment-448252695

...

msmith-techempower:
The rules predate the use of Docker. As such, I will update the rules to include that we require a specific image (typically, locked to a version) to facilitate reproducibility from run to run.

We are specifically trying to ensure that someone using `rust:latest` in a run from 6 months ago has a _known_ version that anyone can test against _tomorrow_ and produce like results.

@fafhrd91 If you have a compelling reason to require `rust:latest` _and_ have a solution to the reproducibility issue brought up in this thread, please explain; otherwise, I will be merging this PR in as-is.

fafhrd91:
it is strange to try to get reproducible builds for test runs. it is up to you, you will need to review prs.

PR #10657, opened 2026-02-02
[swift/swift-nio] Mark tests as stripped

p8:
Swift precalculates Content-Length HTTP headers, so these tests should be marked as stripped:

https://github.com/TechEmpower/FrameworkBenchmarks/blob/d36a2ede0db5c69d70da421e2f48aea53973c0c5/frameworks/Swift/swift-nio/app/Sources/main.swift#L13-L18

https://github.com/TechEmpower/FrameworkBenchmarks/blob/d36a2ede0db5c69d70da421e2f48aea53973c0c5/frameworks/Swift/swift-nio/app/Sources/main.swift#L68

https://github.com/TechEmpower/FrameworkBenchmarks/blob/d36a2ede0db5c69d70da421e2f48aea53973c0c5/frameworks/Swift/swift-nio/app/Sources/main.swift#L74

### DIFF
--- a/frameworks/Swift/swift-nio/benchmark_config.json
+++ b/frameworks/Swift/swift-nio/benchmark_config.json
@@ -5,7 +5,7 @@
       "json_url": "/json",
       "plaintext_url": "/plaintext",
       "port": 8080,
-      "approach": "Realistic",
+      "approach": "Stripped",
       "classification": "Platform",
       "database": "None",
       "framework": "None",


ruroru:
Isn't it better to change it so it would not pre calculate?

p8:
> Isn't it better to change it so it would not pre calculate?

Yes, but my Swift isn't that good, so it would cost me more time than I'm willing to spend on it.

joanhey:
Ping @adam-fowler

adam-fowler:
I can add a PR that doesn't pre-calculate the content-length size

joanhey:
Of course @adam-fowler

adam-fowler:
#10701

p8:
Thanks @adam-fowler !

PR #10701, opened 2026-02-06
Remove pre-calculated values from swift-nio framework.

adam-fowler:
Also cleanup code.


joanhey:
Of course, but please notify the devs than add it !!

But it is yourself !!! :+1: 
Thank you !!

joanhey (approved):

joanhey:

Example:

PR #10704, opened 2026-02-06
[haskell/warp] Marked as broken

joanhey:
Fail in the last runs.

@cptwunderlich @naushadh Please fix. Thank you.


joanhey:
Fixed in #10718

joanhey:
Pull again!!
@msmith-techempower that frameworks use extra time in the runs for nothig.

msmith-techempower (approved):

PR #10712, opened 2026-02-06
[kotlin/pellet] Mark as broken

joanhey:
Fail in the last runs.

@lopcode Please fix it. Thank you.

EDIT: Deleted Pellet as @lopcode recommend.


lopcode:
Please feel free to remove Pellet from the suite, I'm unlikely to work on it any time soon

PR #10715, opened 2026-02-06
[java/quarkus] Mark as broken

joanhey:
Quarkus Vert-x variant marked as broken as fail in the last runs.

@tsegismont @Sanne @franz1981 @hatahet Please fix it. Thank you.

Another big problem is the use of constant for the length header that is not permitted.

https://github.com/TechEmpower/FrameworkBenchmarks/blob/33531b8fcf4ea52f8c885cb288c3ccfc52c8ba73/frameworks/Java/quarkus/vertx/src/main/java/io/quarkus/benchmark/resource/PlaintextHttpHandler.java#L15

https://github.com/TechEmpower/FrameworkBenchmarks/blob/33531b8fcf4ea52f8c885cb288c3ccfc52c8ba73/frameworks/Java/quarkus/vertx/src/main/java/io/quarkus/benchmark/resource/PlaintextHttpHandler.java#L20

### DIFF
--- a/frameworks/Java/quarkus/benchmark_config.json
+++ b/frameworks/Java/quarkus/benchmark_config.json
@@ -70,7 +70,8 @@
         "database_os": "Linux",
         "display_name": "Quarkus, Vert.x",
         "notes": "",
-        "versus": "Vert.x"
+        "versus": "Vert.x",
+        "tags": ["broken"]
       },
       "reactive-routes-pgclient": {
         "json_url": "/json",


franz1981:
I think pretty much all the frameworks do it: this is not a fully encoded header but the values for the maps which has still to be encoded into the proper response ^^

joanhey:
The problem is that don't start in the bench runs, for that is marked as broken.
https://tfb-status.techempower.com/unzip/results.2026-02-03-23-09-26-607.zip/quarkus-vertx


About the `Content-Length` header, need to be calculated for each request:
- https://github.com/TechEmpower/FrameworkBenchmarks/pull/10661
- https://github.com/TechEmpower/FrameworkBenchmarks/pull/10658
- https://github.com/TechEmpower/FrameworkBenchmarks/pull/10657
- https://github.com/TechEmpower/FrameworkBenchmarks/pull/10643
- https://github.com/TechEmpower/FrameworkBenchmarks/pull/10642
- ...

PD: if you can find another framework using constants, please mark it as "Stripped"

PR #10720, opened 2026-02-07
[php] Symfony, Laravel setting adjust

kitrio:
changes:
* set nginx keepalive 40 to 256
* opcache.interned_strings_buffer 8 to 32

### DIFF
--- a/frameworks/PHP/laravel/deploy/conf/php.ini
+++ b/frameworks/PHP/laravel/deploy/conf/php.ini
@@ -1773,7 +1773,7 @@ opcache.enable_cli=1
 opcache.memory_consumption=256
 
 ; The amount of memory for interned strings in Mbytes.
-opcache.interned_strings_buffer=32
+;opcache.interned_strings_buffer=8
 
 ; The maximum number of keys (scripts) in the OPcache hash table.
 ; Only numbers between 200 and 1000000 are allowed.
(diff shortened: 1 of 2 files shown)


joanhey:
None of both settings have any benefit.
But try them, later i'll revert this PR.

joanhey:
Nginx keepalive is the time in seconds to maintain a connection.
Each test use 15 seconds !!

The `opcache.interned_strings_buffer` is the Mb to use as string cache. 
This benchmark is small enough to don't use more than 8 Mb.
And if is the case, show a message _"interned string buffer overflow"_ than never was shown.

ruroru:
> Please close this PR. Thank you.

Chillax, your PRs bring as much value this PR

kitrio:
> Please close this PR. Thank you.

> Nginx keepalive is the time in seconds to maintain a connection. Each test use 15 seconds !!
> 
> The `opcache.interned_strings_buffer` is the Mb to use as string cache. This benchmark is small enough to don't use more than 8 Mb. And if is the case, show a message _"interned string buffer overflow"_ than never was shown.

opcache.interned_string_buffer ... your opinion right i'll modify code.

but, keepalive directive is not second.
https://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive

> Activates the cache for connections to upstream servers.

>The connections parameter sets the maximum number of idle keepalive connections to upstream servers that are preserved in the cache of each worker process. When this number is exceeded, the least recently used connections are closed.


In current benchmark configuration  php -fpm pm.max_children is set to 1024. 
So, a keepalive value 40 is insufficient to handle the concurrent load, so I have increased it to match the scale of the test.

joanhey:
> It should be particularly noted that the keepalive directive does not limit the total number of connections to upstream servers that an nginx worker process can open. The connections parameter should be set to a number small enough to let upstream servers process new incoming connections as well.

PR #10722, opened 2026-02-09
[kotlin/vertx-web-kotlinx-exposed-r2dbc] Add 4 variant configurations to compare R2DBC connection pool strategies for `vertx-web-kotlinx-exposed-r2dbc-postgresql`

ShreckYe:
After #10560 was merged and its [Continuous Benchmarking results](https://www.techempower.com/benchmarks/#section=test&runid=acc1ad82-ae2c-4d1d-a600-c7ff9d0c5917&l=zhxjwf-pa7&f=zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zhb2tb-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-z62j9b-zik0zj-zik0zj-pa7&test=db) came out, I noticed that some results appear to be significantly poorer relatively than what I got on my device. Specifically, the `vertx-web-kotlinx-exposed-r2dbc-postgresql` results are poorer than those of `ktor-netty-exposed-r2dbc-dsl`, which is unusual because they share basically the same DB operation logic, while Vert.x is more performant than Ktor. I suspect its connection and pooling configurations are not optimal for your server.

Since the results look fine on my device, I'd like to benchmark these different configurations on your server via Continuous Benchmarking to determine the optimal configuration. The code is mostly generated by Copilot as it's simple repetitive code and has been reviewed quickly by me. I will open a follow-up PR to revert these changes and apply the optimal configuration once the results are available. Thanks.


joanhey:
Please add the maintainers in the `benchmark_config.json`.

So with any change or marked as broken, it'll be notified to them.

Example:

ShreckYe:
Thanks! I understand the limit. These 4 benchmarks `vertx-web-kotlinx`, `vertx-web-kotlinx-exposed-r2dbc`, `vertx-web-kotlinx-r2dbc`, and `vertx-web-kotlinx-exposed-vertx-sql-client` are actually different implementations using different libraries. I consolidated them into one Gradle project to reduce code duplication and also just because I maintain all of them. Would an exception be possible in this case?

Otherwise, I could remove some benchmark portions unrelated to the connection pool comparison purpose in this PR temporarily, and add them back in the follow-up revert PR I mentioned.

joanhey:
I don't merge the PRs, @msmith-techempower will decide it.

I understand that is really good to test better configurations and need to be a way to do it.
The problem is when the people test configurations and later left in the bench for years, than make the runs to need days for finish without any benefit.

msmith-techempower:
This really isn't how you should be testing configuration. I sympathize that it takes a long time to benchmark on our hardware (and thank you @joanhey in your pruning broken tests to speed that up), but allowing this many permutations will just add even more turnaround time. What I recommend is to try a configuration, wait for the benchmark, then open a new PR to test in the next run.

ShreckYe:
OK thanks. I have removed the `vertx-web-kotlinx-r2dbc` variants in this PR.

msmith-techempower (approved):

PR #10734, opened 2026-02-11
[java] Fix TODAY broken config

TAKETODAY:


frameworks/Java/today/benchmark_config.json
```diff
@@ -24,8 +24,7 @@
         "database_os": "Linux",
         "display_name": "TODAY",
         "notes": "",
-        "versus": "None",
-        "tags": ["broken"]
```
joanhey, line 28: Please maintain the _tags_ field.

```
"versus": "None",
"tags": []
```

frameworks/Java/today/benchmark_config.json
joanhey, line 6: Please add yourself to the maintainers in the `benchmark_config.json`.

Example:

joanhey (changes requested):

frameworks/Java/today/benchmark_config.json
TAKETODAY: ok

joanhey (approved):

TAKETODAY:
@joanhey Is HTTP/2 allowed?

joanhey:
Not.
Actually the bench use `wrk` and only bench`HTTP/1` and `HTTP/1.1`.
Exist `wrk2` than also work with `HTTP/2`.

It's a good question to discuss.
You can open a issue asking for the use of `HTTP/2`, but I suspect than some frameworks don't support it.
And the difference between fw using `HTTP/1` and `HTTP/2` will be noticeable, so we need to mark it in some way.

PD: another problem is that `HTTP/2` need always use TLS Application-Layer Protocol Negotiation (ALPN) and it'll complicate the bench. And first need to try `HTTP/2`  and fallback to `HTTP/1.1` after failing to negotiate `HTTP/2` via ALPN. So the  frameworks using `HTTP/1.1` will have a double penalty.

PD: we are testing the frameworks performance, not the network performance, than with `HTTP/2` change a lot.

PR #10653, opened 2026-02-02
[javascript] Create new Fortune object in scope of request

p8:
The new fortune object must be constructed in the scope of the request, not as a constant:

> Within the scope of the request, a new Fortune object must be
constructed and added to the list.

https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#fortunes

PR #10735, opened 2026-02-11
[javascript/koa] Create new Fortune object in scope of request

p8:
The new fortune object must be constructed in the scope of the request, not as a constant:

> Within the scope of the request, a new Fortune object must be
constructed and added to the list.

https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#fortunes

This also disables the failing update test.


joanhey:
Ping @maximelkin

msmith-techempower:
@p8 Changes look good to me, but was `updates` not working? Why did you remove it?

p8:
@msmith-techempower yes, it was failing, also on [the current run](https://tfb-status.techempower.com/results/bf916986-96a5-4912-b2e7-45c2afa14165).
I'll gladly restore it though (I thought this was holding back merging https://github.com/TechEmpower/FrameworkBenchmarks/pull/10653, so I extracted this PR out of that one).

joanhey:
The bench don't separate the frameworks than don't start. Very bad !!
But why we don't inform about the tests than fail, and they don't run again !!

About the frameworks than don't star I'll create a issue.

PR #10736, opened 2026-02-11
Repair Quarkus Vert.x benchmark

tsegismont:
Follows-up on #10715 

The Ongres SCRAM dependency is required for authentication with the database.

### DIFF
--- a/frameworks/Java/quarkus/benchmark_config.json
+++ b/frameworks/Java/quarkus/benchmark_config.json
@@ -71,7 +71,7 @@
         "display_name": "Quarkus, Vert.x",
         "notes": "",
         "versus": "Vert.x",
-        "tags": ["broken"]
+        "tags": []
       },
       "reactive-routes-pgclient": {
         "json_url": "/json",
--- a/frameworks/Java/quarkus/vertx/pom.xml
+++ b/frameworks/Java/quarkus/vertx/pom.xml
@@ -25,6 +25,10 @@
             <artifactId>vertx-pg-client</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.ongres.scram</groupId>
+            <artifactId>client</artifactId>
+        </dependency>
+        <dependency>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-vertx</artifactId>
         </dependency>
--- a/frameworks/Java/quarkus/vertx/src/main/java/io/quarkus/benchmark/repository/PgClientFactory.java
+++ b/frameworks/Java/quarkus/vertx/src/main/java/io/quarkus/benchmark/repository/PgClientFactory.java
@@ -1,5 +1,6 @@
 package io.quarkus.benchmark.repository;
 
+import io.quarkus.logging.Log;
 import io.vertx.core.Vertx;
 import io.vertx.pgclient.PgConnectOptions;
 import jakarta.annotation.PreDestroy;
@@ -36,7 +37,7 @@ public class PgClientFactory {
         try {
             pgConnectionPool = new PgConnectionPool(vertx, pgConnectOptions());
         } catch (final Exception e) {
-            // TODO LOG ME: usually means inability to connect to the database
+            Log.error("Failed to create pgConnectionPool", e);
         } finally {
             this.pgConnectionPool = pgConnectionPool;
             return pgConnectionPool;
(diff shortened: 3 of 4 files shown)


frameworks/Java/quarkus/benchmark_config.json
```diff
@@ -70,8 +70,7 @@
         "database_os": "Linux",
         "display_name": "Quarkus, Vert.x",
         "notes": "",
-        "versus": "Vert.x",
-        "tags": ["broken"]
```
msmith-techempower, line 74: Please leave `"tags": []`

joanhey:
Not only that, they need to calculate the length in each request !!!
https://github.com/TechEmpower/FrameworkBenchmarks/pull/10715#issue-3908810042

EDIT: or will be marked as stripped !!!

frameworks/Java/quarkus/benchmark_config.json
```diff
@@ -70,8 +70,7 @@
...
+        "versus": "Vert.x"
```
joanhey: ???

joanhey:
Another big problem is the use of constant for the length header that is not permitted.

https://github.com/TechEmpower/FrameworkBenchmarks/blob/33531b8fcf4ea52f8c885cb288c3ccfc52c8ba73/frameworks/Java/quarkus/vertx/src/main/java/io/quarkus/benchmark/resource/PlaintextHttpHandler.java#L15

https://github.com/TechEmpower/FrameworkBenchmarks/blob/33531b8fcf4ea52f8c885cb288c3ccfc52c8ba73/frameworks/Java/quarkus/vertx/src/main/java/io/quarkus/benchmark/resource/PlaintextHttpHandler.java#L20

frameworks/Java/quarkus/benchmark_config.json
tsegismont: Not sure what you mean

tsegismont:
@joanhey thanks for taking a look as well

Regarding your comments about the length header, I'm not a maintainer of this benchmark, consequently not in a position to discuss whether this should be changed.

The scope of this PR is repairing a broken benchmark, so not related to your issue anyway.

joanhey:
Without the header change, this PR will not be merged.
Or It'll be marked as Stripped.

frameworks/Java/quarkus/benchmark_config.json
joanhey: It's OK

tsegismont:
Except for plaintext in quarkus-reactive-routes-pgclient and quarkus-vertx, no Quarkus benchmark computes the content length beforehand.

My understanding is that this is compliant with the rule as explained in https://github.com/TechEmpower/FrameworkBenchmarks/pull/10611#issuecomment-3791641354

@joanhey @msmith-techempower  if I didn't get the statement wrong, can we get this PR merged after one of the maintainers has approved the PR?

msmith-techempower:
@tsegismont I have to agree with @joanhey - we have been really pushing to be more equitable about the stripped implementations. Without the header changes _or_ marking the implementation as 'Stripped', we won't merge. I recommend simply ensuring that the headers are computed on every request.

tsegismont:
Thanks for your answer @msmith-techempower 

>tl;dr - plaintext is the "fun" one where frameworks can do whatever they want to try and go brrrrr, but the rest need to follow the rules or the framework will be marked "Stripped".

So, this is no longer true?

joanhey:
If you want to play in the bench, you need to be a fair player !!
Thank you.

If you find another fw than make the same, please mark it as Stripped.

tsegismont:
> If you want to play in the bench, you need to be a fair player !! Thank you.

If this is a general statement, I fully agree.

> If you find another fw than make the same, please mark it as Stripped.

Sorry I'm not comfortable with sending PRs to other participants marking their implementation as stripped. I don't believe this would cultivate a positive competitive spirit.

msmith-techempower:
I looked at the Quarkus implementations and I think that it aligns with what I said in https://github.com/TechEmpower/FrameworkBenchmarks/pull/10611#issuecomment-3791641354 - only the [Plaintext](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/quarkus/vertx/src/main/java/io/quarkus/benchmark/resource/PlaintextHttpHandler.java) implementation uses the static length of the response body in the header. [JSON](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/quarkus/vertx/src/main/java/io/quarkus/benchmark/resource/JsonHttpHandler.java), for instance, does not set the content length manually and leverages the framework to do that work.

That is the intent of the 'Realistic vs. Stripped' classification. If a developer would actually write the code for non-Plaintext test implementations as such, then that is a 'Realistic' approach. Plaintext is allowed to be non-realistic. We only mark the specific framework permutation as 'Stripped' if a) they only implement Plaintext (and this is disallowed moving forward) or b) the implementations are unrealistic (e.g., manually setting header values like content length to a static known amount etc.).

Sanne:
Thanks @msmith-techempower and @tsegismont !

franz1981:
Thanks :pray:

tsegismont:
Thank you @msmith-techempower !

PR #10738, opened 2026-02-11
[ruby/grape] Don't install nginx

p8:
nginx is no longer used.


joanhey:
I never will ask you to add yourself to the frameworks.
But perhaps we need to add a maintainer dev  for language !!

msmith-techempower (approved):

Issue #10238, opened 2025-11-03
Cutelyst exceeds number of test mutations


p8:
According to https://github.com/TechEmpower/FrameworkBenchmarks/issues/8420 test mutations should be limited.
> The number of test mutations will be limited to 10.

Cutelyst has 14 test mutations:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/f99bb89276455eb3f90eb3cac8253bfa70b73af3/frameworks/C%2B%2B/cutelyst/benchmark_config.json

As Cutelyst also has build times of 45 minutes it is using up a disproportionate amount of time.
Maybe we should limit the amount of time per framework?

volyrique:
Let's ping @dantti. Also, it seems that the DB and the non-DB tests are covered by different configurations, which could probably be combined.

dantti:
> As Cutelyst also has build times of 45 minutes it is using up a disproportionate amount of time.

The issue here is lacking of docker caching, building the real code shouldn't take more than one minute.

> Also, it seems that the DB and the non-DB tests are covered by different configurations, which could probably be combined.

The config_socket.ini is not used anymore, I should remove it, uwsgi support was removed as it was slower and an additional dep.

volyrique:
My own profiling with PR #10125 showed that by far the slowest step was `apt install`. I tried an optimization in #10173 (use `buildpack-deps` instead of `ubuntu` as the base image), but there hasn't been a complete run with it yet, so that I could look at the logs and check if it has had any effect; YMMV.

BTW the issue was completely unreproducible locally and in CI; it seems to affect specifically the continuous benchmarking environment.

**P.S.** The optimization shaved off around 10 minutes off the build time, which is a pretty good result.

joanhey:
> As Cutelyst also has build times of 45 minutes it is using up a disproportionate amount of time.

The issue here is lacking of docker caching, building the real code shouldn't take more than one minute.

#10802

PD: still is only permitted 10 test mutations, try to join json/plaintext with a db test.
Now only for build all your dockerfiles is +12 hours, and later tests and benchmark. And this is only for a framework, not a language.

NateBrady23:
Closing as part of bulk cleanup.

PR #10748, opened 2026-02-13
[ruby/agoo] Marked as broken

joanhey:
Fail in the last 3 runs.

https://tfb-status.techempower.com/unzip/results.2026-02-11-23-41-23-147.zip/results/20260204071442/agoo/run/agoo.log

Ping @p8

### DIFF
--- a/frameworks/Ruby/agoo/benchmark_config.json
+++ b/frameworks/Ruby/agoo/benchmark_config.json
@@ -20,7 +20,8 @@
       "os": "Linux",
       "database_os": "Linux",
       "display_name": "agoo",
-      "notes": ""
+      "notes": "",
+      "tags": ["broken"]
     }
   }]
 }


p8:
@joanhey Yes, thanks, I'm aware of it but hadn't had time to fix it.
https://github.com/TechEmpower/FrameworkBenchmarks/pull/10750

There probably are other frameworks that have been failing for a longer time.
I think @volyrique only marked frameworks as broken if the test failed for 10 runs.

volyrique:
The [value](https://github.com/TechEmpower/FrameworkBenchmarks/blob/af3c0d89ee3cb5f15aeae7bd2e0fc4327d062e39/scripts/tfb-fail-detector.py#L17) that the TechEmpower team has set in the `tfb-fail-detector.py` script is 3.

The reason why I chose 11 was because I couldn't be bothered to research the people that had to be pinged for 60 implementations, so I focused on code that hadn't been changed in years, that had been failing for 3 months or more, and that had no maintainers set in its configuration file. In other words, implementations that for all intents and purposes have been abandonеd and nobody cares about.

joanhey:
For me it's worst.
I try to mark as broken all than fail, and I use my time to identify the devs, to inform them.
And later I ask they add the maintainers, but the bad is me !!!

If anybody don't want to add the maintainers, for me it's OK, but the next time I will not spent my time !!
And the fw will be marked as broken directly !!

joanhey:
This was added some years ago, because I ask it.
But almost no framework maintainer know it !!

joanhey:
I think that is failing, for more than 3 runs, we need to check it.
If the fail it's about not to network, limits in github or dockerhub, ....  we need to gain time for the runs.
Thank you !!

volyrique:
> This was added some years ago, because I ask it. But almost no framework maintainer know it !!

To be fair, some people may simply prefer not to be spammed with notifications. I have added myself in only one place, even though I have contributed to a couple of other implementations because I don't truly care about them (but so far I have fixed them if necessary).

joanhey:
I want to create a new issue for the toolset.
Now the runs mix the 2 problems: don't start or stop and failing a test.
This need to be changed, the fw than don't start need to be show separately.

And still better, have a page for the fw failing, when (time), how much, ....
Because it's very important know when start to fail.

joanhey:
@volyrique I understand you.
But the same than you, another people are angry when we marked as broken, and nobody is notified.

The problem is that the fw can fail for years, without any benefit, and the runs now need more than 1 week.

joanhey:
Now checking the actual run, Cutelyst is terrible, only this framework need more than 12 hours only for build, later tests and bench. 
And have more than 10 mutations !!!

volyrique:
We have issue #10238 for that.

joanhey:
They (we) need to automatize more, if a framework have more than 10 mutations, explain it with a github action.
If json/plaintext are separate from the database tests, notify it with a github action.
No maintainers, explain about it, not mandatory, but later don't cry if your fw is marked as broken without any notification !!

p8:
@joanhey can you close this PR in favor of: https://github.com/TechEmpower/FrameworkBenchmarks/pull/10750 ?

joanhey:
https://github.com/TechEmpower/FrameworkBenchmarks/pull/10750

p8:
Thanks @joanhey !

joanhey:
Of course !!

joanhey:
But I said before, the next time I'll not close a PR.
They start with the older PRs and later with the new.
And it is NOT our work this !!!

PR #10762, opened 2026-02-15
[php] Swoole update to PHP 8.5

joanhey:
Use their docker image for faster builds.
When are ready the image by v6, we'll mark the fixed version.

Ping @NathanFreeman
#10303

### DIFF
--- a/frameworks/PHP/swoole/benchmark_config.json
+++ b/frameworks/PHP/swoole/benchmark_config.json
@@ -1,5 +1,6 @@
 {
   "framework": "swoole",
+  "maintainers": ["NathanFreeman"],
   "tests": [{
     "default": {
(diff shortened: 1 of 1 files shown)


NathanFreeman:
Thank you.
Currently, only Swoole 6.2 supports PHP 8.5, but this version has not yet been officially released. This image is built on Alpine, and the performance of the Alpine image in stress tests is not superior to that of the Ubuntu environment.

joanhey:
When images are created with Swoole version and Debian/Ubuntu, I'll change it.
Because build the images and compile, need a lot of time in each run.

PD: I updated others frameworks with this image, because some still were using PHP 8.3 and use time to compile.
All the fw were working OK with this image.

joanhey:
image

PR #10767, opened 2026-02-17
[ntex] Update ntex framework

fafhrd91:
* Update to latest version

### DIFF
--- a/frameworks/Rust/ntex/benchmark_config.json
+++ b/frameworks/Rust/ntex/benchmark_config.json
@@ -1,5 +1,6 @@
 {
   "framework": "ntex",
+  "maintainers": ["fafhrd91"],
   "tests": [{
     "default": {
         "json_url": "/json",


joanhey:
[    b"HTTP/1.1 200 OK\r\nServer: N\r\nContent-Type: application/json\r\nContent-Length: 27\r\n";](https://github.com/fafhrd91/FrameworkBenchmarks/blob/24a1f3d6b62a027a5bef54a326eb61512b54add2/frameworks/Rust/ntex/src/main_plt.rs#L13)
Only with this line this variant need to be marked as Stripped !!

Please add the maintainers in the `benchmark_config.json`
Thank you !!

joanhey:
And this one too:
https://github.com/fafhrd91/FrameworkBenchmarks/blob/24a1f3d6b62a027a5bef54a326eb61512b54add2/frameworks/Rust/ntex/src/main_plt.rs#L15

joanhey:
Please fix it, before is marked as Stripped !!

joanhey:
They are marked as Stripped now.

joanhey:
#10650

msmith-techempower:
The affected tests are marked stripped, so these changes - while valid - will not be shown by default on the results page. You can fix the Json implementation to abide the rules and submit a PR changing it back to Realistic.

PR #10766, opened 2026-02-16
[kotlin/ktor] Variant R2bc fail in all test, except the first.

joanhey:
Removed unused URLs and updated approach type.

This variant only work with the first bench (db) and later do nothing.

I wanted to delete only the tests than fail, but I left the first one, the db test.
But later I marked as Stripped, because a fw need to return a response good or bad.
And this variant return nothing:

https://tfb-status.techempower.com/unzip/results.2026-02-03-23-09-26-607.zip/results/20260127014105/ktor-r2dbc/json/raw.txt
https://tfb-status.techempower.com/unzip/results.2026-02-03-23-09-26-607.zip/results/20260127014105/ktor-r2dbc/query/raw.txt
https://tfb-status.techempower.com/unzip/results.2026-02-03-23-09-26-607.zip/results/20260127014105/ktor-r2dbc/fortune/raw.txt
...

Return nothing, but the run need the time to bench it !!
A fw need to return a response good or bad, for that I marked it as Stripped.

But @msmith-techempower will decide it.

Ping: @ShreckYe @inemtsev  please fix it and add a maintainer in the `benchmark_config.json`.
Thank you !!

PR #10768, opened 2026-02-17
Add myself as a maintainer of the Ktor Exposed portions as requested in #10766

ShreckYe:


ShreckYe:
@joanhey Is it possible to add me as maintainers only to the `ktor-exposed-*` portions instead of all the Ktor portions? I don't know if the `config.json` format accepts that.

joanhey:
The maintainers it's for the full directory framework. An small change can break all variants.

PR #10769, opened 2026-02-17
[rust/ohkami-rt_nio] Marked as broken

joanhey:
https://tfb-status.techempower.com/unzip/results.2026-02-11-23-41-23-147.zip/results/20260204071442/ohkami-rt_nio/run/ohkami-rt_nio.log

@kanarus Please also check ohkami default:
https://tfb-status.techempower.com/unzip/results.2026-02-11-23-41-23-147.zip/ohkami

https://tfb-status.techempower.com/unzip/results.2026-02-11-23-41-23-147.zip/results/20260204071442/ohkami/fortune/raw.txt
https://tfb-status.techempower.com/unzip/results.2026-02-11-23-41-23-147.zip/results/20260204071442/ohkami/plaintext/raw.txt
...

* [ohkami] failed: json db query fortune update plaintext
* [ohkami-rt_nio] had problems starting or stopping

PR #10776, opened 2026-02-17
[rust/ohkami] Fix broken benchmarks

kanarus:
- close #10769 

## Core Fix (resolving the breaks)

- `rt_nio`: Updates `nio` dependency version from `0.0` into `0.1`.
- `default`: Handle database errors instead of panicking.

## Additional Updates

- Bumps package version and docker images.
- Bump/delete some dependencies, with related codes updated.
- Updates the maintainer information.


kanarus:
@joanhey Thanks for telling me they're broken. This PR will be the fix for #10769

PR #10784, opened 2026-02-18
Added jstach template variant, small Int cache optimization

inemtsev:
- By doing some benchmarks on my machine, I found that kotlinx.html DSL while flexible, seems to have a strong performance cost. I added a ktor variant that uses JStach instead (same templating as Spring uses here). 

- Some database flags that other good performaners seem to be using added.
- Also added Integer caching flag, which some top JVM performers here seem to use.


joanhey:
A bench is to find problems, It's NOT to be faster than others.

inemtsev:
> A bench is to find problems, It's NOT to be faster than others.

Yes, I have found a problem in the kotlinx.html library. 

The other tweaks are copied from other JVM frameworks here, so it is a more fair comparison

PR #10790, opened 2026-02-19
Fix connection exhaustion and closing of connections on errors

inemtsev:
- I think I found the problem, there were many requests made against one connection, which would kill the server under high load. 
- Claude also recommended closing connections on errors (doubt this was the problem, but why not).

@ShreckYe Can you please give blessings?


joanhey:

Example:

PR #10797, opened 2026-02-20
fix cached queries

HassanSharara:


HassanSharara:
That happend when i switched to io using use i think its low level issue i am working on it right now

PR #10847, opened 2026-03-06
Add tests for new general-purpose scripting language with compile-time type checking, safe concurrency

oakulikov:
Add new `Funxy/funxy-stdlib` implementation for Plaintext and JSON tests.

Includes minimal framework files: `worker.lang`, `supervisor.lang`, `funxy.dockerfile`, `benchmark_config.json`, and `README.md`.

Validated locally with `curl` (status/body/headers) and `wrk` smoke runs.


frameworks/Funxy/funxy-stdlib/worker.lang
```diff
@@ -0,0 +1,26 @@
...
+httpServe(8080, fun(req) {
+    match req.path {
+        "/plaintext" -> {
+            status: 200,
+            headers: headers_plain,
+            body: msg_plain
+        }
+        "/json" -> {
+            status: 200,
+            headers: headers_json,
```
msmith-techempower: JSON headers must be constructed on every request along with its object. Ideally, an ergonomic implementation for the developer should be provided by the framework (e.g., most devs wouldn't expect to have to set these headers manually in code when they are responding with json; the framework would do that lift automatically), but this is not strictly required and you may manually build the headers here.

msmith-techempower (changes requested):

frameworks/Funxy/funxy-stdlib/worker.lang
oakulikov: Thanks, fixed.

PR #10853, opened 2026-03-08
[Dart] Update `Shelf` framework and add AOT variant

iapicca:
## Summary
This PR bumps the dependencies for the [Dart](https://dart.dev/) [shelf](https://pub.dev/packages/shelf) framework and adds an AOT variant.

## Changes
* Bumped dependencies for the existing implementation.
* Added all necessary files for the AOT implementation and updated `benchmark_config.json`.

### DIFF
--- a/frameworks/Dart/shelf/benchmark_config.json
+++ b/frameworks/Dart/shelf/benchmark_config.json
@@ -20,9 +20,7 @@
         "display_name": "shelf",
         "notes": "",
         "versus": "None"
-      }
-    },
-      {
+      },
       "aot": {
         "json_url": "/json",
         "plaintext_url": "/plaintext",


frameworks/Dart/shelf/shelf_aot/bin/server.dart
```diff
@@ -0,0 +1,143 @@
...
+
+/// Completes the given [request] by writing the [response] as plain text.
+Response _sendText(String response) => _sendResponse(
+  HttpStatus.ok,
+  type: ContentType.text,
+  bytes: utf8.encode(response),
+);
+
+/// Responds with the JSON test to the [request].
+Response _jsonTest() => _sendJson(const {'message': 'Hello, World!'});
```
msmith-techempower, line 140: Just noting that I spent some time learning about Dart and this `const` bit. Seems to do the work of instantiating a new object from that constant map every request.

iapicca: @msmith-techempower 
it's awesome that you are learning Dart <3

AFAIK  `const` is calculated at `compile time`, so there _shouldn't_ be any perfermance difference between
```dart
Response _jsonTest() => _sendJson(const {'message': 'Hello, World!'});
```
and
```dart
const _message =  {'message': 'Hello, World!'};
/// ...
Response _jsonTest() => _sendJson(_message);
```
literals like `String`s without runtime variables (both `var` and `final`) _should_ be automatically compiled as `const` like below
```dart
Response _plaintextTest() => _sendText('Hello, World!');
```

side note: I tried to leave most of the `native` implementation` "untouched" and the `aot` version as close as possible to `native` to keep the comparison relevant.

PR #10868, opened 2026-03-11
Upgrade to Quarkus 3.32.2

tsegismont:
It was required to add the `os-maven-plugin`. Otherwise, the container image build fails with:

```
Could not find artifact io.netty:netty-transport-native-unix-common:jar:${os.detected.name}-${os.detected.arch}:4.1.130.Final in central
```

Also, Hibernate no longer has the update method on the regular session.

And, for convenience, added `quarkus.hibernate-orm.log.sql` in application.properties

### DIFF
--- a/frameworks/Java/quarkus/README.md
+++ b/frameworks/Java/quarkus/README.md
@@ -13,6 +13,10 @@ There are currently 2 implementations:
 
     ./tfb --mode verify --test quarkus quarkus-hibernate-reactive
 
+After the database container image is created, it can be started independently.
+
+    docker run -d --name tfb-database -e POSTGRES_DB=hello_world -e POSTGRES_PASSWORD=benchmarkdbpass -e POSTGRES_USER=benchmarkdbuser -p 5432:5432 techempower/postgres:latest
+
 ## Versions
 
 * [Java OpenJDK 17](http://openjdk.java.net/)
--- a/frameworks/Java/quarkus/pom.xml
+++ b/frameworks/Java/quarkus/pom.xml
@@ -8,15 +8,15 @@
     <packaging>pom</packaging>
 
     <properties>
-        <compiler-plugin.version>3.14.0</compiler-plugin.version>
-        <maven.compiler.release>21</maven.compiler.release>
+        <compiler-plugin.version>3.15.0</compiler-plugin.version>
+        <maven.compiler.release>25</maven.compiler.release>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
         <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
-        <quarkus.platform.version>3.21.2</quarkus.platform.version>
+        <quarkus.platform.version>3.32.2</quarkus.platform.version>
         <skipITs>true</skipITs>
-        <surefire-plugin.version>3.5.2</surefire-plugin.version>
+        <surefire-plugin.version>3.5.4</surefire-plugin.version>
         <!-- Check https://github.com/netty/netty-incubator-transport-io_uring/tags -->
         <netty.io_uring.version>0.0.26.Final</netty.io_uring.version>
         <rocker.version>1.3.0</rocker.version>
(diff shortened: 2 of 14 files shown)


frameworks/Java/quarkus/README.md
```diff
@@ -13,6 +13,10 @@ There are currently 2 implementations:
 
     ./tfb --mode verify --test quarkus quarkus-hibernate-reactive
 
+After the database container image is created, it can be started independently.
+
+    docker run -d --name tfb-database -e POSTGRES_DB=hello_world -e POSTGRES_PASSWORD=benchmarkdbpass -e POSTGRES_USER=benchmarkdbuser -p 5432:5432 techempower/postgres:latest
+
```
joanhey, line 19: Sorry, why do you need that ?

joanhey (commented):
Why ? 

PD: it's pure curiosity !!

frameworks/Java/quarkus/README.md
tsegismont: It was convenient to have the database started as `tfb` would do it, while running the Quarkus server in dev mode.

tsegismont:
@Sanne @lucamolteni can you please take a look at the changes in `WorldRepository`? Hibernate Reactive no longer has the `update` method on a regular `Session`, so I changed to `StatelessSession`. It's similar to the implementation in the benchmark with Hibernate ORM

Sanne:
The reason I used "update" is to benefit from batching - not using a batch in a `StatelessSession` would be a disaster for efficiency.
@lucamolteni or @DavideD can you check how to do it correctly in the new API?

DavideD:
I will have a look, but batching requires the use of the other `update` methods when using the stateless session:
```java
	/**
		 * Update multiple rows, using the number of the
		 * given entities as the batch size.
		 *
		 * @param entities detached entity instances
		 *
		 * @see org.hibernate.StatelessSession#update(Object)
		 */
		Uni<Void> updateAll(Object... entities);

		/**
		 * Update multiple rows.
		 *
		 * @param batchSize the batch size
		 * @param entities detached entity instances
		 *
		 * @see org.hibernate.StatelessSession#update(Object)
		 */
		Uni<Void> updateAll(int batchSize, Object... entities);

		/**
		 * Update multiple rows, using the size of the
		 * given list as the batch size.
		 *
		 * @param entities detached entity instances
		 *
		 * @see org.hibernate.StatelessSession#update(Object)
		 */
		Uni<Void> updateMultiple(List<?> entities);
```

I think it's the same for ORM.
The reason is that batching doesn't allow the immediate execution of the queries and that's a prerogative of the stateless session

DavideD:
I'm not sure if saving each entity separately is a requirement of the benchmark, but, if you want to have batching and call update on each entity, you cannot use a `StatelessSession`.

DavideD:
I noticed that in the test we are already passing a list of entities to update, so it should be ok to change it. I will create a PR.

DavideD:
@tsegismont, I applied the changes I think are needed here: https://github.com/DavideD/FrameworkBenchmarks/commits/quarkus-upgrade/

* 7bfb45e5ea389ff44bed2842cace66ea1575b256 With the stateless session we should use `.updateMultiple(...)` for both Hibernate Reactive and ORM
* ec368ac8905cee18656cd444183ebca2617bd7e1 In the Hibernate Reactive benchmark, we shouldn't create data using `Uni.combine().all()`, even if it seems to work in this case

I'm trying to figure out how to test it, though. But let me know if you want me to send a new PR or you are ok with cherry picking that changes you need.

tsegismont:
@DavideD can you please send a PR for the `quarkus-upgrade` branch in my fork?

DavideD:
> @DavideD can you please send a PR for the `quarkus-upgrade` branch in my fork?

Sure, I noticed that the change don't pass CI. I will try to fix it and send the PR

DavideD:
> @DavideD can you please send a PR for the `quarkus-upgrade` branch in my fork?

See https://github.com/tsegismont/FrameworkBenchmarks/pull/4

tsegismont:
Thank you @DavideD , the changes are in.

@Sanne can you please review again and add your :+1: so that the PR can me merged? Thanks

Sanne:
> Thank you @DavideD , the changes are in.
> 
> @Sanne can you please review again and add your 👍 so that the PR can me merged? Thanks

Did someone verify that the performance doesn't change substantially when compared to the previous versions?
Sorry I won't be able to do that this week.

If metrics look fine then you can consider me happy! No need to look into the details of the code.

franz1981:
@Sanne at worse we are way distant from an offical round, so we can use the nightly on techempower itself to verify such

tsegismont:
@Sanne I didn't verify because I shared this PR after discussing the idea with @franz1981 and he advised to take a look at nightly build results as explained above.

If that's fine with everyone, can you a :+1: so that the PR is merged and we see the data? Thank you!

PR #10902, opened 2026-03-19
Add TurboAPI framework [Python + Zig]

justrach:
## New Framework: TurboAPI

**Language:** Python 3.14t (free-threaded, no GIL)
**HTTP Core:** Zig 0.15 native server (24-thread pool)
**Tests:** json, plaintext

### What is TurboAPI?

[TurboAPI](https://github.com/justrach/turboAPI) is a drop-in FastAPI replacement that moves the entire HTTP layer (TCP, parsing, routing, JSON serialization, CORS) into Zig. Python only runs your business logic.

- FastAPI-compatible API (`@app.get`, `@app.post`, Depends, middleware)
- 287 tests passing
- Zero-alloc response pipeline (stack buffers, single write syscall)
- Response caching (body cached after first Python call, subsequent requests skip Python)
- Zig-native CORS (pre-rendered headers, 0% overhead)

### Test endpoints

| Test | URL | Response |
|------|-----|----------|
| JSON | `/json` | `{"message":"Hello, World!"}` |
| Plaintext | `/plaintext` | `Hello, World!` |

### Local verification

```
Verification Summary
| turboapi
|       json          : PASS
|       plaintext     : PASS
```

### Files

- `frameworks/Python/turboapi/benchmark_config.json`
- `frameworks/Python/turboapi/turboapi.dockerfile`
- `frameworks/Python/turboapi/app.py`
- `frameworks/Python/turboapi/README.md`

### Docker

Uses `uv` to install Python 3.14t free-threaded + Zig 0.15 for the native HTTP backend. Builds and runs on both x86_64 and aarch64.


remittor:
Great project! I'm looking forward to the results of the tests!

frameworks/Python/turboapi/app.py
```diff
@@ -0,0 +1,18 @@
...
+os.environ["TURBO_DISABLE_RATE_LIMITING"] = "1"
+
+from turboapi import TurboAPI
+from turboapi.responses import PlainTextResponse
+
+app = TurboAPI()
+
+@app.get("/json")
+def json_test():
+    return {"message": "Hello, World!"}
```
joanhey, line 12: For each request, an object mapping the key message to Hello, World! must be instantiated.

PD: the bench is about serialize a json object, not to send raw text directly, for that is the plaintext bench.

joanhey (changes requested):

frameworks/Python/turboapi/turboapi.dockerfile
```diff
@@ -0,0 +1,36 @@
...
+RUN curl -LsSf https://astral.sh/uv/install.sh | sh
+ENV PATH="/root/.local/bin:$PATH"
+RUN uv python install 3.14t && uv venv --python 3.14t /venv
+ENV PATH="/venv/bin:$PATH" VIRTUAL_ENV="/venv"
+
+WORKDIR /turboapi
+
+# Bust Docker cache for git clone
+ARG CACHEBUST=1
+RUN git clone --depth 1 --branch main https://github.com/justrach/turboAPI.git . && \
```
joanhey: Which version ??

frameworks/Python/turboapi/turboapi.dockerfile
```diff
@@ -0,0 +1,36 @@
...
+    git clone --depth 1 https://github.com/justrach/dhi.git /dhi
```
joanhey: Which version ??

joanhey (changes requested):

joanhey:

Example:

frameworks/Python/turboapi/turboapi.dockerfile
```diff
@@ -0,0 +1,36 @@
...
+# uv + Python 3.14t free-threaded
+RUN curl -LsSf https://astral.sh/uv/install.sh | sh
+ENV PATH="/root/.local/bin:$PATH"
+RUN uv python install 3.14t && uv venv --python 3.14t /venv
+ENV PATH="/venv/bin:$PATH" VIRTUAL_ENV="/venv"
+
+WORKDIR /turboapi
+
+# Bust Docker cache for git clone
+ARG CACHEBUST=1
```
joanhey: This is unnecessary !!
When you add the version in Git clone.

joanhey (changes requested):

frameworks/Python/turboapi/app.py
remittor: Here, it's not a string that's returned from the function, but a dictionary. The framework internally serializes the dictionary (a collection of objects) into a byte array.
Example: https://github.com/TechEmpower/FrameworkBenchmarks/blob/a087d02c3bb78d973f0af92b3b1989ad3adc24da/frameworks/Python/emmett/app.py#L41-L45

frameworks/Python/turboapi/turboapi.dockerfile
remittor: @justrach 
For tests to be repeatable, there is no need to specify the repository HEAD!
By the way, the socketify project also abuses this.

frameworks/Python/turboapi/app.py
justrach: Fixed — added `TURBO_DISABLE_CACHE=1` to ensure the dict is created and serialized fresh on every request. No caching.

frameworks/Python/turboapi/turboapi.dockerfile
justrach: Pinned to `v1.0.17`.

frameworks/Python/turboapi/turboapi.dockerfile
justrach: Pinned to `v1.2.1`.

frameworks/Python/turboapi/turboapi.dockerfile
justrach: Removed.

frameworks/Python/turboapi/app.py
justrach: Thanks for clarifying — yes, it returns a dict that the framework serializes internally. Also disabled response caching to ensure it's fresh per request.

frameworks/Python/turboapi/turboapi.dockerfile
justrach: Fixed — both repos now pinned to specific tags (turboAPI v1.0.17, dhi v1.2.1).

PR #10909, opened 2026-03-20
Vertx db parameter optimization

doxlik:
This PR adds caching of Tuples in Vertx. 

Tuple is immutable wrapper of arguments for prepared query. Considering it is Vertx specific abstraction it should be fully legit to cache them because effectively it is the same as if library would not have this abstraction and would accept just primitive values for arguments.


doxlik:
@vietj @tsegismont

vietj:
Looks good to me

franz1981:
Well done @doxlik ! Can you do the same for the vertx based variants in Quarkus as well? 🙏

doxlik:
@franz1981 done

PR #10923, opened 2026-03-22
refactor: update codeigniter4 load model once

ddevsr:
Refactor to load the World model once instead of multiple times in loops.

### DIFF
--- a/frameworks/PHP/codeigniter/app/Controllers/Full.php
+++ b/frameworks/PHP/codeigniter/app/Controllers/Full.php
@@ -40,7 +40,7 @@ class Full extends Controller
         $modelWorld = model('World');
 
         while ($queries--) {
-            $row = model('World')->find(mt_rand(1, 10000));
+            $row = $modelWorld->find(mt_rand(1, 10000));
             $row['randomNumber'] = mt_rand(1, 10000);
             $world[] = $row;
             $modelWorld->save($row);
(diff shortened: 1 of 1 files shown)


ddevsr:
Sorry mentions you @joanhey, i will update codeigniter4 to latest version (4.7) in next PR with using opcache + jit too.

joanhey:
Welcome to the bench @ddevsr.
Always is better that the bench code is optimized by a user specialized for each framework.

Actually the bench use CI 4.7 and OPCache + JIT.

`codeigniter:   - Downloading codeigniter4/framework (v4.7.0)`

https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/PHP/codeigniter/deploy/conf/php.ini#L1765-L1767

https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/PHP/codeigniter/deploy/conf/php.ini#L1912-L1915

Thanks.

joanhey (approved):

frameworks/PHP/codeigniter/app/Controllers/Full.php
```diff
@@ -33,11 +36,14 @@ public function update($queries = 1)
         $queries = is_numeric($queries) ? min(max($queries, 1), 500) : 1;
         $world = [];
 
+        // Load model once
+        $modelWorld = model('World');
+
         while ($queries--) {
             $row = model('World')->find(mt_rand(1, 10000));
```
joanhey: You could change it here too.

ddevsr: Done

PR #10929, opened 2026-03-23
Upgrade Carter to net10

jchannon:
This updates the Carter version


frameworks/CSharp/carter/Benchmarks/JsonModule.cs
joanhey, line 24: It is NOT permitted in the json tests, constants or pre-calculated header for the `Content-Length`.
Need to be calculated for each request.

joanhey (changes requested):

joanhey:

Example:

